PR #29, opened 2019-03-21
Add information messages in sessions nodes

### LINKED ISSUE #3: Indication in dataset view that pattern didn't match any files
When I first started using the plugin, I had some difficulty understanding that I didn't have a pattern set for my dataset filters. This caused me some pain as I was able to see my zowe profiles but couldn't figure out why there were no datasets listed.

Perhaps there should be some indication that the pattern specified doesn't match anything on the mainframe. At the very least, there should be a more clear indication that you need to specify a pattern when it is missing.

crshnburn:
Resolves #3 by adding information nodes to a session when it is first added to indicate a search should be performed, and then if the search returns an empty list of datasets a message to indicate this has happened.


MikeBauerCA:
@ChrisBoehmCA @zFernand0 could you please review this PR?

ChrisBoehmCA:
Nice. Worked well for me when entering an invalid pattern
image

Issue #36, opened 2019-04-08
Missing USS functionality


Colin-Stone:
The USS extension requires menu options for Create, Delete and Rename

Colin-Stone:
For reference the API's already exist for create and delete and are:
zowe.Create.uss, zowe Delete.ussFile

Other zosmf utility functions such as move and copy files are not available in the CLI so Copy/Rename functionality would not be available until added in the core

PR #52, opened 2019-04-18
USS add create and remove node functionality

kristinochka:
Related to: https://github.com/zowe/vscode-extension-for-zowe/issues/36

image
Delete node
create file

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,8 @@ All notable changes to the "vscode-extension-for-zowe" extension will be documen
 
 Check [Keep a Changelog](http://keepachangelog.com/) for recommendations on how to structure this file.
 
-
+## 0.11.0
+ - Create and delete functionality for USS Files and directories added as menu items.
 ## 0.10.4
  - Add additional log messages
 ## 0.10.3
--- a/__tests__/extension.test.ts
+++ b/__tests__/extension.test.ts
@@ -74,6 +74,7 @@ describe("Extension Unit Tests", async () => {
     const delDataset = jest.fn();
     const Create = jest.fn();
     const dataSetCreate = jest.fn();
+    const ussCreate = jest.fn();
     const Download = jest.fn();
     const dataSet = jest.fn();
     const ussFile = jest.fn();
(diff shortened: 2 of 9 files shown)


MikeBauerCA:
Just from the commit list, I am very eager to try this functionality! Thank you for the contribution!!

kristinochka:
@MikeBauerCA I have signed all my commits now so the checks are passing. Please, take a look.

MikeBauerCA:
@Colin-Stone @ChrisBoehmCA could you please review this PR?

ChrisBoehmCA:
Hi

Pretty cool stuff! Thanks for the changes. 

I noticed a few things. When creating a USS directory or file, the text in the background says "Name of Member". It should probably say Name of USS File  or Name of USS Directory depending on which thing the user clicked. 

I'm also getting this error sometimes when deleting USS directories, but then the delete seems to succeed as the files are gone. In fact if I delete a directory with nested directories, I get this error twice from one click of "Delete". Ever seen this? I wonder if it's coming from the call to delete or the call to refresh the directory after the delete? 

image

Colin-Stone:
Shouldn't be adding the extra slash being prepended to the directory. This is because of a flaw in the zosmf API and corrupts the saved value. URLEncodoing is an option

kristinochka:
@ChrisBoehmCA I made a fix Collin suggested, could you try again? The issue should be gone.

ChrisBoehmCA:
Thanks. 
I still got the error a few times deleting directories though, but can't reliably reproduce it. I got it a few times when I first started and then have been trying to reproduce it for a while to no avail. I guess it's a timing based thing where if a refresh  (or something) is already happening when I delete, it causes an error. But it doesn't seem like an issue with deleting itself because I've tried many times now without the error. 

I noticed that if you delete a file or folder at the top level, the view doesn't refresh until you manually click the refresh button. But for folders nested deeper it automatically refreshes. could this be fixed? 

Also it seems like there's no way to add a new file or directory to the top level folder you specify. For example if I enter /u/users/myuser I can't create /u/users/myuser/myfile.txt by right clicking on the profile name.  I guess you can get around this by specifying the parent directory though

Colin-Stone:
I'm happy with these changes, thanks Kristina.

I saw some few issues with directory refresh similar to what Chris described but I don't feel is related to these changes.

I vote we should Approve and Publish

Issue #54, opened 2019-04-23
Unit tests appears broken


Colin-Stone:
Since adding the additional debug logs some of the unit tests are failing. Mock the appropriate missing function

PR #55, opened 2019-04-23
Add logger mock

Colin-Stone:
One test is failing because we don't mock the log.getAppLogger function
https://github.com/zowe/vscode-extension-for-zowe/issues/54


ChrisBoehmCA:
All unit tests passing for me now :+1:

PR #58, opened 2019-04-25
Toggle download as binary to avoid conversion

### LINKED ISSUE #57: USS files are downloaded without taking into account the file encoding tag extended attribute in zFS
I have a file that is encoded and tagged as ISO8859-1 (ASCII), eg notice the `t ISO8859-1` and `T=on`:

```
/u/cicprov/mnt/CICPY000/bundles/cics-nodejs-invoke_1.0.0/META-INF:>ls -lAT
total 16
t ISO8859-1   T=on  -rw-r--r--   1 COCKERM  TSOUSER      338 Apr 25 11:24 cics.xml

/u/cicprov/mnt/CICPY000/bundles/cics-nodejs-invoke_1.0.0/META-INF:>cat cics.xml
<manifest xmlns="http://www.ibm.com/xmlns/prod/cics/bundle" bundleVersion="1" bundleRelease="0" id="cics-nodejs-invoke" bundleMajorVer="1" bundleMinorVer="0" bundleMicroVer="0"><define name="cics-nodejs-invoke" type="http://www.ibm.com/xmlns/prod/cics/bundle/NODEJSAPP" path="nodejsapps/cics-nodejs-invoke.nodejsapp"></define></manifest>
```
When I use the Zowe extension for Visual Studio Code to edit this file it is not shown correctly:
image.

crshnburn:
Resolves #57 as a workaround.

Adds a new menu item which allows to toggle between downloading the file
as text which will do data conversion even if the file doesn't require it
or binary which will just get the contents of the file allowing editing
of non EBCDIC files in an editor.

### DIFF
--- a/__tests__/extension.test.ts
+++ b/__tests__/extension.test.ts
@@ -286,7 +286,7 @@ describe("Extension Unit Tests", async () => {
                     getChildren: mockGetUSSChildren,
                 }
         });
-        expect(registerCommand.mock.calls.length).toBe(27);
+        expect(registerCommand.mock.calls.length).toBe(29);
         expect(registerCommand.mock.calls[0][0]).toBe("zowe.addSession");
         expect(registerCommand.mock.calls[0][1]).toBeInstanceOf(Function);
         expect(registerCommand.mock.calls[1][0]).toBe("zowe.addFavorite");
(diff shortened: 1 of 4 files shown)


ChrisBoehmCA:
Hi! Thanks for adding this. I think you might have missed at least one instance of the context value being checked to see if it's "file" rather than your new binaryFile/textFile

If I open the file first, and then try to toggle binary, it doesn't seem to work. It seems I have to toggle binary before the first time I click on the file.  Is there a way to fix this? Also when I do the toggle on a root level node, the text toggles back and forth between "Toggle Text" and "Toggle Binary", but this doesn't seem to happen for nodes within child folders. It just says "Toggle Binary" over and over. Any idea why this would be?

crshnburn:
@ChrisBoehmCA, thanks for the review comments, I've resolved the context value checking.

The behaviour is updated so that toggling between text and binary automatically opens the file with the new setting.

The issue with files in directories was due to reloading the content of the directory from the server and losing the knowledge of which files had been toggled to binary. Local state is now cached to prevent this.

Colin-Stone:
I've got this installed and it is really useful. 

Question for a future enhancement but I'm wondering if anyone knows if there is a way we can put some sort of indicator or flag against the file name that shows the state of the toggle. So for example I'm looking at a series of files that all need treating this way and to view them I will need to toggle them one at a time. It would be good to see which ones I have toggled

ChrisBoehmCA:
Cool idea Colin. It'd be neat if we could have an icon or just part of the label that marked a file as binary mode

ChrisBoehmCA:
I think I'll publish this and #59 together after merging both

PR #78, opened 2019-05-09
Add the ability to display data set attributes by right clicking on a data set

ChrisBoehmCA:


Colin-Stone:
This is going to be very useful.

Still considering location of the view.. Was the original intention to put this in the main explorer container? In other words on the left? Not sure I like it in the editor section. In another scenario we also had a view in the bottom section where the terminal is for properties. Maybe it should be an overview view on the right hand side?

If we go with a view at the bottom I am wondering if it would be nicer to have it respond to what is selected rather than a menu item. 

Unit test failing. We need an entry for zowe.showDSAttributes and function count incremented. Maybe add that function to the end of the list as the unit test expect list is massive and we don't really want to shuffle everything up.

Should this be a read only editor for now as you cant save anything?. 

At some stage maybe we should remove the throw new Error("hi fernando") in extension.ts ;-)

ChrisBoehmCA:
Updated it to be a read only WebView window

Colin is out now. @zFernand0  can you review when you get the chance? Thanks

PR #88, opened 2019-05-23
JCL Submission improvements

### LINKED ISSUE #81: Add ability to obtain the raw JCL from a job on spool.
The z/OSMF REST API set seems to support getting the raw input JCL: 

```
GET /zosmf/restjobs/jobs/<jobname>/<jobid>/files/JCL/records
```

Some of the spool viewer products on z/OS allow you to get the raw JCL from a previously submitted job. They usually open it in an ISPF edit session, where you can then adjust and re-submit easily. This functionality would be useful in the zowe extension.

crshnburn:
Resolves #81 by adding a menu item to jobs allowing the user to get the JCL.

Enhances the ability to submit JCL by allowing the content of any editor to be submitted by providing a list of available profiles to use to submit the Job if it can't be determined programatically.

Exploits the functionality made available in the [April 2019 release](https://code.visualstudio.com/updates/v1_34#_command-links-in-notifications) to provide a link in the information message which will open the spool for the newly submitted job.

### DIFF
--- a/__tests__/extension.test.ts
+++ b/__tests__/extension.test.ts
@@ -99,7 +99,7 @@ describe("Extension Unit Tests", async () => {
     const mkdirSync = jest.fn();
     const getAllProfileNames = jest.fn();
     const createTreeView = jest.fn();
-    const Uri = jest.fn();
+    // const Uri = jest.fn();
     // const parse = jest.fn();
     const pathMock = jest.fn();
     const registerCommand = jest.fn();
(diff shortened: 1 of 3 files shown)


crshnburn:
Leaving this PR in draft until #85 is merged so more tests can be added for this function.

crshnburn:
@Colin-Stone yep, that was my plan, needed #85 to land into master so that there was a good base to add the extra tests onto.

PR #109, opened 2019-06-12
SpoolsinJobsTree

Colin-Stone:

Asking @crshnburn to have a look and pass comments first


crshnburn:
@Colin-Stone The visual changes look good, hopefully makes it easier to work with multiple jobs. However it looks like the links when submitting the JCL to the submitted job aren't working because the `zowe.setJobSpool` command has been removed as part of this change. 

This previously set the spool view to the recently submitted job, so as part of this change it would make sense to have it now refresh the jobs tree and set the new job as the selected node (which is expanded).

zFernand0:
Sorry about this, but would it be possible to resolve the conflicts?

Colin-Stone:
That's fine I need to fix some other bits up anyway

Colin-Stone:
The updates added reinstate the setJob code. There is an edge case I would like to fix later which is when the new job is submitted using a different profile to the one in the explorer view It won't be highlighted. It would be neat to automatically add that profile into the view.
I don't understand the conflicting files. The Changelog is simple additions and the two tests files have been reworked because in the case of zosspool the view no longer exists. ANy ideas hiow I can clean these up?

tjohnsonBCM:
Sorry for the conflicts, looks like they've all been merged correctly except for one: `package.json` needs an updated version number.

PR #110, opened 2019-06-12
CI/CD initial steps

zFernand0:
- resolve all vulnerabilities
- baby steps on the versioning and changelog automation


zFernand0:
Subsequent PRs will at least run tests and auditing. :smile:

@Colin-Stone, We will need to work on "sharing" the publishing rights of the VSCE.
The pipeline will use my personal token for now, but maybe we should have an azure account for the zowe-robot so that we can use that token instead.

Colin-Stone:
> Subsequent PRs will at least run tests and auditing. 😄
> 
> @Colin-Stone, We will need to work on "sharing" the publishing rights of the VSCE.
> The pipeline will use my personal token for now, but maybe we should have an azure account for the zowe-robot so that we can use that token instead.

I hadn't considered this as I hadn't published to market place before. Thanks for letting us use your personal token for now but your suggestion for a zowe-robot one seems the right way to go. Who's palm do we have to cross with silver to make this happen?.

zFernand0:
> > Subsequent PRs will at least run tests and auditing. 😄
> > @Colin-Stone, We will need to work on "sharing" the publishing rights of the VSCE.
> > The pipeline will use my personal token for now, but maybe we should have an azure account for the zowe-robot so that we can use that token instead.
> 
> I hadn't considered this as I hadn't published to market place before. Thanks for letting us use your personal token for now but your suggestion for a zowe-robot one seems the right way to go. Who's palm do we have to cross with silver to make this happen?.

We can coordinate with @MarkAckert for this.
The VSCode Marketplace Support team is already waiting for use to give them the "VSID" that gets generated for the user (i.e. zowe-robot) :smile:

zFernand0:
Thanks to @MarkAckert, who already helped offline.

Please be aware that there may be a few changes to the versioning and changelog automated mechanisms in the future.

Merging. ;)

PR #116, opened 2019-06-17
USS Favorite profile name fix

lauren-li:
Related to issue zowe/zowe-cli#297

Fixes the following issues with USS Favorites:
- Refresh/update file content when opening a file from Favorites section.
- Use correct profile name when favoriting the child of an already-favorited directory.

No new features in this pull request.


lauren-li:
To test this fix:
1. In the USS section of the extension, under the Profiles section, favorite a directory that contains files.
2. Once the favorited directory appears in the USS Favorites section, expand it to reveal its child items (i.e. files).
3. Favorite one of the child files.
4. Observe the label of the newly-favorited child file. *The profile prefix in brackets should match the name of the parent directory's profile prefix.*
5. Open the child file that was just favorited and make some changes.
6. Open the same child file from the Profiles section of the USS explorer. *The document that is opened should be the same as in step 5, and the changes made in step 5 should also be reflected here.*
- (Changes made to the child file when opening the file from the Profiles section should also be reflected when the child file is opened from the Favorites section, as well.)

rodney-wilson:
The failing integration test above has been resolved.

Colin-Stone:
I tried following the test above but it didn't work for me. First point 4 doesn't appear to work
image
Secondly the files don't seem to be in synch as described in point 6

Issue #98, opened 2019-06-02
Overhaul VSCode testing process


Colin-Stone:
Rethink the testing strategy should be for VSCode

Colin-Stone:
1. Added in timeout fixes to installation tests

Colin-Stone:
This is complete although we should still be adjusting the code to reduce the the volume of code in the extension file and assign it to the more appropriate file.

PR #121, opened 2019-06-19
Include lint and update files

Colin-Stone:
Added lint support to the run scripts in package.json with the obvious effect of creating a lot of fixes for a lot of files. All this is part of the testing issue https://github.com/zowe/vscode-extension-for-zowe/issues/98


package.json
```diff
@@ -762,15 +762,18 @@
     }
   },
   "scripts": {
-    "vscode:prepublish": "npm run build && npm run markdown",
+    "vscode:prepublish": "npm run build && npm run lint",
```
zFernand0: I'd suggest adding `npm run lint` as part of the `build` script. That will make sure we run the linter every time we build the source.
i.e. `"build": "npm run lint && npm run license && tsc --pretty",`

Colin-Stone: Good idea, I've been reorganising a little since too for similar reasons so I'll include that suggestion, cheers.

Colin-Stone: Good thing you suggested that. I needed to add tslint into package.

PR #127, opened 2019-06-24
Context menu grouping for MVS and USS

lauren-li:
Grouped context menu items in MVS and USS explorers according to the design proposed by our internal UX team.

Below is the grouping scheme I followed. The numberings keep the commands in the desired order:

`0_mainframeInteraction`:
- Pull from Mainframe
- Safe Save - Merge if Necessary

`1_create`:
- Create New Member
- Upload Member...
- Create Directory
- Create File

`2_information`:
- Show Data Set Attributes

`3_submitJob`:
- Submit Job

`4_workspace`:
- Add Favorite
- Remove Favorite

`5_modification`:
- (0) Toggle Binary
- (1) Toggle Text
- (2) Rename
- (3) Delete/Delete Data Set/Delete PDS/Delete Member


lauren-li:
Thanks for reviewing this, Colin! I'll look into unit tests for context menu groupings today.

lauren-li:
@Colin-Stone or others, any suggestions for improvements to the category names or the organization of the groupings would be greatly appreciated! 

In particular:
- I am looking for a more general category name to use over `3_submitJob`.
- I am unsure which category `Toggle Binary` and `Toggle Text` should go under. We could create a new category for it, but I'm hesitating to create too many categories.

Colin-Stone:
Toggle binary and toggle text is the same menu that switches it's label. I notice it also opens the file if that can be considered it's function. When we did this type of thing for Eclipse there were always specific categories and everything else we put in a category called "Others".. At the bottom

lauren-li:
Thank you for your input on this, Colin! I think it makes sense to maintain consistency with how VS Code orders their context menu items with `Rename` and `Delete` always in the last (bottom) group. However, per your suggestion, I have moved `Toggle Binary` and `Toggle Text` into a different group, named `3_systemSpecific`, which they share with `Submit Job`. This grouping and ordering would effectively match what our UX team originally suggested.

Below is an updated proposed organization of all the context menu items. (`updated` items are shown outside of the quotes):
>`0_mainframeInteraction`:
>- Pull from Mainframe
>- Safe Save - Merge if Necessary
>
>`1_create`:
>- Create New Member
>- Upload Member...
>- Create Directory
>- Create File
>
>`2_information`:
>- Show Data Set Attributes

`3_systemSpecific`:
- Submit Job
- Toggle Binary
- Toggle Text

>`4_workspace`:
>- Add Favorite
>- Remove Favorite

`5_modification`:
- Rename
- Delete/Delete Data Set/Delete PDS/Delete Member

Colin-Stone:
> Thank you for your input on this, Colin! I think it makes sense to maintain consistency with how VS Code orders their context menu items with `Rename` and `Delete` always in the last (bottom) group. However, per your suggestion, I am thinking to move `Toggle Binary` and `Toggle Text` into a different group, named `3_systemSpecific`, which they would share with `Submit Job`. This grouping and ordering would effectively match what our UX team originally suggested.
> 
> Below is an updated proposed organization of all the context menu items. Feedback is welcome! (`updated` items are shown outside of the quotes):
> 
> > `0_mainframeInteraction`:
> > 
> > * Pull from Mainframe
> > * Safe Save - Merge if Necessary
> > 
> > `1_create`:
> > 
> > * Create New Member
> > * Upload Member...
> > * Create Directory
> > * Create File
> > 
> > `2_information`:
> > 
> > * Show Data Set Attributes
> 
> `3_systemSpecific`:
> 
>     * Submit Job
> 
>     * Toggle Binary
> 
>     * Toggle Text
> 
> 
> > `4_workspace`:
> > 
> > * Add Favorite
> > * Remove Favorite
> 
> `5_modification`:
> 
>     * Rename
> 
>     * Delete/Delete Data Set/Delete PDS/Delete Member

Thanks Lauren that looks good. The main thing is that we keep the odd ones separate. Although the subject is MVS and USS will you do something similar for Jobs whilst you are there?. All we need is a separator between Get JCL and the commands.

lauren-li:
@Colin-Stone I can add context menu grouping for Jobs, as well. Do you have a suggestion for what categories `Get JCL`, `Issue Modify Command`, and `Issue Stop Command` should go under?

I was thinking of adding them to these categories, but I think you may have a better idea of where they should go:

`0_mainframeInteraction`:
- Get JCL

`3_systemSpecific`:
- `Issue Modify Command`
- `Issue Stop Command`

Colin-Stone:
> @Colin-Stone I can add context menu grouping for Jobs, as well. Do you have a suggestion for what categories `Get JCL`, `Issue Modify Command`, and `Issue Stop Command` should go under?
> 
> I was thinking of adding them to these categories, but I think you may have a better idea of where they should go:
> 
> `0_mainframeInteraction`:
> 
>     * Get JCL
> 
> 
> `3_systemSpecific`:
> 
>     * `Issue Modify Command`
> 
>     * `Issue Stop Command`

Exactly how I would do it.

lauren-li:
Thanks Colin! I will work on implementing the updated context menu groupings and respective unit tests.

lauren-li:
@Colin-Stone @zFernand0 I have been looking at unit testing the context menu groupings in this PR, but am having some difficulty figuring out how to approach these tests, as all of my changes are in the `package.json` file, and I do not see anywhere in our code where we make a function call to retrieve or manipulate the contents of `package.json`. I did not add any new functionality to create the groupings, since VS Code's API already provides a [documented way](https://code.visualstudio.com/api/references/contribution-points#Sorting-of-groups) to group context menu commands via `package.json`.

It seems that VS Code just reads the `package.json` file and handles the context menu groupings completely within their own code (perhaps in their [menuExtensionPoint.ts file](https://github.com/microsoft/vscode/blob/acb1e2cb3a24e10068c369976ea5c87fbfdb73ec/src/vs/workbench/api/common/menusExtensionPoint.ts)). Does it still make sense to write unit tests for this addition?

zFernand0:
> @Colin-Stone @zFernand0 I have been looking at unit testing the context menu groupings in this PR, but am having some difficulty figuring out how to approach these tests, as all of my changes are in the `package.json` file ...

Hey @lauren-li,
I don't know if it's a "best practice" but we've seen and used "snapshots" for deterministic testing of some piece of code. For instance, if you believe the `commands` array is good as it is on this PR, you could add a small unit tests (for regression purposes only) that matches a snapshot.
Here is an example, https://github.com/zowe/zowe-cli/blob/master/packages/zosjobs/__tests__/api/JobsConstants.unit.test.ts

lauren-li:
Thanks, @zFernand0! I will look into giving snapshots a try.

lauren-li:
I created a snapshot test, but it seems to only capture the ZoweUSSNode itself, and not the attached context menu. There does not appear to be any information related to commands on the ZoweUSSNode. I have pasted the resulting snapshot below:
```
// Jest Snapshot v1, https://goo.gl/fbAQLP

exports[`Unit Tests (Jest) Checks Zowe USS Node snapshot 1`] = `
ZoweUSSNode {
  "binary": false,
  "binaryFiles": Object {},
  "children": Array [],
  "collapsibleState": 1,
  "contextValue": "directory",
  "dirty": true,
  "fullPath": "",
  "label": "testDir",
  "mCollapsibleState": 1,
  "mLabel": "testDir",
  "mParent": ZoweUSSNode {
    "binary": false,
    "binaryFiles": Object {},
    "children": Array [],
    "collapsibleState": 1,
    "contextValue": "uss_session",
    "dirty": true,
    "fullPath": "",
    "label": "root",
    "mCollapsibleState": 1,
    "mLabel": "root",
    "mParent": null,
    "mProfileName": undefined,
    "parentPath": null,
    "profileName": "",
    "session": Session {
      "ISession": Object {
        "hostname": "fake",
        "password": "fake",
        "protocol": "https",
        "type": "basic",
        "user": "fake",
      },
    },
    "shortLabel": "",
  },
  "mProfileName": undefined,
  "parentPath": null,
  "profileName": "",
  "session": null,
  "shortLabel": "",
}
`;
```
The code I used to create the snapshot was in ZoweUSSNode.test.ts:
```
it("Checks Zowe USS Node snapshot", async () => {
    const rootNode = new ZoweUSSNode("root", vscode.TreeItemCollapsibleState.Collapsed, null, session, null);
    rootNode.contextValue = "uss_session";
    const subNode = new ZoweUSSNode("testDir", vscode.TreeItemCollapsibleState.Collapsed, rootNode, null, null);
    const member = new ZoweUSSNode("testFile", vscode.TreeItemCollapsibleState.None, subNode, null, null);
    await expect(subNode).toMatchSnapshot();
});
```

Testing this showed that it passes the unit test even when the expected grouping for a directory's context menu item is removed:

**Before removing:**
- package.json contains the grouping for `zowe.uss.addFavorite`:
    group_present

- Resulting unit test where snapshot is first created:
    group_present_snap

**After removing:**
- package.json does not contain the grouping for `zowe.uss.addFavorite`:
    group_removed

- Resulting unit test using comparison snapshot:
    group_removed_snap

Another teammate and I were looking into other extensions to see if there might be examples where a VS Code extension context menu is tested. For example, Microsoft's [vscode-docker](https://github.com/microsoft/vscode-docker) extension uses context menu groupings in its [package.json](https://github.com/microsoft/vscode-docker/blob/master/package.json#L262), but for all the VS Code extensions we looked at, we were unable to find any examples of context menu testing. This might be because the only input VS Code extension developers have into the context menu is via package.json, rather than function calls or other any other references to context menus in the extension code.

@zFernand0 Do you have any other suggestions, or would it be okay to go ahead and merge in this PR? We could add unit tests in a separate PR if we later find out how VS Code extension context menus are tested.

zFernand0:
I find value in the snapshot tests you added.
It will identify if there was any regression in how the ZoweUSSNode is formatted.
Yeah, we can merge this PR, no problem. :smile:

Hmm, you mentioned that `package.json` is the only place where the developer has any input into what VSCode does with the Context menus, we could check that the `package.json#contributes.menus`  did not change. For example:
```javascript
it("should not change the existing context menus", async () => {
    const packageJsonContent = require("../package.json");
    expect(packageJsonContent.contributes.menus).toMatchSnapshot();
});
```
The only downside with this approach is that every desired update to the context menus, you have to be conscious that the tests need updated. So it's up to you if you think it should be added or not 🎉

lauren-li:
Thank you @zFernand0 for your advice! I took your suggestion of snapshotting the package.json file, and tweaked it to just focus on the menus in `"view/item/context"`, as that was where all of my changes to the context menus groupings were made.

I actually had not pushed my changes for the ZoweUSSNode snapshot test, as it didn't achieve my goal of testing the context menus. However, I have since added them in as you mentioned finding value in them.

Please let me know if you have any additional feedback. Thanks again!

zFernand0:
> I actually had not pushed my changes for the ZoweUSSNode snapshot test, as it didn't achieve my goal of testing the context menus. However, I have since added them in as you mentioned finding value in them.

Awesome! :smile:
Thanks for addressing things so promptly.

Issue #49, opened 2019-04-18
Provide a refresh option for USS files


Colin-Stone:
if you view a file on USS, then update it on z/OS (eg. its a log file and is constantly being updated by a server), then close and view the file again in VSC it shows a cached old copy.

Refresh could be performed by clicking on the item again or a dropdown menu item.

Customer contact: Mark Cocker IBM

PR #129, opened 2019-06-25
Add USS File Refresh and USS Safe Save

adambattenburg:
Implements current MVS features in USS. Addresses issue #49.

### DIFF
--- a/__tests__/extension.test.ts
+++ b/__tests__/extension.test.ts
@@ -389,7 +389,7 @@ describe("Extension Unit Tests", () => {
                     getChildren: mockGetUSSChildren,
                 }
         });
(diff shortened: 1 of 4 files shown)


adambattenburg:
Ah sorry about that Colin! It adds the hover options available in MVS, to USS. This will only work for files. For directories, users will continue to use the refresh button. 


image



I had some questions about the current UX (I feel `saveSafe` should be default for instance), but thought that was something better addressed in a future story.

jelaplan:
Adam
Is saveSafe default now for MVS? My recollection of saveSafe is that when saving a file / data-set to the Mainframe, it checks the date and won't over right a newer file. I agree that that seems like a wise default.

PR #131, opened 2019-06-26
USS add root children

### LINKED ISSUE #119: USS add ability to create folders and files on the root node
When user adds a filter to view files in USS, they can only create files and folders on the child nodes, not on the very top node.

kristinochka:
image

Resolves #119


kristinochka:
Resolves https://github.com/zowe/vscode-extension-for-zowe/issues/119

kristinochka:
@Colin-Stone I wasn't sure if you wanted to add the path string next to the profile. Sounds like you do. That should b a simple change.

kristinochka:
Added changes to show path next to profile
image

PR #135, opened 2019-07-01
Add ability to upload in the favorites section

kristinochka:
Signed-off-by: KRISTINA MAYO <ktopchi@us.ibm.com>

image
image

### DIFF
--- a/src/mvs/mvsNodeActions.ts
+++ b/src/mvs/mvsNodeActions.ts
@@ -33,9 +33,19 @@ export async function uploadDialog(node: ZoweNode, datasetProvider: DatasetTree)
     datasetProvider.refresh();
 }
 
+export function getDatasetLabel(node: ZoweNode) {
+    if (node.mParent && node.mParent.contextValue === "favorite") {
+        const profileEnd = "]: ";
+        const profileIndex = node.label.indexOf(profileEnd);
+        return node.label.substr(profileIndex + profileEnd.length, node.label.length);
+    }
+    return node.label;
+}
+
 export async function uploadFile(node: ZoweNode, doc: vscode.TextDocument) {
     try {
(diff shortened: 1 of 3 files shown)


src/mvs/mvsNodeActions.ts
```diff
@@ -33,9 +33,19 @@ export async function uploadDialog(node: ZoweNode, datasetProvider: DatasetTree)
     datasetProvider.refresh();
 }
 
+export function getDatasetLabel(node: ZoweNode) {
+    if (node.mParent && node.mParent.contextValue === "favorite") {
+        const profileEnd = "]: ";
```
kristinochka: We do need to refactor this code at some point to store dataset name and profile name as separate attributes on the node object instead of concatenating the strings.

src/mvs/mvsNodeActions.ts
```diff
@@ -33,9 +33,19 @@ export async function uploadDialog(node: ZoweNode, datasetProvider: DatasetTree)
...
+        const profileIndex = node.label.indexOf(profileEnd);
+        return node.label.substr(profileIndex + profileEnd.length, node.label.length);
+    }
+    return node.label;
+}
```
kristinochka: We do need to refactor this code at some point to store dataset name and profile name as separate attributes on the node object instead of concatenating the strings.

zFernand0: That's a good observation.
Is there an issue open about it?

kristinochka: No, but I can create one.

kristinochka: https://github.com/zowe/vscode-extension-for-zowe/issues/136

PR #142, opened 2019-07-03
Set jes-jcl language type on spool files for syntax highlighting

kristinochka:
Signed-off-by: KRISTINA MAYO <ktopchi@us.ibm.com>


crshnburn:
I'm not sure you can infer the language type of every spool file, especially as they might be logs from a process. If a user wants to use some syntax highlighting they can use the `⌘+K M` key combination to set the language on the current editor.

(Other key combinations available on other platforms)

kristinochka:
OK, I agree it shouldn't be applied to everything. Could we look at the name of the spool file (and other attributes) and only set jes-jcl on appropriate files? Here is a sample output: 
image
I believe the middle file would be appropriate to highlight with jcl-like syntax.

kristinochka:
The language won't be JCL, but a custom one.

PR #143, opened 2019-07-04
Correctly determine if file is binary for saving

### LINKED ISSUE #123: USS File Save Error: Cannot read property 'binary' of undefined
When I toggle file format to binary VSC does not allow me to save to USS directory. It is necessary for zOSMF workflow files to be treated as binary during transfer so the zOSMF accept them. Maybe Toggle binary is meant for a different usage but so far it looks to me like a bug.
Capture_toggleBinary
VSC Save error - Cannot read property binary of undefined

crshnburn:
Resolves #123

The saveUss function will now correctly determine if a file is binary for saving.

### DIFF
--- a/__tests__/extension.test.ts
+++ b/__tests__/extension.test.ts
@@ -1607,7 +1607,7 @@ describe("Extension Unit Tests", () => {
 
     it("Testing that saveUSSFile is executed successfully", async () => {
         const testDoc: vscode.TextDocument = {
-            fileName: path.join(extension.USS_DIR, "testFile"),
+            fileName: path.join(extension.USS_DIR, ussNode.mLabel, "testFile"),
             uri: null,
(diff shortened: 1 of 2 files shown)


Colin-Stone:
So we are testing to see if the user has opened as binary ?. Seems a sound plan. 

Would you be able to add additional tests in this area? In the Files changed there are two "TODO missed testing" that are affected by your change so would be good to get these. Those references would be for unit tests but if an integration test could be added too that would help us towards compliance. Thanks Andy

crshnburn:
@Colin-Stone I've enhanced one of the existing Unit Tests to cover the if statement, the actual Zowe CLI call is too wrapped up in mock calls to call via Unit Tests but the integration tests should hit this with the current ones.

Issue #105, opened 2019-06-07
Store local version of dataset with a suffix if appropriate to enable syntax highlighting


Colin-Stone:
Since the server files are downloaded to a local folder the path for which contains additional information about the download, the filename can be further modified to include a suffix that represents a file type. In turn this will integrate with an appropriate syntax highlighter.

The following are examples of how a potential suffix can be identified 
```
/**
 * Provides a label with suffix for the ZoweNode which activates syntax support
 *
 * @param {string} label - the name of the Data Set
 */
function appendSuffix(label: string) {
    let suffix = "";
    if (label.indexOf("CBL.") >-1 || label.indexOf(".COBOL") >-1) {
        suffix = ".cbl";
    } else if (label.indexOf("ASM") >-1 ) {
        suffix = ".asm";
    } else if (label.indexOf("JCL") >-1 ) {
        suffix = ".jcl";
    }
    return label+suffix;
}
```

At the same time remove the square brackets for https://github.com/zowe/vscode-extension-for-zowe/issues/26

PR #144, opened 2019-07-05
Dataset suffix

Colin-Stone:
https://github.com/zowe/vscode-extension-for-zowe/issues/105

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -955,28 +955,33 @@ export function getUSSProfile(node: ZoweUSSNode) {
 }
 
 function appendSuffix(label: string): string {
-    if (label.indexOf(".ASM") > -1 ) {
-        return label.concat(".asm");
+    const limit= 5;
+    const bracket = label.indexOf("(");
+    const split = (bracket > -1) ? label.substr(0, bracket).split(".", limit) : label.split(".", limit);
+    for (let i = split.length - 1 ; i > -1; i--) {
+        if (split[i].indexOf(".ASM") > -1 ) {
+            return label.concat(".asm");
+        }
+        if (split[i].indexOf("JCL") > -1 || split[i].indexOf("CNTL") > -1) {
+            return label.concat(".jcl");
+        }
+        if (split[i].indexOf("COBOL") > -1 || split[i].indexOf("CBL") > -1) {
+            return label.concat(".cbl");
+        }
+        if (split[i].indexOf("PLI") > -1 || split[i].indexOf("PL1") > -1 || split[i].indexOf("PLX") > -1) {
+            return label.concat(".pl1");
+        }
+        if (split[i].indexOf("SCRIPTS") > -1 || split[i].indexOf("SHELL") > -1) {
+            return label.concat(".sh");
+        }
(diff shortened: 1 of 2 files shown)


src/extension.ts
```diff
@@ -954,6 +954,31 @@ export function getUSSProfile(node: ZoweUSSNode) {
     return profile;
 }
 
+function appendSuffix(label: string): string {
```
zFernand0: I can see this being a problem if someone has a dataset named `JASMINE.APPLIED.COBOL.SCRIPTS`
Maybe I'm trying to determine if there is a way in which we can limit this to one suffix only. Maybe go from LLQ to HLQ (i.e. backwards) trying to match an expected suffix.

Colin-Stone: I agree, it's never going to be an exact science and there isn't a way to override in this situation. I did toy with the idea of using the LLQ itself as the suffix and that way it would be down to the individual using VSCode how they would map that suffix to a syntax. For example mapping JASMINE.APPLIED.COBOL(member).**COBOL** the same as if it read .cbl

Maybe the compromise is to put the LLQ as the suffix as well (in lower case) but recognize certain ones like COBOL and replace with the suffix of .CBL like this
JASMINE.APPLIED.COBOL(member).**cbl**

Colin-Stone: Going back to your example. If I handle a default approach as 
`JASMINE.APPLIED.COBOL.AMISC(member)`
gets saved as 
`JASMINE.APPLIED.COBOL.AMISC(member).amisc`

your example above becomes
`JASMINE.APPLIED.COBOL.SCRIPTS(member)`
gets saved as 
`JASMINE.APPLIED.COBOL.SCRIPTS(member).sh`

and for shorter
`JASMINE.APPLIED.COBOL(member)`
gets saved as 
`JASMINE.APPLIED.COBOL(member).cbl`

That way as all items have a suffix users would always be able to associate with a syntax highlighter

Colin-Stone: The only question is do we still traverse backwards to find an overriding suffix in the first place

Colin-Stone: Okay implemented as above. Working backwards as priority.

src/extension.ts
```diff
@@ -955,6 +955,44 @@ export function getUSSProfile(node: ZoweUSSNode) {
...
+ *  2b. If a conventional file (no members) default is to return label
+ */
+function appendSuffix(label: string): string {
+    const limit= 5;
+    const bracket = label.indexOf("(");
+    const split = (bracket > -1) ? label.substr(0, bracket).split(".", limit) : label.split(".", limit);
+
+    const doUntil = 0 ; // (bracket > -1) ? 0 : split.length - 2;
+    for (let i = split.length - 1 ; i > doUntil; i--) {
+        if (split[i] === "ASM" || split[i].indexOf("ASSEMBL") > -1 ) {
```
zFernand0: I see you also made everything more deterministic.
Now `jasmine.applied.rexx.scripts` will show as `file-name.rexx`

Colin-Stone:
Hi Fernando. I've paused on merging this one. I've since discovered some optional parameters in VSCode.openTextDocument which allow you to set the language (.e.g jcl) rather than relying on a suffix. Since the suffix was a hack really I would like to try this language parameter and see if it gives me the same result. 
Sorry about messing you about but I think this will be a lot tidier

Issue #90, opened 2019-05-24
Add support for localization


kristinochka:
As an extension user I would like to see the plugin in my local language. 

My team is interested in exploring this item.

kristinochka:
My team is looking at how to localize the plugin.

PR #147, opened 2019-07-09
Externalized Strings and added relevant documentation to docs/README.md

evannwu20:
Response to Zowe issue: "Add support for localization #90"

Using vscode-nls library to implement localization. Strings from the package.json and src/ files have been externalized. Translated strings will be included in the 'i18n' directory through a series of pull requests. The 'i18n' directory currently contains no translated strings. docs/README.md contains instructions on how to add/edit new strings as well as how to donate translations.

Signed-off-by: Evann Wu <Evann.Wu@ibm.com>


evannwu20:
New PR to clean up the commits.

evannwu20:
> 1. The package gulp is included as both a dev and production dependency.
> 2. I don't think I fully appreciate what you intend with this or how it all works but why is this simply a sample and not a default implementation for say en_US ?.
> 3. Are all files required, for example, those included in the **mocks** directory?

@Colin-Stone 
1. Removed gulp from production dependency.
2. A default directory like en_US is not necessary since the gulpfile creates default _filename_.nls.json files which contain the strings passed into the localize function. For instance, if I externalized this string, localize("example.key", "Example String"), then "Example String" would be in my _filename_.nls.json file. Whenever a locale which has not been translated is used, the strings from the _filename_.nls.json will be displayed. For this reason, to ensure we don't have two versions of English strings, one from an en_US directory and one from the default _filename_.nls.json files, I chose to not include an en_US directory. The sample is there purely for the purpose of reference for how a directory of an actual locale should look.
3. Only the files which contain strings that you want to translate are required to be in the i18n directory (under their specific locale). If you don't include a file that has externalized strings in it, then the default strings from _filename_.nls.json will be displayed.

Issue #97, opened 2019-05-31
Setting for temporary "Resources" folder location


adambattenburg:
Currently when remote files are being edited, they are downloaded to a temporary `resources` folder within the extension root. These files are inaccessible outside of the extension, and limit what can be built on top of the `zowe-vscode` extension.

I’d like to propose this folder location be configurable in user settings, so that other extensions can make use of it. Alternatively, temporary `resources` could be downloaded to a common location provided by the CLI directly. 

If people agree defining the temporary folder location is a good approach, I can submit a PR to add an option in settings.

adambattenburg:
As an update, I'm now starting work on this, and will have a PR soon. @Colin-Stone you can assign me the issue if you like.

Colin-Stone:
If you swap this folder does the old one get cleared?.

adambattenburg:
That's how I have it setup for now. Moves the `temp` folder from one location to another.

Still need to make progress getting testings working, but afterward I can submit the PR.

Colin-Stone:
@adambattenburg For the sake of the next release is this issue closed?.

adambattenburg:
I'd like to fix the the false errors that appear with multiple instances that I [mention here](https://github.com/zowe/vscode-extension-for-zowe/issues/140#issuecomment-508158685), as that is a bit of a regression. I'll try to get a PR in for that very soon!

PR #151, opened 2019-07-11
Error message on multiple instances patch 

adambattenburg:
Previously when two editor windows were open, moving the temp folder via the `Zowe-Temp-Folder-Location` preference would throw an error. This was because one instance would move the temp folder, and the second instance would see that the folder does not exist. 

The [block here](https://github.com/zowe/vscode-extension-for-zowe/compare/master...adambattenburg:temp-errors?expand=1#diff-45327f86d4438556066de133327f4ca2R307), removes the check for subsequent instances, and reverts to the behavior before the temp preference was added. 

This closes [issue 97](https://github.com/zowe/vscode-extension-for-zowe/issues/97).


adambattenburg:
Ah sorry always forget to signoff. Just a moment.

PR #156, opened 2019-07-16
Test restructure

Colin-Stone:
Restructured testing specifically unit testing to make it more clear. Added more testing simply to get the coverage numbers up.

Will be more unit tests to come but this gets the restructuring in place


Colin-Stone:
@zFernand0 Do you happen to know if there anything we can add to a build that would reject a change if it reduced the average code coverage value?.

zFernand0:
> @zFernand0 Do you happen to know if there anything we can add to a build that would reject a change if it reduced the average code coverage value?.

Yes, In fact there are a few ways that this can be achieved.
The "old" way was to add the coverage targets to the Jenkinsfile like:

https://github.com/zowe/zowe-cli/blob/0d063d9b948d6a77e9b09489519b1e0045d0e033/Jenkinsfile#L86-L118

Now, we are working towards adding badges thanks to @jellypuno :smile:
And we will be adding status checks, so that PRs fail if coverage goes below the target.

PR #167, opened 2019-07-24
aligns with @zowe

Colin-Stone:
Placeholder. Do we want to switch VSCode to work off @zowe and @latest rather than lts-incremental?.


Colin-Stone:
Any ideas why the build is barfing regarding the download of ```@zowe/cli```

Colin-Stone:
@MikeBauerCA I think we agreed yesterday to proceed with the switch to VSCode being based off @latest/@zowe rather than @brightside.

zFernand0:
Before merging this PR, we should make sure profiles will continue to work even when using a plugin to securely store their credentials.

Colin-Stone:
Closing this as getting stale but leaving branch for now

PR #175, opened 2019-07-25
Add command to issue tso command

crshnburn:
Allow the user to submit a TSO command and view the response from VS Code.

'Zowe: Issue TSO Command ...' is available from the command palette. This then prompts the user to select the profile to use and then the command to issue. As a convenience function a leading / is removed from the command if present.

The results of the command are sent to a new OutputChannel (along with the command) making it easier for the user to review the output.

Signed-off-by: Andrew Smithson <smithson@uk.ibm.com>

### DIFF
--- a/__tests__/__unit__/extension.unit.test.ts
+++ b/__tests__/__unit__/extension.unit.test.ts
@@ -87,6 +87,16 @@ describe("Extension Unit Tests", () => {
         "subsystem": ""
     };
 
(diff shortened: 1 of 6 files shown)


Colin-Stone:
@crshnburn Thanks for this Andy. I wonder if we should add a menu option off one of the profiles session objects that would go directly to the command input dialog

PR #180, opened 2019-07-30
Add icons to files and folders

Colin-Stone:
Adds all of the icons in
Also includes an update to the suffixes tests to ensure they work in windows.


zFernand0:
Thanks for updating the changelog👍 
There are a few unit tests that fail on Windows because of how paths are handled (`\\` vs `/`)
```
- Tests that temp folder handles default preference
- Tests that temp folder is moved successfully
- Tests that temp folder does not update on duplicate preference
```

Icons look great though 😸

Colin-Stone:
> Thanks for updating the changelog👍
> There are a few unit tests that fail on Windows because of how paths are handled (`\\` vs `/`)
> 
> ```
> - Tests that temp folder handles default preference
> - Tests that temp folder is moved successfully
> - Tests that temp folder does not update on duplicate preference
> ```
> 
> Icons look great though 😸

Thanks for finding those. I had been intending to fix those up suspecting they may have been around for a while. I have also found a slight bug when refreshing so need to spend more time on this.

zFernand0:
Question:
Should the folder icon appear as opened on the Jobs node?
image

Colin-Stone:
> Question:
> Should the folder icon appear as opened on the Jobs node?
> image

I fancied putting a gear or cog icon here but thought it better to stay consistent. I'm more than happy to find something else if you think it more appropriate

MikeBauerCA:
I think @zFernand0 is just pointing out that the opened folder under Jobs is not appearing as opened.
image

Colin-Stone:
There is still an edge case here. The icons work fine when collapsing and expanding except when performing a refresh. 
Starting from here..
Screenshot 2019-08-05 at 11 36 15
If you refresh you get the node collapsed but arrow indicating expanded..
Screenshot 2019-08-05 at 11 37 54
Alternatively I can show expanded but the icon is incorrect
Screenshot 2019-08-05 at 11 36 28
I have been unable to address this so would prefer to create a defect to resolve at a later stage

Colin-Stone:
Hi Fernando, thanks for reviewing but I am no longer happy with this. The edge case I am concerned with relate to what happens between favorites and the other sessions and has been broken by my changes. What I would prefer to do is to cancel this PR and completely start again with plain icons. No dynamic open icons as it's the painting of these that are the root of the issue. There appears to be a VSCode bug which relates to having to change the node label to make this happen and even though I have created a solution (hack) which manages for that situation I am rewriting too much core code. 
My intention is to resubmit a new PR with a very simple single icon approach for now and have the dynamic solution as a background task. Trust you are okay with this approach.

Colin-Stone:
Dropping this pull request for now

Issue #170, opened 2019-07-24
Opening job spool files in untitled windows results in annoying save dialogs


CoderCoco:
When I am looking at a Job, I quite frequently open every spool file at the same time when I have a JCLERROR. This results in having multiple windows opened in an unsaved state.

The problem is that when I am done with the job, I have to close each window and respond to vscode that I do not wish to save the file contents.

A solution might be to open spool files as read only and require a specific action in the jobs display to save it to the local PC.

crshnburn:
Looking through the VSCode API reference we should be using a TextDocumentContentProvider to show the Spool content rather than treating them as a new text file each time.

I've prototyped something which does this and the spool content now opens like this
<img width="575" alt="Screenshot 2019-07-28 at 23 05 25" src="https://user-images.githubusercontent.com/860280/62013592-9ed7ee80-b18c-11e9-9712-6243e82f9c36.png">

I'll tidy the code up and raise a PR shortly

PR #184, opened 2019-08-01
Add a content provider for job spool

crshnburn:
Change how the spool content is displayed from a new text document to a custom provider which shows the content and sets the title based on the job and spool.

Signed-off-by: Andrew Smithson <smithson@uk.ibm.com>


crshnburn:
Resolves #170

Colin-Stone:
For some unknown reason the DCO server was off. It's now been restarted but the only way I can think of getting it rerun for this PR is to commit a no-op update. e.g make a tiny change to a file

PR #204, opened 2019-08-20
Fixed icon mixup

### LINKED ISSUE #203: Icon inconsistency with refresh
Two issues with USS/Dataset trees where the icon expanded state is inconsistent with the twistie state. 
1. When collapsing the session object and expanding again, the tree should be reinstated exactly how it was. However lower down expanded icons have been reset to expanded.
2. When using the refresh buttons folders/directories should all be collapsed. However the icon may not be reflecting that state

Colin-Stone:
Cleans up the icon behavior for USS and Datasets. Fixes #203


Colin-Stone:
For Jobs we have the same effect for collapsing the root and refresh all. They have another function which is refresh individual which gives the same effect as collapsing and expanding the root but does so by resubmitting jobs.
Also found that the link to submitted job was failing so fixed that

Issue #188, opened 2019-08-07
Incorrect profile name in favorites crashing Zowe extension


gplambert:
When navigating to the Zowe Extension in VS Code, the Favorites and zosmf-profile do not show up on any of the tabs, data sets, uss, or jobs. Instead, you get the following message: There is no data provider registered that can provide view data.

Steps to Reproduce

    Create a zosmf-profile
    List datasets and/or USS files
    Select a 'favorite', I had one for both Data sets and USS
    Delete the zosmf-profile
    Create new zosmf-profile with a new name
    Open Zowe view
    Should see the error message listed above

Looking at the Persistent Favorites in the settings.json, I see the zosmf-profile name is still pointing to the old profile.

Expected behavior

I should still be able to see the new profile and use it regardless of the profile name tagged in the favorites. Maybe some error message indicating the profile name used in the favorites is no longer found so that I know to change it. Regardless, the current message give no help as it is a generic VS Code message.

Screenshots
zoweCLI profile was deleted.
![62644219-9a8f9c00-b917-11e9-81a5-8a76b9c80431](https://user-images.githubusercontent.com/43450888/62647891-9bc4c700-b91f-11e9-89f9-2e23ff0734e2.png)

Logs
If applicable, add server logs collected at the time of your problem.

Details

    Version and build number: [e.g. 0.4.4-SNAPSHOT build # 155]
    Test environment: [e.g. zD&T, zPDT, MONOPLEX, SYSPLEX, z/OS Rev 2.2, TSS, RACF, ACF2, or any other germane detail]

Web Browser Details (if the bug relates to Zowe Desktop usage):

    OS: [e.g. macOS, Windows]
    Browser [e.g. Chrome, Safari]
    Version [e.g. 71.0.3578.98]

REST API client (in case of REST API issue):

    Technology: [e.g. Spring Boot, Node.js, Postman]
    OS: [e.g. Windows 10]

Shell Environment Details (if the bug relates to CLI):

    Technology: [e.g. zsh, cygwin, cmd, powershell, bash]
    OS: [e.g. MacOS 10.13, Windows 10, Ubuntu 16.04 LTS]

Additional context

lauren-li:
Planned fix results: When the user has a favorite with an invalid profile name...
- Tree data for Data Sets, USS, and Jobs will still load.
- Favorites with valid profile names will still load in the tree.
- Favorites with invalid profile names will *not* load in the tree.
[...]

PR #206, opened 2019-08-22
Fix incorrect profile name in favorites crashing Zowe extension

lauren-li:
Addresses issue #188 (Incorrect profile name in favorites crashing Zowe extension)

When the user has a favorite with an invalid profile name...
	•	Tree data for Data Sets, USS, and Jobs will still load.
	•	Favorites with valid profile names will still load in the tree.
	•	Favorites with invalid profile names will not load in the tree.
	•	The user will receive an error message for either USS or Data Sets depending on which setting(s) contains the invalid favorites.

Error messages for MVS and USS:
Screen Shot 2019-08-19 at 10 56 13 AM


zFernand0:
> Why did you change the `@brightside` references to `@zowe`?. In the ussNodeActions unit test. It doesn't affect anything but the brightside reference always reminds me that we are using the old brightside stream (lts-incremental) Rather than the zowe stream (latest)

I don't see a problem with changing how the package is being addressed.
However, the `from` has to remain as `@brightside`.
So `import * as zowe from "@brightside/core";` shouldn't be an issue. :smile:

lauren-li:
@Colin-Stone @zFernand0 Thank you for the feedback! I made the change from `import * as brightside` to `import * as zowe` when I was trying to get the unit test for `initializeUSSFavorites` in ussNodeActions to work, but I don't think it affects anything. I can change it back if it's preferred that I use `import * as brightside`. Please let me know if you'd like me to go ahead and change it back, and I can make an additional commit to address that.

However, I noticed in most of the code in our src and integration tests folders, we use `import * as zowe`. But in our unit tests, we use `import * as brightside` (all from `@brightside/core`). Is there a reason for the different naming in different places?

PR #215, opened 2019-08-28
Progress Bar animation for opening large files from the mainframe

rodney-wilson:
When files are pulled and opened from the mainframe, sometimes there may be delay in opening files (particularly for large files). Therefore, a progress bar animation was implemented for opening files from MVS & USS. 

Below is an example of the progress bar animation when opening a large `COBOL` file from MVS:

largeds

Below is an example of the progress bar animation when opening a large `COBOL` file from USS:

animation3

**Note:** The animation is implemented to show only on the initial opening of the file.


Colin-Stone:
As per my slack response to your question: When you commit you have to sign your changes off or you get this DCO error. I always do `git commit -s -m "My changes"` but there may be other ways to do this. The problem is that you now have x’s against DCO which wont go away. Yes you can override them or at least an administrator can but that won’t be accepted for open source Zowe code. The simplest thing to do is to cancel the PR and do it again with the correct sign-offs. Alternativel you can compress all changes but I have never performed the latter so you would need to look into it. I would just cancel the PR and create a new one.

rodney-wilson:
Thanks @Colin-Stone. I am closing this PR due to the DCO error and I will open a new one.

PR #216, opened 2019-08-29
Progress Bar animation for opening large files from the mainframe

rodney-wilson:
When files are pulled and opened from the mainframe, sometimes there may be delay in opening files (particularly for large files). Therefore, a progress bar animation was implemented for opening files from MVS & USS. 

Below is an example of the progress bar animation when opening a large `COBOL` file from MVS:

largeds

Below is an example of the progress bar animation when opening a large `COBOL` file from USS:

animation3

**Note:** The animation is implemented to show only on the initial opening of the file.


Colin-Stone:
Can you resolve the conflicts please?. You can use the resolve conflicts link on the GitHub page top achieve this

rodney-wilson:
The conflicts should now be resolved.

PR #225, opened 2019-09-03
Revised the Icons

Colin-Stone:

Changed the root folders to have a subtle color difference so favorites and profiles are immediately distinguishable. Was part of a later change but that was getting massive so started breaking them down.

Also fixes an issue that the favorite search was broken.


Colin-Stone:
Thanks for reviewing Fernando. 

The brown shade is darker than the blue but not massively. I switched to greyscale on my machine to have a look and side by side you could see the difference. I took a screenshot but that came out in color so pointless sharing.
My motivation was to make it more obvious if you have multiple profiles without being too colorful. The colors used are web safe colors but we can lighten one and darken the other if preferred.

zFernand0:
I think it's Ok as it is.
As long as it's noticeable, it should be fine 👍

PR #242, opened 2019-09-17
Add Edit to context menu of tree browser for MVS and USS

### LINKED ISSUE #245: Double-click to open file in separate tab action
Currently it's very hard to open and view files in separate tabs when working with VSCode Zowe extension.
In VSCode file explorer it is simply done by double clicking on a file.
It would be nice to have same functionality in the Zowe extension.

rodney-wilson:
Added functionality so that files can be opened in edit mode, instead of preview mode, by using right-click > Edit operation in the context menu:

### MVS
Screen Shot 2019-09-17 at 5 55 28 PM

### USS
Screen Shot 2019-09-17 at 5 54 19 PM

Fix #245


MikeBauerCA:
I would recommend implementing this behavior upon double-clicking a data set or USS file in our extension instead of upon right-click > Edit.

This would be consistent with the behavior of the VS Code built-in Explorer. Single clicking a file in the Explorer will open the file in Preview mode. The following will open the file in edit mode:

- Make a change to a file.
- Double-click a file's header.
- Double-click a file in the Explorer.
- Open a file that is not part of the current folder.

rodney-wilson:
Thank you for your recommendation @MikeBauerCA. Double-click operation was our initial implementation idea but it poses some challenges:

The click events are not accessible because of VS Code's restricted access to the DOM. Additionally, no [VS Code API's](https://code.visualstudio.com/api/references/vscode-api) were found that are related to changing the click behavior for opening files in an extension.

An issue (https://github.com/Microsoft/vscode/issues/51536) was also raised about handling events in a tree which linked to another issue (https://github.com/microsoft/vscode/issues/39601#issuecomment-398751109) where Microsoft decided to not expose the double-click command. 

If there is a known way to implement the double-click behavior, please advise. Thanks again!

MikeBauerCA:
@rodney-wilson, perhaps it isn't possible. I am wondering how frequently this "right-click > Edit" operation would be used. If you open a file in preview mode in the extension, you can change it to edit mode by making a change to the file or double-clicking the header. If users want to open all files in edit mode, they may have this configured in their preferences with:
`"workbench.editor.enablePreview": false`

I am cautioning against overpopulating the context menus because it may make other operations more difficult to find.

zFernand0:
I'd tend to agree that having too many options may make things a little harder to find. So I'm leaning towards having as few options as possible.

In this scenario, I do see the value of opening things in edit mode. However, as Mike mentioned, users can disable the preview mode and have this behavior on all views. 

I also understand that people may not be aware of this setting and having the edit option proves useful for them.

Maybe a solution is to add a specific setting on the extension for this?

Based on the [`microsoft/vscode#39601 (comment))`](https://github.com/microsoft/vscode/issues/39601#issuecomment-398751109):
`There should be only one command and executing this command on single click or double click is up to application.`
I believe we could cover both cases by keeping the default (Preview mode) and having a Zowe setting to change the default "command"/behavior.

phaumer:
@MikeBauerCA @zFernand0 I don't agree that it is either-or, so that preference option is not a valid alternative. 

Developers need both ways when they navigate their code as some files they want to preserve in their own tab and others they want replaced while they browsing avoiding that they get dozens of tabs in the editor that they do not need. It is not true that they only want it in edit mode when they actually want to edit. It is more about open a file to keep open (e.g. a COBOL Copybook or a program that is similar to the one I am build to use for read-only references only) vs opening a file to have a quick glance, but then have it can go away again. I am sure you experience this as well when you use VS Code to write Typescript code.

So while the user is inside the explorer they want to explicitly decide to open in preview or edit mode without the need to actually type into the editor. There is only one hidden way to work around the limitation at the moment, which is double-clicking the tab itself. The problem with that it is not discoverable for inexperienced users and I have to move the cursor to the right to double-click, which is really annoying; I am in the process of opening all the files I need for my current task and want to it quickly inside the explorer. 

We really wanted the double-click to open the files in the same way as in the VS Code file explorer, but unfortunately MS does not give us an API for that (see Rodney's links). So to have a more discoverable way we had the choices of context menu and icon. We opted for both as the icon are quicker to reach with the mouse vs the context menu is better for accessibility and discoverability.

MikeBauerCA:
@phaumer I agree with your reasoning. Upon technical review, let's accept this PR and include Edit in the context menu. We can revisit if VS Code ever offers an API to respond to double-click events separately from single-click events.

You mentioned "we opted for both as the icon are quicker to reach with the mouse vs the context menu is better for accessibility and discoverability." I have not seen the icon yet. Could you share a screenshot? I think this will be a good solution as it avoids the extra mouse click and having to read through the context menu options.

rodney-wilson:
@MikeBauerCA Please see the screenshot below for the icon @phaumer was referring to:

Screen Shot 2019-09-19 at 10 43 10 AM

PR #247, opened 2019-09-23
Total recall datasets

Colin-Stone:
Adds filters - Still need to add save search favorite for USS and expect some rework of this regarding the flow that filters work


Colin-Stone:
Have completed intended work for USS search favorites. Added new integration tests and reworked literals into constants. 
Although I think we will rework the filters dialogs I think we're good to go

Colin-Stone:
> Awesome features here!
> One change-request and one question:
> 
>     * Change: I believe we need to remove the `Pull from Mainframe` and `Safe Save` buttons on the USS searches saved as favorites.
> 
>     * Question:
>       
>       * Say the users saves job A as favorite. Then the user deletes job A (after a few days or so)...
>         Should the user receive an error when clicking on the saved/favorite job?
>         I don't think's it's wrong, just curious what if that's the expectation. I'm just asking in case we could also remove the favorite (if we knew the deleted job was on their favorites list as well)
>       * What's the purpose of the `Remove Search` option in the USS context menu? 

Thanks Fernando. Sometimes you just don't see these things even when you're in looking straight at them :-) 
You are right the Safe Save and Pull from mainframe need to go. As does the Remove Search on the USS Context menu. 

Regarding the job question if the user deletes a job it should also delete the favorite I forgot that scenario. It doesn't cover the case that the job is purged elsewhere but in that case you would not be able to open it up. I think I will address the Job question as a bug if okay with you and I can test some other edge cases. I will fix the menu options now though so I can get master up to date

zFernand0:
Thanks for fixing those.
Sorry I missed this as well.
The same options need to be removed from the context menu of the saved USS searches. And instead of `Remove Profile` I believe we want a `Remove Search` option.
image

Last comment:
The `Copy Path` option on that context menu copies the profile name to the clipboard as well, instead of just the path.
So you get `[my-profile]: /u/users` instead of just `/u/users`

Colin-Stone:
> Thanks for fixing those.
> Sorry I missed this as well.
> The same options need to be removed from the context menu of the saved USS searches. And instead of `Remove Profile` I believe we want a `Remove Search` option.
> image
> 
> Last comment:
> The `Copy Path` option on that context menu copies the profile name to the clipboard as well, instead of just the path.
> So you get `[my-profile]: /u/users` instead of just `/u/users`

Thanks once again Fernando. I hadn't thought to look at Copy Path. The first one "Remove Favorite" was intentional because I felt it specifically indicated it was a Favorite. However that is inconsistent with the others so I will change to Remove Search. Ideally I would like "Remove Search Favorite" but this looks a bit big in the context menu. 

The Copy path is a regression.

I'm also going to start updating the Readme prior to publishing a new release. I suggest we do that once I have merged everyone else's changes in and we have all had a good play with it.

Colin-Stone:
> log info for searchPrompt in ZosJobsProvider specifies a USSPath. Is that correct?
> 
> ` public async searchPrompt(node: Job) { if (this.log) { this.log.debug(localize("ussFilterPrompt.log.debug.promptUSSPath", "Prompting the user for a USS path")); }`

Although incorrect it's also unnecessary so I have removed it. But whilst I was there I added the appropriate language sample for the provider file which was incorrect. So good call :-)

PR #257, opened 2019-10-03
Add release tags

### LINKED ISSUE #207: Provide Release tags
As the VS Code extension is under heavy development there are breaking changes from time to time requiring that one can go back and manually build an older version that was still working. 

In the VS Code marketplace the Changelog does not tell me when a version was release and the repo also does not tell me. The version number updates in the package.json also do not seem to be a good indicator. 

To make the extension consumable and releasable it is essential to baseline the code for each marketplace release so that one can go back and branch of any earlier release in case of breakages. There are other advantages to baselining as well.

Github provides Release Tags for that: https://github.com/zowe/vscode-extension-for-zowe/releases. See here how to use them: https://help.github.com/en/articles/creating-releases.

zFernand0:
On every new release (major, minor or patch) the pipeline will create a tag and push it to the repository.

Fixes #207

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,7 +2,7 @@
   "name": "vscode-extension-for-zowe",
   "displayName": "%displayName%",
   "description": "%description%",
-  "version": "0.25.0",
+  "version": "0.0.1",
   "publisher": "Zowe",
   "repository": {
     "url": "https://github.com/zowe/vscode-extension-for-zowe"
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -1146,7 +1146,7 @@
     },
     "globals": {
       "ts-jest": {
-        "tsConfigFile": "tsconfig.json",
+        "tsConfig": "tsconfig.json",
```
zFernand0: This was to remove warnings when testing the code :smile:

PR #282, opened 2019-10-25
Fix vulnerabilities

zFernand0:


phaumer:
@Colin-Stone was `"@brightside/imperative"` removed by accident from package.json as that now causes linting errors wherever it is used?

zFernand0:
Hey @phaumer,
The reason why we are able to remove `imperative` is because we are able to pull those packages directly from the `@brightside/core` dependencies.
I may be wrong though. I will double check this in a second. Apologies for the confusion.

phaumer:
For example in `USSTree.ts` I get the "no indirect imports" linting error for `import { IProfileLoaded, Logger } from "@brightside/core";`.

zFernand0:
> @Colin-Stone was `"@brightside/imperative"` removed by accident from package.json as that now causes linting errors wherever it is used?

Please see https://github.com/zowe/vscode-extension-for-zowe/pull/288

Issue #226, opened 2019-09-03
Allow VS Code Extension Settings of Connection Information


MikeBauerCA:
We should consider removing the strict dependency on Zowe CLI Profiles. As part of this effort, we would need to provide a means of entering connection details directly within the VS Code extension.

However, as a heavy Zowe CLI user I would like to continue having the ability to use profiles I defined in the CLI. Therefore, we should not remove the current integration.

- [x] Add Connection Details using (+) Button
~~- [x] Set-up Default JSON Format for Profile Configuration~~
- [x] Add "Create New" Dialog boxes for profile creation
~~- [x] Update JSON file "Create New" once the dialog boxes are successfully filled - @jellypuno~~
- [x] "Create New" dialog boxes should always be an option. Whether you have loaded profiles or not. - @jellypuno 
- [x] New Profiles/ existing VSC Profiles are displayed in the containers - @crawr 
- [x] Update JSOn file for the new error/informational messages - @crawr
~~- [x] If you have existing profiles in the settings.JSON, it should be loaded as well when the add profile is selected. Follow the process with loading of CLI profiles~~
~~- [ ] Edit profiles available in the containers -@jellypuno~~
~~- [ ] Delete profiles available in the containers -@jellypuno~~
- [x] Refresh Connection - @jellypuno 
- [x] We are asking for the URL, hostname and port should be trimmed if we use the new profile to call zowe functions 
~~- [x] Research VSCode Extension guidelines on how to modify settings~~
~~- [x] Add connection details in settings.JSON~~
- [x] Password; Don't store PW > don't ask at creation of connection > prompt at 1st usage > store (in session variables?)
~~- [ ] Implement order of precedence when loading default profiles (Zowe CLI followed by VSCode)~~
- [x] Test coverage

Notes:

VSCode Docs
- https://github.com/microsoft/vscode-extension-samples
- https://code.visualstudio.com/api/references/vscode-api
- https://code.visualstudio.com/api/references/contribution-points

IgorCATech:
@jalel01 Does removing the dependency on CLI profiles mean that having Zowe CLI as a prerequisite is no longer needed?

MikeBauerCA:
@IgorCATech yes, but we will want to document the CLI profile versions we have integration with.

IgorCATech:
[...]

PR #287, opened 2019-10-29
Create new profiles

crawr:
First part of issue #226 :  Enable adding of new Zowe CLI profiles from the extension using the (+) icon


zFernand0:
Hey @jellypuno, 
Could you fix the DCO Signoffs? :smile:

jellypuno:
DCO Sign-off fixed :D  @crawr @zFernand0

crawr:
Thanks for the password obfuscation hint @Colin-Stone. I added it and it works well. 

I also added a small change so that the new profile created immediately appear in the explorer view.

phaumer:
Would it make sense to develop this more with a Model-View-Control approach? As this PR adds an optional view for creating a profile graphically, perhaps this should be in a separate view class as the Profile class I would see as the model for profiles.  Then when we add support for additional profile types such as zftp we can subclass the view or make it more generic to support other variants.

Colin-Stone:
I've been playing around with this functionality and am currently running on a machine which does not have CLI installed except in it's embedded form within the VSCode extension. Running the VSCode extension will create the .zowe folder so the only other item we need to consider is that the VSCode extension initialization needs to load the default profile for zosmf, which of course is set to null. I think if we modify this startup area so that it does not have to load a default we can be decoupled from CLI. The code I added recently for the persistence of profiles means we don't need to rely on the default so all looking really good

Colin-Stone:
Based upon my previous comment and provided it's acceptable @zFernand0 @MikeBauerCA. Can we disable the dependency on a default profile as follows: In extension.js where we add a session USS, Dataset etc.. Remove the profileNamesList length check and maybe initialize like this rather than using shift `    const chosenProfile = await vscode.window.showQuickPick([createNewProfile, ...profileNamesList], quickPickOptions);` 
The other thing is to remove the this.defaultProfile initialilization (refresh in Profiles) or catch the error that is thrown when one isn't present.

I've made these changes locally (albeit I have not considered the path in which we spawn node modules because that obviously means CLI exists already) and VSCode extension is truly running independently of CLI

crawr:
> The other thing is to remove the this.defaultProfile initialilization (refresh in Profiles) or catch the error that is thrown when one isn't present.

Personally, I prefer the second option: keep the loading of default profile when it exists but to continue processing when none exists.

MikeBauerCA:
@Colin-Stone your suggestion makes sense to me, especially since we have enhanced the extension to start back up with the set of profiles you had previously added to your view. For someone running only the VS Code extension, I do not currently see a need for designating a default profile.

crawr:
@MikeBauerCA, ok, this makes sense.

@Colin-Stone, you mentioned you made some changes locally, are you planning on pushing them to this PR?

Colin-Stone:
@crawr My local changes were my experimentation. I can tidy them up a little and push them if you prefer

Colin-Stone:
I have added the necessary code so this works independently of CLI but we need to increase the level of test coverage before commiting to master

Colin-Stone:
Thanks for adding the extra tests but we still have a slightly negative effect on overall coverage. The console output shows Profiles.ts coverage down to 83% from 100% which looks as if we need to get tests to cover the lines where CliProfileManager saves a new profile. One good way to check this is by looking at the test files produced in your results/coverage/lcov-report/src/Profiles.ts.html folder. If you view the file in a browser you will see the red colored indicators where the lines haven't been run. I also notice that the extensions.ts files percentage coverage has dropped so we will need to ensure that the "add...Session" functions all have proper coverage. 

Sometimes it is virtually impossible to get complete coverage which is why we set the target for 90+% coverage. In situations where this can't be achieved we can also consider providing additional integration testing. 

Overall for this repo there are still a number of areas that unit testing could be better so we have the ability to get to the 90% target before GA. I always think if I find I'm commiting code that is impossible to unit test I look to address some other still outstanding areas of coverage that can be applied. Sort of trading off to ensure that overall code coveage increases.

Colin-Stone:
As part of this change we should remove the following line from the readme.txt file as it no longer applies.

`Important! To use the VSC Extension for Zowe, you must install Zowe CLI version 2.0.0 or later.`

IgorCATech:
Hey @jalel01 , you are right, the doc that will be published once we are ready to go GA already has that part of text removed.

Colin-Stone:
That's fine. We will likely have this going out as part of version 0.28 next week, which will be good to get in front of users

MikeBauerCA:
@IgorCATech @jalel01 looking at the branch this PR is coming from, it appears that the README has not yet been updated to remove `Important! To use the VSC Extension for Zowe, you must install Zowe CLI version 2.0.0 or later.` (https://github.com/zowe/vscode-extension-for-zowe/tree/CreateNewProfiles)

Upon removal, we may want to say that the installation/use of the CLI is highly encouraged as part of the greater solution, note that profile information is shared seamlessly across these technologies, and note a few benefits of the CLI (scripts, interactive use for actions not yet available in the extension, etc.)

IgorCATech:
@MikeBauerCA yes, I know about that, and the reason is that other people have created a PR to merge new changes. There was no prior communication. I've brought this up with my team and @Colin-Stone and I think we are on the same page now. Btw, as for the doc, I've created another issue https://github.com/zowe/vscode-extension-for-zowe/issues/297

Colin-Stone:
I'm not sure we need to use abbreviations like newp and it would aid clarity if it was newProfile for example.

Please review the saveProfile method. Once agaion the attributes can be fuly named i.e. saveProfile(profile, name, type), then I think your .save method becomes .save({profile, name, type});

This method doesn't appear to have any unit test coverage. In fact the Profiles.ts file appears to have dropped from 100% coverage to 91% and I think this method is the culprit.

jellypuno:
@IgorCATech please sign off your commit :)

zFernand0:
DO NOT use the manual overwrite 

```
DCO
Commit sign-off was manually approved.
```

IgorCATech:
Fair enough. Should I reveret that commit and re-commit properly?

zFernand0:
> Fair enough. Should I reveret that commit and re-commit properly?

If the final decision is for these changes to live in another branch, then I can go ahead and remove that one commit from the history of the createNewProfiles branch.
However, if you want to continue working on this branch, I'd suggest taking care of the signoff, unless we are ok with loosing the history on this one : )

zFernand0:
Suggestion:
How about following the same pattern as the filter/search pattern, where you start typing and if it doesn't match any of your previous entries, it will take the default action (which in this case will be creating a new connection)?

Colin-Stone:
> Suggestion:
> How about following the same pattern as the filter/search pattern, where you start typing and if it doesn't match any of your previous entries, it will take the default action (which in this case will be creating a new connection)?

This would be a better user experience as it would behave the same as behavor as the rest of the extension. It also means that one of the lines missed on coverage will be rewritten also.

PR #288, opened 2019-10-30
Proposal: White-list imperative on the no-implicit-dependencies rule

zFernand0:
The other alternative is to set the rule to `false`. This PR also:
- Updates the package-lock
- Adds Lint stage to the pipeline

https://github.com/zowe/vscode-extension-for-zowe/pull/282#issuecomment-548135400

It's open for discussion.

Signed-off-by: zFernand0 <fernando.rijocedeno@broadcom.com>


phaumer:
Whitelisting that one package I think is ok. I assume you want to do that to avoid older versions accidentally being used here? Generally, I think the rule should be switched on, though, with `true` for other dependencies.

phaumer:
@zFernand0 I see more error when running `npm run package`:

`@types/vscode ^1.34.0 greater than engines.vscode ^1.26.0. Consider upgrade engines.vscode or use an older @types/vscode version` 

which can be fixed by updating engines to `"^1.34.0"`

Colin-Stone:
> @zFernand0 I see more error when running `npm run package`:
> 
> `@types/vscode ^1.34.0 greater than engines.vscode ^1.26.0. Consider upgrade engines.vscode or use an older @types/vscode version`
> 
> which can be fixed by updating engines to `"^1.34.0"`

Thanks for testing with package, I may have missed that one. I updated as Peter suggested setting engines to `"^1.34.0"` and all looks well.

zFernand0:
The engine difference was exactly what prevented from publishing the extension last time around.
Thanks to both for the reminder. Updating now! : )

zFernand0:
> Whitelisting that one package I think is ok. I assume you want to do that to avoid older versions accidentally being used here? Generally, I think the rule should be switched on, though, with `true` for other dependencies.

I agree with keeping the rule `on` (true) as it could prevent unexpected errors when referencing dependencies of dependencies from other packages.
About the reasons:
- Prevent the use of different (or maybe incompatible) versions of imperative
- Remove duplicate versions of imperative since the CLI already has its own
  - Preventing strange behavior when running tests

zFernand0:
Hey @phaumer, 
Any thoughts or comments? Or is this good to go as it is?

tslint.json
```diff
@@ -37,8 +37,9 @@
       "check-module"
     ],
     "no-implicit-dependencies": [
-      true,
-      "dev"
+      false,
```
phaumer: @zFernand0 did you want to can this to `true` still?

zFernand0: My apologies. Thought I already had it to true.

PR #296, opened 2019-11-12
Make sure the error is thrown and reported

Colin-Stone:
The error needs to be rethrown to be reported.

### DIFF
--- a/__tests__/__unit__/extension.unit.test.ts
+++ b/__tests__/__unit__/extension.unit.test.ts
@@ -975,7 +975,7 @@ describe("Extension Unit Tests", () => {
         delDataset.mockRejectedValueOnce(Error("not found"));
         showQuickPick.mockResolvedValueOnce("Yes");
 
-        await extension.deleteDataset(node, testTree);
+        await expect(extension.deleteDataset(node, testTree)).rejects.toEqual(Error("not found"));
 
         expect(showInformationMessage.mock.calls.length).toBe(1);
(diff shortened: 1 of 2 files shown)


Colin-Stone:
Code coverage is still appearing a little wacky. The junit tests coverage is actually a little better as I have added one new line which is covered by testing.

zFernand0:
If we look at the [history of the `extension.ts`](https://codecov.io/gh/zowe/vscode-extension-for-zowe/branch/master/history/src/extension.ts?before=f65b20bcfc62a590a39c4082ea3a651f7da6833a) file it would seem that codecov is reporting a different number of lines tracked on the file fluctuating between 814 and 1017. (See proof below)

Should we submit an issue to @codecov?

[`073dbff`](https://codecov.io/gh/zowe/vscode-extension-for-zowe/src/073dbffbdf4c4b78f0254d1663c91875524fd0f3/src/extension.ts)
image

image


[`f65b20b`](https://codecov.io/gh/zowe/vscode-extension-for-zowe/src/f65b20bcfc62a590a39c4082ea3a651f7da6833a/src/extension.ts)
image

image

Colin-Stone:
> 
> Should we submit an issue to @codecov?
> 


Thanks for the review. Are we seeing similar codecov issues with cli?. I was going to ask if we need to create an example for them but I think they can see this so yes probably worth raising an issue

zFernand0:
Possibly related issue on codecov-batch: https://github.com/codecov/codecov-bash/issues/83
Similar issue on another project: https://github.com/zephyrproject-rtos/zephyr/pull/13608

Issue #307, opened 2019-11-18
Profile URL parsing rejects some valid cases


MarkAckert:
Current URL parsing rejects the following key cases. These cases should be allowed as valid.

https://`host`:443/path
https://`host`/path

http://`host`:80/path
http://`host` 

Current URL parsing also accepts protocols other than http/https. We should constrain the check to just http/https, as z/OSMF profiles require one of these protocols.

PR #308, opened 2019-11-18
Update URL Parsing

MarkAckert:
Fixes https://github.com/zowe/vscode-extension-for-zowe/issues/307

### DIFF
--- a/src/Profiles.ts
+++ b/src/Profiles.ts
@@ -19,9 +19,12 @@ import * as vscode from "vscode";
 import * as nls from "vscode-nls";
 import * as ProfileLoader from "./ProfileLoader";
 const localize = nls.config({ messageFormat: nls.MessageFormat.file })();
-let url: URL;
-let validHost: string;
-let validPort: number;
+interface IUrlValidator {
+    valid: boolean;
+    host: string;
+    port: number;
+}
+
 let IConnection: {
     name: string;
     host: string;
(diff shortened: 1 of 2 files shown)


MarkAckert:
In addition to URL parsing fixes, I've changed URL placeholder text from `http(s)` to `https`. I confirmed that z/OSMF Profiles for the CLI only support HTTPS, so `http(s)` would be confusing to end-users who expect an `http` URL to work. `validateUrl` will now only accept URLs with the `https` protocol.

jellypuno:
@IgorCATech I think we need to change/add something the documentation for the URL. 
- We are only accepting https://
- If port is not specified, the defaut value will be 443.

src/Profiles.ts
```diff
@@ -104,15 +105,40 @@ export class Profiles { // Processing stops if there are no profiles detected
         return this.allProfiles;
     }
 
-    public validateUrl = (newUrl: string) => {
+    public validateAndParseUrl = (newUrl: string): IUrlValidator => {
+
+        let url: URL;
+        const validProtocols: string[] = ["https"];
+        const DEFAULT_HTTPS_PORT: number = 443;
```
zFernand0: No need to implement this now, but what if someone wanted to have a different default port?
Should this be a setting in our set of vscode configurable items?

src/Profiles.ts
```diff
@@ -104,15 +105,40 @@ export class Profiles { // Processing stops if there are no profiles detected
...
-        validURL = url.hostname;
-        validPort = Number(url.port);
-        return url.port ? true : false;
+
+        // if port is empty, set https defaults
+        if (!url.port.trim()) {
+            validationResult.port = DEFAULT_HTTPS_PORT;
+        }
+        else {
+            validationResult.port = Number(url.port);
```
zFernand0: Will this be a valid url `https://10.142.0.23:65536`?

jellypuno: > Will this be a valid url `https://10.142.0.23:65536`?

I tested this one and it is not considered a valid URL

src/Profiles.ts
zFernand0: Should we allow the default port to be handled by nodejs depending on the protocol?

zFernand0:
@MarkAckert, could you resolve the conflicts?

zFernand0:
:+1:

PR #314, opened 2019-11-19
Rename data sets

CForrest97:


CForrest97:
@Colin-Stone This PR only adds functionality for renaming data sets. I think this needs to be added as an issue because renaming uss files doesn't seem to play nicely with favourites.

Colin-Stone:
> @Colin-Stone This PR only adds functionality for renaming data sets. I think this needs to be added as an issue because renaming uss files doesn't seem to play nicely with favourites.

I disagree. The value showing in the input box specifies both the profile and the file name. All you should be asking of the user is the changed name and not the profile.
image
The other thing I notice with the favorite is that you are not considering the full path. So in the above example the user should be renaming /u/stonecc/whizzo1.txt to for example /u/stonecc/whizzo2.txt. In the input box you should only specify whizzo2.txt and the /u/stonecc/ will be prepended

CForrest97:
I agree that the value in the input box should just be the filename, however renaming USS files is existing behaviour already in the extension. This PR is only about allowing data sets to be renamed, so I think this bug is irrelevant. Should this be a separate issue that is resolved in a different PR?

Colin-Stone:
You're right of course. I wonder when this slipped in.

PR #335, opened 2019-11-26
Optional credentials

jellypuno:


Colin-Stone:
I am seeing intehration test failures (timeout) on Creating a session. Is that just my environment?. One other thing can you generate the default i18n texts please?. I always run npm run vscode:prepublish which runs the relevant script to populate the texts

jellypuno:
> I am seeing intehration test failures (timeout) on Creating a session

@Colin-Stone I run the integration test on master branch and I see this error everywhere

Capture

Colin-Stone:
> > I am seeing intehration test failures (timeout) on Creating a session
> 
> @Colin-Stone I run the integration test on master branch and I see this error everywhere
> 
> Capture

I was only seeing two related to session, but this may mean the tests are failing as they are taking too long to complete. You may need to debug them to find out more

Colin-Stone:
Thanks for making the change. I found out the right way to do it now. Use the path.join i.e. expect(extension.BRIGHTTEMPFOLDER).to.equal(path.join(testingPath, "temp")); 
But your latest update fixes it so no need to change it again. Thank you :-)

PR #349, opened 2019-11-28
Safe save by default

Alexandru-Dumitru:
- created new `node` property called etag + setter + getter
- using the value of etag to check if remote file has changed
- triggering a merge conflict if above condition is true
- removed safe save functionality


Colin-Stone:
There is an issue with one of yoursign-offs. Unfortunately I don't know how that can be addressed. It's gthe first WIP - fixing tests

Alexandru-Dumitru:
> There is an issue with one of yoursign-offs. Unfortunately I don't know how that can be addressed. It's gthe first WIP - fixing tests. Also the build is failing because of this..
> Starting 'internalNlsCompileTask'...
> src/extension.ts(1556,34): error TS2304: Cannot find name 'uploadEtag'.

Hi Colin. I fixed the DCO and the missing variable error.

Colin-Stone:
I'm getting a "Cannot read property "setEtag' of undefined" when saving a dataset member or a USS file.
image

Alexandru-Dumitru:
This is weird. I am looking into it, but I cannot duplicate it. 
As for the tests, I will try to add more :)

Alexandru-Dumitru:
Hi @Colin-Stone 
I tried to address your concerns. More info in slack 😄 .

Thanks to review again

Colin-Stone:
Hi Ivan. Thanks for the updates. 

I have got a lint error which I didn't expect whilst attempting to start the extension. Do you know what this means?.

Type '{ etag: any; returnEtag: boolean; }' is not assignable to type 'IUploadOptions'.
  Object literal may only specify known properties, and 'returnEtag' does not exist in type 'IUploadOptions'.
image

I'm also getting an integration test failure as follows
image

Alexandru-Dumitru:
Hi Colin.

Are you on latest brightside/core version? that option was recently introduced (4 days ago).
https://github.com/zowe/zowe-cli/blob/lts-incremental/packages/zosfiles/src/api/methods/upload/doc/IUploadOptions.ts

As for the integration test, I need to check on Monday. VPN doesn't connect for some reason. I will try later, and if I succeed, I will update you.

Thanks

src/utils.ts
```diff
@@ -142,3 +144,31 @@ export class JobIdFilterDescriptor extends FilterDescriptor {
...
+export async function concatUSSChildNodes(nodes: ZoweUSSNode[]) {
+    let allNodes = new Array<ZoweUSSNode>();
+
+    for (const node of nodes) {
+        allNodes = allNodes.concat(await concatUSSChildNodes(await node.getChildren()));
+        allNodes.push(node);
+    }
+
+    return allNodes;
+}
```
zFernand0: Why not "extend" the `concatChildNodes` function instead of duplicating thins?

Colin-Stone: I thought we went to node.children instead of getChildren?

Colin-Stone:
I'm seeing a new integration error 
image

src/utils.ts
Alexandru-Dumitru: @Colin-Stone I switched to `node.children` 😄 
@zFernand0 Seems that now test coverage delta is 0 😄

Alexandru-Dumitru:
> I'm seeing a new integration error
> image

I kind of ignored this one, because I am getting it on master as well 🙁

src/utils.ts
Colin-Stone: @zFernand0 Re: Why not "extend" the concatChildNodes function instead of duplicating thins? 
I hope to abstract the node class in the next PI so this could be a single method in there.

Colin-Stone: Unit test Coverage is an issue in extension.ts. Currently master is showing  extension.ts         
`|    83.21 |    81.33 |    35.35 |    91.55 |... 1738,1783,1795 |`
your new branch has dropped this to  extension.ts         
`|    82.04 |    79.27 |    35.71 |    89.62 |... 1801,1842,1856 |`

Colin-Stone:
I don't know if this is within the scope of this issue or is a defect but I tried to save a dataset member within the extension whilst I had the member in edit more in ispf resulting in an error
image
Algthouhg the message isn't obvious the editor change from a dirty state to clean which would mean the user would think the update was saved.

Alexandru-Dumitru:
I think this is not in scope, because it's a remnant behavior from before, but definitely we need to address it, have a clearer error message and change back the `dirty` flag on the file when needed.

Colin-Stone:
May be worth focussing some testing according to the coverage results `results/coverage/lcov-report/src/utils.ts.html` looks as if of concatChildNodes and concatUSSChildNodes has no unit testing.

src/extension.ts
```diff
@@ -1731,17 +1675,80 @@ export async function saveFile(doc: vscode.TextDocument, datasetProvider: Datase
...
-            return zowe.Upload.pathToDataSet(documentSession, doc.fileName, label);  // TODO MISSED TESTING
+            return zowe.Upload.pathToDataSet(documentSession, doc.fileName, label, uploadOptions);  // TODO MISSED TESTING
         });
-        if (response.success) {
-            vscode.window.showInformationMessage(response.commandResponse);  // TODO MISSED TESTING
+        if (uploadResponse.success) {
+            vscode.window.showInformationMessage(uploadResponse.commandResponse);  // TODO MISSED TESTING
+            // set local etag with the new etag from the updated file on mainframe
+            node.setEtag(uploadResponse.apiResponse[0].etag);
+        } else if (!uploadResponse.success && uploadResponse.commandResponse.includes(localize("saveFile.error.ZosmfEtagMismatchError", "Rest API failure with HTTP(S) status 412"))) {
```
zFernand0: This could fail if the CLI localized message is different than the extension's
For example if their corresponding french translation didn't match, we won't be able to detect that it failed because of the etag stuff.
No need to take action on it.

src/extension.ts
```diff
@@ -1763,29 +1770,81 @@ export async function saveUSSFile(doc: vscode.TextDocument, ussFileProvider: USS
...
+            node.setEtag(uploadResponse.apiResponse.etag);
+        // this part never runs! zowe.Upload.fileToUSSFile doesn't return success: false, it just throws the error which is caught below!!!!!
         } else {
-            vscode.window.showErrorMessage(response.commandResponse);
+            vscode.window.showErrorMessage(uploadResponse.commandResponse);
         }
     } catch (err) {
-        log.error(localize("saveUSSFile.log.error.save", "Error encountered when saving USS file: ") + JSON.stringify(err));
-        vscode.window.showErrorMessage(err.message);
+        if (err.message.includes(localize("saveFile.error.ZosmfEtagMismatchError", "Rest API failure with HTTP(S) status 412"))) {
```
zFernand0: Same as before. We may want to use some sort of message if or unique code that can be easily verified.
Again, no need to take action today

src/extension.ts
Alexandru-Dumitru: I was thinking the same thing at some point, and was wondering if it would fail in that case. Then somehow it slipped my mind... I guess we just need to compare against a literal string, since I assume z/OSMF only returns English error messages. 
Thanks for pointing it out.

Issue #343, opened 2019-11-27
Document how to include the extension in a Theia environment as part of the README


MikeBauerCA:
Zowe Explorer is functional in a Theia environment. We should document how to include it in that environment. Perhaps this information could be provided in the docs folder and the README could link to it.

phaumer:
Work is done here for now: https://github.com/phaumer/vscode-extension-for-zowe/blob/theia-doc/docs/README-Theia.md

phaumer:
I created this PR https://github.com/zowe/vscode-extension-for-zowe/pull/359. 

@Colin-Stone @MikeBauerCA do you want to review it?

phaumer:
Closing this one has the merge commit is not signed.

phaumer:
Created new PR: https://github.com/zowe/vscode-extension-for-zowe/pull/360

zFernand0:
This could very well be included in the GA 1.0 : )

Colin-Stone:
Completed. Thanks Peter

zFernand0:
Thanks @phaumer for adding the doc.
In order to close the issue, should we link to the file from the main README.md?

phaumer:
Which one? The readme in docs or even the one at the top? Perhaps the owners of these files could do it?

zFernand0:
@IgorCATech, @BrandonJenkins14, @JamesBauman
Is it possible to add a link on the top level readme file that directs people to `https://github.com/zowe/vscode-extension-for-zowe/blob/master/docs/README-Theia.md`

It could be as simple as updating the Tip on the first section of the Readme. : )
![image](https://user-images.githubusercontent.com/37381190/70535655-6cbb9d80-1b2b-11ea-8ee8-4f88d3f211e7.png)

Colin-Stone:
I quite fancied an image of Zowe Explorer in it's Theia outfit that could be clicked on.

PR #360, opened 2019-12-02
Added document explaining how to work in Theia

phaumer:

This PR provides the documentation requested in Issue #343.
This PR only adds one MD file and adds a VS Code workspace setting so that people (like me) do not forget to sign their commits. It does not do any code changes.


zFernand0:
I thought the settings.json file was ignored.
@Colin-Stone, is the intention to remove it from source too? Or should we remove it from the gitignore file?

phaumer:
Oh, did not realize that. I think settings like these are useful for standardization.

Colin-Stone:
> I thought the settings.json file was ignored.
> @Colin-Stone, is the intention to remove it from source too? Or should we remove it from the gitignore file?

Caveat. I'm not an expert on VSCode settings but just to clarify from the perspective of runtime this extension uses the `global` user settings.json file which you will see in the code as `vscode.ConfigurationTarget.Global` 

I think in the past we may have had a lot of commits that changed the `project` level settings.json file based on personal preferences and I believe adding to .gitignore was simply because it was "noise" when reviewing changes. I hadn't really considered removing it completely and the addition of  `"git.alwaysSignOff": true` made me think that it may be useful if we do keep it updated.  
So my proposal is keep it and when someone has a good idea of additional parameters we drop the .gitignore flag and let it through.

Sorry that is an "All animals are equal but some are more equal than others" statement but would that be acceptable?.

PR #365, opened 2019-12-04
add try catch on loadnamedprofile during initialization

jellypuno:
Signed-off-by: jellypuno <jessielaine.punongbayan@broadcom.com>

### DIFF
--- a/__tests__/__unit__/DatasetTree.unit.test.ts
+++ b/__tests__/__unit__/DatasetTree.unit.test.ts
@@ -766,6 +766,70 @@ describe("DatasetTree Unit Tests", () => {
 
     });
 
+    it("tests the dataset filter prompt credentials, favorite route", async () => {
+        showQuickPick.mockReset();
+        showInputBox.mockReset();
+        const sessionwocred = new Session({
+            user: "",
+            password: "",
+            hostname: "fake",
+            port: 443,
+            protocol: "https",
+            type: "basic",
+        });
+        const sessNode = new ZoweNode("sestest", vscode.TreeItemCollapsibleState.Expanded, null, session);
+        sessNode.contextValue = extension.DS_SESSION_CONTEXT + extension.FAV_SUFFIX;
+        const dsNode = new ZoweNode("[testSess2]: node", vscode.TreeItemCollapsibleState.Expanded, sessNode, sessionwocred);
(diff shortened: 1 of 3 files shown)


Colin-Stone:
Can you show an example of what the message will look like please?.

jellypuno:
> Can you show an example of what the message will look like please?. No problem I have it..
> image
> Need a continue in as well...
> 
> image

I have this one:
Capture

Colin-Stone:
> > Can you show an example of what the message will look like please?. No problem I have it..
> > image
> > Need a continue in as well...
> > image
> 
> I have this one:
> Capture

Same mistake I made, those are for the section preceding. The real message is the nice small message. I'm think it shouldn't be an error message but maybe this and the messages you have attached should be warning messages only.

Colin-Stone:
image

src/DatasetTree.ts
```diff
@@ -109,7 +109,15 @@ export class DatasetTree implements vscode.TreeDataProvider<ZoweNode> {
                 }
             } else if (favoriteSearchPattern.test(line)) {
                 const sesName = line.substring(1, line.lastIndexOf("]")).trim();
-                const zosmfProfile = Profiles.getInstance().loadNamedProfile(sesName);
```
zFernand0: I'm ok with this. However, it makes me worry since in the above `if` clause, we enclosed every zowe operation as well as the node creation with the same try-catch in case we failed to create the session for any reason.

__tests__/__unit__/DatasetTree.unit.test.ts
```diff
@@ -766,6 +766,70 @@ describe("DatasetTree Unit Tests", () => {
...
+        Object.defineProperty(spyMe, "datasetFilterPrompt", {
+            value: jest.fn(() => {
+                return {
+                    tempNode: dsNode2,
+                    mSessionNodes: {Session: {ISession: {user: "", password: "", base64EncodedAuth: ""}}}
+                };
+            })
+        });
+
+        testTree.initialize(Logger.getAppLogger());
```
zFernand0: Should we expect the error on the try-catch we just added to be shown in this case?
If it doesn't happen, maybe it's because the loadNamedProfile function is returning null instead of throwing an error?

Issue #377, opened 2019-12-08
Cannot add a second session to tree browsers in Theia


phaumer:
There is special code to deal with QuickPicks in Theia for the filters, but not sessions. Hence, the picker always returns undefined. 

The code for the three tree is entirely the same and could be refactored to avoid redundancies.

phaumer:
You can assign me to this bug. It needs the labels `bug` and `theia`.
I started the work in this branch: https://github.com/phaumer/vscode-extension-for-zowe/tree/fix-add-session-theia

phaumer:
Created this PR for the issue: https://github.com/zowe/vscode-extension-for-zowe/pull/382

PR #382, opened 2019-12-09
Fixed adding sessions in Theia

phaumer:
Fix for #377.

In the solution I reapplied the special handling of pickers for Theia that was done for filters.
Instead of updating this for all three addSession methods I unified all three into one as their only difference was the type of the tree provider. I added common interfaces for that as I believe we need those anyway to clean up more redundant cases like these.


phaumer:
@Colin-Stone would you be a good reviewer for this PR?

phaumer:
The coverage is going down as there was a THEIA specific path added. Not sure how much value a Theia mock would provide here as Theia itself is a moving target.

The settings change I thought was adding stability as otherwise you would transpile with any typescript version that is installed on a developer machine and not the commonly agreed one.

jellypuno:
In my opinion, I think it is still valuable to add test for theia just for coverage. Like this one: https://github.com/zowe/vscode-extension-for-zowe/blob/6ab3ef8551b3e657eca96c3bdd66fde74413c2c2/__tests__/__unit__/DatasetTree.unit.test.ts#L487

phaumer:
But what is the value? The only line that matters and that is different is the `await vscode.window.showQuickPick()` call which would have to get mocked. Then all the business logic that you are testing such as the user aborting depends on that mock in the same way it already depends on the other mock for the VS Code picker. So there won't be anything else tested other than what was mocked and tested based on mocks already. So all you get imho is a false sense of security from your coverage that this code was tested for Theia, but it was not.

zFernand0:
>  So all you get imho is a false sense of security from your coverage that this code was tested for Theia, but it was not.

Thanks for sharing.
I do believe that this could be solved by adding integration tests that properly exercise this implementation. 
We could report coverage for these integration test and combine the junit results for a more accurate  report from codecov.

Colin-Stone:
Regarding unit test, I agree with Fernando's suggestion abstracted Theia specific code so that can be excluded from the coverage reports. My concern is that making any exceptions regarding coverage sets a prescedence making it more difficult to ensure every contributor contributes an appropriate level of quality.

PR #387, opened 2019-12-10
Update incorrect profile message

lauren-li:
This is the last bit of polish for issue #188 (Incorrect profile name in favorites crashing Zowe extension), which was resolved before we were considering Theia support:
This PR:
- Updates the incorrect profile message to say "Theia" instead of "VS Code" if a Theia user has invalid profile name(s) in their favorites settings for Data Sets, USS, or Jobs.
- Updates the i18n/sample folder for the revised message.

VS Code message example for Data Sets:
Screen Shot 2019-12-10 at 7 05 34 AM

Theia message example for Data Sets:
Screen Shot 2019-12-10 at 7 08 56 AM


lauren-li:
Thank you for the suggestion @Colin-Stone ! Yes, I think that's a good improvement to make.

Issue #367, opened 2019-12-05
Copy a data set or member name


jalel01:
Add the ability to copy a data set or member name.

Right clicking on the data set / member name would offer an option to copy the name to your clipboard.

Use case: i want to create a new dataset using an existing data set name. I copy the name of the existing data set, create a new data set, then paste the name of the existing data set and modify it with a '2' at the end to give it a unique name.

Alexandru-Dumitru:
Hi @jelaplan. 
Could we have your input on this one?
I was looking into it, and VS Code does not provide this functionality in their context menu by default. Is it something we want to add?

By default we have `copy`, `copy path`, `copy relative path`, and if I want to get the file name, I would do `rename` -> this would open the file in edit mode within the tree with the name pre-selected (without the extension), then I can do the `ctrl+c` to get the name in clipboard -> afterwards, I can paste it wherever I need it.

Thank you.

katelynienaber:
Also an additional question...we could add the `copy name` option to the data set, the member, or both...perhaps both would be overkill, and then we should choose one. What do you think?

MikeBauerCA:
Let's run this by @jelaplan. I am cautious about adding unnecessary options to the right-click menu. In the Zowe Explorer, you can already copy the name by clicking rename and copying from there.

katelynienaber:
Okay. I am new to this...do you think he will respond to the pull request
as it is?

Thanks,
-Katelyn Nienaber
+420 725 503 110
katelyn.nienaber@broadcom.com

jelaplan:
Hi Katelyn
Nice to have you join the team. 

I would prefer to not add an additional menu option. A user can use the rename menu option to achieve the same result. While that label maybe not as clear, adding additional menu options makes scanning the menu that much harder so makes it that much harder to use the menu. I think that over time, we will see a push to add additional menu items and I'd like to carefully manage the tension between additional capabilities with cluttering the menu.

PR #393, opened 2019-12-10
Copy a data set or member name

katelynienaber:
Hello :) First pull request woohoo!

This is related to issue [367: Copy a data set or member name](https://github.com/zowe/vscode-extension-for-zowe/issues/367).

I did the following:
-Added `Copy Name` button to data set/member context menus
-Added function `copyDataSetOrMemberName`
-Added unit/integration tests for function `copyDataSetOrMemberName`

@Alexandru-Dumitru suggested we check with UI before committing this.

Also perhaps we only need `Copy Name` on the data set OR the member, not both?

image

EDIT: Also ignore message on last commit, I deleted the changes to package.json and only added my new button...

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "vscode-extension-for-zowe",
-  "version": "0.29.0",
+  "version": "1.0.0",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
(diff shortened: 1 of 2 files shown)


zFernand0:
Should this operation be grouped with the `Rename Data Set`?
I'm just asking since you can click on Rename `[Member|Data Set]` and do `Ctrl + C`.

zFernand0:
Also, don't forget to sign-off your commits : )

Alexandru-Dumitru:
> Should this operation be grouped with the `Rename Data Set`?
> I'm just asking since you can click on Rename `[Member|Data Set]` and do `Ctrl + C`.

My point exactly! Currently, VS Code doesn't offer this context menu option. This doesn't mean that we cannot do it, ofc, but it goes against the UX. I've asked John to tell his opinion on this, but no reply so far. 

Hi @jelaplan, have you had any chance to look on this?

PR #400, opened 2019-12-11
Keytar support and removal of all sync process

Colin-Stone:
- Keytar is required if using the Default Credential Manager or SCS.
  - VSCode now loads Keytar dynamically.

- The CLI Credential Manager interfaces with the OS key chain using a reference that is related to the node job. This is the "name" value in the CLI package.json file.
 - VSCode will now initialize the CLI Credential manager (in it's own process) and can control the reference used to access the user/password credentials from the keychain/store created by VSCode or CLI. (@zowe/cli)
- If the Secure Credential Store plugin is installed in CLI, VSCode can initialize the CLI Credential manager to use the same reference as SCS and can access these credentials. Note it interacts using the default credential manager not SCS.

All this is happening within the VSCode process. No spawning and therefore no performance hit.
The only negative so far is that if accessing (from VSCode) credentials created by CLI. A (one time) pop up will appear from the Mac OSx credential manager requesting access to each of the user and password values maintained by it. I expect the reverse will be true (CLI accessing VSCode created credentials) Not an issue on windows


package.json
```diff
@@ -1162,10 +1169,12 @@
   "dependencies": {
     "@brightside/core": "^2.36.0",
     "@types/fs-extra": "^7.0.0",
+    "@zowe/imperative": "^4.4.4",
```
zFernand0: Do we need to import imperative?
I thought we got it already from `@brightside/core`

src/Profiles.ts
```diff
@@ -12,12 +12,10 @@
 import { IProfileLoaded, Logger, CliProfileManager, Imperative, ImperativeConfig, IProfile, Session, ISession } from "@brightside/imperative";
 import * as nls from "vscode-nls";
 import * as os from "os";
-import * as fs from "fs";
 import * as path from "path";
 import { URL } from "url";
 import * as vscode from "vscode";
 import * as zowe from "@brightside/core";
-import * as ProfileLoader from "./ProfileLoader";
```
zFernand0: I assume that the profile loader and some other files (i.e. `getDefaultProfile.ts` `getAllProfiles.ts`) can be deleted?

src/Profiles.ts
```diff
@@ -45,13 +43,15 @@ export class Profiles { // Processing stops if there are no profiles detected
         return Profiles.loader;
     }
     private static loader: Profiles;
+    private static spawnValue: number = -1;
     public allProfiles: IProfileLoaded[] = [];
     public defaultProfile: IProfileLoaded;
 
-    private spawnValue: number = -1;
-    private initValue: number = -1;
+    private initValue: number = 0;
```
zFernand0: initvalue not used anymore?

src/Profiles.ts
```diff
@@ -45,13 +43,15 @@ export class Profiles { // Processing stops if there are no profiles detected
...
+    private profileManager;
```
zFernand0: Should we think about exposing the profileManager from this class since it's where it's initialized?

src/Profiles.ts
```diff
@@ -292,53 +264,22 @@ export class Profiles { // Processing stops if there are no profiles detected
...
-            }
+    private async getCliProfileManager( type: string ) {
+        if (!this.profileManager) {
+            this.profileManager = await new CliProfileManager({
+                profileRootDirectory: path.join(os.homedir(), ".zowe", "profiles"),
+                type
+            });
         }
-        return this.spawnValue;
+        return this.profileManager;
```
zFernand0: I suggest making this a public getter for the profileManager.
That way you can avoid having to initialize it in multiple places (and guarantee a single instance of it)

src/extension.ts
```diff
@@ -115,6 +115,23 @@ export async function activate(context: vscode.ExtensionContext) {
 
         log = Logger.getAppLogger();
         log.debug(localize("initialize.log.debug", "Initialized logger from VSCode extension"));
+
+        const service: string = vscode.workspace.getConfiguration().get("Zowe-Builtin-Security")["service"];
+        if (service) {
+            const keytar = defineSecurity("keytar");
+            if (keytar) {
+                try {
+                    CredentialManagerFactory.initialize(
```
zFernand0: Perfect :+1:

src/extension.ts
```diff
@@ -382,6 +399,21 @@ export function defineGlobals(tempPath: string | undefined) {
...
+    try {
+        return require(`${vscode.env.appRoot}/node_modules.asar/${moduleName}`);
+        // tslint:disable-next-line: no-empty
+      } catch (err) {}
+    try {
+        return require(`${vscode.env.appRoot}/node_modules/${moduleName}`);
+      // tslint:disable-next-line: no-empty
+      } catch (err) {}
+    return undefined;
+}
```
zFernand0: If possible, I will convert this to a getter as well (to provide a more natural way of accessing/defining keytar.

I see where [this](https://code.visualstudio.com/api/advanced-topics/remote-extensions#persisting-secrets) was pulled from. However I do not believe we need to check for `node_modules.asar` since we are not using the asar node_module to package the extension.

That said, we may need to revisit this and confirm that this will work with webpack

src/extension.ts
```diff
@@ -115,6 +115,23 @@ export async function activate(context: vscode.ExtensionContext) {
 
         log = Logger.getAppLogger();
         log.debug(localize("initialize.log.debug", "Initialized logger from VSCode extension"));
+
+        const service: string = vscode.workspace.getConfiguration().get("Zowe-Builtin-Security")["service"];
```
zFernand0: :+1:

__tests__/__unit__/Profiles.unit.test.ts
```diff
@@ -390,40 +389,40 @@ describe("Profile class unit tests", () => {
...
-                    allProfiles: undefined,
-                    defaultProfile: undefined,
-                    map: jest.fn()
-                };
-            })
-        });
-        (await Profiles.createInstance(log)).listProfile();
-        expect(showErrorMessage.mock.calls.length).toBe(1);
-    });
+    // it("should call listProfile", async () => {
```
zFernand0: We will need to figure out testing for this implementation.
And clean-up some other test files maybe?

src/Profiles.ts
```diff
@@ -61,45 +61,17 @@ export class Profiles { // Processing stops if there are no profiles detected
...
-            this.allProfiles = ProfileLoader.loadAllProfiles();
-        } catch (error) {
-            vscode.window.showErrorMessage(error.message);
-        }
-        this.allProfiles.map((profile) => {
-            return profile.name;
+    public async refresh() {
+        const profileManager = await this.getCliProfileManager("zosmf");
+        this.allProfiles = (await profileManager.loadAll()).filter((profile) => {
+            return profile.type === "zosmf";
```
zFernand0: We may want to parameterize the profile type in order to make it easier in case we need to expose it in the settings, (FTP and SSH in the future?)

package.json
```diff
@@ -1073,6 +1073,13 @@
           "description": "Zowe-Temp-Folder-Loacation",
           "scope": "window"
         },
+        "Zowe-Builtin-Security": {
+          "default": {
+            "service": ""
+          },
+          "description": "%Zowe-Builtin-Security",
+          "scope": "window"
+        },
```
zFernand0: This provide enough customization for the user. :+1:

I'd like to propose a quality-of-life enhancement here (no need to be done now).
Since we support certain services (`@zowe/cli` and `@brightside/core`) I say we could detect whether a profile has secured options by looking at the profile_schema ( properties with `secure:true`) and the property value (assuming it starts with `managed by`)
https://github.com/zowe/imperative/blob/8ca1488259295f4bc40ed5c328c90463f659c86e/packages/profiles/src/constants/ProfilesConstants.ts#L17

package.json
Colin-Stone: No I thought we got it from imperative but that was just a dev dependency. I trust this is till okay. It only gets pulled in if there is a need to go secure

src/Profiles.ts
Colin-Stone: Yes they can but I got nervous deleting so much code until we were all comfortable with this solution

src/Profiles.ts
Colin-Stone: Correct. I'll remove

src/Profiles.ts
Colin-Stone: I didn't really want it accessed from outside of this instance of Profiles. It is only referenced in here and can only be initialized once per type.

src/Profiles.ts
Colin-Stone: I don't understand why we would want it accessed from outside of this instance so I was locking it down.

src/extension.ts
Colin-Stone: I suppose I consider a getter should return something. The only thing this returns is undefined hence I didn't use get. I will remove asar because we have the code documented in your extract if we need to put it back.
Does getSecurityModules sound correct?.

__tests__/__unit__/Profiles.unit.test.ts
Colin-Stone: Are you thinking testing such as integration testing?. I will check the testing for more stuff to remove

src/Profiles.ts
Colin-Stone: Yes I thought we would probably end up going that route at some point but thought we would simply re-engineer then.

__tests__/__unit__/Profiles.unit.test.ts
Colin-Stone: Removed all of the redundant tests. I didn't create any new ones as I still seem to have the coverage I need and the tests were for now removed code

package.json
Colin-Stone: I don't believe we need it now and will take it out.

Colin-Stone:
To use this implementation, please note the new critical settings 
image
In order for Zowe explorer to perform the necessary keytar initialization it checks this value first. If a service has been set then keytar initializes based on that value. In this example it happens to be based upon the name of the application that would have been the creating application. In other words the Zowe CLI. If no value is set keytar is not initialized and passwords are visible.

If the Secure plugin is used that simply overrides everything.

The other item to note is that the .zowe/settings/imperative.json values. These are primarily used by Zowe CLI rather than by Zowe Explorer but obviously need to be in synchronization

Colin-Stone:
It's still a problem if the user tries mixing plain text and secured profiles. I think it may be worth creating an issue to allow this situation to be recognized and to either fail more gracefully by warning of the situation but allowing continuation of the extension with whatever profiles it is able to use

zFernand0:
> The other item to note is that the .zowe/settings/imperative.json values. These are primarily used by Zowe CLI rather than by Zowe Explorer but obviously need to be in synchronization

Do we have an issue to synchronize this?

Colin-Stone:
> > The other item to note is that the .zowe/settings/imperative.json values. These are primarily used by Zowe CLI rather than by Zowe Explorer but obviously need to be in synchronization
> 
> Do we have an issue to synchronize this?

Sorry I hadn't given this much thought as it didn't appear to bring much to the party :-). In imperative.json I see credential-manager or CredentialManager (I can never remember which one does which) we either get false or a Managed by message. False is useful because it should tell me we don't want to use any security but I still need a value from somewhere that indicates the key prefix. This is the @zowe/cli or broadcom/plugin I need to work with the keystore/vault. I suggest this is part of Zowe Explorer settings.   
So i'm still unsure which way to move just now. What's your opinion?.

Colin-Stone:
> > > The other item to note is that the .zowe/settings/imperative.json values. These are primarily used by Zowe CLI rather than by Zowe Explorer but obviously need to be in synchronization
> > 
> > 
> > Do we have an issue to synchronize this?
> 
@zFernand0 Added some additional code in extension that not only checks the new settings but also imperative.json for a non false value of some kind. IS this the type of synchronisation you were thinking of?.

Colin-Stone:
Modified a setting so that it is a UI setting rather than a json edited setting. More user friendly.
image

src/Profiles.ts
```diff
@@ -300,59 +278,27 @@ export class Profiles { // Processing stops if there are no profiles detected
...
-                const value = settings.overrides.CredentialManager;
-                this.spawnValue = value !== false ? 0 : 1;
-            } catch (error) {
-                // default to spawn
-                this.spawnValue = 0;
-            }
+    private async getCliProfileManager( type: string ) {
+        if (!this.profileManager) {
+            this.profileManager = await new CliProfileManager({
+                profileRootDirectory: path.join(os.homedir(), ".zowe", "profiles"),
```
tjohnsonBCM: It's not always safe to use `os.homedir()` here. If the environment variable `ZOWE_CLI_HOME` is set, it should take precedence.

Previously `ImperativeConfig.instance.cliHome` was used to handle this, but it is no longer convenient to use because it requires that `Imperative.init` has already been called.

The following hack would work by mocking out the `loadedConfig` object, but hopefully someone has a better idea :)
```
ImperativeConfig.instance.loadedConfig = {
    defaultHome: path.join(os.homedir(), ".zowe"),
    envVariablePrefix: "ZOWE"
};
const zoweDir = ImperativeConfig.instance.cliHome;
```

src/extension.ts
```diff
@@ -323,6 +340,39 @@ export function defineGlobals(tempPath: string | undefined) {
...
 
+/**
+ * function to check if imperative.json contains
+ * information about security or not and then
+ * Imports the neccesary security modules
+ */
+export function getSecurityModules(moduleName): NodeRequire | undefined {
+    let imperativeIsSsecure: boolean = false;
+    try {
+        const homedir = os.homedir();
```
tjohnsonBCM: See above comment about the problem with using `os.homedir`

src/Profiles.ts
Colin-Stone: Thanks Timothy I can add that into the code. My tests so far show that Imperative had not been initialized when the Config is referenced. Is there a way that I can do that prior to calling these functions?

src/extension.ts
Colin-Stone: Created an exported function for this.

src/Profiles.ts
tjohnsonBCM: Discussed this with @zFernand0, it's good that the code initializing Imperative was removed in this PR because it improves performance :) This is really a defect in Imperative, it shouldn't be necessary to call its initialize function just to access constants like the CLI home directory.

tjohnsonBCM: Created an issue in Imperative to track this: https://github.com/zowe/imperative/issues/357

Colin-Stone:
@tjohnsonBCM Implemented as per your suggestion. Seems to initialize correctly in my testing and I would be keen to get something delivered in v1.2 to be compatible with the SCP even though there will be subsequent changes to imperative.

zFernand0:
Pretty exiting stuff. :) 
I'm sure people will be happy with secure profiles by default on the next release.
Zowe Explorer 1.2 (coming soon) 😄

Issue #398, opened 2019-12-10
Tree view is not refreshed for PDS member actions


crawr:
The tree view is not updated when the following actions are performed on a PDS:
- Upload Member
- Delete Member

Note: This applies to both Theia and VSCode

zFernand0:
Workaround:
Close and reopen the "folder" (PDS) :)

crawr:
Yep, I agree with this workaround. I raised the issue because Rename Member action refreshes the tree view and I was expecting the same behavior for Upload and Delete actions.

I will tag it as enhancement request :)

PR #401, opened 2019-12-11
Fix for Tree view is not refreshed for PDS member actions (issue #398)

katelynienaber:
Fix for [#398](https://github.com/zowe/vscode-extension-for-zowe/issues/398)

Added some small bits of code in functions `deleteDataset `and `uploadDialog`.

### DIFF
--- a/resources/testProfileData.example.ts
+++ b/resources/testProfileData.example.ts
@@ -2,14 +2,14 @@ import { IProfile } from "@brightside/imperative";
 
 export const profile: IProfile = {
     type : "zosmf",
-    host: "",
-    port: 0,
-    user: "",
-    password: "",
+    host: "192.86.32.67",
+    port: 10443,
+    user: "Z40642",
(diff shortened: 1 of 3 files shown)


resources/testProfileData.example.ts
```diff
@@ -2,14 +2,14 @@ import { IProfile } from "@brightside/imperative";
 
 export const profile: IProfile = {
     type : "zosmf",
-    host: "",
-    port: 0,
-    user: "",
-    password: "",
```
zFernand0: I'd suggest not changing the example file, but rather creating a `testProfileData.ts` with the desired information.

We should document this better (maybe within this same file?)

Colin-Stone: Additionally you should try and remove this file from this PR sooner rather later as it contains sensitive data. Maybe close this and submit another

katelynienaber:
Okay, will pull again...

PR #402, opened 2019-12-11
Publish a 1.0.1 version

zFernand0:
Signed-off-by: zFernand0 <fernando.rijocedeno@broadcom.com>


zFernand0:
Going forward, fixes should be ported to the `v1.0.x` branch if we want to publish them.

We can discuss more about this on Friday :smile:

MikeBauerCA:
@zFernand0 thanks for separating and getting these fixes out!!

PR #414, opened 2019-12-17
Fix Issue 413 according to comments

jellypuno:
Signed-off-by: jellypuno <jessielaine.punongbayan@broadcom.com>

### DIFF
--- a/__tests__/__unit__/extension.unit.test.ts
+++ b/__tests__/__unit__/extension.unit.test.ts
@@ -372,7 +372,6 @@ describe("Extension Unit Tests", () => {
     Object.defineProperty(vscode.env, "clipboard", { value: clipboard });
     Object.defineProperty(Rename, "dataSetMember", { value: renameDataSetMember });
 
-
     beforeEach(() => {
         mockLoadNamedProfile.mockReturnValue({profile: {name:"aProfile", type:"zosmf"}});
         Object.defineProperty(profileLoader.Profiles, "getInstance", {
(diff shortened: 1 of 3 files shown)


Colin-Stone:
This may help https://code.visualstudio.com/api/extension-guides/command

jellypuno:
> This wont work. You have added the datasetProvider as an optional field
> `export async function openPS(node: ZoweNode, previewMember: boolean, datasetProvider?: DatasetTree)`
> In your code you will get an undefined error thrown when opening from favorites
> `await datasetProvider.refreshElement(node); await datasetProvider.refresh();`
> 
> Your solution may be around the command
> `vscode.commands.registerCommand("zowe.ZoweNode.openPS", (node) => openPS(node, true));`
> This needs the datasetProvider added. Maybe then it's possible to call the command with the node

I tested it. It is not having an error except for refreshJobsServer() that's why I passed the provider there: `vscode.commands.registerCommand("zowe.refreshJobsServer", async (node) => refreshJobsServer(node, jobsProvider));`

I could still add the providers as you said but this code already works.

Colin-Stone:
image

jellypuno:
@Colin-Stone I find it weird that I did not experience this during my testing. I was in the impression that the values for the provider is propagated during initialization so I don't need to pass it anymore. At least that is what I saw. But anyways, thank you for testing it and capturing this error.

jellypuno:
> That's sorted it. Thanks Jelly. No idea why code coverage is down, I thinkit may be a glitch with codecov itself rather than a reduction in testing

Yeah... I don't know as well. 😕  Thank you for your reviewing. 👍

PR #430, opened 2019-12-24
Rebased on interface

Colin-Stone:
Introduced interfaces. See api/ZoweTree

Would still like to refine a little more but was hoping to get a perspective from @zFernand0 before going any further


Colin-Stone:
Lint errors and ambiguity over mProfileName

PR #434, opened 2019-12-30
Added URL template to valid URL error message

katelynienaber:


Colin-Stone:
Just an observation but if you alter text that a user would receive you should really generate the NLS version of that text. I usually run `npm run vscode:prepublish` before commiting as this regenerates any nls texts. If you do this you will see that more files will be altered.

katelynienaber:
> Did you want to localize this one too?.

Yes it should be fixed now ^^'

Issue #263, opened 2019-10-09
Filters behavior wording and reworking


Colin-Stone:
Feedback from https://github.com/zowe/vscode-extension-for-zowe/issues/102

~~1. For consistency, should label "Search for Jobs using attributes" be "Search Jobs Using Attributes"?
   Compare to casing and style of "Search Data Sets by Entering Patterns".
2. Label "Choose "Create new..." to define a new filter alternatively select a previously defined filter" needs to have a comma, be two sentences or rephrased
   eg. "Choose "Create new..." to define a new filter or select a previously defined filter".~~
3. When creating a new filter and entering the owner, it allows for multiple words eg. "COCKERM BLA" when only one word is actually used.

Colin-Stone:
Reworded for points 1 and 2 rather than miss them later. Issue still good as a first issue adding the callback function

thushanwasa:
Can I call

parschina:
[واردات از چی](https://parschina.ir/)

katelynienaber:
Changes were pushed today (pull request [#435](https://github.com/zowe/vscode-extension-for-zowe/pull/435))

PR #435, opened 2019-12-30
Disallowed multi-word profile names in job search.

katelynienaber:
Please check and see if this is what you need.

image


Colin-Stone:
What defect does this relate to?. The code relates to the user enter an owner id but the text added describes a profile.

katelynienaber:
> What defect does this relate to?. The code relates to the user enter an owner id but the text added describes a profile.

This was in relation to issue [#263](https://github.com/zowe/vscode-extension-for-zowe/issues/263). Sorry didn't notice there was no reference number.

Colin-Stone:
Thanks @katelynienaber, so maybe the text should be different ?. Owner rather than profile?.

katelynienaber:
> Sorry but I had missed the fact that no localize appears around the text you have updated so it hasn't being translated you should probably have put `validateInput: (value: string) => (value.match(/ /g) ? localize("jobs.enter.valid.owner", "Please enter a valid owner name (no spaces allowed).") : "")`

Should I also add this localize function to my edits on pull request [#434](https://github.com/zowe/vscode-extension-for-zowe/pull/434)?

Issue #422, opened 2019-12-17
Add to filter tip to convey ability to add multiple filters entries. 


jelaplan:
@morridt commented in [385](https://github.com/zowe/vscode-extension-for-zowe/issues/385) that he didn't realize that the extension supports adding multiple filter strings. To help users to get the most our of search, we could change the tip to inform them. 

![image](https://user-images.githubusercontent.com/3207260/71035402-6314d580-20e9-11ea-8877-5e9a838de96c.png)

katelynienaber:
Cosed in pull request [#437](https://github.com/zowe/vscode-extension-for-zowe/pull/437).

zFernand0:
Should this go under the 1.2 milestone?

PR #437, opened 2020-01-03
Add to filter tip to convey ability to add multiple filters entries (#422)

katelynienaber:
Changed the `ussFilterPrompt.option.prompt.search` text in `DatasetTree.ts` & `USSTree.ts`

In response to issue [#422](https://github.com/zowe/vscode-extension-for-zowe/issues/422).


katelynienaber:
> This functionality is not implemented on USS filters. It does work on Datasets. Maybe that's what you were thinking of?.

Ya, whoops. It was using the same USS text reference for both of them. Is this better?

PR #440, opened 2020-01-07
CLI Extensibility API 

Colin-Stone:
PR created initially to assess potential impact but now in state appropriate for general review.


Colin-Stone:
Looking good. I too had the integration tests working but did not have to change Profiles.ts to do it. It's a sensitive area which I am about to change again with the keytar PR which will go in for 1.2
Did you have any reason to add this refresh to loadNamedProfile ?
`        
    public loadNamedProfile(name: string): IProfileLoaded {
        if (!this.allProfiles || this.allProfiles.length === 0) {
            this.refresh();
        }
` 
Other changes I have from yours I have the IProfileLoaded with fail not found false and none of the 
`
    Profiles.createInstance(Logger.getAppLogger());
`
you have in the integration tests. I think it's worth pushing my change as it is a reduction in lines of code. The other question I have is the fact that we have inconsistency between USS and Datasets nodes. I think MVS needs IProfileLoaded and USS just uses the profile name. Is this an area we can rationalize?

phaumer:
Thanks a lot @Colin-Stone. 

I added commits with the changes you request: I removed the FTP implementation and placed it into a standalone repository here: https://github.com/phaumer/vscode-extension-for-zowe-api-sample

The README of that repo explains how to build and use it to test out that the Extension API is working.

image

phaumer:
Oh, about the profile parameters, @Colin-Stone: I agree 100%. My goal with this PR is to introduce only one new set of abstractions at the time and keep the change of refactoring to a minimum by just adding for now and not rewriting everything. 

A second round of refactoring should look at the right abstractions to use. I think the use of Session in Zowe Explorer should be entirely encapsulated as what we really need in almost all cases is the profile. The Session has already lost all information about the profile, such as the profile name needed to display in the tree browser and hence a hacky parameter name such as profileName that you mentioned; it is also entirely http specific, which is not the abstraction we need when we want to support ftp or ssh.   I would propose to store a profile reference with each node and derive the session object from that (as I have implemented in the common interface). Then we can get rid of session in constructors and parameters and do need to create them in all those many places, but just in one central place.

Colin-Stone:
Just pushed an update because of an issue with sctringUpdate and ZoweVscApiRegister references @lauren-li would you take a quick look to see if this is correct please

phaumer:
@lauren-li @Colin-Stone I see what the problem was. I can fix it.

lauren-li:
@Colin-Stone @phaumer I submitted a PR into this branch to update the localization for ZoweExplorerApiRegister and added both of you as reviewers: https://github.com/zowe/vscode-extension-for-zowe/pull/451

In the future, it might be nice to automate the updating of stringUpdateScript.js so that developers will not have to remember to update it every time they create a new file that needs localization.

lauren-li:
Closing PR #451 as the ZoweExplorerApiRegister localization issue was fixed in commit: https://github.com/zowe/vscode-extension-for-zowe/commit/8cedf733298c0f56a4b39b5f4e5ae729788d80c4

Thank you @phaumer for taking care of it!

jellypuno:
I am experiencing this error when I am viewing my newly added favorites. I am using a zosmf profile.

Capture

After I refreshed my workspace, I can view the favorites again. The error happens only when you are trying to view the added favorites on the same session.

jellypuno:
USS session is set to null after I created a new file.

Steps to reproduce:
1. Use a profile without any credentials
2. Login to your credentials
3. Search for a USS path.
4. Create a new File
5. Open that created file - 401 error is encountered.

zFernand0:
Based on the discussion last week I wanted to follow up on the action items. : )
Are we planning to add the profile type next to the name as part of this PR?

phaumer:
Thanks @jellypuno I have a look and will fix. @zFernand0 yes, I will add it.

zFernand0:
I believe there is a problem introduced in this PR.
When uploading members to a PDS, it doesn't respond with an error message (so we can assume it was good)
However, the new member does not appear in the list.

As a matter of fact, all PDS appear to be empty. 😢

phaumer:
@jellypuno I fixed the issue with the Favorites. 

Can you provide more information about the no credentials case? Isn't that the whole point of profiles? To stored your credentials? When I remove them from the profile yaml file and try to add them to the tree I get this error:

image

phaumer:
@zFernand0 Can you try this again with my latest commit? The issue of no progress report and no refresh of the favorite (for adding or deleting) or the main tree I see in master as well, so we could fix that in a separate Issue. But uploading works for me from the Fav.

jellypuno:
> @jellypuno I fixed the issue with the Favorites.
> 
> I cannot reproduce the Delete member case. Do you have some steps to reproduce?
> The only issue I see is that after Delete the tree does not get refreshed, but I see the same behavior in the master branch as well, so this could be fixed with a separate PR.
> 
> Can you provide more information about the no credentials case? Isn't that the whole point of profiles? To stored your credentials? When I remove them from the profile yaml file and try to add them to the tree I get this error:
> 
> image

Today, we are able to store profiles with optional profiles. The yaml file will look like this:

```
host: host
port:  123
user: ""
password: ""
rejectUnauthorized: false
```

If you use a profile with optional credentials, it will prompt you for your username and password then it will save it as a session. Then you can use Zowe Explorer with those credentials.

phaumer:
@jellypuno got it. Thanks. I can patch my yaml files and see the behavior you described. 

Just as a side questions, however: I do not seem to be able to create such a profile like this via command line. How would I do that? Are we are using profiles here in a way that is not really supported by the CLI?

I also ran into the problem that when you enter the incorrect credentials that you are stuck and do not get re-prompted. You have to remove the session and add it again or developer-reload the entire VS Code window. 

I am working on a fix for profiles that will keep the behavior consistent for now with the session prompts; including the perceived flaw of not being able to re-prompt. I will file another issue for proposing a refactoring to only use profiles once we get this PR through the reviews.

jellypuno:
> @jellypuno got it. Thanks. I can patch my yaml files and see the behavior you described.
> 
> Just as a side questions, however: I do not seem to be able to create such a profile like this via command line. How would I do that? Are we are using profiles here in a way that is not really supported by the CLI?

This is a feature available only for Zowe Explorer. If you are going to use the profile in CLI, you need to supply --user --password in the command. 

> 
> I also ran into the problem that when you enter the incorrect credentials that you are stuck and do not get re-prompted. You have to remove the session and add it again or developer-reload the entire VS Code window.
> 

There is an issue open for this enhancement.

> I am working on a fix for profiles that will keep the behavior consistent for now with the session prompts; including the perceived flaw of not being able to re-prompt. I will file another issue for proposing a refactoring to only use profiles once we get this PR through the reviews.

Great! Thanks :)

jellypuno:
I added a PDS in Favorites and when I clicked it, I experience this error.

Capture

I tested in master and the normal behavior is that it should show the members inside the PDS.

jellypuno:
> I cannot reproduce the Delete member case. Do you have some steps to reproduce?

Me too. I can't reproduce it anymore.

jellypuno:
This happened to me twice now. I am not sure why.

pds not opening

Steps I did:

1. I uploaded a file as a PDS member.
2. Nothing happens so I tried to refresh it.
3. After Refresh, I cannot open any PDS anymore. but I can open PS files.

zFernand0:
☝️ https://github.com/zowe/vscode-extension-for-zowe/pull/440#issuecomment-577108178
That's similar to what I experienced before, seems to be intermittent since I got it to work after refreshing the instance of VSCode. 🤔

phaumer:
Sorry, I was away for a couple of days. Will resume work today, latest tomorrow..

Colin-Stone:
Can references to node.profile field become node.getProfile() ?. 

It is an intention to clean up interfaces and optional parameters with private fields and public getters in classes. Implementing the getProfile() method in the node classes now should avoid additional rework.

phaumer:
@jellypuno I see the same problem in the master branch as it is a bug in the DatasetTree.addFavorite() method. It was testing the wrong data set type mixing the cases of a partitioned vs a serial dataset. I fixed that in the commit above.

> I added a PDS in Favorites and when I clicked it, I experience this error.
> I tested in master and the normal behavior is that it should show the members inside the PDS.
> I also added a sequential file in the favorites and when I clicked it, it doesn't open the text file

src/api/ZoweExplorerApi.ts
```diff
@@ -0,0 +1,472 @@
...
+
+        /**
+         * Deletes the USS file at the given path.
+         *
+         * @param {string} fileName
+         * @param {boolean} [recursive]
+         * @returns {Promise<zowe.IZosFilesResponse>}
+         */
+        delete(
+            fileName: string,
```
zFernand0: I hope I'm not disrupting too much here.
I was just thinking maybe some consistency between path vs name would be nice.
I'm not really requesting for this to be changed at this time though.

phaumer:
@zFernand0 thanks so much for fixing the tests. Was just about to get started on those after my daily allotment of meetings got finished ;-) 

Will review the naming now...

src/api/ZoweExplorerApi.ts
phaumer: Done.

phaumer:
Ok, I think I fixed all the issues and feedback. Are we ready to merge?

I propose to show the profile type in a separate Pull Request now as postponing this PR causes continuous issues with other changes being merged into master (spent a whole day catching up on Tuesday). I also would like to think about the best way of doing it as well. It would probably make sense to only show the type if there is more than one profile type registered.

phaumer:
@jellypuno I can reproduce the problem also in the master branch. I think this is more related to the redundant business logic and handling of special cases inconsistently? I will do some debugging, but I think it is independent of my PR.

phaumer:
I files this issue for that case. Seems I cannot create any data sets in master right now without the tree getting broken: https://github.com/zowe/vscode-extension-for-zowe/issues/485

zFernand0:
Thanks to everyone who made this possible. 👍

Issue #441, opened 2020-01-08
Create vsix deployment pipeline


katelynienaber:
As a Zowe Explorer developer, i want to build and package the latest development version of the Zowe Explorer in the Zowe Artifactory.

zFernand0:
Do we want these snapshots to be taken right after the build?
This way, even if the test fail, you stilll have a snapshot-vsix uploaded to the zowe artifactory for debugging purposes (if needed).

katelynienaber:
Yes I can move it, updates will be in the pull request in like 5 minutes...

jellypuno:
Referencing #456

katelynienaber:
This was addressed in PR #443

PR #443, opened 2020-01-09
Jenkins publish to artifactory

katelynienaber:
Fix for issue [#441](https://github.com/zowe/vscode-extension-for-zowe/issues/441).

Publishes the .vsix file to `https://zowe.jfrog.io/zowe/libs-release-local/org/zowe/vscode`


crawr:
Suggestion: Add a new step to publish a vsix file to a libs-snapshot-local folder so that we can check the changes outside VSCode even before they are merged to the master branch.

katelynienaber:
> Do we still need the npm_login.expect file?

Noop...committing again...

zFernand0:
Suggestion: When updating pipelines, try using the `replay button on the Jenkins server.
It should save you some time since you don't have to make a commit and push every single change to test it. 😸

Jenkinsfile
```diff
@@ -9,6 +9,9 @@
 *
 */
 
+@Library('shared-pipelines') import org.zowe.pipelines.nodejs.NodeJSPipeline
+def lib = library("jenkins-library").org.zowe.jenkins_shared_library
+
```
zFernand0: Do we need to import the shared libraries here?

katelynienaber: No, I forgot to remove them.

So they are removed and also @crawr's solution is added in the Smoke Test step.

Jenkinsfile
```diff
@@ -98,6 +115,26 @@ pipeline {
         } }
       }
     }
+    stage('Smoke Test') {
```
zFernand0: I'm a little confused as to why we are calling this stage "Smoke test"
Are we planning to crate a draft snapshot release on GH? 
Is the intention to do some pass-fail automated test with the vsix file in this stage?

Jenkinsfile
```diff
@@ -98,6 +115,26 @@ pipeline {
...
+      when { allOf {
+        expression { return !PIPELINE_CONTROL.ci_skip }
+      } }
+      steps {
+        timeout(time: 10, unit: 'MINUTES') { script {
+          def vscodePackageJson = readJSON file: "package.json"
+          def version = "v${vscodePackageJson.version}"
+          sh "git tag ${version}"
```
zFernand0: Is the intention to push a snapshot git-tag to GH?
If not, I believe this can be removed from the stage.

Jenkinsfile
```diff
@@ -98,6 +115,26 @@ pipeline {
...
+
+          sh "npx vsce package -o ${version}.vsix"
+
+          // Release to Artifactory
+          withCredentials([usernamePassword(credentialsId: ARTIFACTORY_CREDENTIALS_ID, usernameVariable: 'USERNAME', passwordVariable: 'PASSWORD')]) { script {
+            def uploadUrlArtifactory = "https://zowe.jfrog.io/zowe/libs-snapshot-local/org/zowe/vscode/${version}.vsix"
+            sh "curl -u ${USERNAME}:${PASSWORD} --data-binary -H \"Content-Type: application/octet-stream\" -X PUT ${uploadUrlArtifactory} -T @${version}.vsix"
+          } }
```
zFernand0: If the purpose of this stage is to "Archive Build artifacts in Zowe artifactory", do we need a separate stage, or could we move this logic to the build stage.
I'm on the fence on this one since it doesn't really affect too much to have lots of stages :)

zFernand0: @jackjia-ibm, It'd be nice if we could overload the `archiveArtifacts` method to upload files to the zowe.jfrog.io site, (maybe on a default artifacts location, or the user-specified location in artifactory)

Jenkinsfile
```diff
@@ -176,6 +213,12 @@ pipeline {
 
           sh "npx vsce package -o ${version}.vsix"
 
+          // Release to Artifactory
+          withCredentials([usernamePassword(credentialsId: ARTIFACTORY_CREDENTIALS_ID, usernameVariable: 'USERNAME', passwordVariable: 'PASSWORD')]) { script {
+            def uploadUrlArtifactory = "${DL_ARTIFACTORY_URL}/${version}.vsix"
+            sh "curl -u ${USERNAME}:${PASSWORD} --data-binary -H \"Content-Type: application/octet-stream\" -X PUT ${uploadUrlArtifactory} -T @${version}.vsix"
+          } }
```
zFernand0: :+1:

Jenkinsfile
katelynienaber: Yes I believe the goal here was, to quote @crawr:
> Suggestion: Add a new step to publish a vsix file to a libs-snapshot-local folder so that we can check the changes outside VSCode even before they are merged to the master branch.

So that we can still get the vsix from the libs-snapshot-local folder, even if all the tests don't pass, for testing purposes.

Jenkinsfile
katelynienaber: Should remove these...

Jenkinsfile
jackjia-ibm: @zFernand0 good idea. I created an issue https://github.com/zowe/jenkins-library/issues/57 to track it.

jellypuno:
@katelynienaber @zFernand0 @jackjia-ibm I have an idea. Is it possible to add the PR# if the VSIX file is created during testing? The scenario that I have in mind is that what if there are 3 people enhancing vscode and all of them wanted to test their changes in theia. How are they going to get the right vsix file where their changes are included?

zFernand0:
> Is it possible to add the PR# if the VSIX file is created during testing? 

Definitely possible. 😄

jackjia-ibm:
@jellypuno I didn't read the Jenkinsfile before your cooment, just realize it's not using jenkins library we have. CLI is widely using https://github.com/zowe/zowe-cli-version-controller, and ibm projects and part of zlux are using https://github.com/zowe/jenkins-library. They can all handle the versioning for you.

For example, each branch of explorer-jes will be published to `libs-snapshot-local` : https://zowe.jfrog.io/zowe/webapp/#/artifacts/browse/tree/General/libs-snapshot-local/org/zowe/explorer-jes. And releases are built by checking parameter `Perform Release on release branches (https://wash.zowe.org:8443/job/explorer-jes/job/master/build?delay=0sec), and will be published to `libs-release-local`: https://zowe.jfrog.io/zowe/webapp/#/artifacts/browse/tree/General/libs-release-local/org/zowe/explorer-jes. You can see the version scheme put into the build to distinguish the build from different branches and purpose. If you are interested in `jenkins-library`, here is how the pipeline looks like: https://github.com/zowe/explorer-jes/blob/master/Jenkinsfile.

I and Fernando are happy to answer your questions if you wish to use the library.

tjohnsonBCM:
> Is it possible to add the PR# if the VSIX file is created during testing?

Also, could we include the extension's name in the VSIX filename? This would be consistent behavior with how VSIX files are named on the VSCode Marketplace. For example, instead of `v1.1.0.vsix` the name would be `vscode-extension-for-zowe-1.1.0.vsix`.

zFernand0:
I would definitely like to see the combined version of the libraries being adopted everywhere. :) 

I should say that I'm a bit at fault here for not using it since the very start.
We can certainly open an issue and record the effort of migrating to one of the two libraries 👍 

Thanks for the suggestion Jack

katelynienaber:
Okay, so I added the PR# and the extension name to the Artifactory snapshot .vsix file, and just the extension name to the release .vsix file.

Also removed: sh "git tag ${version}" as mentioned by @zFernand0 

The idea to use the Jenkins Shared Libraries in this Jenkinsfile has been added to issue [#456](https://github.com/zowe/vscode-extension-for-zowe/issues/456) by @jellypuno

zFernand0:
I don't mind if the publisher name is not included in the vsix name too much. i.e. `zowe.vscode-extension-for-zowe...vsix` instead of `vscode-extension-for-zowe...vsix`.

katelynienaber:
@jackjia-ibm Our Jenkins user needs permission to delete (overwrite) files with the same name in Artifactory. Specifically in the snapshot folder...the release folder should only receive a .vsix file once, when the pull request is merged into master. Please let me know if you need more information ^^

jellypuno:
Looks good to me. The right file size is uploaded 👍

Issue #370, opened 2019-12-06
Re-Prompting of Credentials


jellypuno:
If I'm prompted for credentials and I entered wrong information, I do not know how to correct it. 
A button for Re-Prompting for credentials might be good.

Workaround: refresh the vscode workspace

jellypuno:
It's better to implement #354 first then this one.

MikeBauerCA:
Instead of a button, the user should be automatically re-prompted if the service informs of invalid credentials.

sladyn98:
I could give this a shot, any pointers on how to begin ?

zFernand0:
Hey @sladyn98, 
As @jellypuno mentioned, it'd be ideal to centralize the prompting mechanism so that we don't implement the same logic in multiple places (should we need to).

Also, do we want to re-prompt if we where not able to verify the credentials?

How about the host and port? (could be validated with a `zosmf check status` which doesn't require credentials)

sladyn98:
Thanks @zFernand0 
> Also, do we want to re-prompt if we where not able to verify the credentials?

Yeah it would be good practice to re prompt if we aren't able to verify the credentials.
 
>How about the host and port? (could be validated with a zosmf check status which doesn't require credentials)

I am not too familiar with zowe but is this related to the zowe cli ?

zFernand0:
`zosmf check status` is related to the CLI, but I meant to say the `/zosmf/info` endpoint to verify that the host and port are Ok :)

sladyn98:
Okay so basically we want to reprompt for credentials on these conditions
1) Invalid Credentials
2) Human Error in entering Credentials.

jellypuno:
> `zosmf check status` is related to the CLI, but I meant to say the `/zosmf/info` endpoint to verify that the host and port are Ok :)

Issuing zosmf check status in every profile can result to slowness of the vscode extension. I really do not like using this command to check the host and port. That is why I am leaning more for the edit button or just re-prompt it if 401 is encountered.

jellypuno:
#580

Issue #388, opened 2019-12-10
Centralized error handling


jellypuno:
It would be nice to have a new function that handles error messages. 

For Example:
```
If (err.errorCode === 401) {
    await ErrorHandling("ZWEEX401E")
}
```

```
public async ErrorHandling(errCode) {
    switch(errCode) {
        case "ZWEEX401E": {
             vscode.window.showErrorMessage(localize("getChildren.error.response", "Invalid 
             Credentials for profile") + `\n${this.label}\n` + ". Please make sure that the 
             credentials are valid.");
```

And then we could have a document like the [api-layer troubleshooting](https://github.com/zowe/docs-site/blob/master/docs/troubleshoot/troubleshoot-apiml-error-codes.md) for easier troubleshooting.

zFernand0:
This will definitely be helpful if the decision is to move forward with Message IDs.
If someone has any concerns about this, please feel free to bring it up.

This centralized location could also be used for handling any error mapping as well (if we decide to go that route too).

My only concerns are:
- The extra maintenance (which probably isn't too bad if we had scripts that produced said troubleshooting document for example)
- The temporary inconsistency (at least until we go back and create Message IDs for every other error message that we create, which could be a bit harder to automate)

jellypuno:
I created a POC for this one. The message ID format is:

ZE = Zowe Explorer
ZN = Zowe Node (The class where the error is triggerred)
401 = https Status Code
S = Severe Error

![Capture](https://user-images.githubusercontent.com/45042840/71979379-dbb90e80-321d-11ea-966f-22fa1cf6a933.PNG)

zFernand0:
I like where this is going! 👍 

Questions:
1. What happens if the Class name is too long or has more than two distinct words? (e.g. ZoweAwesomeNode)
2. Since we already have the HTTP Status Code, do we need the severity level?
3. Can there be Different types of message with different "severity" levels? (i.e. Trace, Debug, Info, Warning, Failure)

jellypuno:
> I like where this is going! 👍
> 
> Questions:
> 
> 1. What happens if the Class name is too long or has more than two distinct words? (e.g. ZoweAwesomeNode)
> 2. Since we already have the HTTP Status Code, do we need the severity level?
> 3. Can there be Different types of message with different "severity" levels? (i.e. Trace, Debug, Info, Warning, Failure)

[...]

Issue #373, opened 2019-12-06
Optional Credentials: Prompt* as value for Username and Password


jellypuno:
I don't want to save my credentials in ZOWE CLI so I created a profile with --user prompt* and --pass prompt* so that whenever i use it, I will be prompted for my credentials. 

The issue is that Zowe Explorer reads it as a regular user and password so I have a 401 error. 

There are 2 possible fix for this:
1. To handle "prompt*" in the validation of promptCredentials()
2. Change the approach of Optional Credentials and create profiles with prompt* as values for user and password. So that whenever you use the profile with Optional Credentials you will be prompted both in Zowe Explorer or Zowe CLI.

jellypuno:
Limitation: prompt* doesn't work in secure credential store 😞

zFernand0:
I added an issue in the secure credential store plugin so that we don't forget about this. Will reference it once the plugin is contributed to Zowe 👍

MikeBauerCA:
The Zowe CLI should be enhanced so that all profile fields are optional and when running a command if an option is not supplied and there is no default option, the user is prompted.

jellypuno:
This issue was handled in v.1.6.0 "Fix failure to load optional secure fields from profiles."

Originally I have an issue with SCS so I wanted to implement this. But since @tjohnsonBCM implemented his change, this is no longer a problem.

Issue #354, opened 2019-11-30
Centralized solution for prompting of credentials


jellypuno:
Today, prompting of credentials are written in several functions like flipState, datasetFilterPrompt, OpenPS, searchPrompt etc. I would like to have a centralized solution for this so that any new function added in Zowe Explorer  will be validated for credentials.

zFernand0:
I'd recommend trying to get this done by v1.1 since this could result in quite a bit of technical debt. : )

jellypuno:
The change for this issue was implemented on v1.6.0

PR #447, opened 2020-01-10
Centralizing  Re-Prompting Logic

sladyn98:
Basically we want to re-prompt for credentials on these conditions:

a) Invalid Credentials
b) Human Error in entering Credentials.

Fixes : https://github.com/zowe/vscode-extension-for-zowe/issues/370

### DIFF
--- a/src/Profiles.ts
+++ b/src/Profiles.ts
@@ -18,6 +18,7 @@ import { URL } from "url";
 import * as vscode from "vscode";
 import * as zowe from "@brightside/core";
 import * as ProfileLoader from "./ProfileLoader";
+import { ECONNREFUSED, ECONNABORTED } from "constants";
 const localize = nls.config({ messageFormat: nls.MessageFormat.file })();
 interface IUrlValidator {
     valid: boolean;
@@ -249,6 +250,14 @@ export class Profiles { // Processing stops if there are no profiles detected
             newProfile = await this.saveProfile(IConnection, IConnection.name, "zosmf");
         } catch (error) {
             vscode.window.showErrorMessage(error.message);
+            if(error === ECONNREFUSED || error === ECONNABORTED) {
+            const reEnterCredentials = "Re-Enter Credentials";
+            vscode.window.showInformationMessage("Re-Enter", reEnterCredentials)
+            .then((selection) => {
+            if (selection === reEnterCredentials) {
+               this.rempromptProfile();
+            }});
+        }
         }
         await zowe.ZosmfSession.createBasicZosmfSession(newProfile);
         vscode.window.showInformationMessage("Profile " + profileName + " was created.");
@@ -305,7 +314,7 @@ export class Profiles { // Processing stops if there are no profiles detected
         return [updSession.ISession.user, updSession.ISession.password, updSession.ISession.base64EncodedAuth];
     }
 
-    private async rempromptUsername() {
+    private async rempromptProfile() {
         let userName: string;
         let passWord: string;
         let options: vscode.InputBoxOptions;
@@ -315,12 +324,6 @@ export class Profiles { // Processing stops if there are no profiles detected
             value: userName
         };
         userName = await vscode.window.showInputBox(options);
-        return userName;
-    }
-
-    private async rempromptPassword() {
-        let passWord: string;
-        let options: vscode.InputBoxOptions;
         options = {
             placeHolder: localize("promptcredentials.option.prompt.password.placeholder", "Password"),
             prompt: localize("promptcredentials.option.prompt.password", "Enter a password for the connection"),
@@ -328,7 +331,6 @@ export class Profiles { // Processing stops if there are no profiles detected
             value: passWord
         };
         passWord = await vscode.window.showInputBox(options);
-        return passWord;
     }
 
     private async saveProfile(ProfileInfo, ProfileName, ProfileType) {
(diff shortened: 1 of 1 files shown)


sladyn98:
CC @jellypuno @zFernand0

src/Profiles.ts
```diff
@@ -305,6 +305,32 @@ export class Profiles { // Processing stops if there are no profiles detected
         return [updSession.ISession.user, updSession.ISession.password, updSession.ISession.base64EncodedAuth];
     }
 
+    private async rempromptUsername() {
```
sladyn98: Does having two separate functions for re prompting make life easier because then the logic can be centralized and reused if even one of the two are wrong.

src/Profiles.ts
```diff
@@ -305,6 +305,32 @@ export class Profiles { // Processing stops if there are no profiles detected
         return [updSession.ISession.user, updSession.ISession.password, updSession.ISession.base64EncodedAuth];
```
sladyn98: Also is where is the invalid error being returned :/ in case of invalid credential entry

jellypuno:
hmmm. this is a tricky thing to explain. But the idea is this:

If I create a profile, I would input the url, username, password and reject-unathorize. If I made a mistake, I want to correct it hence I wanted an edit profile button. But @MikeBauerCA is suggesting that if you experience a 401, there will be an automatic call for that function.

Now, the profiles are called and processed in every function like datasetFilterPrompt,OpenPS etc. Some functions do not have this and I wanted to revise it and make it a centralize. Basically, I want to re-factor the whole process. 

So for re-prompting of credentials, I would suggest that you create an edit button for manual editing and then call that function if 401 is experience. I would like to reference several issues that is possibly linked to this: #320 #354 #373

src/Profiles.ts
jellypuno: Invalid error can be returned anywhere. #388 is an open issue that consolidates this error message. I am partially working on this in handle-invalid-credentials branch

src/Profiles.ts
jellypuno: I think it would be easier to re-prompt the whole profile.

sladyn98:
> So for re-prompting of credentials, I would suggest that you create an edit button for manual editing and then call that function if 401 is experience.

Okay I kind of understood what is the problem and how we are trying to solve it, I will give it a shot.

src/Profiles.ts
```diff
@@ -315,20 +324,13 @@ export class Profiles { // Processing stops if there are no profiles detected
...
-        let passWord: string;
-        let options: vscode.InputBoxOptions;
         options = {
             placeHolder: localize("promptcredentials.option.prompt.password.placeholder", "Password"),
             prompt: localize("promptcredentials.option.prompt.password", "Enter a password for the connection"),
             password: true,
             value: passWord
         };
         passWord = await vscode.window.showInputBox(options);
-        return passWord;
```
sladyn98: I was unsure about calling the `createBasicZosmfSession` because it would be code-duplication and because I am unsure if we want to just store the credentials in the profile without creating the session.

jellypuno: You will probably need this when updating a profile:

await new CliProfileManager({
        ...
    }).update

sladyn98: Yeah updated the code.

sladyn98:
> I'm afraid this is not going to be an easy change. Have you tested this?

It definitely isn't going to be easy, but this would teach me a lot about zowe and typscript in general, if I can get over the hurdles.
No I haven't tested it , a normal `Ctrl + F5`  doesn't quite run the extension.Am I missing something.

jellypuno:
> It definitely isn't going to be easy, but this would teach me a lot about zowe and typscript in general, if I can get over the hurdles.
> No I haven't tested it , a normal `Ctrl + F5` doesn't quite run the extension.Am I missing something.

I understand. To test this, go to Debug and Run Tab, In the drop-down selection (on top) click extension. It will run the extension.

Capture

jellypuno:
Also, just take note of these things:
- DCO signoff. Do not forget to sign-off your commits.
- Run `npm run test` and make sure that your code has coverage. You can check the coverage in here: `results\coverage\lcov-report\index.html`
- Run Integration Test

sladyn98:
```
Activating extension 'Zowe.vscode-extension-for-zowe' failed: Cannot find module '/home/sladyn/OMP-20/zowe/vscode-extension-for-zowe/out/src/extension'
Require stack:
- /usr/share/code/resources/app/out/vs/loader.js
- /usr/share/code/resources/app/out/bootstrap-amd.js
- /usr/share/code/resources/app/out/bootstrap-fork.js.
```
@jellypuno

jellypuno:
> ```
> Activating extension 'Zowe.vscode-extension-for-zowe' failed: Cannot find module '/home/sladyn/OMP-20/zowe/vscode-extension-for-zowe/out/src/extension'
> Require stack:
> - /usr/share/code/resources/app/out/vs/loader.js
> - /usr/share/code/resources/app/out/bootstrap-amd.js
> - /usr/share/code/resources/app/out/bootstrap-fork.js.
> ```
> 
> @jellypuno

https://github.com/zowe/vscode-extension-for-zowe/blob/master/docs/README.md
https://github.com/zowe/vscode-extension-for-zowe/blob/master/docs/README-Theia.md

zFernand0:
@sladyn98, don't forget to signoff your commits

PR #448, opened 2020-01-13
Tso commands

Colin-Stone:
In reality this isn't a TSO command. It's an MVS command that may be able to do TSO via the zosmf console. I've altered most of the references to reflect this.

New function now allows 10 previous commands to be stored. 
New settings section stores commands
User requested they be able to edit before submitting a command so addede as another setting parameter.
Unit/Integration tests


Colin-Stone:
Renamed maost of references to "TSO" as this is should be described either as Mvs or "Issue Console Command". Left the zowe.issueTsoCmd" as I think this would represent a high level version number change

Colin-Stone:
Thanks for reviewing @zFernand0. I won't merge just yet as I would like see if my colleague who is trialing it comes back with any suggestions

Issue #220, opened 2019-09-03
Jobs root folder not expanding automatically when search run [Consistency]


Colin-Stone:
The Job view is inconsistent with the other two Zowe views because it does not automatically expand when it has criteria applied to it. Either as owner/prefix or by manually applying search criteria. The user instead has to manually activate the twistie to see the results.

Colin-Stone:
Seen this issue before. VSCode not refreshing root tree nodes because the label stays the same. Label is analgous to id btw. SO fix is to kid it into thinking it has changed to a different object. 

Jobs now working the same as USS/Datasets

zFernand0:
Thanks Colin! 👍

PR #450, opened 2020-01-14
Add Label hack

Colin-Stone:
Added the rather horrible labelHack to the Jobs tree so that the session node will have the label subtley changed and therefore vscode decides to refresh it. Without this VSCode thinks it's the same object that hasn't changed so would only refresh if manually pressing the expand button.


zFernand0:
This seems to fix #220 😄

PR #462, opened 2020-01-21
Emulate Recent File Picker (NOT including USS files)

katelynienaber:
Re-opening pull request from new branch off zowe/vscode (instead of my own fork).


src/PersistentFilters.ts
```diff
@@ -69,20 +70,24 @@ export class PersistentFilters {
      *
      * @param {string} criteria - a line of search criteria
      */
-    public async addHistory(criteria: string) {
+    public async addHistory(criteria: string, isMember?: boolean) {
```
Colin-Stone: addHistory is Search history. I don't think we should be confusing it with edit recall. Is the intention for the "edit recall" to persist?. I would prefer to see a new section in settings called something llike "recall" if it is.

katelynienaber: Yes I think it should persist because in the vscode recent file picker (CTRL+R) the list persists too. So I will add this.

Also added a fix for empty selection case which you mentioned above.

Thanks for your time reviewing this!

katelynienaber:
Took me a minute to get to this. No more conflicts now :)

Issue #486, opened 2020-02-01
Computing new filters when creating data sets does not cover important cases


phaumer:
When you create a new data set the `createFile()` method checks if the new data set actually matches the current filter so that it would actually be shown to the user after creation. I see several issue with the current test and update method.

1. The test only checks for `.*` wildcards, but just providing a HLC is also a valid filter, which is not covered and hence redundant entries will be created
2. The code that parses the current filter assume spaces between each filter, which also not required. It fails when users specify the comma separated list without spaces
3. The for loop that checks all the current filters does not work as the last entry's test in the filter array will overwrite the previous findings.
4. The newly created filter adds the entire data set name, which will lead to a long list if multiple data sets are created.  It might be more efficient if the filter would use a wildcard.  The downside could be that suddenly many more data sets previously hidden would be displayed, but I wanted to propose this  solution to see what people think. A better solution would use perhaps a user preferences or would prompt the user if suddenly a certain additional number of data sets would be displayed. An item for future exploration.

jelaplan:
Thx Peter for catching all these details.

zFernand0:
resolved by #487

PR #487, opened 2020-02-01
Refined filter generation in to cover more cases

phaumer:

This is a proposal for Issue https://github.com/zowe/vscode-extension-for-zowe/issues/486 for computing the new filter when creating data sets.


phaumer:
> This works much better! I missed several cases.
> 
> One question, on lines 855/856 why not add the full name of the new DS to the filter list? When you replace the last segment of the name with a star, it could filter for more than just the newly-added DS.
> 
> For example if no filter is applied and I add USER.NEW.PDS.TEST1, but DS with name USER.NEW.PDS.TEST2 also exists, then both will show up in the list after the filter is added.

Yes, agreed @katelynienaber. That is what I mentioned under topic (4) in the issue. If I create 5 datasets for an application the filter gets impossibly long and looks super redundant, so I prefer this solution with the use cases I had in mind.  But I agree this should probably be a user preference or something that the user has more control over. So we should discuss this.

PR #488, opened 2020-02-03
v1.2 Changelog and package updates. This will publish the release when merged

Colin-Stone:
Will need updating with any additional function we add this week
@zFernand0 Readyfor when you want to say "Go" :-)


zFernand0:
The failure on the tests is because Jest doesn't recognize that the snapshot file already exists.
[`__test__/__unit__/__snapshots__/ZoweUSSNode.unit.test.ts.snap`](https://github.com/zowe/vscode-extension-for-zowe/blob/master/__tests__/__unit__/__snapshots__/ZoweUSSNode.unit.test.ts.snap)
```
New snapshot was not written. The update flag must be explicitly passed to write a new snapshot.

This is likely because this test is run in a continuous integration (CI) environment in which snapshots are not written by default.
```

Ideas:
- Delete it from source, and add it on a subsequent commit?
- Rename it to `something` and back to `ZoweUSSNode` on separate commits?

zFernand0:
Thanks to everyone who made this possible 👍 

Zowe Explorer `v1.2.0`
🥂 🍰

PR #492, opened 2020-02-04
Fix issue 485

### LINKED ISSUE #485: Creating data sets causes tree to loose expand behavior
- Have a connection and filter that shows at least two existing PDS with members
- Expand the first and you see the members listed.
- Create a new data set such as a sequential one.
- Expand the second PDS

--> Nothing happens. The members are not shown. You have to restart VS Code.

katelynienaber:
Fix #485 

Added public variable to Dataset Tree, USS Tree, and Job Tree so that one vscode Tree View can be accessed & controlled for each tree.


Colin-Stone:
There are 2 integration test failures around "Cannot read property 'reveal' of undefined"

katelynienaber:
> I don't really like accessing fields in another class like this and would rather it was a getTreeView() method on the TreeProvider class. In this case I think we could pass the tree view ID in the static TreeProvider create...Tree(log, viewId) call and let it instantiate the tree view and make it available via getTreeView.
> 
> I intend to push that down as a common function when I refactor for interfaces

Okay I can add these changes tomorrow. I am already home tonight.

Colin-Stone:
Didn't eman toapprove just agree changes you made. But you have two integration test failures
  2 failing
base.js:383
  1) Extension Integration Tests - USS
       Enter USS Pattern
         should output path that match the user-provided path

base.js:266
  2) TreeView
       should create the TreeView:

katelynienaber:
> All seems to work for me thanks for this change

Ya thanks for your help :)

zFernand0:
Thanks for fixing! 👍

Issue #74, opened 2019-05-08
Bundle VSCode extension with webpack


ChrisBoehmCA:
Microsoft recommends bundling VSCode extensions before publishing 

https://code.visualstudio.com/api/working-with-extensions/bundling-extension

Colin-Stone:
Should we link with https://github.com/zowe/vscode-extension-for-zowe/issues/34

Colin-Stone:
@zFernand0 I think this means altering the packaging to incorporate webpack but is this something that we would also need to do some updates to the CI/CD process too?.

zFernand0:
> @zFernand0 I think this means altering the packaging to incorporate webpack but is this something that we would also need to do some updates to the CI/CD process too?.

Hey @Colin-Stone,
I believe it'd be good if we could incorporate webpacking into the publishing step.

Colin-Stone:
Maybe do in conjunction with https://github.com/zowe/vscode-extension-for-zowe/issues/5

lauren-li:
I have been working with my teammate to try integrating webpack into Microsoft's basic [i18n sample](https://github.com/microsoft/vscode-extension-samples/tree/master/i18n-sample), just to be sure we can get the implementation right for a more basic set of files before bringing it into the Zowe VS Code extension. The i18n sample by itself runs as expected. I've created a repo for work on bringing webpack into the i18n sample: https://github.com/lauren-li/vscode-extension-samples/tree/add-webpack/i18n-sample

We are currently using vscode-nls-dev's [webpack-loader](https://github.com/microsoft/vscode-extension-samples/tree/master/webpack-sample#more) in our webpack config, but we are also modifying the gulp file to pull in the webpack config as part of the build task. (VSCode-nls is our localization library.)

With webpack, the resulting file structure looks like it is getting close to what it should be, and the command titles are getting localized, but I am getting errors when running the commands themselves:

Localized command titles:
<img width="599" alt="Screen Shot 2019-11-14 at 10 41 53 AM" src="https://user-images.githubusercontent.com/45975633/68899792-6c80db80-0700-11ea-9a19-6ca2b3ef01ff.png">

Errors when running commands:
<img width="467" alt="Screen Shot 2019-11-14 at 10 25 29 AM" src="https://user-images.githubusercontent.com/45975633/68899701-39d6e300-0700-11ea-9ff5-7503d56cff38.png">

Resulting file structure with webpack*:
[...]

PR #493, opened 2020-02-04
Webpack with localization

lauren-li:
Work in progress. This branch has webpack working with localization, but integration tests need some additional work. Addresses issue #74.
- You can check wepback by running `npm run package`, renaming the resulting .vsix file to a .zip, and checking this .zip folder's contents.
  - Under the **dist** folder, there should be a single **extension.js** file and **extension.js.map** file (as opposed to .js files named after source code files). There should be several .json files as well, including .json files in a **src** folder that mirror the soure code's **src** folder hierarchy, and these are used for localization.
- You can check the localization in this branch by either installing the extension from the .vsix file built from this branch, or by just pulling the branch and running it with the Extension Development Host.
  - Localization can be tested by switching the [VS Code display language](https://code.visualstudio.com/docs/getstarted/locales) to French. The following text of the extension should be in French:
    - Hover text for the "Create New Data Set" icon. (Tests package.json i18n.)
    - QuickPick text that appears when deleting a data set. (Tests i18n for TypeScript files.)

Some notes about the files changed:
- An **i18n/fra** folder full of French language files was added as part of this PR so that localization functionality can be tested with webpack. Changed files beginning with **i18n/fra** do not need to be code-reviewed.
- Changes in **src/\*\*/*.ts** files have to do with adding and/or updating the localization config/call at the top of the file. No changes in actual code logic were made.


Colin-Stone:
get an error trying to run the integration tests Error: Cannot find module './items/document' Is there something else that's needed for the build to produce these. The extension itself appears to run okay

lauren-li:
@Colin-Stone Thanks for pointing this out! I also ran into the same issue starting yesterday, and have pushed a commit to address it.

lauren-li:
I removed the majority of the node_modules folder from vsix generation, as webpack should already include necessary node modules in the extension.js file. The only one I left in was node_modules/vscode-nls/**, since the extension (installed from the generated vsix file) does not seem to function properly without it.

**Comparison of generated Zowe Explorer vsix file sizes:**

|vsix file generation source|Resulting vsix file size|
|-------------------------------------------------------|--------|
|Master branch                                                                  |51.6 MB|
|webpack-separate-dist, including all node_modules.    | 10.8 MB|
|webpack-separate-dist, vscode-nls node_module only | 3.8 MB |

tjohnsonBCM:
Attempted unsuccessfully to get the integration tests working with webpack in this branch. Have summarized my findings below.

Since the integration tests depend on the extension's source code (e.g., `src/Profiles.ts`), they need to load those modules from the same place as the extension. When trying to run the integration tests against the webpacked extension, they fail because they are loading the source code from a different location that isn't webpacked (`out/*.js` instead of `dist/extension.js`).

I believe we have two options: webpack both the extension and the integration tests, or don't webpack either of them. The first option seems ideal, it would be nice to run tests with the webpacked extension to simulate the published VSIX as closely as possible.

I looked into using the NPM package [mochapack](https://www.npmjs.com/package/mochapack) to make Mocha and Webpack work together. However, I couldn't find a way to make the VS Code extension host launch webpacked tests, as it seems to require that the tests be supplied through `extensionTestsPath`.

Unless we can find a way to run the integration tests webpacked, it seems the extension may need to be built without webpack in the NPM `pretest` script. This approach seems recommended in [this GitHub comment from a Microsoft employee](https://github.com/microsoft/vscode-docs/issues/2254#issuecomment-469752205), which states that if tests for a VS Code extension depend directly on source code, they shouldn't be webpacked.

lauren-li:
Closing this PR in favor of #860.

PR #512, opened 2020-02-11
501 migrate data sets

### LINKED ISSUE #501: Allow data sets to be migrated

Pranay154:
Resolves #501, added functionality for data sets to be migrated


Pranay154:
> "I'm pressing the context menu and nothing seems to happen". does this mean that's nothing is happening or that we have just fired off the request. We need either a wrapper if it's a log running synchronous job or if asynch we need a showInformationMessage to indicate that the request has been sent

Hey @Colin-Stone, Ive added in a message to say that the for the chosen data set, a migration request is sent.

Colin-Stone:
Sorry to be such a pedant but could we have the following message. `Migration of dataset: MyHLQ.nn.nn requested`

Pranay154:
> Sorry to be such a pedant but could we have the following message. `Migration of dataset: MyHLQ.nn.nn requested.`
> When you come to the recall it would then be `Recall of dataset: MyHLQ.nn.nn requested.`

Hey @Colin-Stone, made this change 😄

Colin-Stone:
@Pranay154 Try and avoid publishing package-lock unless absolutely necessary. The conflict is due to the switch to `@zowe`. I think your best solution now is to commit an exact copy of the master/package-lock file into your branch

Issue #514, opened 2020-02-12
Zowe 1.2.1 does not activate when ~/.zowe directory is not present


phaumer:
In Zowe Explorer v1.1.0 we had a ProfileLoader class that would actually spawn a node process to run the imperative main from node_modules to basically run Zowe CLI to create the .zowe directory in the user’s home directory and if already present read the profiles already available.  This was useful for users who do not have Zowe CLI installed as they could just use Zowe Explorer without installing the CLI. 

I see that this code got removed in this PR https://github.com/zowe/vscode-extension-for-zowe/pull/400. So now when I install Zowe Explorer without having a ~/.zowe directory activation of Zowe Explorer even fails. The only workaround it seems is installing Zowe CLI and calling any command.

![image](https://user-images.githubusercontent.com/4391934/74381171-d2344080-4d9f-11ea-80c4-0c873110846e.png)

phaumer:
I created this PR as a quick work-around using the previous technique of initializing the .zowe directory. I am hoping there is a different way. See the commented out lines of code in the PR, which was the other way I tried this unsucessfully.

Colin-Stone:
The spawn function was originally added to profile management because of issues loading a profile that used a security mechanism such as keytar. It required that the original cli node app to load the keys correctly. With the keytar implementation complete the spawn function was considered redundant and hence all mechanisms around it removed.
It sounds that this function served a secondary purpose of instantiating the .zowe folder if CLI was not already installed. 
I'm not sure we need to spawn a process to achieve this but need to reinstate code in some way to satisfy the issue at least temporarily

tjohnsonBCM:
@phaumer Sorry, I didn't understand exactly what the issue was when initially discussing this on Slack.

It appears that the extension is only expecting the `.zowe` folder to exist in order to access `~/.zowe/settings/imperative.json`. I believe the extension is already handling the case that the `~/.zowe/profiles/zosmf` folder may not exist when creating a profile.

We should be able to fix this by setting `imperativeIsSecure` to false if the `settings.json` file doesn't exist.

tjohnsonBCM:
Previous comment is incorrect, the `~/.zowe/profiles/zosmf` folder is also an issue. Discussed this with @zFernand0 and we should be able to make a patch in @brightside/core to allow a more ideal fix.

tjohnsonBCM:
[...]

PR #515, opened 2020-02-13
Added spawn mechanism back to create .zowe if missing

phaumer:

This is a quick hack for https://github.com/zowe/vscode-extension-for-zowe/issues/514
Would like to learn how to do this properly.  Would imperative have to be extended? I see a challenge that this will not work once we add webpack support, unless we explicitly exclude the file ImperativeInit.js from the webpack bundle. (@lauren-li )

### DIFF
--- a/__tests__/__unit__/extension.unit.test.ts
+++ b/__tests__/__unit__/extension.unit.test.ts
@@ -21,6 +21,7 @@ import * as brightside from "@brightside/core";
 import * as os from "os";
 import * as fs from "fs";
 import * as fsextra from "fs-extra";
+import * as child_process from "child_process";
 import * as profileLoader from "../../src/Profiles";
 import * as ussNodeActions from "../../src/uss/ussNodeActions";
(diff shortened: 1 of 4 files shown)


Colin-Stone:
Pushed a more appropriate location for the initialization. Prior to your added code block there is the check for the security modules. It checks for the imperative settings file to be present and if not throws an error. I have now included your spawnSync code here. 
This seems to be the logical location since the security step is always performed and if an error is thrown indicating that imperative.json is not present and consequently re-initalized that will be fine.
I left your original code in commented out. If this all works fine in your scenarios please delete and re commit

phaumer:
Thanks. That is a much better place and I like that it is now called only on failure, which should be speeding up startup for n+1.

zFernand0:
Moving this PR to the v1.2.x branch :)

zFernand0:
Please see #517

zFernand0:
We can delete the `call-imperative-init` branch after merging #517

PR #516, opened 2020-02-13
Don't require ~/.zowe folder to exist

### LINKED ISSUE #514: Zowe 1.2.1 does not activate when ~/.zowe directory is not present

zFernand0:
Signed-off-by: Timothy Johnson <timothy.johnson@broadcom.com>
(cherry picked from commit 3392940dcd4b43837ac87e6502d3a4491ad8ee4b)

Fixes #514

### DIFF
--- a/package.json
+++ b/package.json
@@ -1211,7 +1211,7 @@
     "vscode-nls-dev": "^3.2.6"
   },
   "dependencies": {
-    "@brightside/core": "^2.36.0",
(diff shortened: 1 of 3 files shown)


zFernand0:
We will go out with Peter's fix (#517) since there are still failing tests in this PR

zFernand0:
Closing in favor of #517

You can find the fix without spawning a process in #518

Issue #489, opened 2020-02-03
Intermittent Jest failure because of missing snapshot


zFernand0:
The failure on the tests is because Jest doesn't recognize that the snapshot file already exists.
[`__test__/__unit__/__snapshots__/ZoweUSSNode.unit.test.ts.snap`](https://github.com/zowe/vscode-extension-for-zowe/blob/master/__tests__/__unit__/__snapshots__/ZoweUSSNode.unit.test.ts.snap)
```
New snapshot was not written. The update flag must be explicitly passed to write a new snapshot.

This is likely because this test is run in a continuous integration (CI) environment in which snapshots are not written by default.
```

Ideas:
- Delete it from source, and add it on a subsequent commit?
- Rename it to `something` and back to `ZoweUSSNode` on separate commits?

_Originally posted by @zFernand0 in https://github.com/zowe/vscode-extension-for-zowe/pull/488#issuecomment-581546861_

zFernand0:
See https://github.com/zowe/vscode-extension-for-zowe/pull/607#issue-386264166 for more deatils

PR #517, opened 2020-02-14
Ph fix 514 spawn

### LINKED ISSUE #514: Zowe 1.2.1 does not activate when ~/.zowe directory is not present

zFernand0:
fixes #514

### DIFF
--- a/__tests__/__unit__/extension.unit.test.ts
+++ b/__tests__/__unit__/extension.unit.test.ts
@@ -21,6 +21,7 @@ import * as brightside from "@brightside/core";
 import * as os from "os";
 import * as fs from "fs";
 import * as fsextra from "fs-extra";
+import * as child_process from "child_process";
 import * as profileLoader from "../../src/Profiles";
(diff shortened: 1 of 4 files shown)


zFernand0:
Replaying build because of #489 again

zFernand0:
Personally, I'm not concerned about Codecov failing here since this may be just a temporary fix.

zFernand0:
> This looks fine. Are we treating as a temporary fix?. Replacing with Tim's more comprehensive solution?

That's correct. We want to fix this as soon as we possibly can, in case new users download the extension over the weekend and they don't have the CLI 😄

tjohnsonBCM:
Tested this branch without a `~/.zowe` folder present and got an error, although Zowe Explorer seemed to work fine afterwards.
image

zFernand0:
We believe that the worst case scenario is that people may not be able to store secure profiles over the weekend (until the next version - 1.3) comes out on Tuesday

PR #518, opened 2020-02-14
Don't require ~/.zowe folder to exist

### LINKED ISSUE #514: Zowe 1.2.1 does not activate when ~/.zowe directory is not present

zFernand0:
Signed-off-by: Timothy Johnson <timothy.johnson@broadcom.com>

Fix #514 Without spawning a process

### DIFF
--- a/__tests__/__unit__/Profiles.unit.test.ts
+++ b/__tests__/__unit__/Profiles.unit.test.ts
@@ -55,17 +55,12 @@ describe("Profile class unit tests", () => {
     const createInputBox = jest.fn();
     const showQuickPick = jest.fn();
     const showErrorMessage = jest.fn();
-//    const cliHome = jest.fn().mockReturnValue(homedir);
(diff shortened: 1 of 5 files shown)


Colin-Stone:
Concidering the build issue, is this awaiting the change to `@zowe`

tjohnsonBCM:
Some of the mock definitions in `extension.unit.test.ts` seemed to be duplicate so I removed them. Now the coverage seems to have gone down significantly even though I didn't remove any tests. Should I restore these mock definitions?

Colin-Stone:
> Some of the mock definitions in `extension.unit.test.ts` seemed to be duplicate so I removed them. Now the coverage seems to have gone down significantly even though I didn't remove any tests. Should I restore these mock definitions?

I think it's a false negative. The actual report from the unit tests looks the same

zFernand0:
Thanks for fixing the tests @tjohnsonBCM

PR #519, opened 2020-02-14
change rename, copy and paste labels for data sets

### LINKED ISSUE #478: UX question - why copy member action is called "Copy Data Set"?
I find it very confusing to right-click a PDS member and have "Copy Data Set" action item.
Could it be changed to "Copy Member"?

CForrest97:
resolves #478


CForrest97:
@Colin-Stone yes this resolves #478

PR #525, opened 2020-02-18
Fix issue 161

jellypuno:
@Colin-Stone @zFernand0 Do you know how I could add coverage for refreshAllJobs?


jellypuno:
> @Colin-Stone @zFernand0 Do you know how I could add coverage for refreshAllJobs?

Nevermind. I just transferred it to a different function.

jellypuno:
> I think if we can rather than add more to extension.ts it would probably serve us better in the long run if we use the nodeActions file to put these classes rather than making extension.ts even bigger.

uhmmmm. Should I create a separate jobsNodeActions.ts and transfer the RefreshAllJobs there? or do you see this as a separate issue that will be handled in the future?

> Oh and we definitely need to make sure we have integration tests as they tend to test around the gaps

:sob: okay...

Colin-Stone:
Yes please add it to the actions file for now. I've spent two days trying to decouple the rename functionality from extension.ts but it has links all over the place. With the unit tests if you can modularize them so they are in a single describe section on there own would help when we put the code into a class

Colin-Stone:
Withv1.4 we would now write this function 
`vscode.commands.registerCommand("zowe.refreshAll", () => datasetProvider.refreshAll());`

datasetProvider will be of type `IZoweTree<IZoweDatasetTreeNode>;`
and refreshAll will be a generic method on `IZoweTree<T>`

Colin-Stone:
:-( 
Unable to run integration tests: TypeError: Cannot read property 'loadNamedProfile' of undefined

jellypuno:
> Withv1.4 we would now write this function
> `vscode.commands.registerCommand("zowe.refreshAll", () => datasetProvider.refreshAll());`
> 
> datasetProvider will be of type `IZoweTree<IZoweDatasetTreeNode>;`
> and refreshAll will be a generic method on `IZoweTree<T>`

I'm confused because in your previous comment you said to add it in Actions file? Am I misunderstanding it?

jellypuno:
> :-(
> Unable to run integration tests: TypeError: Cannot read property 'loadNamedProfile' of undefined

I am not experiencing this error. Is it under "It should call the RefreshALL function"?

Colin-Stone:
> I'm confused because in your previous comment you said to add it in Actions file? Am I misunderstanding it?

Sorry for causing confusion. You are doing exactly what I asked. I was suggesting that once v1.4 is available we can begin rewriting these functions as class methods

Colin-Stone:
> I am not experiencing this error. Is it under "It should call the RefreshALL function"?

No the integration tests just don't get started so it sounds like an issue with my set up then

jellypuno:
> > I'm confused because in your previous comment you said to add it in Actions file? Am I misunderstanding it?
> 
> Sorry for causing confusion. You are doing exactly what I asked. I was suggesting that once v1.4 is available we can begin rewriting these functions as class methods

aha! ok 😄 noted! 👍

Colin-Stone:
@lauren-li Are you able to review whilst @phaumer is unavailable?

src/utils.ts
```diff
@@ -184,9 +186,32 @@ export function errorHandling(errorDetails: any, label?: string, moreInfo?: stri
...
+        if (sessNode.getProfileName() === profNode.name) {
+            sessNode.getProfile().profile = profNode.profile;
+            const SessionProfile = profNode.profile as ISession;
+            if (sessNode.getSession().ISession !== SessionProfile) {
+                sessNode.getSession().ISession.user = SessionProfile.user;
+                sessNode.getSession().ISession.password = SessionProfile.password;
+                sessNode.getSession().ISession.base64EncodedAuth = SessionProfile.base64EncodedAuth;
+                sessNode.getSession().ISession.hostname = SessionProfile.hostname;
+                sessNode.getSession().ISession.port = SessionProfile.port;
+                sessNode.getSession().ISession.rejectUnauthorized = SessionProfile.rejectUnauthorized;
```
zFernand0: Maybe we should have a setter for this?
I know it won't have much of an performance impact, but theoretically this could reduce the amount of session calls to one :smile:

@jellypuno, @Colin-Stone, any thoughts?

No change requested!

__tests__/__integration__/dataset/dsNodeActions.integration.test.ts
```diff
@@ -0,0 +1,86 @@
...
+import * as chaiAsPromised from "chai-as-promised";
+import * as sinon from "sinon";
+import * as testConst from "../../../resources/testProfileData";
+import * as vscode from "vscode";
+import * as extension from "../../../src/extension";
+import { ZoweDatasetNode } from "../../../src/ZoweDatasetNode";
+import { DatasetTree } from "../../../src/DatasetTree";
+import * as dsActions from "../../../src/dataset/dsNodeActions";
+
+const TIMEOUT = 45000;
```
zFernand0: I think this should be exposed in some sort of test-Configuration constants file.
However I don't think adding more stuff to testProfileData is the correct way to go since it already contains things not related to someone's profile (e.g. patterns).

No change requested, just something to think about while refactoring things. :smile:

src/utils.ts
Colin-Stone: I would vote for a setter

PR #531, opened 2020-02-19
Add KeytarCredentialManager class modelled after SCS plugin

tjohnsonBCM:
This PR adds support for profiles created with any version of the SCS plugin, without needing to have a credential key configured in VSCode settings.

The `KeytarCredentialManager` checks for credentials stored securely under any of the following IDs which have been used over time:

* `Zowe-Plugin`
* `@brightside/core`
* `@zowe/cli`
* `Broadcom-Plugin`

Much of the added code was taken from @zFernand0 's credential manager implementation in the SCS plugin. 😄 

When this PR is merged, a 1.2.3 patch release of Zowe Explorer should go out.

### DIFF
--- a/__mocks__/@brightside/imperative.ts
+++ b/__mocks__/@brightside/imperative.ts
@@ -93,6 +93,9 @@ export class DefaultCredentialManager {
      public test: "test";
 }
 
+// tslint:disable-next-line:max-classes-per-file
+export class AbstractCredentialManager {
+}
 
 // tslint:disable-next-line:max-classes-per-file
 export class Logger {
--- a/__tests__/__unit__/KeytarCredentialManager.unit.test.ts
+++ b/__tests__/__unit__/KeytarCredentialManager.unit.test.ts
@@ -9,7 +9,7 @@
 *                                                                                 *
 */
 
-import { KeytarCredentialManager } from "../../src/KeytarCredentialManager";
+const KeytarCredentialManager = require("../../src/KeytarCredentialManager");
 
 describe("KeytarCredentialManager Unit Tests", () => {
     // Use a fake insecure credential store
@@ -20,7 +20,7 @@ describe("KeytarCredentialManager Unit Tests", () => {
         "Broadcom-Plugin_user4": "froyo"
     };
 
-    const credentialMgr = new KeytarCredentialManager("Awesome-Service", "");
+    let credentialMgr: any;
 
     // Mock the Keytar module
     KeytarCredentialManager.keytar = {
@@ -36,6 +36,11 @@ describe("KeytarCredentialManager Unit Tests", () => {
             delete credentialStore[`${service}_${account}`];
         })
     };
+
+    beforeAll(async () => {
+        credentialMgr = new KeytarCredentialManager("Awesome-Service", "");
+        await credentialMgr.initialize();
+    });
 
     afterEach(() => {
         jest.resetAllMocks();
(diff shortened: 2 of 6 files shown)


Colin-Stone:
Circular dependency in unit tests, so using require for KeytarCredentialManager instead of import. Have not started specific manual testing yet however

Colin-Stone:
Working so far with 1. no security, installing 2. SCS v4 and then the 3. default keytar only solution. The keytar only for Zowe Explorer solution is still a little clunky, primarily because it uses the `"credential-manager"` key rather than the `"CredentialManager"` key that SCS v4 uses. This would need to be documented when used stand alone.

Where it gets a bit foggy for me..
Installed a v1.3 SCS tgz that I must have had for testing a while back but this didn't work within the CLI storing passwords in the clear. One thing that may be rlevant was that installing this plugin    updated the imperative.json `"credential-manager"` key rather than the expected `"CredentialManager"` key.

I could benefit from some pointers. What are the pre-existing versions that need to be supported?. Am I able to access these?.

tjohnsonBCM:
> The keytar only for Zowe Explorer solution is still a little clunky, primarily because it uses the `"credential-manager"` key rather than the `"CredentialManager"` key that SCS v4 uses.

Shouldn't Zowe Explorer work with both "CredentialManager" and "credential-manager" because of the lines below?

https://github.com/zowe/vscode-extension-for-zowe/blob/4eb7b9aa6f1624e69a7346a8661876e763eb46a5/src/extension.ts#L381-L384

Colin-Stone:
> > The keytar only for Zowe Explorer solution is still a little clunky, primarily because it uses the `"credential-manager"` key rather than the `"CredentialManager"` key that SCS v4 uses.
> 
> Shouldn't Zowe Explorer work with both "CredentialManager" and "credential-manager" because of the lines below?
> 
> https://github.com/zowe/vscode-extension-for-zowe/blob/4eb7b9aa6f1624e69a7346a8661876e763eb46a5/src/extension.ts#L381-L384

Zowe explorer does work with both but something needs to be added manually so we need to let users know that they need to put something into the credentials-manager key field so keytar gets activated

tjohnsonBCM:
> Installed a v1.3 SCS tgz that I must have had for testing a while back but this didn't work within the CLI storing passwords in the clear. One thing that may be rlevant was that installing this plugin updated the imperative.json `"credential-manager"` key rather than the expected `"CredentialManager"` key.
> 
> I could benefit from some pointers. What are the pre-existing versions that need to be supported?. Am I able to access these?.
> 
> > Update 4. v2.2 working fine :-)

Not sure how many scenarios we need to test, seems you've already verified that the common cases are handled. Also had @MikeBauerCA test this, it worked on his system to load credentials stored with multiple different service IDs.

To test the v1.3 SCS plugin that uses "credential-manager", you'd need to downgrade your version of Zowe CLI to lts-incremental for it to be compatible.

Colin-Stone:
The last sceanrio to consider is if we need to do something for the keytar only solution by way of swapping the credential-manager key string either to CredentialManager or something more obvious. 
The recommended way for anyone with CLI installed would be to use SCS v4+ and both will work in harmony. 
For standalone Zowe Explorer situations do we really need to have the extra indicator in imperative.json or maybe just using the `@zowe/cli` key in settings ?

tjohnsonBCM:
> For standalone Zowe Explorer situations do we really need to have the extra indicator in imperative.json or maybe just using the `@zowe/cli` key in settings ?

Discussed with @zFernand0. I believe the plan is that for now, the extra indicator in imperative.json will always be required. But in a future version (1.4 I think?), we can make Zowe Explorer secure by default, so that it will store credentials securely, even if imperative.json does not exist.

PR #532, opened 2020-02-20
V1.2.3 including readme updates

Colin-Stone:
Revised documentation for Readme links to new docs section for secure credentials. Will tidy this up tomorrow but welcome input etc.

Only the one element for the changelog. Do we want to consider cherry picking the Don't require ~/.zowe folder to exist PR as well ?.

Changelog and package all set ready to publish so not to be merged until approved etc


Colin-Stone:
Aborting, have v1.3 code here too

PR #533, opened 2020-02-20
Updates for release 123

Colin-Stone:
Revised documentation for Readme links to new docs section for secure credentials. Will tidy this up tomorrow but welcome input etc.

Only the one element for the changelog. Do we want to consider cherry picking the Don't require ~/.zowe folder to exist PR as well ?.

Changelog and package all set ready to publish so not to be merged until approved etc


Colin-Stone:
Added an extrenal link for information about the plugin relating to the npm download site. Unsure if I can use this or should use the Readme in Github source but that isn't as user friendly to users as say developers

tjohnsonBCM:
> Only the one element for the changelog. Do we want to consider cherry picking the Don't require ~/.zowe folder to exist PR as well ?.

I think only one item in the changelog is OK since this is a patch release. The other PR will make it into 1.3.

PR #535, opened 2020-02-21
Port #531 to Master

awharn:
Ports the change in PR #531 to master for v1.3.0 of the extension.


Colin-Stone:
Merged prior to everyone reviewing because the docs are referenced on the master branch only meaning that the 1.2.x fix was getting a 404

PR #540, opened 2020-02-24
await needed on CredentialManagerFactory.initialize

Colin-Stone:


tjohnsonBCM:
Should we consider using the TSLint rule "no-floating-promises", to prevent this kind of issue from happening again?

zFernand0:
> This all appears to work in the `@brightside` environment. I assume we don't have to test the `@zowe` (V4) SCS in the `@brightside` environment?

No need 😉

PR #542, opened 2020-02-24
Support "Broadcom-Plugin" as credential key

### LINKED ISSUE #541: Cannot create profiles with Credential Key if Credential Key already in allServices List
When creating a profile with a credential key specified that is already an element of the allServices list, the profile created will have the "Zowe-Plugin" key.

Steps to Reproduce:

1) Set Credential Key setting to "@brightside/core", "@zowe/cli", or "Broadcom-Plugin"
2) Restart VSCode
3) Create Profile as normal
4) Observe Credential Key associated with profile in Credential Manager

The key that appears will be Zowe-Plugin (element zero of the allServices list)

tjohnsonBCM:
Fixes #541 

This makes Zowe Explorer compatible with the lts-incremental version of the Secure Credential Store plugin, which doesn't support the credential key "Zowe-Plugin".

This fix is also present in patch release 1.2.4, as part of #540

### DIFF
--- a/__tests__/__unit__/KeytarCredentialManager.unit.test.ts
+++ b/__tests__/__unit__/KeytarCredentialManager.unit.test.ts
@@ -45,6 +45,10 @@ describe("KeytarCredentialManager Unit Tests", () => {
         })
     };
 
+    afterEach(() => {
+        jest.clearAllMocks();
+    });
+
     it("Test loading passwords from credential store", async () => {
         let secret = await credentialMgr.load("user1");
         expect(KeytarCredentialManager.keytar.getPassword).toHaveBeenLastCalledWith("Zowe-Plugin", "user1");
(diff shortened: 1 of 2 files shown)


awharn:
Thanks, @tjohnsonBCM. Do you think you could add a unit test around the issue raised in #541 ?

tjohnsonBCM:
> Do you think you could add a unit test around the issue raised in #541 ?

It has been done 😄

__tests__/__unit__/KeytarCredentialManager.unit.test.ts
```diff
@@ -99,4 +103,14 @@ describe("KeytarCredentialManager Unit Tests", () => {
         expect(KeytarCredentialManager.keytar.getPassword).toHaveBeenLastCalledWith("Awesome-Service", "user1");
         expect(secret).toBe("cupcake");
     });
+
+    it("Test saving passwords to credential store under previous service name", async () => {
+        credentialMgr = new KeytarCredentialManager("Broadcom-Plugin", "");
```
zFernand0: Could we use a more generic service name, instead of a company specific one?

PR #548, opened 2020-02-26
Additional refactoring

Colin-Stone:
This represents the next refactoring stage which allows us the use generic references for similar methods on each tree class.

In this first commit I have converted the renameNode function to be rename and implemented across the different classes

The next example addZoweSession applies to all tree views and can therefore become discrete methods on each of the trees although mostly inheriting the behavior from their shared super class (not yet committed)


src/DatasetTree.ts
```diff
@@ -63,6 +62,52 @@ export class DatasetTree extends ZoweTreeProvider implements IZoweTree<IZoweData
...
     }
 
+    /**
+     * Rename data sets
+     *
+     * @export
+     * @param {ZoweDatasetNode} node - The node
+     * @param {DatasetTree} datasetProvider - the tree which contains the nodes
+     */
+    public async rename(node: IZoweDatasetTreeNode) {
```
zFernand0: Thanks for keeping this PR at one command.
And for adding the placeholders for the other functions. 👍

Colin-Stone:
Thanks for handling this for me @zFernand0. My subsequent refactoring changes will be incremental with one or two commands addressed each time. Because of the volume of code that gets moved around the changes appear more difficult to review than they are.

Issue #480, opened 2020-01-30
Icons retrieval logic refactoring


stepanzharychevbroadcom:
Currently all of the icons' paths are generated in `utils.ts` file via huge if-else construction which is not extendable at all. I suggest to fully rework it using strategy pattern: so each icon will define it's own object with methods and properties for retrieving it.

Pilot implementation is provided in #461 and connected PR.

In short each Icon is going to have it's own implementation to search for it + some necessary properties. You can check infographics below to understand:
![Icon Generator](https://user-images.githubusercontent.com/60000150/73469574-dbc4ae00-4386-11ea-80c7-52065cdb9073.png)

So the idea is generator will pick an icon for us, and it should be easily extendable.

Colin-Stone:
I haven't yet seen how we can get away from the fact that we have to hardcode contextvalues with specific values in the package.json file. That is the only place where VSCode allows the definition of the context menus and operations available to a business entity. 
So I read the purpose of the generator as providing the mapping of contextvalue to id and type to state which if you are implementing as a strategy pattern would be much cleaner and nicer than a big if-then block. Is that how you envisage this or do we still have different perspectives?.

stepanzharychevbroadcom:
Created PR to resolve this issue: https://github.com/zowe/vscode-extension-for-zowe/pull/549

PR #549, opened 2020-02-26
Issue #480 - Icons Refactoring

stepanzharychevbroadcom:


src/abstract/ZoweTreeProvider.ts
```diff
@@ -80,6 +81,8 @@ export class ZoweTreeProvider {
      * @param isOpen the intended state of the the tree view provider, true or false
      */
     public async flipState(element: IZoweTreeNode, isOpen: boolean = false) {
+        element.collapsibleState = isOpen ? vscode.TreeItemCollapsibleState.Expanded : vscode.TreeItemCollapsibleState.Collapsed;
```
Colin-Stone: Why have you added this?. The collapsible state is controlled by vscode and this is reacting to it being set. 
Although I agree with what you are setting is correct it feels like something that isn't broken (unless there are Theia issues?)

stepanzharychevbroadcom: By setting this property I eliminate necessity to pass this flag to generator of icons, I just want to rely on internal properties of the node. Hope it makes sense.

__tests__/__unit__/DatasetTree.unit.test.ts
```diff
@@ -171,8 +176,14 @@ describe("DatasetTree Unit Tests", () => {
...
-        sessNode[0].iconPath = utils.applyIcons(sessNode[0]);
-        sessNode[1].iconPath = utils.applyIcons(sessNode[1]);
+        let targetIcon = getIconByNode(sessNode[0]);
+        if (targetIcon) {
+            sessNode[0].iconPath = targetIcon.path;
+        }
+        targetIcon = getIconByNode(sessNode[1]);
+        if (targetIcon) {
+            sessNode[1].iconPath = targetIcon.path;
+        }
```
zFernand0: I feel like having a TestEnvironment will be helpful for handling this kinds of things in the future.
It may help to reduce not only volume of test files but also complexity. :smile:

No changes requested!

__tests__/__unit__/extension.unit.test.ts
```diff
@@ -617,9 +618,18 @@ describe("Extension Unit Tests", () => {
...
+            sampleFavorites[0].iconPath = targetIcon.path;
+        }
+        targetIcon = getIconByNode(sampleFavorites[1]);
+        if (targetIcon) {
+            sampleFavorites[1].iconPath = targetIcon.path;
+        }
+        targetIcon = getIconByNode(sampleFavorites[2]);
+        if (targetIcon) {
+            sampleFavorites[2].iconPath = targetIcon.path;
+        }
```
zFernand0: Same comment as before.
No changes requested though! 😸

src/ZoweUSSNode.ts
```diff
@@ -101,7 +101,10 @@ export class ZoweUSSNode extends ZoweTreeNode implements IZoweUSSTreeNode {
             this.setProfile(Profiles.getInstance().loadNamedProfile(mParent.mProfileName));
         }
         this.etag = etag ? etag : "";
-        utils.applyIcons(this);
+        const icon = getIconByNode(this);
```
zFernand0: No need to specify the `index.ts` file when importing things :smile:
`import { getIconByNode } from "./generators/icons/index";`
I know it could've been an auto-import functionality from VSCode, so no worries : )

No need to change it here since most imports will change based on the cyclic dependency issue.

No changes requested!

src/extension.ts
```diff
@@ -9,34 +9,34 @@
 *                                                                                 *
 */
 
+// tslint:disable-next-line: no-duplicate-imports
```
zFernand0: Not sure if there is an issue to capture this as part of the refactoring, but I do believe that we need to reduce/eliminate the use of `tslint-disable`.
Unless we think we need to update the `tslint-config` (which I don't mind either way)

I know it's not even needed here since the references to the interfaces/types were properly addressed 👍 

Again, No changes requested!

src/generators/icons/index.ts
```diff
@@ -36,7 +43,14 @@ const items = [
...
     require("./items/downloadedDocument"),
-    require("./items/documentBinaryDownloaded")
+    require("./items/documentBinaryDownloaded"),
+    require("./items/pattern"),
+    require("./items/session"),
+    require("./items/sessionOpen"),
+    require("./items/sessionFavourite"),
+    require("./items/sessionFavouriteOpen"),
+    require("./items/folder"),
+    require("./items/folderOpen")
```
zFernand0: It'd be nice if we could read all icons in a directory and load them all instead of having to modify this file every time a new icon is added.

No changes requested at this time though! :smile:

Issue #399, opened 2019-12-10
Inconsistent behavior for renaming USS files and directories


crawr:
In VSCode, the following error is encountered when trying to rename USS file or directory:

![rename USS](https://user-images.githubusercontent.com/45845502/70549594-c300fe80-1b74-11ea-9d4d-d5c61e33f9a1.PNG)

This works fine in Theia.

zFernand0:
It might be just me, but I cannot perform a single rename operation. whether it's a text file, a binary file or a directory.
Should we look into this?

jellypuno:
> It might be just me, but I cannot perform a single rename operation. whether it's a text file, a binary file or a directory.
> Should we look into this?

Same with me. I can't rename anything in USS.

stepanzharychevbroadcom:
Added fix in: https://github.com/zowe/vscode-extension-for-zowe/pull/550

zFernand0:
Fixed, but I'm still getting a bit of an issue with it. See #567

PR #550, opened 2020-02-26
Issue #399 Fix

stepanzharychevbroadcom:


stepanzharychevbroadcom:
> Thanks for fixing this Stepan. Quite an annoying bug really from a user perspective. What I am mulling over still is that you now fire an open action which makes sense in order to reopen any currently open view on the file, pick up the new name etc.
> 
> The edge case is that the file isn't open which may be the case that you want to rename but you don't want to open the file as it's big for example. Right hand button context menu allows this action without opening the file.
> 
> There may be a mechanism to identify if the file is open in the editor container but an easier solution may be to see if the file is present locally on the file system. If it is perform and open or download to refresh it

@Colin-Stone I will move this to the separate issue.

stepanzharychevbroadcom:
> ```
>         newNamePath = newNamePath.replace(/\\/g, "/"); // Added to cover Windows backslash issue
> ```
> 
> Isn't this where path.normalize should be used

@Colin-Stone Based on what I see here: https://nodejs.org/api/path.html#path_path_normalize_path it won't work. Normalise sets slashes and structure based on **running system**, which in our case means that path stays unchanged and when we pass it to USS it's going to fail.

stepanzharychevbroadcom:
> Please add integration test

@jellypuno, integration test is added.

stepanzharychevbroadcom:
@Colin-Stone Your proposal was moved to the separate ticket: https://github.com/zowe/vscode-extension-for-zowe/issues/563

zFernand0:
Even though the functionality works as expected, I'm having issues with the folders opening properly after being renamed. The open operation occurs and the icon reflects the folder as open, but no files show up and it throws the following error(s)

image


A few things that may help reproducing it are:
- Rename folder `abc` to abc1`, and then rename it back to `abc`
- Rename a binary file inside a folder, rename the folder, rename the file back

Issue #536, opened 2020-02-21
Terminology consistency


venkatzhub:
When opening a data set member - the UI shows "Opening Dataset.." 
When data set member is edited, and autosave kicks in - the UI shows "Uploading Dataset.." 

But the object is NOT a dataset, its a member. 

The UI should be consistent and portray accurate information to the user.

stepanzharychevbroadcom:
Added fix using generator, so we can extend this approach later: https://github.com/zowe/vscode-extension-for-zowe/pull/558

jellypuno:
Ooops. I thought the fix is already merged 😝

venkatzhub:
This is still not fixed. When Saving a member - it says "Saving Dataset Member" - which is good. But immediately after that you see a message "Dataset uploaded successfully" - It is NOT a dataset that is being uploaded, it is a member still... so requesting the review of the messages for consistency.

PR #558, opened 2020-02-27
Issue #536 Terminology consistency

stepanzharychevbroadcom:


stepanzharychevbroadcom:
@Colin-Stone It's fixed.

src/generators/messages/index.ts
```diff
@@ -0,0 +1,91 @@
...
+    type: MessageHierarchyType;
+    generic?: IMessageItem;
+    messages: { [index: string]: string };
+    check: (node: CombinedNode) => boolean;
+}
+
+const items = [
+    require("./items/dataset"),
+    require("./items/datasetMember")
+].map((item) => item.default) as IMessageItem[];
```
zFernand0: This messaging structure looks kind of cool.
I'm just a bit curious about the future impact that extending this generator could have in the long run.
Do you think it will be a performance issue if we have tongs of requires for handling dynamic messages based for different nodes?

stepanzharychevbroadcom: @zFernand0 we need to use it wisely, only if string really needs to be picked up dynamically based on the node type. If we obey those rules - we should be fine.

zFernand0: In the event that we decide to change this in the future, it's possible that this could become a breaking change in case extenders of the Explorer decided to rely on this dynamic mechanism (which I don't think we will be advising any time soon).

That being said, it's something to keep in mind though.
I'm cool with the changes.
Thank you for acknowledging the future implications : )

zFernand0:
Good to merge from my end.
 @jellypuno, @Colin-Stone  any final thoughts?

Colin-Stone:
> @jellypuno, @Colin-Stone any final thoughts?

Possibly over-engineered with regards to the original issue but it addresses the requirement

PR #565, opened 2020-02-28
Use Registry

Colin-Stone:

Add in the registry based implementation


Colin-Stone:
@zFernand0 Thank you for reviewing. If we do a v1.3.1 we should include this fix if possible. Thanks again

Issue #470, opened 2020-01-27
Copy and Rename is taking a long time to re-load


jellypuno:
This is related to issue #327 

I have a dataset with 30k members. I am able to load and edit it normally but when I tried to copy and rename a member, the re-load for the tree is taking a long time. It took at least 5 mins before this error message appears:

![Capture2](https://user-images.githubusercontent.com/45042840/73161116-16aab580-40eb-11ea-8277-0e5a7b210adc.PNG)

stepanzharychevbroadcom:
@jellypuno With **rename** the reason is we do parent node refresh, instead of just element refresh. So I guess it should be reworked.
`renameDataSet`(of extension.ts) method is the target for such changes.
Also what should be refactored - the fact we do such changes even if rename request has failed, you can see it in the same method.

stepanzharychevbroadcom:
@jellypuno I checked how rename works for the dataset, in the case if it fails further rename operation will be prevented, so only refresh should be refactored.

CForrest97:
Is the issue only when renaming members? If so that is handled by `renameDataSetMember`. The parent node is refreshed here too but it looks like that isn't needed

jellypuno:
@CForrest97 In issue #327 , there is a BOX link (in the end) of the research that I have concerning this.

CForrest97:
@jellypuno Ive created a branch which should fix renaming members for large data sets, once Ive done some more testing I can create a PR for this and take a look at copying too https://github.com/zowe/vscode-extension-for-zowe/tree/470-rename-members-fix-for-large-sets

PR #571, opened 2020-03-02
470 rename data sets and members performance fix

CForrest97:
Fixes rename problems found in #470 (not copy/paste)
When renaming a member, just the member is refreshed instead of the entire data set
Similary when renaming a data set, only the data set is refreshed instead of the parent node


zFernand0:
Thinking purely about test refactoring: `Maybe we could capture this in a unit test?`
It may be too idealistic, however it'd be nice if we were able to restructure the tests in a way that we can isolate the specific module/function we are testing (in this case the refresh of the node instead of the parent). :smile:

jellypuno:
I will run this PR again, I can't see the coverage on this. Please don't merge it yet.

jellypuno:
@zFernand0 For some reason, coverage is not available :(

zFernand0:
> For some reason, coverage is not available :(

Am I missing something here?
https://wash.zowe.org:8443/job/zowe-cli-vscode-extension/job/PR-571/8/Zowe_20Explorer_20-_20Unit_20Test_20Coverage_20Report/

If you are referring to the codecov report, it seems that the hash of the commit is no longer valid do to a force-push (not necesarely on this repo but on the base/master branch in this case)

Colin-Stone:
> and0 For some reason, coverage is not available :(

Because of this type of issue I always check the unit test results in the job to ensure that we don't have any reduction in coverage. The number I always check is the one that is currently 91.39%. We can allow a small deviation but I don't think it should drop below 90%

image

jellypuno:
Thank you @Colin-Stone @zFernand0. I'm looking for the results in codecov.io but I can't find it so I was confused 😄 

But anyways, @CForrest97 Is it ok to right some test to cover this code?

Capture

CForrest97:
> Thank you @Colin-Stone @zFernand0. I'm looking for the results in codecov.io but I can't find it so I was confused 😄
> 
> But anyways, @CForrest97 Is it ok to right some test to cover this code?
> 
> Capture

Hi Jelly, I have added a couple of tests which cover these lines

Issue #563, opened 2020-02-28
USS Rename - File Opening flow


stepanzharychevbroadcom:
Here I will try to describe few cases which should be properly covered.

**1. Renaming opened file**
- User opens file;
- User renames file with any name;
- After successful renaming current tab is closed and temporary file related to the old filename is deleted;
- File is reopened;

**2. Renaming closed file**
- User renames file with any name;
- After successful renaming temporary file related to the old filename is deleted;
- File keeps it's closed state;

____
For **(1)** we already have function which ensures that the file is currently opened, so I see no problems to detect if we should reopen or not. In the same place we should resolve issues with removed state of node which happens immediately after rename.

stepanzharychevbroadcom:
@Colin-Stone After checking of VSCode APIs I found out that there's no proper way to check if **binary** file is opened. Though we can check it for **text** file. So I would suggest the following way, we do smart opening for text files, but always open binary files. I guess it's not a big problem because binary files are less common case.

Colin-Stone:
Agree with text file approach I wouldn't want to open a binary file if it was massive. So my feeling is that we could check if it is already downloaded and existing in local storage. If it is I would say download it. If not don't download it

stepanzharychevbroadcom:
@Colin-Stone @zFernand0 Comment above is taken into account. Implementation is available here: https://github.com/zowe/vscode-extension-for-zowe/pull/576
Also I implemented separate function to ensure that text file is really opened in workspace, because I found out that **textDocuments** collection is not updated properly according to visibility state. Approach may look strange but that's what some extensions do due to poor VS Code API nature.

PR #576, opened 2020-03-02
Issue #563

stepanzharychevbroadcom:


src/utils/workspace.ts
```diff
@@ -0,0 +1,27 @@
+import * as vscode from "vscode";
+
+interface IExtTextEditor extends vscode.TextEditor {
+    id: string;
+}
+
+export async function checkTextFileIsOpened(path: string) {
```
VitGottwald: Are we sure this does not interfere with the user work flow - like blinking a window or activating some other window than currently selected?

stepanzharychevbroadcom: It will be, as I showed you, but unfortunately there's no other way for us to achieve the requested functionality.

stepanzharychevbroadcom:
> Thanks Stepan an interesting solution. Needs test coverage please
> image

@Colin-Stone Should be fixed now.

stepanzharychevbroadcom:
@Colin-Stone @VitGottwald @zFernand0 
What deserves special attention in this review: **workspace.ts**.
I introduced the fix which omits empty/binary tabs up to 3 times, and after stops. (this condition required not to go to the infinite loop) It's reflected in tests.

Also I added **constants**, so magic values can be used across the application, not stay hardcoded in 1 file.

stepanzharychevbroadcom:
New implementation is in https://github.com/zowe/vscode-extension-for-zowe/pull/817. PR should be closed.

Issue #172, opened 2019-07-25
Emulate vscodes "workbench.action.quickOpenNavigateNextInFilePicker" (ctrl+p on windows) file picker in data set list


tucker01:
I hope this isn't a duplicate issue. I think this idea has been discussed before. 

In vscode, using `ctrl+p` (the command is `workbench.action.quickOpenNavigateNextInFilePicker`) I'm presented with a dialog box that allows me to type the name (or part) of a file to quickly open/navigate:

![image](https://user-images.githubusercontent.com/37446232/61891770-6d36fd00-aed8-11e9-9f04-301fb20533f3.png)

This is normally how I navigate a project I'm familiar with. 

It would take some design/thinking, but it would be great if I could apply this concept to data-sets. For example, I could type something like:
```
jcl ie
```

And the file picker would show the following for selection:
```
DATA.SET.JCL(IEFBR14)
DATA.SET.JCL(IEBCOPY)
PROJECT.JCL(IEBGENER)
```

Then selecting one would open the member.

ChrisBoehmCA:
:fire:

katelynienaber:
Looks like shortcuts can be added like [this](https://code.visualstudio.com/api/references/contribution-points#contributes.keybindings). I'll give it a try next week ^^

katelynienaber:
@tucker01 We talked about this today.

It seems this functionality is available in z/OS using `searchfor`, but there's no API for it.

Would it be possible for someone at IBM to build this API at some point? From your perspective is there a more effective way of running this search?

jellypuno:
@tucker01 Do you think this VSC feature will suffice to fix this issue? Or file picker is still better?

![VSC Search](https://user-images.githubusercontent.com/45042840/73541487-dde44680-4432-11ea-9ea0-61eda7bf516c.gif)

![VSC Search 2](https://user-images.githubusercontent.com/45042840/73541494-e3da2780-4432-11ea-926d-e6d7f5c067c5.gif)

tucker01:
This seems like a good alternative to the file picker. However, if it was possible, I'd definitely prefer the file picker method.

Would this method work for data sets in my list that I haven't expanded the member set for?

jellypuno:
> Would this method work for data sets in my list that I haven't expanded the member set for?

[...]

PR #592, opened 2020-03-05
Emulate VSCode quick open (issue 172)

katelynienaber:
Emulate VSCode's quick open shortcut (workbench.action.quickOpenNavigateNextInFilePicker)

- Allows use of CTRL+ALT+P to search files currently loaded in workspace
- For data set & USS trees
- Issue [#172](https://github.com/zowe/vscode-extension-for-zowe/issues/172)


Colin-Stone:
Interesting that the hotkey for mac has no equivalent usage in VSCode so we don't have to worry about what the label says: Now I think "Filter file list" would read better. 
Can we limit the number of rows stored?.
Would also be good to be able to filter on USS or Dataset. There is a parameter in the QuckPick that allows you to filter on description as well as label

katelynienaber:
> Found some issues and sent it via email

Fixed 3 issues according to your email ⚡️

katelynienaber:
> @katelynienaber I love it! ❤️
> 
> Do you think it will be a difficult change if you allow asterisk in the filtering? like this tes*.data 😅 . So basically if I input tes*.data the result would be [tesa.data, tesb.data, tesc.data, /tmp/tesA.data]
> 
> It is just nice to have. you don't have to do it if it will be a big change. I can open another issue, if ever.

I added this :) And I put it into a separate function so we can add more ways to filter files in the future.

Issue #593, opened 2020-03-06
Direct reference to zosmf function rather than via ZoweExplorerApiRegister


Colin-Stone:
Possible regression. Within ussNodeActions the uploadBinaryFile function refers directly to 
`        await zowe.Upload.fileToUSSFile(node.getSession(), filePath, ussName, true);`

This needs to be via the Api Register as per the uploadFile function

Colin-Stone:
@zFernand0 If we create a v1.3.1 this will need inclusion as it was included in 1.3
@stepanzharychevbroadcom would you be able to address as you are familiar with this area

Alexandru-Dumitru:
I would also suggest to use the new API `zowe.Upload.fileToUssFile()`:
https://github.com/zowe/zowe-cli/blob/f5d82918db5cce5aa8eec4a4c4fbf63bbcd49503/packages/zosfiles/src/api/methods/upload/Upload.ts#L580
- this one takes advantage of `IUploadOptions` and it is consistent with the mvs API
- we can use still use `zowe.Upload.fileToUSSFile()` but we deprecated it for the other one on the CLI level

stepanzharychevbroadcom:
@Colin-Stone @zFernand0 @Alexandru-Dumitru All of the things which you've mentioned here are fixed in https://github.com/zowe/vscode-extension-for-zowe/pull/587/commits/40d9fd5c14eb1660ea2a61723045cb90c8fb0b37 related to issue  #472. Please, take a look, and I think this issue can be closed once we merge #472 into master.

stepanzharychevbroadcom:
Added fixes in: https://github.com/zowe/vscode-extension-for-zowe/pull/600

phaumer:
@Colin-Stone there is one more regression in deleteUSSNode() that breaks zFTP. I created another PR here: https://github.com/zowe/vscode-extension-for-zowe/pull/617

Colin-Stone:
Addressed

Issue #589, opened 2020-03-05
Renaming datasets doesn't update an open tab


Colin-Stone:
Noticed that the tab name stayed the same

stepanzharychevbroadcom:
Added fix in: https://github.com/zowe/vscode-extension-for-zowe/pull/603

PR #594, opened 2020-03-06
Prepare v1.3.1

zFernand0:
Prepare for a v1.3.1 release :smile:

- [x] #525
- [x] #593 => #599 
- [x] #550
- [ ] #589 similar to `#550` but for datasets?
- [x] #565
- [x] #597 
- [x] #609

### DIFF
--- a/__tests__/__integration__/dataset/dsNodeActions.integration.test.ts
+++ b/__tests__/__integration__/dataset/dsNodeActions.integration.test.ts
@@ -1,0 +1,86 @@
+/*
+* This program and the accompanying materials are made available under the terms of the *
+* Eclipse Public License v2.0 which accompanies this distribution, and is available at *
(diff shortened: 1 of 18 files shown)


Colin-Stone:
Re: Issue #593 There is some confusion about where and how it is fixed. Issuges a specify PR to fix only this otherwise we need to pull in PR's with a lot of additional functionality

lauren-li:
@Colin-Stone @zFernand0 I would like to suggest adding PR #609 into v1.3.1 - It seems that the latest version of VS Code (1.43.0) broke the folder expand for Zowe Explorer's tree views, such that the user has to click on the same folder 3 times for it to expand and properly display its contents (see issue #608). PR #609 addresses that issue.

zFernand0:
Merging.
Will open a PR for #603 after it's ready to go (merged in master or someone else approves it?)

PR #597, opened 2020-03-06
bug fix for triggering uploading of USS files and using correct Interface in saveFile()

### LINKED ISSUE #595: wrong interface used
https://github.com/zowe/vscode-extension-for-zowe/blob/0c333cfced10c5817eff29f02b1ed3caf3fd293f/src/extension.ts#L1644

It should be `IZoweDatasetTreeNode`. We only use `.getEtag()` and `.setEtag()` methods, which are common, but still we should have proper type.

### LINKED ISSUE #596: Upload USS file doesn't work anymore
https://github.com/zowe/vscode-extension-for-zowe/blob/0c333cfced10c5817eff29f02b1ed3caf3fd293f/src/extension.ts#L192

Needs to call `saveUSSFile()`.

Alexandru-Dumitru:
fixes #595
fixes #596

Signed-off-by: Alexandru-Paul Dumitru <alexandru.dumitru@broadcom.com>

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -189,7 +189,7 @@ export async function activate(context: vscode.ExtensionContext): Promise<ZoweEx
                 await saveFile(savedFile, datasetProvider); // TODO MISSED TESTING
             } else if (savedFile.fileName.toUpperCase().indexOf(USS_DIR.toUpperCase()) >= 0) {
                 log.debug(localize("activate.didSaveText.isUSSFile", "File is a USS file -- saving"));
-                await saveFile(savedFile, ussFileProvider); // TODO MISSED TESTING
+                await saveUSSFile(savedFile, ussFileProvider); // TODO MISSED TESTING
             } else {
                 log.debug(localize("activate.didSaveText.file", "File ") + savedFile.fileName +
                     localize("activate.didSaveText.notDataSet", " is not a data set or USS file "));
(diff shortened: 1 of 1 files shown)


Alexandru-Dumitru:
don't really understand why snapshot test is failing in pr-merge.. 😞

stepanzharychevbroadcom:
For me it looks okay. @zFernand0 could you check what's going on with snapshot? I think it's the same with the issue which I used to have.

Alexandru-Dumitru:
Thanks @jellypuno for the magic touch. I replayed the job myself and still got the error 😛 .

Alexandru-Dumitru:
> I don't think you should change the USSNode reference to Dataset (line 1644)
> I also included the other fix in #598 as I addressed a few similar ones.

Hi Colin. Why not? we are using the wrong interface because we are in `saveFile()` function which addresses only data sets.

If you would capture also https://github.com/zowe/vscode-extension-for-zowe/issues/595 in your PR, then we can close this one.

Colin-Stone:
> Hi Colin. Why not? we are using the wrong interface because we are in `saveFile()` function which addresses only data sets.

My mistake, you're quite right. I was looking at the saveUSS function

jellypuno:
everything looks good and coverage is over 91%. I can merge this.

PR #598, opened 2020-03-06
Implement Delete rather than DeleteJob

Colin-Stone:
This is another refactoring that was missed associated with the generics work.
I will test all of these again next week to confirm there are no more. Only affects the master code stream


Colin-Stone:
Identified two more disconnects. saveSearch for USS was the wrong way around and relates to the generics issue whilst the uploadBinaryFile relates to the issue concerned with flipping between binary and text files

Colin-Stone:
fixes: #596

PR #599, opened 2020-03-09
Fix uploadBinaryFile redirection

Colin-Stone:

Fixes regression that bypassed the API registry


zFernand0:
TL;DR: This won't happen on master.

Vulnerabilities found in devDependencies. Technically, they don't get shipped with the extension.

In order to help with this, the Pipeline libraries have a `checkVulnerabilities` function that bypasses any devDependency vulnerability by default. However the v1.3.x still uses regular `npm audit` in the declarative pipeline definition.

PR #600, opened 2020-03-09
Issue #593

stepanzharychevbroadcom:
Added fixes for incorrectly implemented and used USS APIs.

### DIFF
--- a/src/api/ZoweExplorerApi.ts
+++ b/src/api/ZoweExplorerApi.ts
@@ -86,23 +86,15 @@ export namespace ZoweExplorerApi {
         /**
          * Uploads the file at the given path. Use for Save.
          *
-         * @param {string} inputFilePath
-         * @param {string} ussFilePath
-         * @param {boolean} [binary]
-         *      Indicates if a conversion should be attempted or treated as binary.
-         * @param {string} [localEncoding]
-         *      Optional encoding that can be used by an implementation to overwrite defaults
-         * @param {string} [etag]
-         * @param {boolean} [returnEtag]
-         * @returns {Promise<zowe.IZosFilesResponse>}
+         * @param inputFilePath - Path to local file
+         * @param ussFilePath   - Name of USS file to write to
+         * @param options   - Uploading options
+         * @returns {Promise<IZosFilesResponse>} - A response indicating the outcome
          */
         putContents(
             inputFilePath: string,
             ussFilePath: string,
-            binary?: boolean,
-            localEncoding?: string,
-            etag?: string,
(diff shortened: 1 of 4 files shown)


src/api/ZoweExplorerApi.ts
```diff
@@ -86,23 +86,15 @@ export namespace ZoweExplorerApi {
...
-         *      Indicates if a conversion should be attempted or treated as binary.
-         * @param {string} [localEncoding]
-         *      Optional encoding that can be used by an implementation to overwrite defaults
-         * @param {string} [etag]
-         * @param {boolean} [returnEtag]
-         * @returns {Promise<zowe.IZosFilesResponse>}
+         * @param inputFilePath - Path to local file
+         * @param ussFilePath   - Name of USS file to write to
+         * @param options   - Uploading options
+         * @returns {Promise<IZosFilesResponse>} - A response indicating the outcome
```
VitGottwald: We delete the since it duplicates the information available in type annotation.

src/api/ZoweExplorerApi.ts
```diff
@@ -86,23 +86,15 @@ export namespace ZoweExplorerApi {
...
          */
         putContents(
             inputFilePath: string,
             ussFilePath: string,
-            binary?: boolean,
-            localEncoding?: string,
-            etag?: string,
-            returnEtag?: boolean
+            options?: zowe.IUploadOptions
```
phaumer: You cannot change the API as it will break all implementers. We need to define a change process for doing such changes.

phaumer: I suggest that you build the IUploadOptions object in the implementation for now only.

stepanzharychevbroadcom: @phaumer Could you explain why other implementation will be broken by this change, if that's a local project typing file?

zFernand0: Deprecation will be required in this case.
We have multiple example in other Zowe repos (some CLI ones for instance) where we do this.

zFernand0: As we start refactoring functionality, we need to make sure we give people enough time to adjust. (deprecation)
In this case we could create an overloaded function that behaves the same way and maybe issue a deprecation warning if we detect that some of the old parameters were set, (maybe in the imperative OR zowe-cli log since this is a developer thing?). Thanks @awharn and @tjohnsonBCM for taking the time to discuss this 😄 
Here is an example: https://howtodoinjava.com/typescript/function-overloading/

PR #603, opened 2020-03-10
Issue #589

stepanzharychevbroadcom:


__tests__/__unit__/utils/workspace.unit.test.ts
```diff
@@ -0,0 +1,157 @@
...
+
+jest.mock("vscode");
+
+const activeTextEditor = jest.fn();
+const executeCommand = jest.fn();
+
+Object.defineProperty(vscode.window, "activeTextEditor", {get: activeTextEditor});
+Object.defineProperty(vscode.commands, "executeCommand", {value: executeCommand});
+
+describe("Checking Workspace file status method", () => {
```
VitGottwald: How about shortening the descriptions for better readability in test summary like this
```suggestion
describe("Workspace file status method", () => {
```

__tests__/__unit__/utils/workspace.unit.test.ts
```diff
@@ -0,0 +1,157 @@
...
+    beforeEach(() => {
+        activeTextEditor.mockReset();
+        executeCommand.mockReset();
+    });
+
+    it("Tests if method runs correctly with no tabs available", async () => {
```
VitGottwald:
```suggestion
    it("Runs correctly with no tabs available", async () => {
```
and similarly for other test titles ...

stepanzharychevbroadcom: Sounds logical.

src/extension.ts
```diff
@@ -1023,8 +1026,17 @@ export async function renameDataSet(node: IZoweDatasetTreeNode, datasetProvider:
             datasetProvider.renameFavorite(node, afterDataSetName);
             node.label = temp;
         }
-        datasetProvider.refreshElement(node.getParent());
+
+        datasetProvider.refreshElement(node);
         datasetProvider.updateFavorites();
+
+        if (fs.existsSync(beforeFullPath)) {
```
VitGottwald: In general `sync` methods are not a good idea, because they block the event loop. Perhaps someone from the team can comment on whether that is or is not the case here?

VitGottwald:
Good job @stepanzharychevbroadcom. Especially with the nicely readable unit tests, thank you!

src/extension.ts
zFernand0: 🤔 Interesting comment.
While I agree that synchronous event block the event loop, it's also true that we logically need to `await` for the code to execute.

I don't see it as a bad practice to use `async` methods whenever possible, even though the logical structure seems the same.
In the long run, we might get a slight performance increase thanks to the JS engine and someone could make the case that it's theoretically more secure because of blocked threads due to malicious input.

As a general rule, (personally) I don't request changes if the functionality does not what's supposed to, or conflicts with other existing design paradigm.

Hope this helps 😄

VitGottwald: Looking at the code around and thinking a bit more about this, I believe we should stick to a single paradigm - unless there is a reason to do things otherwise. All of our code uses async/await so I believe it would be better to use it here as well.
```suggestion
import {promisify} from 'util';
const exists = util.promisify(fs.exists);
        if (await exists(beforeFullPath)) {
```
and similarly for `fs.unlink` bellow

zFernand0: How about we tackle this as a separate (refactoring) ticket.

That way we can thoroughly go through the code and focus on the places where we don't need to be blocking the event loop. :smile:

Will open another issue for this.

And maybe we can start using the wiki? (Credits to @Colin-Stone for suggesting 👍 )

zFernand0:
Will migrate to `v1.3.x` after merging into master.

PR #604, opened 2020-03-10
Moving rename functionality into appropriate tree classes

Colin-Stone:
Correctly implements the rename for both datasets and USS files


Colin-Stone:
@CForrest97 would you be able to take another look please?. Had to make a lot of mods because of the other chnages to rename

PR #617, opened 2020-03-12
Remove zosmf specific code

phaumer:
deleteUSSNode() in src/ZoweUSSNode.ts was modifying the path for a zosmf bug that would break other implementations such as zFTP.  The same fix was already present in src/api/ZoweExplorerZosmfApi.ts (line 90).

### DIFF
--- a/src/ZoweUSSNode.ts
+++ b/src/ZoweUSSNode.ts
@@ -282,7 +282,12 @@ export class ZoweUSSNode extends ZoweTreeNode implements IZoweUSSTreeNode {
         this.tooltip = injectAdditionalDataToTooltip(this, newFullPath);
 
         vscode.commands.executeCommand("zowe.uss.refreshUSSInTree", this);
-        vscode.commands.executeCommand("zowe.uss.ZoweUSSNode.open", this);
+        if (this.contextValue === extension.DS_TEXT_FILE_CONTEXT ||
+            this.contextValue === extension.DS_FAV_TEXT_FILE_CONTEXT) {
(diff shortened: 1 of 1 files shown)


src/ZoweUSSNode.ts
```diff
@@ -282,7 +282,10 @@ export class ZoweUSSNode extends ZoweTreeNode implements IZoweUSSTreeNode {
         this.tooltip = injectAdditionalDataToTooltip(this, newFullPath);
 
         vscode.commands.executeCommand("zowe.uss.refreshUSSInTree", this);
-        vscode.commands.executeCommand("zowe.uss.ZoweUSSNode.open", this);
+        if (this.contextValue === extension.DS_TEXT_FILE_CONTEXT ||
```
stepanzharychevbroadcom: I would remove those changes, because the same logic is introduced in fix for rename:
https://github.com/zowe/vscode-extension-for-zowe/pull/619

lauren-li:
I think we can merge this PR to address the deleteUSSNode() issue.
- PR #626 can be merged afterwards to address the USS Folder rename issue. Although the post-rename refresh for that PR collapses the user's open USS folders, it also refeshes the renamed folder's children so the user will be able to work with those children after the rename without getting errors.

src/ZoweUSSNode.ts
lauren-li: @stepanzharychevbroadcom I've removed the USS folder rename changes from this PR to avoid merge conflicts when your changes from #619 come in.

zFernand0:
Dev Dependency vulnerability. Merging

Issue #616, opened 2020-03-11
New Rename USS function causes HTTP 500 error


Colin-Stone:
The rename code causes a HTTP 500 error which is related to a subsequent chtag call when opening the file again.
![image](https://user-images.githubusercontent.com/30794003/76448723-b635b600-63c2-11ea-882d-3d6ce85301e9.png)

The flow of events is to execute the rename command and then follow with execute an open command. I wonder if the use of execeute rather than a promise and then strategy is presenting a sequencing problem

Colin-Stone:
"HTTP(S) error status "500" received.\nReview request details (resource, base path, credentials, payload) and ensure correctness.\n\nHost:      winmvs3b.hursley.ibm.com\nPort:      32070\nBase Path: \nResource:  /zosmf/restfiles/fs/u%2Fstonecc%2F.test123553\nRequest:   PUT\nHeaders:   [{"Content-Type":"application/json"},{"Content-Length":"35"},{"X-CSRF-ZOSMF-HEADER":true}]\nPayload:   { request: 'chtag', action: 'list' }"
causeErrors:"{"category":7,"rc":266,"reason":0,"stack":"  1 15425A20+000003EE JsonParser::serialize(bool)\n  2 15403E50+000008D4 FileUtilServlet::doPut()\n  3 1540FF58+000000D6 TsoServlet::service()\n  4 1543EB08+00000190 TsoServlet::run()\n  5 1543E468+00000396 main\n  6 06809570+0000127E CEEVROND\n  7 14BFF2A0+000000B4 EDCZHINV\n  8 066A5CA0+000001C6 CEEBBEXT\n"}

Colin-Stone:
Related to issue: https://github.com/zowe/vscode-extension-for-zowe/issues/567 ???

Original PR https://github.com/zowe/vscode-extension-for-zowe/pull/597 ??

1.3.x PR https://github.com/zowe/vscode-extension-for-zowe/pull/613

lauren-li:
I can actually rename USS files without any error, but I do get this error upon renaming a USS folder. (I do not need to open the renamed folder for the error to occur.)

phaumer:
@Colin-Stone the problem is in src/ZoweUSSNode.ts, rename(), line 284.  

Issue 1: zowe.uss.refreshUSSInTree does not refresh the children as their absolute path is changed
Issue 2: zowe.uss.ZoweUSSNode.open tries to open a directory as if it is a file

[...]

PR #619, opened 2020-03-12
Issue #616

stepanzharychevbroadcom:


src/ZoweUSSNode.ts
```diff
@@ -278,14 +278,21 @@ export class ZoweUSSNode extends ZoweTreeNode implements IZoweUSSTreeNode {
...
-        vscode.commands.executeCommand("zowe.uss.refreshUSSInTree", this);
-        vscode.commands.executeCommand("zowe.uss.ZoweUSSNode.open", this);
+        if (this.isFolder) {
+            await vscode.commands.executeCommand("zowe.uss.refreshAll");
+        } else {
+            await vscode.commands.executeCommand("zowe.uss.refreshUSSInTree", this);
+        }
+
+        if (!this.isFolder) {
+            await vscode.commands.executeCommand("zowe.uss.ZoweUSSNode.open", this);
```
lauren-li: Does this line need to be in its own `if (!this.isFolder)` condition, or could it be part of the `else` block at line 289, after the `refreshUSSInTree` command?

lauren-li:
With this branch, I no longer see the HTTP 500 error upon USS folder rename. I did have a question about the code (see my comment above). Thanks @stepanzharychevbroadcom for working on this!

src/ZoweUSSNode.ts
stepanzharychevbroadcom: I did it in this form to align with more advanced implementation in another PR.
https://github.com/zowe/vscode-extension-for-zowe/blob/af3719e4aff62fb7bd0fa470a0b3ca6470b1d35a/src/ZoweUSSNode.ts#L292

If it's a problem I can change it - no worries

stepanzharychevbroadcom: @lauren-li Please, let me know if you have other concerns

lauren-li: @stepanzharychevbroadcom To me, it makes more sense to have this line as part of the `else` block for now, and then later move it into its own `if (!this.isFolder) &&...` block in the PR for the more advanced implementation.

stepanzharychevbroadcom: @lauren-li Changed it.

lauren-li: Thanks @stepanzharychevbroadcom !

lauren-li:
Since this PR is a fix, should we change it to be merging into the v1.3.x branch rather than master?

Issue #553, opened 2020-02-26
Create New Data Set error: Cannot Read property 'toUpperCase' of undefined.


Alexandru-Dumitru:
Create New Data Set gives error while moving out of focus from the input box.

Steps to reproduce:

1. Click `Create New Data Set` 
2. Pick `Data Set Sequential` (haven't tested with other formats)
3. (optional) Fill in the data set name in the input box
4. Click outside to remove focus from the box. Anywhere will do.

Error received:
![err1](https://user-images.githubusercontent.com/45846673/75363080-ec027700-58b9-11ea-802d-5c979fbcc720.PNG)

Tried it in USS tree, and the action is canceled silently when doing the same steps (expected behavior from my point of view).

Using: 
Zowe Explorer v1.3.0
VSCode v1.42.1

PR #620, opened 2020-03-12
Issue #553

stepanzharychevbroadcom:


src/extension.ts
```diff
@@ -830,43 +840,43 @@ export async function createFile(node: IZoweDatasetTreeNode, datasetProvider: IZ
         }
 
         // get name of data set
-        let name = await vscode.window.showInputBox({placeHolder: localize("dataset.name", "Name of Data Set")});
-        name = name.toUpperCase();
+        let name = await vscode.window.showInputBox({ placeHolder: localize("dataset.name", "Name of Data Set") });
+        if (name) {
+            name = name.toUpperCase();
```
jellypuno: can you add .trim() so I will also be prevented to add datasets with spaces at the beginning.

stepanzharychevbroadcom: @jellypuno Added.

Issue #578, opened 2020-03-02
Deleting a Job, the interface requires me to reflesh to make it go away


jelaplan:
When I delete a job, the ZE shows me a message that it has been deleted but it remains in the list of jobs. I can click the refresh button and it will go away. This seems like a bug. If I delete something, it should go away. 
<img width="1174" alt="image" src="https://user-images.githubusercontent.com/3207260/75716006-66593f80-5c9c-11ea-8f80-201429ff2b17.png">

stepanzharychevbroadcom:
Added fix in: https://github.com/zowe/vscode-extension-for-zowe/pull/622
Also moved implementation to the declared with interface.

JillieBeanSim:
this is no longer an issue.  the profile folder will now collapse after deleting a job.  I believe there is another issue to handle that so it will work like the other view trees on delete

PR #622, opened 2020-03-13
Issue #578 Deleting a Job, the interface requires me to refresh

stepanzharychevbroadcom:


katelynienaber:
> It works, however the node/folder/profile just gets collapsed. I believe that we may want to keep it open.
> So instead of refreshing all jobs, should we consider doing the same as the refresh button next to the profile name?
> @stepanzharychevbroadcom @Colin-Stone, @jellypuno any thoughts?

It works for me as well, and I like @zFernand0's idea about refreshing but keeping the job list expanded.

jelaplan:
I have Zowe Explorer 1.5.2 and the behavior is the same. If I delete something, it should go away without having to manually refresh. The panel shouldn't close. I'm not hoping to add an additional refresh icon. 
https://www.screencast.com/t/Cy0bin0YG5dl

stepanzharychevbroadcom:
> It works for me, but like @jelaplan and @zFernand0 said, the tree collapses again. Is it possible to just refresh the job list and leave it expanded?
> 
> I was having trouble with VSCode's reveal function, so maybe this won't work right now....if you get it to work that's awesome and I will copy you because I opened an issue on their repo for this P:

I suspect that I tried to do this behaviour back then, but it didn't work.
I will spend some time investigating it, but since it didn't work for you after all attempts I don't see too many reason to dig here.

katelynienaber:
> > It works for me, but like @jelaplan and @zFernand0 said, the tree collapses again. Is it possible to just refresh the job list and leave it expanded?
> > I was having trouble with VSCode's reveal function, so maybe this won't work right now....if you get it to work that's awesome and I will copy you because I opened an issue on their repo for this P:
> 
> I suspect that I tried to do this behaviour back then, but it didn't work.
> I will spend some time investigating it, but since it didn't work for you after all attempts I don't see too many reason to dig here.

If it doesn't work for you either, could you maybe help me bug VSCode about it here: [https://github.com/microsoft/vscode/issues/98095](https://github.com/microsoft/vscode/issues/98095)

Because I am being ignored over there P:

crawr:
closing in favor of #1128

Issue #273, opened 2019-10-16
Upload Member action fails with cryptic message when no file is selected


zFernand0:
When uploading a member to a data set, the OS File Browser comes up as expected. If you cancel the action, you get:
![image](https://user-images.githubusercontent.com/37381190/66920682-cacb7900-eff1-11e9-812f-e2520b740161.png)

Would it be possible to handle this the same way we do with some (if not all) other inputs?
For example, selecting a filter and leaving the input box by clicking outside, we get:
![image](https://user-images.githubusercontent.com/37381190/66920841-17af4f80-eff2-11e9-9526-b1c89be51459.png)

stepanzharychevbroadcom:
@Colin-Stone @VitGottwald @jellypuno Added fix in: https://github.com/zowe/vscode-extension-for-zowe/pull/632

PR #632, opened 2020-03-17
Issue #273

stepanzharychevbroadcom:
Added fix for no member selected scenario of DS Member upload.


src/mvs/mvsNodeActions.ts
```diff
@@ -27,22 +29,30 @@ export async function uploadDialog(node: ZoweDatasetNode, datasetProvider: IZowe
...
     const value = await vscode.window.showOpenDialog(fileOpenOptions);
 
-    await Promise.all(
-        value.map(async (item) => {
-            // Convert to vscode.TextDocument
-            const doc = await vscode.workspace.openTextDocument(item);
-            await uploadFile(node, doc);
-        }
-    ));
+    if (value && value.length) {
```
Alexandru-Dumitru: Not sure if `value.length` is required here. I think we cannot pass an empty string via the Open File dialog, but I am not opposed to leave it in as an extra security measure.

Other than this, I have tested the implementation, and works as expected.

Alexandru-Dumitru: Should be fine to go after all checks pass

PR #649, opened 2020-03-31
Context uses Regular Expressions

Colin-Stone:
Signed-off-by: Colin Stone <30794003+Colin-Stone@users.noreply.github.com>ViewItem in package.json and node.contextValue in ts files is now referenced using regular expressions. Created new helper functions because this big change is to allow additional elements to a context such as the return code from a Job and these elements can be added to the end of the context string. 
Will write this up but developers need to consider that apart from the base context which starts strictly at the beginning of the string that any additional element could appear any where in the string so for example job_rc=CC 0000_fav is the same as  job_fav_rc=CC 0000
The underscore is used as a separator


src/utils/context.ts
```diff
@@ -0,0 +1,247 @@
...
+    return isFavorite(node) ? node.contextValue : node.contextValue + extension.FAV_SUFFIX;
+}
+
+/**
+ * Helper function to retrieve the base context of a node
+ * @param node
+ * @return The inital element of the context.
+ */
+export function getBaseContext(node: TreeItem): string {
+    return node.contextValue.indexOf(extension.CONTEXT_PREFIX) > -1 ?
```
zFernand0, line 244: Thanks for checking for the `_` 👍

src/extension.ts
```diff
@@ -49,7 +49,9 @@ export let ZOWE_TMP_FOLDER;
 export let USS_DIR;
 export let DS_DIR;
 export let ISTHEIA: boolean = false; // set during activate
-export const FAV_SUFFIX = "_fav";
+export const CONTEXT_PREFIX = "_";
```
zFernand0, line 52: @Colin-Stone, question...
Should we document that `_` is not allowed when providing context information by extenders?
I'm curious what will happen with the `getBaseContext` function if extenders had a `_` as part of their context name. 
I don't think we would be able to properly update favorites.

__tests__/__unit__/utils/context.unit.test.ts
```diff
@@ -0,0 +1,306 @@
...
+    const ICON_STATE_OPEN = "open";
+    const ICON_STATE_CLOSED = "closed";
+    const JOBS_JOB_FAVORITE1 = "job_fav";
+    const JOBS_JOB_FAVORITE2 = "job_fav_rc=cc 000";
+    const JOBS_JOB_FAVORITE3 = "job_rc=cc 000_fav";
+    const JOBS_SESSION_CONTEXT_FAV = "server_fav";
+    const USS_SESSION_CONTEXT_FAV = "uss_session_fav";
+    const DS_SESSION_CONTEXT_FAV = "session_fav";
+
+    const testList: string[] = [
```
zFernand0, line 45: Clever way to address most (if not all) permutations.

src/extension.ts
```diff
@@ -1038,7 +1040,7 @@ export async function createMember(parent: IZoweDatasetTreeNode, datasetProvider
 export async function showDSAttributes(parent: IZoweDatasetTreeNode, datasetProvider: IZoweTree<IZoweDatasetTreeNode>) {
 
     let label = parent.label.trim();
-    if (parent.contextValue === DS_PDS_CONTEXT + FAV_SUFFIX || parent.contextValue === DS_DS_CONTEXT + FAV_SUFFIX) {
+    if (contextually.isFavoritePsDs(parent)) {
```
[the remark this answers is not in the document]
Colin-Stone: Yes I wanted just oine regexp rather than two

src/mvs/mvsNodeActions.ts
```diff
@@ -40,7 +41,7 @@ export async function uploadDialog(node: ZoweDatasetNode, datasetProvider: IZowe
 
         // refresh Tree View & favorites
         datasetProvider.refreshElement(node);
-        if (node.contextValue.includes(extension.FAV_SUFFIX) || node.getParent().contextValue === extension.FAVORITE_CONTEXT) {
+        if (contextually.isFavorite(node) || contextually.isFavoriteContext(node)) {
```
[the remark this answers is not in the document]
Colin-Stone: Correct I tried to be too clever and fell flat on my face

src/extension.ts
Colin-Stone: I will document this as an addendum to the contributors guide

Colin-Stone: I think this is only an issue for uss_session. I can either globally rename it as ussSession or put in a specific test in the getBaseContext for it. The rename is the right thing to do (28 replaces) but the quick fix has less impact. Which would you prefer @zFernand0 ?

zFernand0: I think I'd go for consistency here 😞 Sorry.
So, Change it to `ussSession` if it's not too disruptive.

Colin-Stone: Thinking about this some more I will do option 2 for now and option1 should be a separate issue which will be implemented after Katelyn has merged her big PR

zFernand0: Good catch! :) 
Thanks!

src/utils/context.ts
```diff
@@ -0,0 +1,247 @@
...
+
+/**
+ * Helper function which identifies if the node is a session but not a favorite
+ * @param node
+ * @return true if a session, false otherwise
+ */
+export function isSessionNotFav(node: TreeItem): boolean {
+    return new RegExp("^((?!.*" + extension.FAV_SUFFIX + ")(" + extension.JOBS_SESSION_CONTEXT + "|" + extension.USS_SESSION_CONTEXT + "|"
+                        + extension.DS_SESSION_CONTEXT + "))").test(node.contextValue);
+}
```
CForrest97, line 216: A lot of these functions could reuse the other context check functions For example, could this function be
```js
export function isSessionNotFav(node: TreeItem): boolean {
    return isSession(node) && !isFavorite(node);
}
```

Colin-Stone: They could, yes but I am concerned about performance so have tried to combine regular expression tests so they only run once. It may look more complex but is far more efficient

src/utils/context.ts
```diff
@@ -0,0 +1,247 @@
...
+
+/**
+ * Helper function which identifies if the node is a session favorite
+ * @param node
+ * @return true if a session favorite, false otherwise
+ */
+export function isSessionFavorite(node: TreeItem): boolean {
+   // return isSession(node) && isFavorite(node);
+   return new RegExp("^(" + extension.FAVORITE_CONTEXT + ")").test(node.contextValue);
```
CForrest97, line 225: If we only need to check if the node has Favorite Context can we remove this comment?

Colin-Stone: Yes it is the root. Ironically I do have it removed in my follow up branch so will take it out now

src/utils/context.ts
```diff
@@ -0,0 +1,247 @@
...
+export function isFavoritePds(node: TreeItem): boolean {
+    return new RegExp("^(" + extension.DS_PDS_CONTEXT + ")(.*" + extension.FAV_SUFFIX + ")").test(node.contextValue);
+}
+
+/**
+ * Helper function which identifies if the node is a Favorite binary or text file
+ * @param node
+ * @return true if a Favorite binary or text file, false otherwise
+ */
+export function isFavoriteTextorBinary(node: TreeItem): boolean {
```
CForrest97, line 79: Should this have a capital `O` for `Or`

src/utils/context.ts
CForrest97: Ah I didnt think about that.
I ran some simulations to test this and found these results. (Times are for all iterations)
```
iterations: 1,000,000
- 1 single regex: 244.296ms
- 2 regex with boolean logic: 376.171ms
```
So you are right doing one larger regex is faster. So we dont need to add this change if performance is a high priority.
I only found a difference of ~130ms over 1 million regex tests so its not a large difference

src/utils/context.ts
Colin-Stone: I'll fix that. Helps me remember to do it next time.

src/utils/context.ts
Colin-Stone: Comment removed

src/extension.ts
```diff
@@ -1575,17 +1577,17 @@ export async function openPS(node: IZoweDatasetTreeNode, previewMember: boolean,
     if (validProfile === 0) {
         try {
             let label: string;
-            switch (node.getParent().contextValue) {
-                case (FAVORITE_CONTEXT):
+            switch (true) {
```
zFernand0, line 1580: Could we change this to regular `if...else` logic?
I don't know if `switch(true)` would be confusing for others reading this, but it confused me for a second 😋

Colin-Stone: Seemed to be a recognised approach but am happy to change it if preferred

zFernand0:
Merging this one before #661 
Will address all merge conflicts over there.

PR #655, opened 2020-04-01
Include migrate icon

Colin-Stone:
Appropriate fix for the 1.4.x branch. Alternative solution for v1.5


zFernand0:
Thanks for using `v1.4.x`  as teh base branch here.
Even if we decide not to have a 1.4.1 this time around, we can certainly update master with the v1.4.x version 😋

Issue #606, opened 2020-03-10
Editing PDS member from "Favorites" list only saves locally


fritzc1:
I believe this started after the zowe plugin got an error when saving. The PDS was filled and needed compression. Ever since, if I open a member from favorites, it saves locally.

Editing the same member from my profile section saves normally to the mainframe.
<img width="310" alt="zowesave" src="https://user-images.githubusercontent.com/8673780/76331349-691fea00-62c5-11ea-8ec8-9106369b2a29.png">

bearou:
Same story here... A v1.9.0 issue ?

FALLAI-Denis:
When reopen a file or PDS member from synchronised files list with mainframe, current changes (not saved, and even saved) made in file or PDS member opened from "Favorites" are discarded without warning.

Same issue if reopen from "Favorites" : file or PDS member is refresh from mainframe, discarding current changes (not saved) made on file or PDS member from synchronized files list.

Issue #660, opened 2020-04-03
Issue with saving USS files back to mainframe


Alexandru-Dumitru:
After editing a USS file, the save process only happens on local file system, but the upload is not triggered. Happens in Favorites as well. There is no error in VS Code.

Issue reported on Slack #zowe-explorer by Dan Jast. 

Possibly related with (didn't do any RCA yet): 
https://github.com/zowe/vscode-extension-for-zowe/issues/555
https://github.com/zowe/vscode-extension-for-zowe/issues/606

zdmullen:
I think this is quite a significant issue and should be looked into sooner than later. I will mark it for the 1.7 release to give the squad time to research it and work on it as needed. We can update it if we deem it more pressing.

Rappel87:
This is one of the main functions for many of our devs using the Zowe extension. When will you release 1.7 and plan to fix the bug? Because we are already looking into alternatives to save our data to USS.

stepanzharychevbroadcom:
@Rappel87 Hello, no worries, I will discuss it with my colleagues and keep you posted.

zdmullen:
@Rappel87, I've adjusted the milestone to the 1.5 Release, which will go out at the end of April. As @stepanzharychevbroadcom mentioned, we will present this to the team during the stand-up today and ensure it is ready.

Rappel87:
Thanks for the heads-up.

phaumer:
@Colin-Stone @zFernand0 I think this should be an emergency 1.4.1 fix as users would loose all their edits without warning!

I created a PR for 1.4.x here: https://github.com/zowe/vscode-extension-for-zowe/pull/669

tjohnsonBCM:
@Rappel87 A patch release is now available on the VSCode Marketplace (1.4.1) that fixes the issue with saving USS files.

impervachris:
I am using 1.4.1 and see the same behavior with saving PDS members.  Did no one verify this behavior before the patch was released?  This is a major issue.

phaumer:
@impervachris I did actually. I just tried it again:
- opened a PDS member with a jcl file
- edit, save
- close 
- Pull from Mainframe
> I see my edits.

Do you see any errors in the logs or output windows?

impervachris:
No.  there are no messages in the output window.  It just simply saves it locally.

impervachris:
Maybe it is specific to using WSL?

impervachris:
[...]

PR #664, opened 2020-04-05
Issue #555, #606 Fixes for saving of Datasets from Favourites section

stepanzharychevbroadcom:
Fixes for saving of Datasets from Favourites section.


src/extension.ts
```diff
@@ -1710,7 +1710,50 @@ export async function saveFile(doc: vscode.TextDocument, datasetProvider: IZoweT
...
+            const targetParentNode = datasetProvider.mFavorites.find((favNode) => {
+                return nameRegex.test(favNode.label);
+            });
+
+            if (targetParentNode) {
+                const memberRegexResult = datasetName.match(memberRegex);
+                const memberName = memberRegexResult && memberRegexResult[0];
+
+                const hasMember = targetParentNode.children.some((memNode) => {
+                    return memNode.label === memberName.replace("(", "").replace(")", "");
```
katelynienaber, line 1732: Could replace the double replace hehe
memberName.replace(/\(|\)/g, "");

stepanzharychevbroadcom: @katelynienaber It's fixed for both occurrences. Valid point.

src/extension.ts
```diff
@@ -1710,7 +1710,50 @@ export async function saveFile(doc: vscode.TextDocument, datasetProvider: IZoweT
     }
     const start = path.join(DS_DIR + path.sep).length;
     const ending = doc.fileName.substring(start);
-    const sesName = ending.substring(0, ending.indexOf(path.sep));
+    let sesName = ending.substring(0, ending.indexOf(path.sep));
+
+    // Added special handling for attempt to save favourite node
+    if (sesName === "Favorites") {
```
Colin-Stone, line 1716: I think "Favorites" should be a global constant rather than hardcoded. However when created it's localized which is compounding the situation and potential error. Suggest we should address this oversight in USSTree and DatasetTree

Colin-Stone:
I can see #660 gets fixed, but the other two also sounded like they were regressions from 1.3. The code being added here seems complex and was expecting to be looking at something more straight forward if this was indeed a regression.

stepanzharychevbroadcom:
> I can see #660 gets fixed, but the other two also sounded like they were regressions from 1.3. The code being added here seems complex and was expecting to be looking at something more straight forward if this was indeed a regression.

@Colin-Stone I think that regression is connected with a way how we get profile now - using `(await Profiles.getInstance()).loadNamedProfile(sesName)`, because for favourites it couldn't work correctly before with logic which we apply.

src/extension.ts
stepanzharychevbroadcom: @Colin-Stone Good idea, I will move it.

Colin-Stone:
So my perspective had been to look for nodes based upon the tree structure and make it do the work however I now wonder if we should let the local file system manage this similar to what you have been doing. Because we know where the favorites are located and where the real files would be located we can use a common function or better still a method on ZoweTreeNode to: 
1. Calculate the logical location of it's twin.
2. If the twin is present return that location.

This function should work on either type and may also be relevant to a number of other areas of functionality such as copy rename etc where a check has to be made if anything needs to happen to the equivalent favorite etc.

Colin-Stone:
The alternative to my previous comment is to base everything off a node in the tree structure. To enable this we would have a generic function that relates a file back to a tree item. This can be done based upon the local path names which as you indicate in your code would also describe the node as a favorite or not. 

Since every other action/operation we perform is on a node it just makes code more consistent. What do you think?.

Colin-Stone:
Please also run `npm run vscode:prepublish` to regenerate the i18n language samples with your new strings.
I've analysed unit test coverage for this file. There is no unit test coverage of your new code but I notice that since the code has been moved a lot of unit test coverage has been lost overall e.g. renameDataset and renameDatasetMember has no test coverage and that did exist before?

stepanzharychevbroadcom:
> > Please also run npm run vscode:prepublish to regenerate the i18n language samples with your new strings.
> > I've analysed unit test coverage for this file. There is no unit test coverage of your new code but I notice that since the code has been moved a lot of unit test coverage has been lost overall e.g. renameDataset and renameDatasetMember has no test coverage and that did exist before?

@Colin-Stone I will run prepublish, no worries. My only question is: can we increase the coverage after new tests are merged? It will be way more logical, because structure and approach will be changed a lot. We can create additional issue for that.

stepanzharychevbroadcom:
> I don't like the practice of not unit-testing (code-coverage) such a big addition.
> That being said, I'm Ok with the functionality since it solves the problem in question AND tests will be included as part of the refactoring. Isn't that right @stepanzharychevbroadcom?

@zFernand0 I can add coverage in my **refactor-datasets** branch once we merge this one. So no worries.

PR #669, opened 2020-04-07
Emergency fix for USS files do not get saved (Issue #660)

phaumer:


jellypuno:
Is this different from PR #664 ?

stepanzharychevbroadcom:
> Is this different from PR #664 ?

@jellypuno No, I fixed it in the same way already.

phaumer:
> > Is this different from PR #664 ?
> 
> @jellypuno No, I fixed it in the same way already.

@stepanzharychevbroadcom  I did not realize. I think it would be good to provide explicit descriptions into the PR name for other people to follow along. We cannot ask people to drill into each issue and pr to find out what is going on.  The PR list should give me a quick overview. That will improve transparency making it easy for everyone to follow along who is not in the Scrum timezone.

PR #675, opened 2020-04-09
Include all "nls.metadata.json" files to generate i18n sample

tjohnsonBCM:
This makes the following unnecessary:
> If the new string is in a newly-created typescript file, update the stringUpdateScript.js file with an entry for the new typescript file.

It appears the i18n sample folder had already gotten out of date, this should hopefully ensure it stays up to date in the future.

Additional changes to "stringUpdateScript.js":
* Empty the "i18n/samples" folder before generating files
* Create directories in the file path if they don't exist yet


scripts/stringUpdateScript.js
```diff
@@ -0,0 +1,37 @@
...
+* Copyright Contributors to the Zowe Project.                                     *
+*                                                                                 *
+*/
+
+// This script creates the i18n/sample template used for adding/updating i18n files
+const fse = require("fs-extra");
+const path = require("path");
+
+// Clean out i18n samples directory
+fse.emptyDirSync("./i18n/sample");
```
tjohnsonBCM, line 17: Is it bad to be doing this? Not sure if we expect users might create files in that folder which should be preserved.

phaumer: Developers needs to understand how nls works. If we do not clean the entire dir then it will be hard to catch and remove the files for renamed modules.

phaumer:
@lauren-li @zFernand0 would it make sense to add a PR automated check that runs `npm run package` and checks that there are no changed files to reject the PR? Somehow we should enforce developers to run this before creating a PR.

scripts/stringUpdateScript.js
lauren-li: @tjohnsonBCM  Is there a specific use case you have in mind for when developers would want to manually create files in i18n/sample? In my mind, the contents of i18n/sample should be automatically generated based on wherever the `localize` function is called, and developers should not have to touch anything in that folder. Therefore, it shouldn't be detrimental to clean out i18n/sample before generating an updated set of files for it.

lauren-li:
@phaumer I think it would be great to add something in the PR pipeline that runs `npm run package` (or some other npm script that executes the stringUpdateScript) and checks the result for no changed files compared to the PR. That would help ensure that at least for each PR, the i18n/sample strings are kept up-to-date.

scripts/stringUpdateScript.js
tjohnsonBCM: No I didn't have a specific use case in mind, was just unsure since the i18n/sample folder wasn't being cleaned in this script previously. Thanks for the explanation, it confirms that emptying the contents of the folder is the desired behavior.

zFernand0:
> This check could either be added as part of this PR, or in a future PR.

I agree that we should add a PR check for this. 
@tjohnsonBCM, Maybe we could track this effort as a separate issue?

Issue #648, opened 2020-03-31
ContextValue of undefined


jellypuno:
![Capture](https://user-images.githubusercontent.com/45042840/78013063-38702500-7346-11ea-9eef-bcf333a15726.PNG)

Steps I did:
1. Create a new member
2. Paste the JCL from clipboard
3. Save it
4. Error appears
5. After this, my profiles are now Invalid and my credentials are replaced by "managed by Zowe Explorer". I need to refresh my workspace

Note: I am using the code from master branch and I have SCS installed

stepanzharychevbroadcom:
@jellypuno I tried to reproduce this bug, but no luck. As well as I have no ideas why it might happen. May I ask you to try the same flow out without this additional extension?

jellypuno:
@stepanzharychevbroadcom Do you have secure credential store installed?

stepanzharychevbroadcom:
@jellypuno I tried it without extensions on clear master. Could you share the link with me and may be some guideline?

jellypuno:
I installed this plugin: https://github.com/zowe/zowe-cli-scs-plugin

stepanzharychevbroadcom:
I verified saving of _.jcl_ file and submitting it as well. Everything went fine with installed  SCS plugin. To be sure I also removed all of the profiles I had and added everything from scratch. Still no luck.

Used dataset:
![Screenshot 2020-04-07 at 14 33 55](https://user-images.githubusercontent.com/60000150/78673237-24966700-78e2-11ea-897f-d836aca2a1d3.png)

Could someone else try the same flow and confirm it works? @Alexandru-Dumitru @katelynienaber @Colin-Stone @zFernand0

jellypuno:
This is the exacts steps that I've done.
1. I opened a JCL in **Profile # 1**. 
2. I Copied the contents of the JCL
3. I opened a dataset in **Profile # 2**
4. I right-clicked then select Create New Member
5. I entered the Member name
6. The editor opens and I pasted the contents that I copied in Step#2
7. I Saved it via Ctrl + S
8. Error Appears.

![contextValueError](https://user-images.githubusercontent.com/45042840/78678255-fec09080-78e8-11ea-8d86-bf42fbf96203.gif)

stepanzharychevbroadcom:
@jellypuno I did the same for the attached dataset and there's no error for me. Could you share full error log from the Debug Console, please?

katelynienaber:
[...]

PR #680, opened 2020-04-14
Fix issue 648 - Context value of undefined on new member save

katelynienaber:
Resolves issue #648.

I changed variable label to uppercase on line 1739 so that the labels will match
in the search starting on line 1765.

Signed-off-by: Katelyn Nienaber <katelyn.nienaber@broadcom.com>


Colin-Stone:
Please update with a meaningful title and link to related issue

Colin-Stone:
This one can be merged into master @zFernand0  is it worth us keeping the branch temporarily and decide if we want a v1.4.2 at some point or shall we just focus on getting v1.5.0 out?.

zFernand0:
> This one can be merged into master @zFernand0 is it worth us keeping the branch temporarily and decide if we want a v1.4.2 at some point or shall we just focus on getting v1.5.0 out?.

We can discuss more tomorrow (during refinement) but given all the merge conflicts, with a lot of PRs, it might be worth focusing more on 1.5.0 instead of 1.4.2 (unless we know of some customer that absolutely needs a fix before Monday or something similar.

zFernand0:
LGTM!
Two questions:
1. Could you resolve the merge conflicts? (I think the function was moved to some utils file) 😋 
2. Can we capture this in a unit test?

katelynienaber:
> LGTM!
> Two questions:
> 
> 1. Could you resolve the merge conflicts? (I think the function was moved to some utils file) 😋
> 2. Can we capture this in a unit test? (I'm just asking since it's labeled as a bug)

Done :3 Added a small unit test for this within saveFile

PR #689, opened 2020-04-15
feat: add support for deleting profiles

### LINKED ISSUE #332: Support deleting a profile
The idea is to let a user delete a profile because it is no longer needed. 

Today, a user can remove a profile which removes it from the Zowe Explorer UI but the profile remains stored in the .zowe > profiles folder. 
image

crawr:
### Addresses:
Fix #332 - Support deleting a profile

### Details:
- New feature: adds `deleteProfile()` under Profiles which can be executed via context menu and command palette
- The function makes use of the `CliProfileManager` to perform the delete
- Via the command palette, the user has to choose from the list of all profiles
- All the trees are updated to remove the successfully deleted profile
- Favorites related to the profile deleted are cleaned up
- "Remove Profile" context menu option is now called "Hide Profile" to avoid confusion

### Release Notes:
**Milestone:** v1.5
**Changelog:** Provide ability to delete profiles via the explorer.


katelynienaber:
Yes, works great for me :) Except for what Colin mentioned ofc

katelynienaber:
Heyoo, 2 things V:

1. The recall list isn't updated when I delete a profile
2. If I delete my default profile, there are a bunch of errors related to not having a default ZOSMF profile. Is there a way to set the default through the graphical interface?
3. If I delete my default/all profiles, the + button doesn't work to add a profile back again. I had to use the CLI to do it.

VitGottwald:
@crawr I am just testing this branch and I am not able to list files in a filter. After a long wait I get a 500 error.

Screenshot 2020-04-24 at 11 30 34

src/Profiles.ts
```diff
@@ -336,6 +338,178 @@ export class Profiles {
...
+                deletedProfile = await this.getDeleteProfile();
+                if (!deletedProfile) {
+                    return;
+                } else {
+                    deleteLabel = deletedProfile.name;
+                }
+            } else {
+                deletedProfile = node.getProfile();
+                deleteLabel = node.getProfileName();
+            }
```
VitGottwald, line 425: Would this work as well? (fewer iffs)
```suggestion
            if (!node){
                deletedProfile = await this.getDeleteProfile();
            } else {
                deletedProfile = node.getProfile();
            }
            if (!deletedProfile) {
                return;
            }
            deleteLabel = deletedProfile.name;
```

src/Profiles.ts
```diff
@@ -336,6 +338,178 @@ export class Profiles {
...
+            datasetTree.mSessionNodes.forEach((sessNode) => {
+                if (sessNode.getProfileName() === deleteLabel) {
+                    datasetTree.deleteSession(sessNode);
+                }
+            });
+            await datasetTree.refresh();
+
+            // Delete from Data Set Favorites
+            datasetTree.mFavorites.forEach((ses) => {
+                const findNode = ses.label.substring(1, ses.label.indexOf("]")).trim();
```
VitGottwald, line 443: This looks like trouble going forward 😁 But I do not know how to do it better at this point.

src/Profiles.ts
```diff
@@ -336,6 +338,178 @@ export class Profiles {
...
+            // Delete from Data Set Recall
+            const recallDs: string[] = datasetTree.getRecall();
+            for (let i = recallDs.length - 1; i >= 0; i--) {
+                const findNode = recallDs[i].substring(1, recallDs[i].indexOf("]")).trim();
+                if (findNode === deleteLabel) {
+                    datasetTree.removeRecall(recallDs[i]);
+                }
+            }
+
+            await datasetTree.refresh();
```
VitGottwald, line 459: I see three datastTree.refresh(); calls here, [here](https://github.com/zowe/vscode-extension-for-zowe/pull/689/files#diff-55ccdcb77f62eddb7a1fbf1079ed6bafR448) and [here](https://github.com/zowe/vscode-extension-for-zowe/pull/689/files#diff-55ccdcb77f62eddb7a1fbf1079ed6bafR439). Is it necessary? Would it not suffice to do just one at the end?

src/Profiles.ts
```diff
@@ -336,6 +338,178 @@ export class Profiles {
...
+            // Delete from USS Recall
+            const recallUSS: string[] = ussTree.getRecall();
+            for (let i = recallUSS.length - 1; i >= 0; i--) {
+                const findNode = recallUSS[i].substring(1, recallUSS[i].indexOf("]")).trim();
+                if (findNode === deleteLabel) {
+                    ussTree.removeRecall(recallUSS[i]);
+                }
+            }
+
+            await ussTree.refresh();
```
VitGottwald, line 486: Similarly to the datset refreshes - would one ussTree refresh ad the end suffice?

src/Profiles.ts
```diff
@@ -336,6 +338,178 @@ export class Profiles {
...
+            await jobsProvider.refresh();
+
+            // Delete from Jobs Favorites
+            jobsProvider.mFavorites.forEach((ses) => {
+                const findNode = ses.label.substring(1, ses.label.indexOf("]")).trim();
+                if (findNode === deleteLabel) {
+                    jobsProvider.removeFavorite(ses);
+                }
+            });
+            await jobsProvider.refresh();
```
VitGottwald, line 503: Same as for dataset and uss - would one refresh at the end be enoug?

src/Profiles.ts
```diff
@@ -336,6 +347,224 @@ export class Profiles {
...
+
+        if (sesName === undefined) {
+            vscode.window.showInformationMessage(localize("deleteProfile.undefined.profilename",
+                "Operation Cancelled"));
+            return;
+        } else {
+            zosmfProfile = allProfiles.filter((temprofile) => temprofile.name === sesName)[0];
+        }
+
+        return zosmfProfile;
```
VitGottwald, line 377: This can be simplified again because of the `return` statement at the end of `if`

```suggestion

        return allProfiles.filter(temprofile => temprofile.name === sesName)[0];
```

__tests__/__unit__/Profiles.unit.test.ts
```diff
@@ -397,6 +457,155 @@ describe("Profile class unit tests", () => {
 
     });
 
+    describe("Deleting Profiles", () => {
+        let profiles: Profiles;
```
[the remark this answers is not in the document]
crawr: I can see that. But I would prefer to defer this change when we refactor the tests for Profiles.ts since this affects not just the profile deletion.

__tests__/__unit__/extension.unit.test.ts
```diff
@@ -685,7 +685,7 @@ describe("Extension Unit Tests", () => {
         expect(createTreeView.mock.calls[0][0]).toBe("zowe.explorer");
         expect(createTreeView.mock.calls[1][0]).toBe("zowe.uss.explorer");
         // tslint:disable-next-line: no-magic-numbers
-        expect(registerCommand.mock.calls.length).toBe(68);
+        expect(registerCommand.mock.calls.length).toBe(72);
```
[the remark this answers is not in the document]
crawr: Agree, this can be fixed with the refactor

src/Profiles.ts
```diff
@@ -336,6 +347,224 @@ export class Profiles {
...
+            }
+
+            // Delete from Data Set Recall
+            const recallDs: string[] = datasetTree.getRecall();
+            for (let i = recallDs.length - 1; i >= 0; i--) {
+                const findNode = recallDs[i].substring(1, recallDs[i].indexOf("]")).trim();
+                if (findNode === deleteLabel) {
+                    datasetTree.removeRecall(recallDs[i]);
+                }
+            }
```
VitGottwald, line 450: This works. But I think there is a better way
```suggestion
            recallDs.slice().reverse()
                .filter(ds => ds.substring(1, ds.indexOf("]")).trim()  === deleteLabel)
                .forEach(ds => {
                    datasetTree.removeRecall(ds);
            });
```
because _reverse_ mutates the array in place, we call _slice_  to get a new copy of the array before reversing it. `[...recallDs]` would also work in place of the `recallDs.slice()`.

src/Profiles.ts
```diff
@@ -336,6 +338,178 @@ export class Profiles {
...
+            return;
+        }
+
+        const profileType = ZoweExplorerApiRegister.getMvsApi(deletedProfile).getProfileTypeName();
+        try {
+            this.deleteProf(deletedProfile, profileName, profileType);
+        } catch (error) {
+            this.log.error(localize("deleteProfile.delete.log.error", "Error encountered when deleting profile! ") + JSON.stringify(error));
+            await errorHandling(error, profileName, error.message);
+            throw error;
```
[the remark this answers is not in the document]
crawr: The error handling utility does not actually throw errors, just shows error messages. So the throw is necessary to stop the processing. There are plans to improve this under the "Centralized error handling" issue.

src/Profiles.ts
```diff
@@ -336,6 +347,224 @@ export class Profiles {
...
+                    datasetTree.removeRecall(recallDs[i]);
+                }
+            }
+
+            // Delete from Data Set Favorites
+            const favoriteDs = datasetTree.mFavorites;
+            for (let i = favoriteDs.length - 1; i >= 0; i--) {
+                const findNode = favoriteDs[i].label.substring(1, favoriteDs[i].label.indexOf("]")).trim();
+                if (findNode === deleteLabel) {
+                    await datasetTree.removeFavorite(favoriteDs[i]);
```
VitGottwald, line 457: 🤔 Interesting, why `await` for `.removeFavorite`, but not for `.removeRecall`?

crawr:
Thanks for the reviews!

> * If I delete my default profile, there are a bunch of errors related to not having a default ZOSMF profile. Is there a way to set the default through the graphical interface?
> * If I delete my default/all profiles, the + button doesn't work to add a profile back again. I had to use the CLI to do it.

This looks like a regression in the handling of the Profile loader. I fixed it so that the extension will still load even if the default profile is not set. An informational message will appear instead.
Note that this does not have impact on the case for new users that do not have any profiles at all.
In both cases, the next profile the user will create will be set as default in the meta yaml file.

> It may be worth quickly running npm run vscode:prepublish and committing Package.json and the i18n file get changed slightly. I don't know if the linter on the pipeline would flag it otherwise

Thanks for the tip, I ran the script and included the i18n file as a result.
 
> I am just testing this branch and I am not able to list files in a filter. After a long wait I get a 500 error.

I tested listing with my just my HLQ using regular DS filter search and `Ctrl + Alt + P` and it loaded the entire list without any problems. Please let me know if you were able to replicate it after the new commit.

VitGottwald:
Checked the latest commit `d4ae34d9609494fb354` and was able to list datasets. Then tried remove the (only) profile by right-clicking it and selecting _Delete Profile_ 
Screenshot 2020-04-27 at 12 42 05.

Got an error message as a result
Screenshot 2020-04-27 at 12 36 13

jellypuno:
> Checked the latest commit `d4ae34d9609494fb354` and was able to list datasets. Then tried remove the (only) profile by right-clicking it and selecting _Delete Profile_
> Screenshot 2020-04-27 at 12 42 05.
> 
> Got an error message as a result
> Screenshot 2020-04-27 at 12 36 13

@VitGottwald Are you sure? I just tried it now, I am not experiencing this error.

VitGottwald:
> > Checked the latest commit `d4ae34d9609494fb354` and was able to list datasets. Then tried remove the (only) profile by right-clicking it and selecting _Delete Profile_
> > Screenshot 2020-04-27 at 12 42 05.
> > Got an error message as a result
> > Screenshot 2020-04-27 at 12 36 13
> 
> @VitGottwald Are you sure? I just tried it now, I am not experiencing this error.

Thanks @jellypuno . Looks like it was a problem on my end. After recompiling and deleting/adding profile from CLI I am not getting the error anymore. 👍

src/Profiles.ts
crawr: I removed the await. I think it was a remnant from when i was debugging it

src/extension.ts
```diff
@@ -153,6 +153,16 @@ export async function activate(context: vscode.ExtensionContext): Promise<ZoweEx
             }
         });
     }
+    if (datasetProvider || ussFileProvider || jobsProvider) {
```
Colin-Stone, line 156: Not worth adding this so close to v1.5 but the three trees have a superclass ZoweTreeProvider so you could have ZoweTreeProvider.
No action required.

src/Profiles.ts
```diff
@@ -336,6 +347,218 @@ export class Profiles {
...
+            let faveJobs: string[] = jobsSetting.favorites;
+            sessJobs = sessJobs.filter( (element) => {
+                return element.trim() !== deleteLabel;
+            });
+            faveJobs = faveJobs.filter( (element) => {
+                return element.substring(1, element.indexOf("]")).trim() !== deleteLabel;
+            });
+            jobsSetting.sessions = sessJobs;
+            jobsSetting.favorites = faveJobs;
+            await vscode.workspace.getConfiguration().update(this.jobsSchema, jobsSetting, vscode.ConfigurationTarget.Global);
```
VitGottwald, line 553: There seems to be a lot of knowledge duplication here. How about this:

```suggestion
    // Delete from Session List
    const isNotDeletedSession = (element) => element.trim() !== deleteLabel;
    const isNotDeletedFavorite = (element) =>
      element.substring(1, element.indexOf("]")).trim() !== deleteLabel;

    const deleteFromSessionList = async (schema) => {
      const setting: {
        settings: string[];
        favorites: string[];
      } = vscode.workspace.getConfiguration().get(schema);
      await vscode.workspace.getConfiguration().update(
        schema,
        {
          ...setting,
          session: setting.favorites.filter(isNotDeletedSession),
          favorites: setting.favorites.filter(isNotDeletedFavorite),
        },
        vscode.ConfigurationTarget.Global
      );
    };

    deleteFromSessionList(this.dsSchema);
    deleteFromSessionList(this.ussSchema);
    deleteFromSessionList(this.jobsSchema);
```

zFernand0:
It had 2 or more approvals before resolving merge conflicts.
Merging!

PR #697, opened 2020-04-16
Manage Vsam files

Colin-Stone:
Introduces a new graphic and management for VSAM clusters


Colin-Stone:
> VSAM Icons works great. Although I cannot open the file. Before, I know can open some VSAM file but now I can't. (Maybe it change. I don't know.) Also, there's no error message telling me that I can't open a VSAM file. (I am also not sure if this is the normal behaviour)

Thanks Jelly. The Vsam files are always garbled just like looking at a database as zosmf returns binary information. The intent is that third parties will add the ability to open them and any other specific functionality. 
It's an interesting point however that you would expect a message to that effect but I have specifically excluded Vsam files from attempting to be oppened in Zowe Explorer because that would stop the third party adding it. To solve this I think we would need an extender specific API that could override any command already defined in Zowe Explorer package.json. This would come later because we should test the reaction from users/extenders rather than adding too much extender API now.
Would be interested in otehr opinions of course

jellypuno:
> Thanks Jelly. The Vsam files are always garbled just like looking at a database as zosmf returns binary information. The intent is that third parties will add the ability to open them and any other specific functionality.

- Due to my limited knowledge, I cannot imagine how this would work (for now) but it would be interesting to see.

> It's an interesting point however that you would expect a message to that effect but I have specifically excluded Vsam files from attempting to be opened in Zowe Explorer because that would stop the third party adding it. 

I was expecting something because nothing is happening. It is just like a list there but it doesn't do anything. But your explanation helps.

> To solve this I think we would need an extender specific API that could override any command already defined in Zowe Explorer package.json. This would come later because we should test the reaction from users/extenders rather than adding too much extender API now.
> Would welcome other opinions of course.

I am really interested in your vision for that extender API and how it will accommodate different 3rd part extension that will access that API. Is it focused CLI Plugins or will it also accomodate different VSCode extensions? I know HCL created a VSCode Extension that uses Zowe Explorer so APIs that allows them to access Zowe Explorer Capabilities would be great. Maybe we chould open an issue and expand this discussion to that issue :)

Colin-Stone:
> Since the design is to just identify the VSAM file and nothing will happen, I don't see any reason not to approve this change. Knowing as well that this is in preparation for future 'e-API' enhancements.

That's a fair assessment and I have already gone ahead and put this functionality into the Profile_link PR making the same judgement as you have. 
It is still here as a place-holder if it is decided to implement vsam but not profile linking

Colin-Stone:
This PR has been superceded by #705 and will be closed once that is merged

__tests__/__unit__/ZoweNode.unit.test.ts
```diff
@@ -90,6 +90,7 @@ describe("Unit Tests (Jest)", () => {
             new ZoweDatasetNode("BRTVS99.CA11.SPFTEMP0.CNTL", vscode.TreeItemCollapsibleState.Collapsed,
                                 rootNode, null, undefined, undefined, profileOne),
             new ZoweDatasetNode("BRTVS99.DDIR", vscode.TreeItemCollapsibleState.Collapsed, rootNode, null, undefined, undefined, profileOne),
+            new ZoweDatasetNode("BRTVS99.VS1", vscode.TreeItemCollapsibleState.None, rootNode, null, VSAM_CONTEXT, undefined, profileOne)
```
zFernand0, line 93: I'm always going to find it a bit strange to have to specify multiple `null` or `undefined` just to provide a 5th or 6th parameter. 😋 

No changes requested though!

__tests__/__unit__/__snapshots__/extension.unit.test.ts.snap
```diff
@@ -160,7 +155,7 @@ Array [
   Object {
     "command": "zowe.showDSAttributes",
     "group": "2_information",
-    "when": "view == zowe.explorer && viewItem =~ /^(pds.*|migr.*)/",
+    "when": "view == zowe.explorer && viewItem =~ /^(pds|migr|ds|vsam).*/",
```
zFernand0, line 158: Thanks for combining! 👍

Colin-Stone:
Superceded by #705

PR #705, opened 2020-04-17
Profile link work

Colin-Stone:
Removed from draft version as hoping to get into release 1.5. 

Currently the only extension API's included relate to retrieving the base profile for a tree item node or an associated profile.

Use this a sample extender https://github.com/Colin-Stone/zowe-explorer-test-mule


Colin-Stone:
Please note this branch is currently broken subsequent to rebasing master on it

Colin-Stone:
This branch contains the link dialog to associate the profiles which will create the yaml file. Or at least should do ;-) Also includes the vsam file updates which merges the three entries into one. You can't do anything with them but the intention is that extenders will be able to. 

If you download the test mule, build and package as a vsix file you can install that into Zowe Explorer and see how things start to integrate.

Colin-Stone:
Question for reviewers. One thing I am unsure of is if the link should be quite as up-front and maybe the inline is getting crowded. The alternative could be to use the meatballs icon but I would also need an additional dropdown to select the profile. 
image

jellypuno:
> Question for reviewers. One thing I am unsure of is if the link should be quite as up-front and maybe the inline is getting crowded. The alternative could be to use the meatballs icon but I would also need an additional dropdown to select the profile.

I liked the icon. For me that made sense. When I saw your demo I immediately understood what it is. But this is also good. Which is easier to code?

Colin-Stone:
> > Question for reviewers. One thing I am unsure of is if the link should be quite as up-front and maybe the inline is getting crowded. The alternative could be to use the meatballs icon but I would also need an additional dropdown to select the profile.
> 
> I liked the icon. For me that made sense. When I saw your demo I immediately understood what it is. But this is also good. Which is easier to code?

Thanks Jelly. The link is easier. I am glad you are happy with it as I know you are also adding menus

Colin-Stone:
Same PR as #724

src/utils/profileLink.ts
```diff
@@ -0,0 +1,134 @@
...
+                                     chosenType,chosenName,aProfile.type,aProfile.name));
+                } catch (err) {
+                    vscode.window.showErrorMessage(localize("profileLink.unableToSave", "Unable to save profile association. ") + err.message);
+                }
+            }
+        }
+    }
+}
+
+async function findLinkedProfile(aProfile: IProfileLoaded, type: string) {
```
zFernand0, line 81: From what I recall, we are trying to be more consistent with the blocking vs non-blocking methods. 
The direction (As far as I remember) was (wherever possible) try going with a non-blocking execution approach.
i.e. use `async-await` over nodejs sync methods.

@VitGottwald, @dkelosky , any thoughts

src/utils/profileLink.ts
```diff
@@ -0,0 +1,134 @@
...
+import { ZoweTreeNode } from "../abstract/ZoweTreeNode";
+import { IZoweTreeNode } from "../api/IZoweTreeNode";
+import { getZoweDir } from "../utils";
+import * as nls from "vscode-nls";
+const localize = nls.config({ messageFormat: nls.MessageFormat.file })();
+
+const LINKS_FOLDER = "profile_links";
+const FILE_SUFFIX = ".yaml";
+
+export async function getLinkedProfile(node: IZoweTreeNode, type: string, logger?: Logger) {
```
zFernand0, line 28: Functional vs OO? 
I believe we talked about this, but I don't want to disrupt the implementation too much (and I don't have a strong opinion in this scenario to go OO here, other than what we have discussed in the past)

src/utils/profileLink.ts
```diff
@@ -0,0 +1,134 @@
...
+        return (node as ZoweTreeNode).getProfile();
+    }
+    throw new Error(localize("profileLink.notTreeItem", "Tree Item is not a Zowe Explorer item."));
+}
+
+export async function linkProfileDialog(aProfile: IProfileLoaded) {
+    let chosenName;
+    let chosenType;
+    if (aProfile) {
+        const possibles = Profiles.getInstance().getAllTypes().filter( (value) => value !== aProfile.type);
```
zFernand0, line 53: Going back to one of my previous comments (having duplicate types), I assume this is where the impact will be reflected?
Maybe it's just not possible to get duplicates and I missed something when looking at the `allTypes` array creation.

src/Profiles.ts
```diff
@@ -336,6 +344,32 @@ export class Profiles {
...
+            return profile.type === type;
+        });
+        return profilesForType.map((profile)=> {
+            return profile.name;
+        });
+    }
+
+    public async directLoad(type: string, name: string): Promise<IProfileLoaded> {
+        let directProfile: IProfileLoaded;
+        const profileManager = await new CliProfileManager({
```
zFernand0, line 363: Would it be possible to reuse `this.getCliProfileManager`? (Please forgive me if I'm missing something here)

__tests__/__unit__/Profiles.unit.test.ts
```diff
@@ -540,4 +542,28 @@ describe("Profile class unit tests", () => {
         await theProfiles.checkCurrentProfile(testIProfile);
         expect(theProfiles.validProfile).toBe(ValidProfileEnum.INVALID);
     });
+
+    it("Tests getAllTypes", async () => {
```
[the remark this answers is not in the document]
Colin-Stone: Yes it was self-explanatory so I thought I would just be waffling more words (as usual)

__tests__/__unit__/utils/profileLink.unit.test.ts
```diff
@@ -0,0 +1,260 @@
...
+    });
+
+    it("Test get profile via the API - Bad input", async () => {
+        let success = false;
+        const extenderApi = ZoweExplorerApiRegister.getExplorerExtenderApi();
+        try {
+            await extenderApi.getProfile(aBadNode as IZoweDatasetTreeNode);
+        } catch (error) {
+            expect(error.message).toEqual("Tree Item is not a Zowe Explorer item.");
+            success = true;
```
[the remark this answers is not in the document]
Colin-Stone: Got a neat fix for this then...

src/utils/profileLink.ts
Colin-Stone: I didn't really feel it was an OO concept and that it was simpler doing it this way. Could have used a singleton but just felt right like this as a support function rather than class

src/utils/profileLink.ts
Colin-Stone: We have async-await throughout. That's what VSCode recommends I believe.

src/api/ZoweExplorerApiRegister.ts
```diff
@@ -216,4 +226,13 @@ export class ZoweExplorerApiRegister implements ZoweExplorerApi.IApiRegisterClie
                 localize("getJesApi.error", "Internal error: Tried to call a non-existing JES API in API register: ") + profile.type);
         }
     }
+
+    /**
+     * Lookup of an API implementation for JES for a given profile.
+     * @param {IProfileLoaded} profile
+     * @returns an instance of the API for the profile provided
+     */
+    public getExplorerExtenderApi(): ZoweExplorerApi.IApiExplorerExtender {
```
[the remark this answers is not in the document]
Colin-Stone: Well the doc comments were totally incorrect to start with. I've tried to fit in with Peter's structure but there is only the one instance of the ExplorerExtenderApi.  
From the extender activate method in extension.ts you call the following
 ```
   const baseExt = vscode.extensions.getExtension('zowe.vscode-extension-for-zowe');
    try {
        if (baseExt && baseExt.exports) {
            EXTENDER_API = baseExt.exports.getExplorerExtenderApi();
```
You are correct that if you were using Peter's profile switching version the 'given profile' would be relevant but in this case it's not and it delivers a singleton instance. The single instance of the ZoweExplorerExtender is automatically initialized at loadtime

File: src/utils/profileLink.ts (hunk 3 above)
Comment (Colin-Stone) [reply src/utils/profileLink.ts:53]: I can't find a previous comment about duplicate types did you mean zosmf to zosmf?. No i wouldn't let you choose that.
Here I am just finding a list of possible alternative types to (Normally zosmf) I did think I maybe should have extended this to exclude types that had no profile instances defined. But I only thought it.

File: src/Profiles.ts (hunk 4 above)
Comment (Colin-Stone) [reply src/Profiles.ts:363]: Good spot. Will update

File: src/utils/profileLink.ts (hunk 1 above)
Comment (zFernand0) [reply src/utils/profileLink.ts:81]: We can address this in a future issue though 😋

File: src/utils/profileLink.ts (hunk 3 above)
Comment (zFernand0) [reply src/utils/profileLink.ts:53]: Sorry, I must've deleted (or not added) my previous comment.

Here is the block of code I was referring to:
https://github.com/zowe/vscode-extension-for-zowe/blob/b8fb7f9e8ecb23ebebee0aa1cb7ee234df6c3375/src/Profiles.ts#L127-L131

So, if I have multiple ftp or tso profiles, will it be mad (i.e. show ftp more than once when selecting the types) because of the possible duplicates?

Comment (zFernand0) [reply src/utils/profileLink.ts:53]: Maybe it's fine since if `.configurations` only contains one element per type?

File: __tests__/__unit__/utils/profileLink.unit.test.ts (hunk 6 above)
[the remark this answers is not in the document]
Comment (zFernand0) [reply __tests__/__unit__/utils/profileLink.unit.test.ts]: Brilliant way to fix it! 😋 
Now we will see the message showing up in case the unit test fails

File: src/utils/profileLink.ts (hunk 3 above)
Comment (zFernand0) [reply src/utils/profileLink.ts:53]: I was being dumb, configurations are the same as what we see for profile creation, a list of profiles types available, no matter the number of profiles.

I should've done my research before complaining!

File: docs/README-Extending.md [hunk 8]
```diff
@@ -0,0 +1,79 @@
...
+		"zowe.vscode-extension-for-zowe"
+	],
+
+```
+The addition of this script also enforces that the Zowe Explorer is a pre-requisite package on the users machine which will install it if necessary 
+when the extender package is installed. This also ties the activate functionality of both packages which enables access to the Zowe Explorer API
+
+## Contextual Hooks
+
+To enable menus, extenders can reference the context values associated with the individual Tree Items. Zowe Extender uses a strategy of adding and removing
```
tjohnsonBCM, line 25: "Zowe Extender" -> "Zowe Explorer"

src/api/ZoweExplorerApiRegister.ts
```diff
@@ -62,6 +63,15 @@ export class ZoweExplorerApiRegister implements ZoweExplorerApi.IApiRegisterClie
         return ZoweExplorerApiRegister.getInstance().getJesApi(profile);
     }
 
+    /**
+     * Static lookup of generic extender API implementation.
+     * @private
```
tjohnsonBCM, line 68: `@private` -> `@static`

docs/README-Extending.md
Colin-Stone: I redeveloped the documentation as https://github.com/zowe/vscode-extension-for-zowe/pull/720. This was really here as a place-holder. Sorry for the confusion.

Colin-Stone:
> One other thing, should the QuickPick to "select a type of alternative profile" only show profile types for which I have at least one profile created?

Thanks Tim yes I thought exactly the same thing. However I was intending not to change anything now so close to release. I will raise an issue to track this and it can go in later.

src/api/ZoweExplorerApiRegister.ts
Colin-Stone: Good spot. I notice that the other API ones (which I copied as a template) are also wrong. I will raise another issue for the original developers to look at this. Raised https://github.com/zowe/vscode-extension-for-zowe/issues/729

zFernand0:
Thanks again for implementing!
Merging.

Issue #631, opened 2020-03-17
Create New Connection needs to work with other profile types


phaumer:
The current Create New Connection UI is hard-coded for z/OSMF only. To support the Extensibility API it needs to support other types of profiles such as zFTP dynamically as each profile type will need to ask different questions to the user. 

This can be done by generating the prompts based on the meta data stored with each profile folder (e.g. `~/.zowe/profiles/zosmf/zosmf_meta.yaml` for z/OSMF). These files have all the information needed, including the text strings that describe each property.

We propose he following changes:
- Re-implement Profiles.ts to support MVC principles: 
  - a model for reading and updating the profiles 
  - a view to render the user prompts dynamically
  - a controller that reads the meta-data of the type requested and drives the control flow between view and model

phaumer:
https://github.com/JillieBeanSim will help with implementation

jelaplan:
I'm wondering if all fields in the meta data will appear as questions in the quick pick? We may want to have some fields that are used in CLI but not appear in the Zowe Explorer quick pick.

jellypuno:
@dkelosky @jelaplan do we need to consider this change in profiles with tokens?

phaumer:
@jelaplan I agree that the text in the meta-data might not always match well, but I see this as a quick fix to get the other profile types working.  

Afterwards I would like to get back to the discussion we started here: https://github.com/zowe/vscode-extension-for-zowe/issues/423 and some of the other ideas mentioned there such as the automation of profile creation from @jalel01. I would like to learn and discuss more around that.

jelaplan:
@jellypuno WRT tokens, need to talk that over with Dan. I've started writing a response here but keep deleting it because my comments are likely not to bring clarity.

jelaplan:
@phaumer How do you anticipate the user choosing between zOSMF and FTP as a connection method?

dkelosky:
sorry @jellypuno - sometimes I miss these.  I don't know if I fully understand the proposal - hopefully the token additions to the profile are transparent to the proposal, just an addition of extra fields in the profile.

jelaplan:
I imagine the choice of FTP or z/OSMF being part of the quick pick. But for many users, especially z/OSMF capable users, it would be nice to not ask that question at all. One way to handle that would be to add a split button. If they click the + icon, they use zOSMF and proceed to setup a connection. If they click the right side arrow, they get a menu to  choose between z/OSMF and FTP or additional choices.

[...]

PR #717, opened 2020-04-22
Create and Edit Connection with any other profile types (#631)

phaumer:
See Issue #631 for details. The goal was to extend the Create Connection UI with the ability to create connections graphically when other API extensions are registered. 

You can test this with the FTP extension here: https://github.com/phaumer/zowe-explorer-ftp-extension

It will generate and prompt new questions based on the FTP profile's meta-data.

This work was implemented by my colleague @JillieBeanSim who is not yet a squad member. I submit this PR on her behalf.


phaumer:
@Colin-Stone @zFernand0 we are proposing this change for the 1.5 release.

jellypuno:
HI, Kindly increase the coverage.

Capture

Also, Please tag the appropriate issue and milestone.

Colin-Stone:
Please use the CodeCov reports and links above to identify those areas particulalrly in Profiles that have not been covered by unit testing. Alternatively the output from npm run test:unit provides a statistical report of those lines not covered or tested. Look at the entry for Profiles.
There is an integration test that tests profile creation, did you consider creating an equivalent for aa alternative type?

Colin-Stone:
I attempted to run the integration tests and had 8 failures and it looks as if the original branch you have based upon is missing some functionality that is now in the master branch. Specifically Profile.ts

JillieBeanSim:
@Colin-Stone I have brought the code coverage up from:
image
to:
image

Colin-Stone:
Thanks Billie-Jean integration tests are now correct. We still need to focus on specific unit testing. If you look at the image below I have highlighted specific lines of new code that is not being touched by testing.
image
You also get a similar report via 
image
image
What is shown is areas where your code is not being exercised and it's critical we do this to protect from any subsequent inadvertent changes that may happen that changes the behavior.

jellypuno:
UX Comment: 

- host.com is not really clear. Especially if I am already a Zowe Explorer user
- If I add https:// in url, it will say ENOTFOUND. Similar to point 1 if I am used to adding https://, it would be great to handle and concatenate this info.
- Thank you for adding the basePath. I would suggest to add Optional in the placeholder information so that means that I am not forced to add it something in it.
- Too much words as well. It is hard to read it. @jelaplan any UX suggestions?

Capture

JillieBeanSim:
@jellypuno if https:// is added it will be trimmed the port as well with no error for user (this was discussed to have), I can add optional in the placeholder for base path but the wording is grabbed from definition in schema

jellypuno:
> @jellypuno if https:// is added it will be trimmed the port as well with no error for user (this was discussed to have), 

"Discussed to have" meaning it will be added soon? ok then.

I can add optional in the placeholder for base path but the wording is grabbed from definition in schema

Ok. I guess. it just looks weird and cluttered. Maybe other are okay with that.

src/Profiles.ts
```diff
@@ -141,113 +131,231 @@ export class Profiles {
...
+                if (!host) {
+                    vscode.window.showInformationMessage(localize("createNewConnection.zosmfURL",
+                        "No valid value for z/OS Host. Operation Cancelled"));
+                    return undefined;
+                }
+                try {
+                    if (host.includes("/") || host.includes(":")) {
+                        const result = this.parseUrl(host);
+                        host = result.host;
+                    }
```
lauren-li, line 222: > "Discussed to have" meaning it will be added soon? ok then.

@jellypuno, I think @JillieBeanSim was meaning say the code already handles trimming of `https://` and port number if the user includes those in the host URL. I highlighted the code I believe is taking care of this, and I am able to successfully create working profiles while entering URLs that both do and do not contain `https://` and/or port number. If you don't mind sharing, what string are you entering to get the ENOTFOUND error for host URL?

Regarding the wordiness of the field description for base path, I agree the text is a bit on the long side, but at the same time, this is something that would be controlled by the profile schema defined in Zowe CLI (for zosmf) or a Zowe Explorer extender (for other profile types), rather than the code in this PR. Grabbing the field descriptions from these defined profile schemas rather than hard-coding them into the Zowe Explorer code allows more flexibility and scalability/maintainability in the types of profiles that can be created and used with Zowe Explorer.

Billie, please correct me if I'm wrong with my interpretations, thanks.

katelynienaber:
@JillieBeanSim I currently still receive ENOTFOUND if I create a profile with https://

Also would prefer a shorter text too. Does anyone have an opinion on shortening it to:
_The base path for your API mediation layer instance. See the [documentation](https://docs.zowe.org/stable/web_help/index.html?p=zowe_profiles_create_zftp-profile) for more info._

And then we would have to add info about the base path to the docs. Because it currently doesn't exist there, as far as I can tell...

Colin-Stone:
> Also would prefer a shorter text too. Does anyone have an opinion on shortening it to:
> _The base path for your API mediation layer instance. See the [documentation](https://docs.zowe.org/stable/web_help/index.html?p=zowe_profiles_create_zftp-profile) for more info._

I agree with Katelyn on the shorter text. Looks a bit of a mess with that much text

src/Profiles.ts
jellypuno: > code already handles trimming of https:// and port number if the user includes those in the host URL. 

@lauren-li I tried doing it today and yes it is working. Couple of things that i saw.

_ If I entered `https://url:port`, I am still going to be ask for a port. If I didn't put anything it will automatically change it to 443. Is it possible to extract the hostname and port if I give the whole URL? 

- If I entered `url:port` (without https) the url ended up being spaces in the yaml file.

src/Profiles.ts
```diff
@@ -141,113 +131,231 @@ export class Profiles {
...
+                schemaValues[value] = rejectUnauthorize;
+                break;
+            default:
+                const description: string = schema[value].optionDefinition.description.toString();
+                switch (schema[value].type) {
+                    case "string":
+                        options = await this.optionsValue(value, schema);
+                        const profValue = await vscode.window.showInputBox(options);
+                        schemaValues[value] = profValue;
+                        break;
```
lauren-li, line 310: > > Also would prefer a shorter text too. Does anyone have an opinion on shortening it to:
> > _The base path for your API mediation layer instance. See the [documentation](https://docs.zowe.org/stable/web_help/index.html?p=zowe_profiles_create_zftp-profile) for more info._
> 
> I agree with Katelyn on the shorter text. Looks confusing with that much text
> _The base path for your API mediation layer instance._ Would suffice

@Colin-Stone @katelynienaber It looks like the base path property is currently being handled by this `default` case, which means the description text is retrieved from the schema defined by Zowe CLI or the Zowe Explorer extender. So as-is, the text shortening would have to be done in those respective repositories where they define the profile schema, rather than by Zowe Explorer.
- Alternatively, a base path `switch` case could be added into Zowe Explorer, similar to what was done for common profile properties such as `host`, `username`, `password`, etc. - In that case, a shortened base path description could be hard-coded there.

(Currently, there are `switch` cases for `host`, `port`, `user`, `password`, and `rejectUnauthorized`. The `default` case is used when the profile type the user is creating contains additional properties.)

katelynienaber: Sorry, my mistake!

jellypuno: @tjohnsonBCM @zFernand0 Do you think it would be possible to handle the basePath description in zowe/CLI ? Can we open an issue for this?

src/Profiles.ts
lauren-li: Thanks @jellypuno for pointing out these cases. I am seeing them as well, and they should be addressed in this PR.

JillieBeanSim: > * If I entered `https://url:port`, I am still going to be ask for a port. If I didn't put anything it will automatically change it to 443. Is it possible to extract the hostname and port if I give the whole URL?
> * If I entered `url:port` (without https) the url ended up being spaces in the yaml file.

@jellypuno thanks I will correct this issue,  I tested with `https://...` and that trim was working but not `url:port`.  Not sure what you are asking about the port being included.  Are you wanting that prompt to not be presented if the port was entered in with the host or did you want the port entered within the host to be the value in the textbox vs it showing the default value?

jellypuno: hmmm. good point. I was thinking that if I entered `https://url:port`, the port dialog box will not be displayed anymore. But you are right. Maybe it is also good to display for confirmation but instead of 443, it will be the value that I provided. I dont know the better UX for this. I prefer the first one(not to display it). but I don't know. what do you think @lauren-li ?

lauren-li: @jellypuno I can see either way working. On one hand, it's more consistent to present the prompt for the port number, using the user's value if they entered it in with the host, or using the default if not. On the other hand, that would make the user have to do an extra step to hit "Enter" again, so it would be more streamlined to just skip the `port number` prompt and use the user's value if they already entered it in as part of the `host` value. I think either option would work.

JillieBeanSim:
@jellypuno those host fixes are in now, could you test it and see if it's what you were looking for? a side note: if `host:port` is entered since port is not a number it will continue to ask for port but if `host:123` is entered it should skip asking for port

JillieBeanSim:
Hey @jellypuno the fix is in, could you give it another test please

jellypuno:
@JillieBeanSim Works great! Thank you for the changes.

I notice something again and apologies for not seeing it earlier,

If I created a profile using zowe-cli without basePath, it looks like this:

```
host: zos123
port: 1443
user: ibmuser
password: myp4ss
rejectUnauthorized: true
```

If I created one using this PR, it looks like this:

```
port: 123
host: url
user: hey
password: ho
rejectUnauthorized: true
basePath: ""
```

I have 2 things:
- host should be first and then port (minor stuff)
- If basePath is spaces, do not write it.

src/Profiles.ts
tjohnsonBCM: The description text is being fetched from the z/OSMF profile meta definition in `~/.zowe/profiles/zosmf/zosmf_meta.yaml`. If you want it to be shortened, we could an open an issue requesting a change in [these lines of code in Zowe CLI where the description originates](https://github.com/zowe/zowe-cli/blob/59820d6cefed5916d4807de78b90c68f4e56e237/packages/zosmf/src/ZosmfSession.ts#L88-L90).

JillieBeanSim:
> @JillieBeanSim Works great! Thank you for the changes.
> 
> * host should be first and then port (minor stuff)
> * If basePath is spaces, do not write it.

Ok, I have a fix for this on my fork.  Let me get @lauren-li to check it out then we will get it pushed in for you.  Thanks @jellypuno

JillieBeanSim:
@jellypuno we got the changes you requested pushed in

src/Profiles.ts
```diff
@@ -143,130 +134,291 @@ export class Profiles {
...
+                        if (host.includes("/")) {
+                            const result = this.parseUrl(host);
+                            if (result === undefined) {
+                                return undefined;
+                            } else {
+                            host = result.host;
+                            if (result.port !== null) {
+                                port = result.port;
+                            }}
+                        } else {
```
tjohnsonBCM, line 260: These if/else branches appear to contain essentially the same logic and could be replaced by:
```typescript
if (!host.includes("/")) {
    host = "https://" + host;
}
const result = this.parseUrl(host);
...
```

src/Profiles.ts
```diff
@@ -143,130 +134,291 @@ export class Profiles {
...
+                                } else {
+                                    schemaValues[value] = undefined;
+                                }
+                            }
+                        break;
+                    default:
+                        options = await this.optionsValue(value, schema);
+                        const defaultValue = await vscode.window.showInputBox(options);
+                        switch (defaultValue){
+                            case "True" || "true":
```
tjohnsonBCM, line 405: Why are Boolean values parsed here, since in the `default` case the type of the profile field is not "boolean"? Sorry, I'm probably missing something 😢

JillieBeanSim: there is a type of `type: -boolean -null`,`type: -string -null`, and/or `type: -number -null` in the zftp cli plugin 
image

Colin-Stone:
The createNewConnection had the overriding profileType field removed. The intention going forward when creating new connections applies to all possible CLI profiles not just those associated with registered API providers. Other extenders use different mechanisms to associate with other profiles. My suggestion is to use,
```javascript
    public async createNewConnection1(profileName: string, requestedProfileType: string): Promise<string | undefined> {
..
const profileType = requestedProfileType ? requestedProfileType : await this.getProfileType();
```

jellypuno:
@Colin-Stone @phaumer Do you think it is a good idea to include creation of SSH and TSO profiles too. When I clicked the Associate profiles icon, it dawned on my that I have no way to create those profile but through CLI. So far the design of this PR is to just create z/osmf profile if you do not have extra plugins.

JillieBeanSim:
@jellypuno I have gotten the PR updated with the changes needed to work with v1.5 release if you want to check it out now.
@tjohnsonBCM majority of the comments you left above have also been addressed

jellypuno:
HI @JillieBeanSim, Can you add trim in the profile name so I won't be allowed to add a profile if it is spaces.

Capture

Question or Clarification, If I wanted to try to add an FTP profile, I should install the FTP VSCode extension and FTP C plugin?

JillieBeanSim:
> Question or Clarification, If I wanted to try to add an FTP profile, I should install the FTP VSCode extension and FTP C plugin?

@jellypuno yes you can get the FTP VS Code ext here https://github.com/phaumer/zowe-explorer-ftp-extension

I will get @lauren-li to push the fix for the profile name once she gets online and ping you.

jellypuno:
Thank you @JillieBeanSim , I saw that you are also fixing issue #746 on deleting recall in settings.json

JillieBeanSim:
> Thank you @JillieBeanSim , I saw that you are also fixing issue #746 on deleting recall in settings.json

Cool, @jellypuno could you begin testing this PR for functionality, after yesterday's scrum we have merged in the Edit/Update work to handle different profile types here as well.  We are working on the DTO error and the theia-test error was green yesterday so we may need to bring in any changes to Master to get it back green again.  Thank you for your hard work and time looking into this PR.

jellypuno:
@JillieBeanSim will do my best! thanks :)

lauren-li:
I noticed that this PR also fixes this error message I was seeing while trying to edit zosmf profiles with the FTP extension installed:
Screen Shot 2020-05-05 at 4 27 50 PM

zFernand0:
Squashing!
Thanks for the contribution!

zFernand0:
See commit history in https://github.com/zowe/vscode-extension-for-zowe/commit/47c8c0fcccdcbac44f0d58b2bc9cc60ddfd50271

PR #719, opened 2020-04-22
Issue #718: Add back addRecall to openPS

katelynienaber:
Added back addRecall to function openPS and fixed unit tests to match.

[Issue #718](https://github.com/zowe/vscode-extension-for-zowe/issues/718)

PR #723, opened 2020-04-23
Recall data sets

### LINKED ISSUE #713: Allow data sets to be recalled

Pranay154:
Resolves #713 
Adds option to recall a data set when it is in a migration state.


__tests__/__integration__/extension.integration.test.ts
```diff
@@ -603,78 +603,198 @@ describe("Extension Integration Tests", () => {
...
+                        zowe.Delete.dataSet(
+                            sessionNode.getSession(),
+                            dataSetName
+                        ),
+                        zowe.Create.dataSet(
+                            sessionNode.getSession(),
+                            zowe.CreateDataSetTypeEnum.DATA_SET_SEQUENTIAL,
+                            dataSetName,
+                        ),
+                    ].map((p) => p.catch((err) => err)));
```
CForrest97, line 619: I think you're doing this because you want the data set to be reset but I dont like the implementation.
I have done this Promise.all trick in the past for promises where I dont care if they fail but because they are both dependent on the same resource I would run them sequentially.
A solution I would try is to await the first promise and return the error instead of throwing in case the ds doesnt exist yet
Then await the second promise but dont change the catch for this promise as if we cant create the ds we cant really run the tests

katelynienaber:
Hello, is it possible to rename this? Because as @jellypuno mentioned, there already exists a recall function in the extension...

https://github.com/zowe/vscode-extension-for-zowe/pull/719#issuecomment-618955299

Pranay154:
> Hello, is it possible to rename this? Because as @jellypuno mentioned, there already exists a recall function in the extension...
> 
> [#719 (comment)](https://github.com/zowe/vscode-extension-for-zowe/pull/719#issuecomment-618955299)

Hey @katelynienaber, thanks for your comment. I had a look at your PR, just for some clarification, what does your recall function do? The recall function I'm implementing here is to recall migrated data sets and it is a part of the hsm commands, for reference this is what I'm talking about: https://www.ibm.com/support/knowledgecenter/SSLTBW_2.2.0/com.ibm.zos.v2r2.izua700/IZUHPINFO_API_PutDataSetMemberUtilities.htm 

I have also implemented the hMigrate and hRecall commands in the CLI and I wanted to keep consistent with my implementation there, where I call the commands migrate and recall respectively.

katelynienaber:
@Pranay154 Recall is part of a functionality which allows users to open recently-opened files with CTRL+ALT+R, similar to opening recent files in VSCode with CTRL+R

Would it be possible for both of us to rename these functions? So it is clear what both are doing, for anyone who comes after us.

Like I would rename mine to `recentFileRecall` and you could use something like `RecallDataSet` or whatever you think makes sense...

jellypuno:
@Pranay154 @katelynienaber Ignore the naming. maybe it is just me. If I read the code, I would understand the purpose of both recalls. This change uses hRecall anyway and I would understand that mainframe reference.

__tests__/__integration__/extension.integration.test.ts
```diff
@@ -603,78 +603,198 @@ describe("Extension Integration Tests", () => {
...
-                it("Should send a migrate request", async () => {
-                    let error;
-
-                    try {
-                        const node = new ZoweDatasetNode(dataSetName, vscode.TreeItemCollapsibleState.None, sessionNode, session);
-                        node.contextValue = DS_DS_CONTEXT;
+                afterEach(async () => {
+                    await Promise.all([
+                        zowe.Delete.dataSet(sessionNode.getSession(), dataSetName).catch((err) => err),
+                    ].map((p) => p.catch((err) => err)));
```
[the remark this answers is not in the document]
Pranay154: Hey @CForrest97 , thanks for reviewing this, as discussed offline, ill remove the `Promise.all`'s

__tests__/__integration__/extension.integration.test.ts
```diff
@@ -603,78 +603,170 @@ describe("Extension Integration Tests", () => {
...
                     await zowe.Delete.dataSet(sessionNode.getSession(), dataSetName).catch((err) => err);
                     await zowe.Create.dataSet(
-                        sessionNode.getSession(),
-                        zowe.CreateDataSetTypeEnum.DATA_SET_SEQUENTIAL,
-                        dataSetName,
-                    );
+                            sessionNode.getSession(),
+                            zowe.CreateDataSetTypeEnum.DATA_SET_SEQUENTIAL,
+                            dataSetName,
+                        ).catch((err) => err);
```
CForrest97, line 614: I think this has been indented once too many and I dont think this catch is needed

__tests__/__integration__/extension.integration.test.ts
```diff
@@ -603,78 +603,170 @@ describe("Extension Integration Tests", () => {
...
+                }
+                expect(error).to.not.equal(undefined);
+            }).timeout(TIMEOUT);
+        });
+    });
+});
 
-                it("Should fail if data set doesn't exist", async () => {
-                    let error;
+    describe("Recalling a data set", () => {
```
CForrest97, line 687: This could be unindented

__tests__/__integration__/extension.integration.test.ts
```diff
@@ -603,78 +603,170 @@ describe("Extension Integration Tests", () => {
...
+                } catch (err) {
+                    error = err;
+                }
+                expect(error).to.be.equal(undefined);
+            }).timeout(TIMEOUT);
+        });
+    });
+    describe("Failure Scenarios", () => {
+        describe("Recall a sequential data set", () => {
+            const dataSetName = `${pattern}.TEST.FAIL`;
```
CForrest97, line 752: I think Ive been recommended before to use a more random name like `pattern.test.A23HBG9` to ensure the ds doesnt exist

src/api/ZoweExplorerApi.ts
```diff
@@ -293,6 +293,16 @@ export namespace ZoweExplorerApi {
             dataSetName: string,
         ): Promise<zowe.IZosFilesResponse>;
 
+        /**
+         * Recalls a data set member.
```
CForrest97, line 297: Doesnt this only recall a data set?

Pranay154: Yep it does, good spot!

__tests__/__integration__/extension.integration.test.ts
Pranay154: @CForrest97 thanks for your comment, I have been told before that using `TEST.FAIL` is a good way to name for failure scenarios, but I would be happy to change this if you think that the way recommended to you is better

__tests__/__integration__/extension.integration.test.ts
Pranay154: Yep, the catch isn't need there, good spot (again)

Pranay154:
> Unable to fully run integration tests due to system issues (and integration tests are fine since we haven't incorporated the standards there yet).
> Also, should the unit tests be part of the dataset actions unit test and not the extension.unit.test.ts file?

Hey @zFernand0, that's correct, my bad, I thought I had done it in the `actions.unit.test.ts` file originally, ill change this now

JillieBeanSim:
> Tested functionality and this works well. All unit and system tests are passing, including the added ones. Thanks @Pranay154 for your hard work on this!
> ~
> One thing to consider in a future issue is that we may want to revise our grouping/ordering of commands in the right-click context menu. To me, this is starting to look a little messy and inconsistent with how VS Code groups/orders their context menu options:
> Screen Shot 2020-05-26 at 11 42 29 AM

I didn't notice the Migrate & Recall showing up in the right-click menu at the same time if the Data set is available (not migrated).  It should only show the migrate option.  Likewise once migrated it should only show Recall option.

Pranay154:
@lauren-li @JillieBeanSim Thanks for taking the time to review this 😄 ! I didn't notice that recall appears when the data set **isn't** migrated, thanks for bringing it to my attention, let me have a look into this now

Pranay154:
> @lauren-li @JillieBeanSim Thanks for taking the time to review this 😄 ! I didn't notice that recall appears when the data set **isn't** migrated, thanks for bringing it to my attention, let me have a look into this now

@JillieBeanSim @lauren-li @zFernand0 The recall option should now only show up when a data set is migrated, could you confirm this?

Pranay154:
> One thing to consider in a future issue is that we may want to revise our grouping/ordering of commands in the right-click context menu. To me, this is starting to look a little messy and inconsistent with how VS Code groups/orders their context menu options:
> Screen Shot 2020-05-26 at 11 42 29 AM

@lauren-li completely agree, it is quite cluttered and can be simplified.

zFernand0:
It seems that we've never had unit tests for the `ZoweExplorerZosmfApi` which is why I believe there were none added here.

I know it may seem silly to add a test just to verify that the proper zowe function (from the cli package) was called with teh expected parameters, but I think that's good enough to ensure code-coverage of the file mentioned above.

Will create an issue for this, and apologies for missing it.

Issue #528, opened 2020-02-18
Refinements for Extensibility API activation


phaumer:
The following issues need to be addressed for extending VS Code extensions using the Zowe Explorer extensibility API such as the FTP extension (https://github.com/phaumer/vscode-extension-for-zowe-api-sample):

- Activation of other extensions happens after the Zowe Explorer. Other extensions need to be allowed to come in so that
  - the Explorer views can get refresh to see the other profile appearing automatically
  - the profiles listed in previous sessions should not be removed, but ignored until the other extensions are activated again
  - Favorites created by other extension should not throw errors as the other extension is not activem yet

phaumer:
Started work here: https://github.com/phaumer/vscode-extension-for-zowe/tree/add-profile-refresh-to-api

phaumer:
Remaining work was extracted to https://github.com/zowe/vscode-extension-for-zowe/issues/742

PR #725, opened 2020-04-24
Refresh profile cache when new type is registered (Issue #528)

phaumer:

This implements the first part of https://github.com/zowe/vscode-extension-for-zowe/issues/528. An extension such as FTP will activate after Zowe Explorer and register its profile type. The user then always have to first click the Refresh button to be able to see the profiles. This PR will automatically refresh when a new type gets registered.


Colin-Stone:
I don't have an environment to check this except by eyeballing it. @lauren-li are you able to test this function is working as intended?.

src/api/ZoweExplorerApiRegister.ts
```diff
@@ -95,6 +96,8 @@ export class ZoweExplorerApiRegister implements ZoweExplorerApi.IApiRegisterClie
     public registerUssApi(ussApi: ZoweExplorerApi.IUss): void {
         if (ussApi && ussApi.getProfileTypeName()) {
             this.ussApiImplementations.set(ussApi.getProfileTypeName(), ussApi);
+            // reload the internal profiles cache to add the newly added profiles
+            Profiles?.getInstance()?.refresh();
```
tjohnsonBCM, line 100: Is the usage of `?` as a safe navigation operator desired here, or should an error be raised if the call to `Profiles.getInstance` fails?

phaumer: @tjohnsonBCM thanks for the review. No, the constructor of this singleton will call these method before the Profiles singleton is initialized. In that case no refresh is needed, yet. So only if Profiles already exist, after activation finished, this is needed to be called.

phaumer:
@lauren-li did some more testing and because we were issuing all these refresh requests without an await we observed a race condition when more than one extender would register their APIs (e.g. ftp and rse).  One extension's profiles would not show up as the other parallel refresh would not load them. We also did not want to break the API signatures by returning promises now. 

Hence, we decided to let the extender decide when to call the refresh and to add a new optional method to the interface that will not break anyone, if they work with an older version of Zowe Explorer and extenders would not have to call it.

lauren-li:
Functionally, this PR works well for me after the refinements @phaumer mentioned above.

Manual tests:
- I tested manually using 2 different extenders that had been modified to call the new optional `reloadProfiles()` method. z/OSMF profiles plus profiles from all extenders were automatically available from their applicable "Add profile" dropdowns (MVS, USS, and/or JES) without me needing to click the "Refresh" button.
- I also tested this PR with no extenders to Zowe Explorer, and the user experience is exactly the same as it currently is when the user has no Zowe Explorer extenders (i.e. no regression to existing functionality).

Automated tests:
- This PR branch passes all existing unit and integration tests.
- **TODO**: Unit test needs to be added to cover the new code.

phaumer:
@lauren-li added the requested tests.

src/api/ZoweExplorerApi.ts
```diff
@@ -488,5 +488,13 @@ export namespace ZoweExplorerApi {
          * @returns the registered API instance
          */
         getJesApi(profile: IProfileLoaded): IJes;
+
```
Colin-Stone, line 491: This method should be part of the ZoweExplorerExtender class along with the getProfile and getLinkedProfile methods. https://github.com/zowe/vscode-extension-for-zowe/pull/705

src/api/ZoweExplorerApi.ts
```diff
@@ -488,5 +488,13 @@ export namespace ZoweExplorerApi {
...
+        /**
+         * After an extenders registered all its API extensions it
+         * might want to request that profiles should get reloaded
+         * to make them automatically appears in the Explorer drop-
+         * down dialogs.
+         */
+        reloadProfiles?(): Promise<void>;
```
Colin-Stone, line 498: This method should be part of the ZoweExplorerExtender class along with the getProfile and getLinkedProfile methods. https://github.com/zowe/vscode-extension-for-zowe/pull/705

lauren-li: @phaumer @Colin-Stone I've created a new branch to try merging the two branches and moving the `reloadProfiles` function as requested: https://github.com/zowe/vscode-extension-for-zowe/tree/profile-link-w-refresh

phaumer:
@Colin-Stone we did the refactoring as you suggested.

I updated the FTP extension now to support that refactoring as well now for testing: https://github.com/phaumer/zowe-explorer-ftp-extension/tree/adoptions-for-next-release

I will do one more PR to suggest splitting the API folder into two: one for the interfaces the extender needs to copy and one with the implementations in Zowe Explorer.

phaumer:
I merged with @Colin-Stone's profile_link branch as we discussed over Slack.
I also updated the FTP extension with the changed API: https://github.com/phaumer/zowe-explorer-ftp-extension/tree/adoptions-for-next-release

zFernand0:
Since this PR only addresses parts of issue #528, should we create a separate issue  with the remaining pieces? I'm afraid this PR could/will close the issue when there is more great stuff to be implemented. 😋

lauren-li:
Functionally, this PR is working well for me after the updates. Thanks @phaumer for keeping this branch up-to-date with all the merges going on. System tests are also all passing.

For some reason, I'm getting unit tests that are inconsistently failing (1 or 2 on every run) with errors like `Test suite failed to run  Cannot find module 'moment' from 'ZoweUSSNode.ts'` or `Test suite failed to run  Cannot find module 'vscode-nls' from 'utils.ts'`. This is after running `npm install` and `npm run build`. I'm not sure if I've missed a step somewhere and this is just local to my machine, or if others might be seeing these failing tests as well. I'd like to conditionally approve this PR if it's just an issue on my end, but if others are also seeing this, perhaps some adjustment needs to be done.

@zFernand0 Re: Issue #528 If we could keep this issue open after this PR is merged in, I would be interested in working on some of the other items there. However, I would also be fine with addressing the remaining pieces in a new separate issue.

phaumer:
> Since this PR only addresses parts of issue #528, should we create a separate issue with the remaining pieces? I'm afraid this PR could/will close the issue when there is more great stuff to be implemented. 😋

Thanks @zFernand0. I created https://github.com/zowe/vscode-extension-for-zowe/issues/742

PR #737, opened 2020-04-27
Move extender implementor down and use TreeItem

Colin-Stone:
Requirements for future proofing API
1. Moved the implementor class into src rather than API, 
2. In the API only use vscode.TreeItem rather than IZoweTreeNode. All compiles and works fine with my test mule


zFernand0:
This PR can be closed after #725 is merged in 😋

zFernand0:
725 was merged. 
Closing! 
And deleting branch

PR #759, opened 2020-04-29
Fix configuration for running unit and integration tests locally

### LINKED ISSUE #710: EPERM error when running unit tests locally
When some of us are running unit tests locally, we get an EPERM error caused by the` jest-stare` reporter

image

Error occurs in fs functions `mkdirsSync` and `mkDirSync`...
image
`const dirs` is set to an array: `['C:', 'Users', 'theUser', ...]`
...and then on line 35 `mkDirSync `is called with `C:/` as the argument, somehow passes the test on line 26, and fs tries to make directory `C:/`
...and then we get the EPERM error shown above P:

The error does _not_ occur on the following unit tests: extension, DatasetTree, ZoweJobNode, USSTree

But all subsequent tests have the error.

### LINKED ISSUE #738: Add ts-node as dev dependency
To run the pre/post-integration test NPM scripts, the package `ts-node` is required, but it is missing from dev dependencies in `package.json`.

tjohnsonBCM:
* Fixes #710 - EPERM error when running unit tests locally
   `jest-stare` fails when unit tests are run with the `--runInBand` option, because it runs in the same Node process as the tests, and tries to use the `fs` module which still has active mocks. Fixed this by restoring mocked functions in the `fs` module to their original definitions.
* Fixes #738 - Add ts-node as dev dependency
   This makes the NPM scripts `pretest:integration` and `posttest:integration` work without needing the `ts-node` package to be installed globally.


tjohnsonBCM:
Closing in favor of #776 which contains a better fix for #710 (EPERM error when running unit tests locally).

For #738 (add ts-node as dev dependency) I can create a separate PR.

Issue #643, opened 2020-03-27
Stand-alone SCS Issue


jellypuno:
I don't know how to name this issue but this is what I am experiencing. Please note that this is applicable for changes in the **master branch only**.

1. I created a profile using Zowe CLI. I entered incorrect values in username: _user_ and password: _pass_. It says "managed by @zowe/secure-credential-store-for-zowe-cli."
2. I tried to search for a dataset and the error 401 appeared with Check Credentials button.
3. I clicked Check Credentials. I saw that the value for username is _user_ and password is _pass_. I updated the username and password to the correct values.
4. Everything is working fine and correctly after this one.
5. I closed my VSCode workspace
6. I used the newly created profile and I experienced a 401.
7. I clicked Check Credentials button and the value for the username and password is "managed by Zowe Explorer"

I added "Zowe-Plugin" in the security section of Zowe Settings.

tjohnsonBCM:
Researched why this is happening, seems there is a bug in Imperative that when the `profile` property is set on the interface `IUpdateProfile`, then secure fields are stored twice. The first time they are stored with the correct values, but the second time they get overwritten with the string `managed by <CredentialManager>`.

The `zowe profiles update` command uses the interface `IUpdateProfileFromCliArgs` which sets the `args` property instead of the `profile` property. This stores the secure fields only once as expected.

Going to create an issue in Imperative to track this. Until it gets fixed, a workaround for Zowe Explorer is to change the following lines:
https://github.com/zowe/vscode-extension-for-zowe/blob/e3e9c10a6c33432014050b9048fd6f9652f65515/src/Profiles.ts#L321-L325

```
        const updateParms = {
            name: this.loadedProfile.name,
            merge: true,
            args: { user: NewProfileInfo.user, password: NewProfileInfo.password }
        };
```

jellypuno:
This issue was fixed in v1.6.0

PR #766, opened 2020-04-30
Fix failure to load optional secure fields from profiles

### LINKED ISSUE #731: Fails to load profiles with optional secure fields when SCS enabled
When the Secure Credential Store plugin is enabled, and I have a profile containing optional secure fields that are not set (e.g., an SSH profile with no `keyPassphrase` value), the extension fails to load my profiles.

This issue was fixed in the `DefaultCredentialManager` of Imperative here: [@zowe/imperative#366](https://github.com/zowe/imperative/pull/366/files#diff-050e978f1da3b12e0441f6c1af9e1433). The same fix (adding an `optional` param to the `loadCredentials` method) should be implemented in the `KeytarCredentialManager` of Zowe Explorer.

tjohnsonBCM:
Fixes #731 by updating `KeytarCredentialManager` to not fail when an optional secure field is missing in a profile. This makes Zowe Explorer work in an environment with the SCS plugin enabled and multiple profile types such as SSH stored securely.


jellypuno:
I tried to recreate this the scenario above in master. And it doesn't happen. I can see my credentials

tjohnsonBCM:
Thanks for catching this @jellypuno, I believe this is #643 rearing its ugly head again 😢

Pushed a commit that seems to work around this for now, until the underlying problem (https://github.com/zowe/imperative/issues/379) is fixed.

jellypuno:
@tjohnsonBCM Cool! I will check it :) Thank you

jellypuno:
Hey @tjohnsonBCM this looks good! Thanks for the change. I am discovering bugs on my end while reviewing this PR 😢

PR #770, opened 2020-04-30
Fix: bugs found for edit session

### LINKED ISSUE #748: Pressing Escape does not abort Edit profile dialogue
When I click the pencil button to edit a profile I get an input box as expected

Screenshot 2020-04-29 at 11 36 10

When I press Escape, the input box does not cancel the wizard but progresses to next input box 
Screenshot 2020-04-29 at 11 52 48

### LINKED ISSUE #767: Editing cred to spaces
If I edit the profiles and change it to spaces, the prompt is not triggered

jellypuno:
### Addresses:
**Fix #767** - Editing of Credentials and setting them to spaces
**Fix #748** - Pressing Escape does not abort Edit profile dialogue

### Details:
The prompt function is triggered if the profile is changed to spaces
When you press escape when editing the URL, it will not go through the next dialogue box

### Release Notes:
**Milestone:** v1.5.1
**Changelog:** Fix bugs for edit session

### DIFF
--- a/src/abstract/ZoweTreeProvider.ts
+++ b/src/abstract/ZoweTreeProvider.ts
@@ -134,7 +134,7 @@ export class ZoweTreeProvider {
             node.getProfile().profile= EditSession as IProfile;
             await setProfile(node, EditSession as IProfile);
             await setSession(node, EditSession as ISession);
-            await this.refresh();
+            this.refresh();
         }
     }
 
--- a/src/abstract/ZoweTreeProvider.ts
+++ b/src/abstract/ZoweTreeProvider.ts
@@ -126,16 +126,15 @@ export class ZoweTreeProvider {
         return undefined;
     }
 
-    public async editSession(node: IZoweDatasetTreeNode) {
+    public async editSession(node: IZoweTreeNode) {
         const profile = node.getProfile();
         const profileName = node.getProfileName();
         const EditSession = await Profiles.getInstance().editSession(profile, profileName);
-
         if (EditSession) {
             node.getProfile().profile= EditSession as IProfile;
-            setProfile(node, EditSession as IProfile);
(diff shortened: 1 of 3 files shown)


src/abstract/ZoweTreeProvider.ts
```diff
@@ -126,16 +126,15 @@ export class ZoweTreeProvider {
...
         const EditSession = await Profiles.getInstance().editSession(profile, profileName);
-
         if (EditSession) {
             node.getProfile().profile= EditSession as IProfile;
-            setProfile(node, EditSession as IProfile);
-            setSession(node, EditSession as ISession);
-            this.refresh();
+            await setProfile(node, EditSession as IProfile);
+            await setSession(node, EditSession as ISession);
+            await this.refresh();
```
Pranay154, line 137: Does the `await` affect the `this.refresh()`?

jellypuno: yes, as far as I could see. It will wait for setProfile and setSession before refreshing. If I remove the await it will execute first and not wait for the other functions. which is something that I do not want.

Pranay154: I see, the reason I asked was because to my understanding, the `this.refresh()` isn't an `async` function so it doesn't return a `Promise`, so I don't think the `await` is needed and it won't affect the order of execution.

jellypuno: removed

zFernand0:
Similar to #748, if you try a search, but you escape out of the username or password input fields, you are still asked for the searchFilter.

As for #767, you still get the error message if you hit enter multiple times in the username or password fields. But it at least ask twice.

I'm good with the fixes though.

Issue #774, opened 2020-05-04
tests failing randomly


VitGottwald:
jest-stare is causing test failures because it is trying to create existing directories. I opened an [issue](https://github.com/dkelosky/jest-stare/issues/300) against it and suggest removing jest-stare.  Once the issue is fixed and someone wants nice html reports, we can add it back.

tjohnsonBCM:
Is this the same issue as #710? If so it should be fixed by #759.

VitGottwald:
Not sure @tjohnsonBCM . When I replace the custom code in https://github.com/dkelosky/jest-stare/blob/740dde23253f3056ef48bcf8fb1d220a5f097441/src/utils/IO.ts#L73 with [mkdirp](https://www.npmjs.com/package/mkdirp) it works.

tjohnsonBCM:
> Not sure @tjohnsonBCM . When I replace the custom code in https://github.com/dkelosky/jest-stare/blob/740dde23253f3056ef48bcf8fb1d220a5f097441/src/utils/IO.ts#L73 with [mkdirp](https://www.npmjs.com/package/mkdirp) it works.

I investigated this error in depth and don't think `jest-stare` is doing anything wrong. It checks if a directory exists before creating it, which mimics the behavior of `mkdirp`.

The error message is puzzling:
`EISDIR: illegal operation on a directory, mkdir '/'`
The root directory has to already exist, so why does `jest-stare` think it doesn't exist and try to create it?

The only answer I could find is that our tests override `fs.existsSync` with a mock that always returns True. This mock is not being reset properly by some of our unit tests.

The error only manifests itself when `jest-stare` runs in the same process as the unit tests (with the option `--runInBand`). If it is running in a different process then it doesn't matter whether the `fs` mock is reset.

This is why the `npm run test:unit` command succeeds, because it doesn't use `--runInBand`. The problem has only happened in the VSCode launch task for unit tests, which does use `--runInBand`.

When I modified our tests to properly reset the mock, this error went away and `jest-stare` worked fine with the `--runInBand` option.

To make sure our mocks get reset properly in the future, I suggested in https://github.com/zowe/vscode-extension-for-zowe/issues/751#issuecomment-621320712 that when refactoring tests we change them to use `jest.spyOn` instead of `jest.mock`, so that `restoreAllMocks` can reset all the mocks easily in one operation.

dkelosky:
[...]

PR #775, opened 2020-05-04
fix: remove jest-stare

VitGottwald:
because it is causing random test failures,
[see](https://github.com/dkelosky/jest-stare/issues/300)

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -84,10 +84,8 @@ node('ca-jenkins-agent') {
         JEST_JUNIT_ANCESTOR_SEPARATOR: " > ",
         JEST_JUNIT_CLASSNAME: "Unit.{classname}",
         JEST_JUNIT_TITLE: "{title}",
-        JEST_STARE_RESULT_DIR: "${UNIT_TEST_ROOT}/jest-stare",
-        JEST_STARE_RESULT_HTML: "index.html"
       ],
-      testResults: [dir: "${UNIT_TEST_ROOT}/jest-stare", files: "index.html", name: "${PRODUCT_NAME} - Unit Test Report"],
+      testResults: [dir: "${UNIT_TEST_ROOT}", files: "results.html", name: "${PRODUCT_NAME} - Unit Test Report"],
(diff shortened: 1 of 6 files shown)


dkelosky:
hey @VitGottwald, as @tjohnsonBCM mentioned, I think `jest-stare` failing is only a symptom of real a mocking problem with `fs`.  I'll add notes to #774.

It might still be valid to remove `jest-stare` but perhaps for another reason 😄 .

VitGottwald:
@dkelosky I agree and do believe there are bigger hidden problems to fix. jest-stare is simply one more thing to worry about at this point, that's why I want it out of the picture.

VitGottwald:
The test suite actually is failing after jest-stare is removed. I will investigate. I guess I will come at similar conclusion as you and @tjohnsonBCM .

VitGottwald:
You are right about `jest.mock('fs')` @dkelosky . Because it is a core node module, explicit jest.mock is [required](https://jestjs.io/docs/en/manual-mocks)
> Warning: If we want to mock Node's core modules (e.g.: fs or path), then explicitly calling e.g. jest.mock('path') is required, because core Node modules are not mocked by default.

dkelosky:
I know I'm being a pain here, but the premise of this change was on a perceived failure in the reporter and not wanting to divert energy from refactoring.  

Perhaps this issue & PR should titles should be changed to reflect the current work: "Removing Unused Reporter" or whatever lines you're thinking.

As I mentioned, I don't care if it's removed, however, the test failures are independent from `jest-stare`.  

The important takeaway was how to handle mocking of core node modules and shouldn't be conflated with removing a reporter.

VitGottwald:
You are certainly not pain here @dkelosky . Not at all. If someone is a pain here it is myself. 

You are right, jest-stare seems not to be an issue in the end. I wanted to eliminate unnecessary code and it became a scapegoat in the effort.

I am closing this PR in favour of https://github.com/zowe/vscode-extension-for-zowe/pull/776

PR #776, opened 2020-05-05
fix: do not modify fs node module

VitGottwald:
to mock `fs` and other core node modules,
explicit call to jest.mock('<module-name>') is required

Signed-off-by: Vit Gottwald <vit.gottwald@gmail.com>

### DIFF
--- a/package.json
+++ b/package.json
@@ -1195,6 +1195,9 @@
     "testPathIgnorePatterns": [
       "<rootDir>/src/decorators"
     ],
+    "watchPathIgnorePatterns": [
+        "<rootDir>/results/unit"
+    ],
     "transform": {
       "^.+\\.(ts|tsx)$": "ts-jest"
     },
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -1195,6 +1195,9 @@
     "testPathIgnorePatterns": [
       "<rootDir>/src/decorators"
     ],
+    "watchPathIgnorePatterns": [
+        "<rootDir>/results/unit"
+    ],
```
VitGottwald, line 1200: This fixes jests `--watch` mode. At the end of every test run, jest reporters are executed and they write their results into `<rootDir>/results/unit` directory. Unless this directory is excluded jest think watcher thinks something has changed and rerunst tests. This results in a flickering screen with nothing running. Thank you @tjohnsonBCM for finding this config option 🎉

VitGottwald:
This PR is ready, the build failure is due to security vulnerability inside zowe-cli which is unrelated to the code changes, please, check the build.

Issue #743, opened 2020-04-28
Delete Profile does not consider all extensibility use cases


phaumer:
The new edit and delete profile capabilities have these issues working with extensions such as https://github.com/phaumer/zowe-explorer-ftp-extension

Right-click a zFTP profile in the USS explorer and select Delete: You see this error

```txt
Error running command zowe.uss.deleteProfile: Internal error: Tried to call a non-existing MVS API in API register: zftp. This is likely caused by the extension that contributes zowe.uss.deleteProfile.
```

This is caused by extender not being required to register APIs for all protocols. FTP currently only support USS for example, but there is a hard-coded check to always load the MVS api that will fail. It needs to load the API based on where Delete was selected (e.g. mvs or uss or jes explorer view)

phaumer:
Notes, the new Edit session is hard coded for zOSMF and assumes https. It needs to be updated to support other profile types.  Something that could be addressed with https://github.com/zowe/vscode-extension-for-zowe/issues/631 now, though.

PR #777, opened 2020-05-05
Fix deleting profiles (issue #743)

phaumer:
We cannot assume that all API extensions implement all APIs. For example the FTP extension currently only implements USS.  However, the Delete Profile code was doing a request to the API Registry for finding a profile under MVS.  However, all the information that was needed from the return value was the profile type, which is already included in the profile itself. So I just removed the call.

Other minor changes:
- renamed deleteProf() to deleteProfileOnDisk() as the name was not expressing what the method was doing.
- made deletePrompt() and deleteProfileOnDisk() private as they are not used anywhere else.


phaumer:
I based this PR on PR https://github.com/zowe/vscode-extension-for-zowe/pull/776 as tests were failing in the master branch.

jellypuno:
Note: Merge after #776

phaumer:
> The only thing I noticed were changes in the i18n files after I ran the `vscode:prepublish` script.

I was surprised by that as well and added that commit. The only thing different was that one sting changed position in the file.

phaumer:
@jellypuno it is merge now and I merged into this branch. Looks like all checks are passing as well.

src/Profiles.ts
```diff
@@ -592,6 +553,44 @@ export class Profiles {
...
+        };
+        // confirm that the user really wants to delete
+        if (await vscode.window.showQuickPick([localize("deleteProfile.showQuickPick.yes", "Yes"),
+            localize("deleteProfile.showQuickPick.no", "No")], quickPickOptions) !== localize("deleteProfile.showQuickPick.yes", "Yes")) {
+            this.log.debug(localize("deleteProfile.showQuickPick.log.debug", "User picked no. Cancelling delete of profile"));
+            return;
+        }
+
+        try {
+            this.deleteProfileOnDisk(deletedProfile, profileName, deletedProfile.type);
```
zFernand0, line 572: If the `deletedProfile` already hast the infromation about the type, why does the `deleteProfileOnDisk` function requires 3 parameters? (is there a `Closure` problem?)

phaumer: You are right. I missed that.

zFernand0: No worries, I was just wondering if there was a problem accessing the value due to closure issues. So, it's fine 😋

Issue #746, opened 2020-04-29
cleanup of recall for setting.json


jellypuno:
Due to added changes in addRecall toUppercase(), Some new data added in recall:[] are not deleted properly.

katelynienaber:
So this is due to this line: https://github.com/zowe/vscode-extension-for-zowe/blob/2d09a64ce3c90b2e4e84e54a6f4700e0ebc926b3/src/PersistentFilters.ts#L106

I added this line because the addRecall() function could receive a path with any sort of capitalization, and this should not cause addRecall() to add a duplicate item to the recall array.

I would like to leave it this way, otherwise we will need to standardize capitalization in all paths...

jellypuno:
@katelynienaber Yah. no worries. with PR #631 this issue is actually fixed 😄

PR #785, opened 2020-05-11
Fix issue #746: Persistent recent file array deals better with received paths

katelynienaber:
Changing `removeRecall()` so it will still find a match in the recent files array (recall array), even if the path it receives is lowercase.

This means that both `addRecall()` and `removeRecall()` can be passed a path capitalized any which way, and they will convert it to capitals and proceed accordingly.

This issue was handled slightly differently in PR [#717](https://github.com/zowe/vscode-extension-for-zowe/pull/717) so I have still added this fix.

Fix for issue [#746](https://github.com/zowe/vscode-extension-for-zowe/issues/746)


zFernand0:
Jenkins server overloaded.
Will replay in a few minutes :yum:

PR #787, opened 2020-05-11
Feat: Validate Profiles

### LINKED ISSUE #309: Validate the z/OSMF Host is a Valid Service
Validate connection: The Zowe Explorer today doesn't check the connection to zOSMF when it is created. Instead it attempts to connect when the user first does a search. A validation check would help to detect if zOSMF is alive and ready to use and this is a valuable check for new users to onboard to the Mainframe. This is a compatibility feature for zOS Explorer.

Note this was available in the original z/OS Explorer extension.

jellypuno:
Addresses:
Fix #309 - Validate the z/OSMF Host is a Valid Service

Details:
Validate all profiles during initialization
Changes icons if the profile is active or not
Error message isdisplayed if you try to use an inactive profile

Release Notes:
Milestone: v1.7
Changelog: Adding Validate Profiles Feature


src/api/ZoweExplorerApi.ts
```diff
@@ -43,6 +43,15 @@ export namespace ZoweExplorerApi {
...
         getSession(profile?: IProfileLoaded): Session;
+
+        /**
+         * Create a session for the specific profile type.
+         *
+         * @param {IProfileLoaded} profile
+         *      will use the profile the API was retrieved with by default
+         * @returns {IZosmfInfoResponse} z/OSMF Check Status response
+         */
+        getZosmfStatus(profile: IProfileLoaded): Promise<zowe.IZosmfInfoResponse>;
```
phaumer, line 54: The API is intended to be independent of zosmf and abstract from it.  So the method be called something generic such as `validateProfile()`.

src/Profiles.ts
```diff
@@ -119,12 +140,19 @@ export class Profiles {
...
         this.allProfiles = [];
         this.allTypes = [];
+        this.profilesForValidation = [];
         for (const type of ZoweExplorerApiRegister.getInstance().registeredApiTypes()) {
             const profileManager = await this.getCliProfileManager(type);
             const profilesForType = (await profileManager.loadAll()).filter((profile) => {
                 return profile.type === type;
             });
             if (profilesForType && profilesForType.length > 0) {
+                for (const validateProfile of profilesForType) {
```
phaumer, line 150: As discussed on Slack. We should avoid validating all profiles automatically as it is potentially a super expensive operation. Refresh is called during activation as well as every time an extender registers itself as well as every time the user clicks the refresh button. We do not want to run expensive network operations on all profiles every time. 

Especially not during activation as it would slow down the editor startup time, which is highly important for the user experience and in environments such as Che could even lead to timeouts.

An alternative would be to validate profiles when the user explicitly requests it or when the use a profile the first time such as doing the first Search query as well as after Create/Edit Profile.

src/api/ZoweExplorerApi.ts
phaumer: Plus we probably need to define a new return type interface. So far all the other return types were zosmf independent such as IZosFilesResponse.  We would need something like IZosServerInfo.

JillieBeanSim:
@jellypuno as is the validation of non zOSMF profiles fail the validation due to not having a ZoweExplorerApi.getStatus() function and this disallows the user of being able to access anything with that profile.
I have some suggestions: 
1. A profile type check for `zosmf` if validation fails and if not zosmf type have it work without a validation check like previously.  It could present user with a message and leave the icon as is. This will prevent the change from breaking extenders.
2. This change will need to be made aware to all extenders, possibly in change log, so they can add the needed functionality to their extension so they can have their profile types validated as well.

JillieBeanSim:
I was fiddling with the code and have this https://github.com/JillieBeanSim/vscode-extension-for-zowe/tree/ValidateProfiles it's hacky but shows the direction I was leaning towards in comment above.

validate-alt-profile

jellypuno:
> I was fiddling with the code and have this https://github.com/JillieBeanSim/vscode-extension-for-zowe/tree/ValidateProfiles it's hacky but shows the direction I was leaning towards in comment above.

Hi @JillieBeanSim , Yes. I am going to change the code to stop that breakage on extender profile. My goal is something like this.

phaumer:
I would return string values, but use either types or enums.

src/api/ZoweExplorerZosmfApi.ts
```diff
@@ -41,17 +41,19 @@ class ZosmfApiCommon implements ZoweExplorerApi.ICommon {
...
 
             if (sessionStatus) {
-                return true;
+                return "active";
             } else {
-                return false;
+                return "inactive";
             }
+        } else {
+            return "unverified";
```
JillieBeanSim, line 56: @jellypuno this is nice but this is still being handled on the extender's side. When I try a different extension's profile that may not yet have the new getStatus() method it breaks like I mentioned above. An extender would have to make sure they have this method in place on their side and tested before this is released, and release their update right after this goes out

jellypuno: @JillieBeanSim Thank you for testing it. I will take a look at it.

src/api/ZoweExplorerApi.ts
```diff
@@ -43,6 +43,15 @@ export namespace ZoweExplorerApi {
...
         getSession(profile?: IProfileLoaded): Session;
+
+        /**
+         * Create a session for the specific profile type.
+         *
+         * @param {IProfileLoaded} profile
+         *      will use the profile the API was retrieved with by default
+         * @returns {IZosmfInfoResponse} z/OSMF Check Status response
+         */
+        getStatus(profile: IProfileLoaded, profileType?): Promise<string>;
```
JillieBeanSim, line 54: @jellypuno the new getStatus should have a ? after it like getStatus?(profile: IProfileLoaded, profileType?): Promise<string>; until v2.0 so if the extender doesn't have it yet it will be optional and the caller cannot just call it, but they need to test if it exists; so on the Zowe Explorer side they need something like if (ZoweExplorerApiRegister.getUssApi(node.getProfile()).getStatus) { ZoweExplorerApiRegister.getUssApi(node.getProfile()).getStatus()}

jellypuno: ok. I will change it as you said.

src/api/ZoweExplorerZosmfApi.ts
jellypuno: @JillieBeanSim I added some changes that I think could fix this issue. Can you please verify? Thank you!

jellypuno:
Note: The star icon is already here. I will fix it once that PR is merged to master

jellypuno:
@JillieBeanSim I need your help. 😞 For some reason if I use FTP Profile, the function getStatus is not recognized as a function that is why FTP profiles ends up as an inactive profile. I don't know why. If I use z/OSMF profile it is recognized as a function.

image

JillieBeanSim:
@jellypuno getStatus() doesn't exist on ftp yet and this is the same with other extensions since it is new to even zosmf.  I mentioned above about having the ? added to getStatus() in ZoweExplorerApi() and having a check to see if it exists on the extension, if it does exist then it can be called but if not it will have to bypassed at least until v2.0 is released and extenders have the time to update their extensions.  The getStatus() shouldn't be called, but tested for existence first. 
image

jellypuno:
> @jellypuno getStatus() doesn't exist on ftp yet and this is the same with other extensions since it is new to even zosmf via zowe explorer. I mentioned above about having the ? added to getStatus() in ZoweExplorerApi() and having a check to see if it exists on the extension/profile type, if it does exist then it can be called but if not it will have to bypassed at least until v2.0 is released and extenders have the time to update their extensions. The getStatus() shouldn't just be called, but tested for existence first.
> image
> 
> I will look into the code as well and see what exactly should be done here along with you.

Thank you! 😭  I really appreciate your help! I have added the ? in [getStatus](https://github.com/zowe/vscode-extension-for-zowe/blob/2c4b1fa76ef67d70f267273622ce0c3887e28919/src/api/ZoweExplorerApi.ts#L54) but I was not able to do the second part. Thank you for looking into the code. 😁

jellypuno:
@JillieBeanSim I added some changes. Seems to be working but can you please check as well? Thank you!

jellypuno:
Performance Test Results for z/OSMF Profiles:

**Loading of Zowe Explorer:** 13-15 seconds depending on how many profiles to load. I have more than 10 so it is around 15s
**Loading with FTP extension installed:** 15 - 17 seconds. I am waiting for the activation of the FTP ext. I cannot click or do anything unless it is activated.
**Validating profiles:** 3-5 seconds for valid profiles. 1-3 for invalid profiles.
**Editing Profiles with validation:** 8-10 seconds

I can't test FTP profiles because of this error. Maybe I need additional set-up.
image

JillieBeanSim:
> @JillieBeanSim I added some changes. Seems to be working but can you please check as well? Thank you!

@jellypuno YES! working good for alternate profiles, I do notice the lag mentioned above but great work getting it to work with profiles that don't yet have the getStatus() yet!

jellypuno:
> > @JillieBeanSim I added some changes. Seems to be working but can you please check as well? Thank you!
> 
> @jellypuno YES! working good for alternate profiles, I do notice the lag mentioned above but great work getting it to work with profiles that don't have the getStatus() yet!

Hooray! 🎉 I am glad it is working now! 😄

jellypuno:
> Functionality works well and unit tests are passing but some integration tests are failing
> 
> 1. should remove the selected favorite search from the treeView
> 2. should delete a data set if user verified

Thank you for testing! :) It's weird because for me it is passing in those tests. But I do have errors in:
- should create a member when zowe.createMember is invoked
- should open a PS
- should download, change, and re-upload a PS
- should download, change, and re-upload a PDS member

I will investigate.

jellypuno:
@JillieBeanSim After running pretest:integration script, the integration tests were successful.

jellypuno:
This is how the icons looks like in different themes:

ValidateTHeme

JillieBeanSim:
Hey @jellypuno everything looks good to me on this PR.  I am still getting some failing integration tests (attached image) but 1 & 3 are failing for me on master branch as well.
image

So I will wait on another reviewer to see what they are getting with integration tests, other than that the functionality works great and all unit tests are passing.

Thanks Jelly for the work done here!

JillieBeanSim:
since Peter is on vacation this week I added @lauren-li to review the PR

src/Profiles.ts
zFernand0: I believe this was addressed 😋

src/Profiles.ts
```diff
@@ -857,6 +876,69 @@ export class Profiles {
...
+                filteredProfile = {
+                    status: profile.status,
+                    name: profile.name
+                };
+            }
+        });
+
+        // If not yet validated, call getStatus and validate the profile
+        // status will be stored in profilesForValidation
+        if (filteredProfile === undefined) {
```
zFernand0, line 896: Just curious what the expected behavior is if the value is `null`.
I understand it can only happen if someone adds additional logic or initializes the valuable to null, but just as a general rule (maybe) we can check for `variable != null`. That way we cover both, undefined and null cases.

No changes requested though.

jellypuno: @zFernand0 I'll take note of this! Thank you :)

jellypuno:
> functionality works well and all unit tests are passing. still getting those 3 failing integration tests mentioned above, where 1 & 3 are failing on master branch but # 2 only fails on this branch.

@JillieBeanSim Yes. I just needed your approval again because it reset when I merged master to my branch

PR #794, opened 2020-05-12
Added getCommonApi() to API registry

phaumer:

In discussion around Issue https://github.com/zowe/vscode-extension-for-zowe/issues/309 and PR https://github.com/zowe/vscode-extension-for-zowe/pull/787 the need came up to provide access to the Common API for a profile independent if it was registered just with MVS, USS, JES (either one or more) to allowing adding generic functions such as `validateProfile()` that are independent of MVS, USSm or JES, but are dependent on the registered protocol providers such as zosmf or zftp.

This PR adds the `getCommonApi(IProfileLoad)` that looks in all three registries for the profile and returns the first it can find.


jellypuno:
@phaumer Is it okay if I merge this in my PR instead? and then merge it together in master afterwards.  I would like to thoroughly test this with my changes. What do you think?

phaumer:
Sounds good @jellypuno

jellypuno:
The code is added in #787. This PR can be closed.

Issue #654, opened 2020-04-01
Refactor Unit Tests


zdmullen:
In conjunction with the refactoring of the extension.ts file, we plan to refactor unit tests and place them in the appropriate folder for their corresponding action. Similarly, we will provide guidance on the creation of unit tests for new functionalities that are implemented. 

The unit tests can be grouped in the following manner, which Stepan will provide greater detail on in his presentation during the April 1 stand-up (and in the discussion here):

- Jobs
- USS
- Data sets
- Shared
- Utils
- Extension.ts (what is kept in the current file)

We will update this issue to provide an example and will work to produce the documentation (guidance) for contributors.

katelynienaber:
To-do list: https://docs.google.com/spreadsheets/d/e/2PACX-1vR0VUpkoWsvVuYSzCeJkCNrE8F7YWYNtSZTVQJFDp3o_76_Lhid2Eh82RV4JwjG3L07dx9ZWlBKaK-g/pubhtml

Here you can see all functions in the entire extension, sorted by file (where they are located AFTER the `extensions.ts` refactoring)

We are working only on functions not marked as EXCLUDED.

I will update this as I work. Unfortunately it is only editable by people at Broadcom. I am not super aware of what Stepan does each day so I will only update his work when I hear about it...

**Feel free to add comments here!**

katelynienaber:
Order of merges for USS test refactoring PRs:
- [Issue #654: Refactoring of ZoweUSSNode.unit.test.ts (refactor-unit-tests-uss)
](https://github.com/zowe/vscode-extension-for-zowe/pull/709)
- [Issue #654: Refactoring of /uss/actions.unit.test.ts (refactor-unit-tests-uss-actions)
](https://github.com/zowe/vscode-extension-for-zowe/pull/714)
- [Issue #654: Refactoring of USSTree.unit.test.ts (refactor-unit-tests-uss-tree)](https://github.com/zowe/vscode-extension-for-zowe/pull/726)

zFernand0:
there should be one or two more PRs left

zFernand0:
Two more PRs are still open

PR #800, opened 2020-05-18
Issue #654: Refactoring of ZoweUSSNode.unit.test.ts

katelynienaber:
In this PR:

- Moved all tests for ZoweUSSNode functions into `__tests__/__unit__/uss/ZoweUSSNode.unit.test.ts`
- Created mock generator file for uss mocks (`__mocks__/generators/uss.ts`)

**There are some commented blocks in the code that should be discussed!**
I want to remove them but need to hear opinions about it first.

Related to issue [#654](https://github.com/zowe/vscode-extension-for-zowe/issues/654)


__tests__/__unit__/job/actions.unit.test.ts
```diff
@@ -60,17 +60,17 @@ describe("Jobs Actions Unit Tests - Function setPrefix", () => {
     let blockMocks;
```
VitGottwald, line 60: The same comment I gave to Stepan yesterday. This should become `const` and to into individual `it` blocks. But we will tackle this in all files at once after we merge the current slate of PRs.

katelynienaber: I can make this change in the last PR

__tests__/__unit__/shared/utils.unit.test.ts
```diff
@@ -12,17 +12,183 @@
...
+        sharedUtils.labelRefresh(rootNode);
+        expect(rootNode.label === "gappy");
+    });
+});
+
+describe("Positive testing", () => {
+    it("should pass for ZoweDatasetTreeNode with ZoweDatasetNode node type", async () => {
+        const dsNode = new ZoweDatasetNode(null, null, null, null);
+        const value = sharedUtils.isZoweDatasetTreeNode(dsNode);
+        expect(value).toBeTruthy();
```
VitGottwald, line 68: Note: I see the use of `.toBeTruthy` a lot. I think it is better to use explicit values when possible like`.toBe(true)`. Not sure if it applies here. No action requied.

PR #860, opened 2020-06-12
Weback with localization and logging

lauren-li:
## Proposed changes

This PR addresses issue #74 (Bundle VSCode extension with webpack).

This PR has webpack working with localization and logging. Integration tests are built without webpack, as decided in the discussion for PR #493 (which was replaced by this one).
- You can check wepback by running `npm run package`, renaming the resulting .vsix file to a .zip, and checking this .zip folder's contents.
  - Under the **extension/out/src** folder, there should be a single **extension.js** file (as opposed to .js files named after source code files). There should be several .json files as well, including .json files in a **src** folder that mirrors the soure code's **src** folder hierarchy, and these are used for localization.
- You can check the localization in this branch by either installing the extension from the .vsix file built from this branch, or by just pulling the branch and running it with the Extension Development Host.
  - Localization can be tested by switching the [VS Code display language](https://code.visualstudio.com/docs/getstarted/locales) to French. The following text of the extension should be in French:
    - Hover text for the "Create New Data Set" icon. (Tests package.json i18n.)
        Screen Shot 2020-06-12 at 2 38 53 PM

    - QuickPick text that appears when deleting a data set. (Tests i18n for TypeScript files.)
        Screen Shot 2020-06-21 at 10 47 56 PM

Some notes about the files changed:
- An **i18n/fra** folder full of French language files was added as part of this PR so that localization functionality can be tested with webpack. Changed files beginning with **i18n/fra** do not need to be code-reviewed.
    - **The i18n/fra folder is currently for testing only, and should removed after this PR is approved. French should also be removed from the `languages` array in the gulpfile until the extension actually has a French translation.**
- Changes in **src/\*\*/*.ts** files mostly have to do with adding and/or updating the localization config/call at the top of the file. No changes in actual code logic were made except in globals.ts, where the path formation for `initLogger` was updated per @tjohnsonBCM 's advice.

## Release Notes
Milestone: 1.7

Changelog: Use Webpack to bundle extension.

## Types of changes

## Checklist

## Further comments

N/A


lauren-li:
I'm not quite sure what Codecov is looking for, as I didn't really add or change source code to bring Webpack into the extension. The only thing I can think of is @tjohnsonBCM 's suggestion for assigning appender paths for the log4js initialization in globals.ts, but globals.ts is already at 100% coverage:
Screen Shot 2020-06-12 at 3 36 15 PM

.vscode/launch.json
```diff
@@ -23,12 +23,14 @@
...
             "args": [
-                "--extensionDevelopmentPath=${workspaceFolder}"
+                "--extensionDevelopmentPath=${workspaceFolder}",
+                "--verbose"
             ],
             "outFiles": [
                 "${workspaceFolder}/out/**/*.js"
             ],
-            "preLaunchTask": "npm: watch"
+            "preLaunchTask": "npm: build",
```
tjohnsonBCM, line 32: Noticed this was changed from "watch" to "build". Could Webpack watch task could be used here?

.vscodeignore
```diff
@@ -6,6 +6,7 @@ out/__tests__/**
 out/src/__mocks__/**
```
tjohnsonBCM, line 6: Perhaps it would make sense to structure the `.vscodeignore` file to ignore everything, and only whitelist what is needed (example [here](https://github.com/Microsoft/vscode-vsce/issues/12#issuecomment-285964473)).

This could prevent new folders/files that are added to the repo in the future from accidentally getting bundled in the VSIX, but no need to change this now 😄

lauren-li: @tjohnsonBCM This is an interesting suggestion and a valid point. I would rather address it as a separate issue, since it doesn't strictly have to do with Webpack. If others on the Zowe Explorer squad like the idea, I'd be happy to go along with it!

lauren-li:
@katelynienaber Thanks for trying this PR out! I still need to address @tjohnsonBCM's comments regarding the Secure Credential Store, so I will be adding commits to support that soon.

Regarding translation, we have a basic infrastructure in place for the community to donate translations via [Zanata](http://zanata.org/). Our documentation for localization can be found here: https://github.com/zowe/vscode-extension-for-zowe/tree/master/docs#adding-a-new-language
Feel free to reach out to me if you give it a try and run into questions or issues. It's a largely manual process for now, and we may need to update the JSON files we have on Zanata to match our current strings.

.vscodeignore
lauren-li: @tjohnsonBCM Issue #883 filed to discuss/address this.

.vscode/launch.json
lauren-li: @tjohnsonBCM Addressed in https://github.com/zowe/vscode-extension-for-zowe/pull/860/commits/bf491f5135f8441fc72c2c3c2e48fdd7555dd93c, based on [CPPTools'](https://github.com/microsoft/vscode-cpptools/tree/master/Extension) configuration for extension launch in watch and development mode.

lauren-li:
> Thanks for adding Webpack to the extension, should make extension install time much faster 👍
> 
> Left a few comments, which don't require changes. I found an issue that may require changes:
> 
> The Webpacked extension fails to load for me, because I have the Secure Credential Store plugin for Zowe CLI installed. I get this error when the extension activates:
> image
> 
> I believe the problem lies in the `getSecurityModules` method of "extension.ts", because Webpack breaks the `require` call used to load Keytar. VSCode docs show a way to work around this by loading Keytar with `__non_webpack_require__`: https://code.visualstudio.com/api/advanced-topics/remote-extensions#persisting-secrets

@tjohnsonBCM Thank you for pointing this out. I was able to recreate the issue, and believe I addressed it in https://github.com/zowe/vscode-extension-for-zowe/pull/860/commits/a9d15690c9094877b6df4fdce7f8eca54706e038 following guidance from the link you provided to VS Code's documentation. The issue no longer occurs for me when using the Secure Credential Store (SCS) after this commit. However, I was unable to check if integration tests work okay with SCS.
- For both this PR branch and `master` branch, despite having the `Zowe-Plugin` key set and the 2 z/OSMF profiles set up (the default one being named `ZoweCLI`), the integration tests close without running for me, giving only the following message in the Debug Console: `Error: Could not find profile named: zoweCLI.` @JillieBeanSim tried as well and got the same issue with both branches. Without SCS, all integration tests appear to run and pass as they do for me on the `master` branch.
- I think I would be able to run integration tests successfully on this PR branch if I were able to also run them in `master` branch, but until then, it would be very helpful if someone who has SCS and can run integration tests with it could check those on this PR branch, as well.

jellypuno:
I ran integration tests for this branch and everything is executed successfully

lauren-li:
> I ran integration tests for this branch and everything is executed successfully

@jellypuno Thank you for running the integration tests and reporting your results!

>Question: I tried to add a new localized error message and run npm run prepublish, should I expect that error message added in fr i18n/nls files?

No, with the way localization is currently set up, your new error message would be only be added to the `i18n/sample` folder upon `npm run prepublish`. We would then have to manually copy the new key:value pair from there into any language folders (e.g `i18n/fra`) to be able to translate it. If the key:value pair is missing from the `fra` folder when the extension is running in French, the extension will just use the default key:value pair that was defined in the source code. I think we could consider automating the addition of new strings into language folders in the future, especially if interest in actually translating Zowe Explorer picks up.

zdmullen:
Let's merge this as the first pull request for the 1.8 release. That will allow for further testing among people in the squad.

tjohnsonBCM:
Tested the latest changes and webpacked ZE works fine with SCS now for me 😄 

> I think I would be able to run integration tests successfully on this PR branch if I were able to also run them in `master` branch, but until then, it would be very helpful if someone who has SCS and can run integration tests with it could check those on this PR branch, as well.

I have the same problem, over 20 integration tests fail for me on `master` branch.

zFernand0:
> Tested the latest changes and webpacked ZE works fine with SCS now for me 😄
> 
> > I think I would be able to run integration tests successfully on this PR branch if I were able to also run them in `master` branch, but until then, it would be very helpful if someone who has SCS and can run integration tests with it could check those on this PR branch, as well.
> 
> I have the same problem, over 20 integration tests fail for me on `master` branch.

I only get 10 Integration tests failures (however they are not related to this PR)
4 of them due to system configuration (Recal and Migrate)
2 of them due to the use of Thenable ??
Others are intermittent due to async functions not properly awaited ?

PR looks good to me.
Just need to remove the french references! : )

zFernand0:
Theia pipelines will probably fail due to not finding certain files.
@crawr do you think this should be fixed right in this PR, or should we not be too concerned if there are plans to refactor the Integration Tests in the near future?

lauren-li:
Thanks for the reviews @tjohnsonBCM and @zFernand0! I will be removing the French testing files shortly, as discussed.

lauren-li:
@zFernand0 French i18n testing has been removed from the PR branch. 👍

zFernand0:
I'm Ok with MErging this on Monday if you do not get any other approvers by then :yum:

PR #879, opened 2020-06-15
feat: load encoding value from z/OSMF profiles 

### LINKED ISSUE #260: National (danish/nordic) characters are not translated correctly when observed in VS Code
The 3 danish national characters are displayed correctly on 3270, but when seen in VS Code they are represented by #@$ ... 
What information should be provided to dig into this issue ? 
Brgds
Claus Bjørno

dkelosky:
## Proposed changes

Fix #260.  Zowe CLI supports an encoding value in a profile to be used for USS file & data set operations (see more [here](https://github.com/zowe/zowe-cli/issues/632)).  This change allows Zowe Explorer to make use of that.  

## Release Notes

Milestone:

Changelog:

## Types of changes

- Use encoding value from profile

## Checklist

- [ ] `npm run vscode:prepublish` has been executed

## Further comments

I introduced a new API method that RSE and FTP extensions might want to implement.  `putContents` in its current form does not accept an options object like other API methods.  `putContent` is a proposal for the new API method.  It's implementation is optional for now and we test within ZE for the existence of the new method before using it (for backward compatability).


jellypuno:
@dkelosky I am not really sure what to check here because the result in master and in your branch is the same. If I will test this, what kind of test should I do?

image

dkelosky:
For a first pass, this requires an `encoding` value set in your z/OSMF profile to alter how characters are transfered.  E.g.

```
encoding: 285
```

If you have a `$` in a z/OS data set (x'5B') or USS file, with encoding `285`, it should be returned as `£`.  This will be `$` in `1047`.

dkelosky:
@jellypuno here's an example of viewing a USS file:
image

With this profile encoding value:
image

dkelosky:
fyi, something equivalent for ftp would be `quote site SBDATACONN=(285,ISO8859-1)`

jellypuno:
Cool Beans! So I updated my profile with encoding = 285 and this is the result :) Some symbols are not handled correctly but maybe it is not included in encoding 285? No idea. 

3270: 
image

VSCode:
image

dkelosky:
Thanks for trying, if you could turn `hex on` we could see what values are expected for [285](https://en.wikipedia.org/wiki/EBCDIC_285#:~:text=IBM%20code%20page%20285%20(CCSID,%E2%82%AC%20in%20that%20code%20page.).

src/api/ZoweExplorerApi.ts
```diff
@@ -106,6 +107,20 @@ export namespace ZoweExplorerApi {
...
+         * @param {string} inputFilePath
+         * @param {string} ussFilePath
+         * @param {zowe.IUploadOptions} [options]
+         * @returns {Promise<zowe.IZosFilesResponse>}
+         */
+        putContent?(
+            inputFilePath: string,
+            ussFilePath: string,
+            options?: zowe.IUploadOptions
+        ): Promise<zowe.IZosFilesResponse>;
```
zFernand0, line 122: Interesting why we never thought of this since we did it the right way for `getContents`
Now we have a deprecated `putContents` and regular `getContents`.
Either way, this serves as a great example for how we will be enhancing the APIs (that we plan to deliver as a separate package)

@phaumer, @lauren-li, Check this out! 😋

src/mvs/mvsNodeActions.ts
```diff
@@ -68,7 +68,10 @@ export function getDatasetLabel(node: ZoweDatasetNode) {
 export async function uploadFile(node: ZoweDatasetNode, doc: vscode.TextDocument) {
     try {
         const datasetName = getDatasetLabel(node);
-        await ZoweExplorerApiRegister.getMvsApi(node.getProfile()).putContents(doc.fileName, datasetName);
+        const prof = node.getProfile();
+        await ZoweExplorerApiRegister.getMvsApi(prof).putContents(doc.fileName, datasetName, {
+            encoding: prof.profile.encoding
+        });
```
zFernand0, line 74: This makes me think that instead of having to forward all properties from the user profile, we could just combine the "already constructed" profile (i.e. whenever we take into account the use of base-profile and furthermore respect the order-of-precedence `Env -> Service Profile -> Base Profile` in the APIs on NPM) with whatever options are passed in these functions.
Just a thought. I sure hope it makes some sense 😋

src/api/ZoweExplorerZosmfApi.ts
```diff
@@ -71,6 +71,19 @@ export class ZosmfUssApi extends ZosmfApiCommon implements ZoweExplorerApi.IUss
         return zowe.Upload.fileToUSSFile(this.getSession(), inputFilePath, ussFilePath, binary, localEncoding, task, etag, returnEtag);
     }
 
+    public async putContent(inputFilePath: string, ussFilePath: string,
```
zFernand0, line 74: @dkelosky, Should we deprecate the `putContents` function in favor of `putContent` here as well?
And maybe combine implementations i.e. have `putContents` create the options object and then call `putContent`, just a thought! 😋

src/shared/utils.ts
```diff
@@ -233,13 +233,32 @@ export async function willForceUpload(node: IZoweDatasetTreeNode | IZoweUSSTreeN
...
+                                binary,
+                                localEncoding: null,
+                                etag: null,
+                                returnEtag,
+                                encoding: profile.profile.encoding
+                            });
+                    } else {
+                        return ZoweExplorerApiRegister.getUssApi(profile).putContents(
+                            doc.fileName, remotePath, binary, null, null, returnEtag);
                     }
```
zFernand0, line 259: Going back to my previous comment, if we deprecate `putContents`, by constructing the object in it and calling `putContent` we could reduce this logic 😋

src/mvs/mvsNodeActions.ts
dkelosky: It would be great if whenever we added the encoding feature into Zowe CLI, that we only needed to update the version of Zowe Explorer for it to immediately make use of it.  

If that's what you mean, I'm onboard 😄

src/api/ZoweExplorerApi.ts
dkelosky: Yes, it would be nice for RSE and FTP extensions to implement `putContent` soon.

src/api/ZoweExplorerZosmfApi.ts
dkelosky: I'll add a `@deprecated` mark here and call `putContent` from `putContents` as you suggested.

src/shared/utils.ts
dkelosky: Is that possible here?  If the new method exists, I use it to pass on encoding.  If the new method doesn't exist I call the old method without encoding.  Or am I missing it entirely?  😄

src/api/ZoweExplorerZosmfApi.ts
dkelosky: I think this is addressed now

src/mvs/mvsNodeActions.ts
zFernand0: Yeah, Thanks for putting it in fewer words 😋

dkelosky:
Closing while I sort out my tests...

dkelosky:
I'm stumped on how to get codecov patch above 75%.  I'm don't _think_ its a problem with codecov itself.  I've tried:

- rebasing on master
- a new single commit on a new branch
- closing & reopening this PR
- opening new PRs

Diff is consistently around 75% coverage.  If anyone has ideas on what function / file to test and increase patch coverage, I'll do it.  Otherwise, I'm stuck.

jellypuno:
@dkelosky this is the explanation for codecov/patch check: https://docs.codecov.io/docs/commit-status#patch-status

dkelosky:
Thanks for sharing - it's a fairly understandable description.  However, I can't interpret the report in such a way to know which specific lines I should be covered to increase my overall patch coverage.  I tried and failed a few times now 😄

jellypuno:
@dkelosky i will take a look

dkelosky:
Thanks @jellypuno for your help; you were spot on about the lines I was missing 😀

zFernand0:
`pr-merge` pipeline was successful, however it couldn't report back due to the issues with the GitHub APIs earlier today.
https://wash.zowe.org:8443/blue/organizations/jenkins/zowe-cli-vscode-extension/detail/PR-879/15/pipeline

PR #920, opened 2020-07-02
FOR REFERENCE: Editable data set attributes & data set templates

katelynienaber:
## Proposed changes

So I was talking to Vit about the webview stuff, and he had this really great idea about how to make DS attributes editable, without using webview! I think this really opens up more options for us while still using VSCode's `QuickPick` and `InputBox`.

The PR contains the following:
- Editable DS attributes. When allocating a new DS, the user is given the option to edit the DS's attributes before allocating.
- Persistent DS templates. The user can choose a template, from which the DS attributes will be prefilled in the editing step. Templates are stored as objects in an array inside `settings.json`.
- When a user clicks _Create New Data Set_, they are given the option to choose a template. If there are no stored templates, this template-selection step is skipped.
- Templates can be created during allocation of a new DS (From _Create New Data Set_ button) or by right-clicking on a node and selecting _Manage Data Set Templates_.
- The user does NOT have to use templates. When allocating a data set, they can simply click "Allocate Data Set" and no template will be created.
- Templates can be edited from the _Manage Data Set Templates_ button.
- Templates can be deleted from the _Manage Data Set Templates_ button.

## Release Notes

Milestone: 1.8
Changelog: Added ability to edit data set attributes before allocation, and to use & save data set templates.

## Types of changes

## Checklist

## Further comments

Please give me alllllll your feedback! Hopefully this will be public for long enough to get all the feedback in and make it a really nice feature by the end of the PI :)

I will eventually need some assistance/more information about what all attributes should be available for editing in the list, and how to send an API request with different attributes...

Shall I add something similar to USS or no? Is there really a need for a template with USS? The only attribute that we manage now for USS is the file name.

Unit tests are not done yet, they will be added tomorrow or next week...

Currently there is NO input validation happening on any part of this functionality. I talked to Jelly and she said: `We have an open RFE about it and Dan wants z/OSMF to handle this rather than in our application`. _But_, if there is any data set attribute which we should validate before firing the API, this is entirely possible! So we can make that decision at some point.


VitGottwald:
Thank you @katelynienaber ! I have only so far looked at the code and plan to try this out tomorrow.

VitGottwald:
There are quite a few files changed with the renames of searchHistory and fileHistory. Would it be possible to have a PR just for these renames? That would allow us to merge that part quickly, get it out of the way and focus on the new features.

VitGottwald:
Just playing with it and noticed these items (see picture). Do they have some meaning or is it something that remained from the first prototyping and should be removed ?

Screenshot 2020-07-03 at 08 50 24

VitGottwald:
OK I see those are the prototypes for attributes.

zdmullen:
And your comment clarifies my question, @VitGottwald. So is the intent to have the list of attributes in the drop-down from the command palette, and one-by-one a user can enter the details before selecting either "allocate data set" or "create template"? Because while at first glance it seems a little strange, I like the fact that you can see each of the attributes you've entered in that way. I'll be curious to hear what @jelaplan says. 

Question then--which attributes do we want to make available in the drop-down?
image

Also, what I see in ISPF and what I see in settings.json is slightly different. Here are the settings from a Type C data set:

"Zowe-Default-Datasets-C": {
  "dsorg": "PO",
  "alcunit": "CYL",
  "primary": 1,
  "secondary": 1,
  "dirblk": 25,
  "recfm": "VB",
  "blksize": 32760,
  "lrecl": 260
}

Then this is what the options look like in Zowe CLI: 
image
image

I guess one issue I see here already, and this is before we even begin to address what we add to what Katelyn is doing here, is that there are different labels for the attributes across ISPF, Zowe CLI, and what is in settings.json for Zowe Explorer. We'll have to ensure consistency (eventually) while at the same time providing the most important attributes for this template creation.

katelynienaber:
> There are quite a few files changed with the renames of searchHistory and fileHistory. Would it be possible to have a PR just for these renames? That would allow us to merge that part quickly, get it out of the way and focus on the new features.

I have created a PR for the renames [here](https://github.com/zowe/vscode-extension-for-zowe/pull/925). Do you think it is alright to leave the renamed functions in this PR as well?

katelynienaber:
> OK I see those are the prototypes for attributes.

To make this more clear I set the filler value to Undefined for any fields that have not yet been filled.

katelynienaber:
@zdmullen I guess I would prefer to match our names & attribute order to the CLI docs, since they have made the options available before us. I will add these in this order and then it will look more similar to a "final preview" and @jelaplan can evaluate it more thoroughly.

zdmullen:
That sounds good, @katelynienaber. I guess I just wanted to take the opportunity to point out the differences here as part of the greater conversation. In general though, we'll stay consistent across the CLI and Zowe Explorer, definitely. Also, I like what you've done with this!

katelynienaber:
> That sounds good, @katelynienaber. I guess I just wanted to take the opportunity to point out the differences here as part of the greater conversation. In general though, we'll stay consistent across the CLI and Zowe Explorer, definitely. Also, I like what you've done with this!

Okay. Thanks for your review @zdmullen !!

katelynienaber:
Also just a note so I don't forget: @VitGottwald and I were discussing the possibility to create default templates for each type of data set. Users would be required to select one, from which they could edit the prefilled attributes. It is possible we would add this feature with only these templates available at first...and not give users the option to create their own custom templates until a later PR.

src/PersistentFilters.ts
```diff
@@ -131,6 +137,35 @@ export class PersistentFilters {
...
+     * Adds the name of one recently-edited file to the local store and
+     * updates persistent store. The store contains a
+     * maximum number of entries as described by `maxFileHistory`
+     *
+     * If the entry matches a previous entry it is removed from the list
+     * at that position in the stack.
+     *
+     * @param {ZoweDatasetNodeTemplate} template - the template to add
+     */
+    public async addTemplate(template: ZoweDatasetNodeTemplate) {
```
VitGottwald, line 150: Types belong to declaration/definition, not to a JSDoc. Also it is clear from the function nam that the template argument is a template to be added.

```suggestion
     */
    public async addTemplate(template: ZoweDatasetNodeTemplate): ZoweDatasetNodeTemplate {
```

src/PersistentFilters.ts
```diff
@@ -205,6 +240,21 @@ export class PersistentFilters {
         await this.updateFileHistory();
     }
 
+    /**
+     * Removes one template from the persistent templates array
+     *
+     * @param {string} templateName - The name of the template to remove
+     */
+    public async removeTemplate(templateName: string) {
```
VitGottwald, line 248:
```suggestion
     */
    public async removeTemplate(templateName: string) {
```

katelynienaber:
Possible ideas:
- Locked-in templates?
- Sharing templates?
- Generate template from existing dataset

dkelosky:
Which `settings.json` is used?  There are user and workspace `settings.json`.  Workspace settings are stored in `.vscode` folder in a project and sourced / shared with the team.  I think workspace settings take precedence over user settings.  Perhaps this is a means to share templates?

jellypuno:
@dkelosky what is not clear for me is the 'sharing' part. I cannot imagine it. Are the templates shared through github? If I have pure Mainframe Background, can I share the templates through datasets as well? Is that a valid scenario? 

@katelynienaber Do you think that it is possible for users to specify the location of the templates? It could be a local file, USS location or Datasets? Does this makes sense?

dkelosky:
I might have missed the point entirely 😄 .  My thought was if templates are stored in a `.vscode/settings.json` and stored in `git`/GitHub, then my team will obtain the templates through source control.  Does that make sense?

jellypuno:
> I might have missed the point entirely 😄 . My thought was if templates are stored in a `.vscode/settings.json` and stored in `git`/GitHub, then my team would obtain the templates through source control. Does that make sense?

Yes it makes sense. My question is do you think that it is possible that these templates can be stored in Mainframe? Like what if we are not using github? Or this scenario is not applicable anymore?

dkelosky:
I'm not sure.  I was only thinking about how _**I**_ would want to share them and that would be in GitHub 😄 

Did you have a place in mind for storing these?  z/OSMF does seem to allow for [storing / retrieving user & application data](https://www.ibm.com/support/knowledgecenter/SSLTBW_2.4.0/com.ibm.zos.v2r4.izua700/izuprog_API_PersistData.htm).   But it also seems sharing via source control would be easier.

jellypuno:
> I'm not sure. I was only thinking about how _**I**_ would want to share them and that would be in GitHub 😄
> 
> Did you have a place in mind for storing these? z/OSMF does seem to allow for [storing / retrieving user & application data](https://www.ibm.com/support/knowledgecenter/SSLTBW_2.4.0/com.ibm.zos.v2r4.izua700/izuprog_API_PersistData.htm). But it also seems sharing via source control would be easier.

@dkelosky In Ops before, we have a TWS application that contains jobs that can be scheduled so it can create specific types of datasets. Or we have a PDS that contains members that can be submitted to generate such datasets. That's why datasets are in my head. But some customers have it SMF. The DS attributes will be populated depending on volume or HLQ1

katelynienaber:
**To-do (in this order)**

First PR to be opened:
- Actually send attributes to MF
- Make function for showing the list of editable attributes reusable (ex. for profiles)

Later PRs (probably multiple)
- Research using source control to store/share data set templates between repos/teams
- Research filling of data set attributes according to HLQ1/volume (use SMF to fill attributes the user leaves undefined)
if attributes !== null then pass that information. If attribute === null then let mainframe fill that :)

katelynienaber:
Possible flows for Create Data Set & Allocate Like:
image
image

jelaplan:
We discussed using these flows for allocate like and create

image

jelaplan:
@dkelosky @jellypuno @tucker01 @MikeBauerCA

We continue to discuss adding additional templates. That is higher priority that allowing user to create new templates are edit templates but those ideas have potential. I have worked on a spreadsheet with Venkat, Katelyn, and Vit to define new templates. I'd like to socialize these ideas and discuss since the origin of the templates came from the early days of the CLI. The feedback we've gotten are that the templates that Venkat proposed are pretty solid but I want to get a wide opinion and maybe share these ideas with some people who write code at a MF shop. 
[Data Set Templates.xlsx](https://github.com/zowe/vscode-extension-for-zowe/files/5411681/Data.Set.Templates.xlsx)

In terms of storing the data, I think that still a matter of discussion. VS Code files, Git files, etc. Mike Bauer had ideas around this too. I think we should provide a solid set of defaults so that out-of-the-box, there are good templates.

jellypuno:
@jelaplan I have a few questions:
- Are we removing the name "Partitioned and Sequential Dataset" in the create dataset and are replacing them with template names?
- Should we create a survey asking users about their opinions about this?
- Are we exposing all the types and allowing users to edit them? (e.g. Data Class, Mgmt Class and Storage Class)

MikeBauerCA:
My opinion is that these templates should be shared amongst Zowe client technologies (SDKs, CLI, VS Code). Currently, we are moving towards merging Zowe settings and Zowe profile information into a JSON config -> https://github.com/zowe/imperative/pull/450. Also, in the CLI, these "templates" are currently represented via different commands with different default option values:
image

We need to leave existing commands in place for backwards compatibility for an extended period of time.

Given all of that, I would suggest placing these templates in the JSON config which can be shared across clients, adding a new `zowe files create data-set` command that also accepts a `--template` option, and deprecating the other commands that leverage the same z/OSMF endpoint. I would also allow the user the ability to fully customize/change/add to the set of default templates provided.

jelaplan:
+@venkatzhub

@jellypuno 
> @jelaplan I have a few questions:
> 
>     1 Are we removing the name "Partitioned and Sequential Dataset" in the create dataset and are replacing them with template names?
> 
>     2 Should we create a survey asking users about their opinions about this?
> 
>     3 Are we exposing all the types and allowing users to edit them? (e.g. Data Class, Mgmt Class and Storage Class)

1. WRT removing names, I don't think so. But it is open to discussion. We disucssed the possibility of supporting the existing names for backward compatibilty and adding the five additional templates. Mike had concerns with that in the sense that if they are similiar then we are creating duplication. 

2. WRT a survey, the additions to this issue was the start of expanding the conversation. I think a survey that reaches beyond our echo chamber is an excellent idea. I'm wondering how to frame the conversation and who to ask. 

3. WRT exposing all types, the CLI supports that. In Katelyn's POC, she also included these values 
image

venkatzhub:
@jelaplan @jellypuno  - our goal is to make things simple and easy. So by default - if the user is doing Allocate Like or selecting a Template - the dataset attributes should be displayed "if the user wants to" - but not displayed all the time. A lot of users (experienced or otherwise) do not really worry about what those attributes are - so the straight line path should be real simple i.e 
Allocate Data set --> Allocate like --> Specify a Data set name to be allocated --> Press Enter (optionally allow them to view the data set attributes)==> Display the allocated dataset (or error message if something failed)
Allocate Data set --> Select a Template --> Specify a Data set name to be allocated --> Press Enter (optionally allow them to view the data set attributes) ==> Display the allocated dataset (or error message if something failed)

Once we are there, then the additional scenarios of someone wanting to change the defaults in the templates or creating additional templates can be addressed. 

For the templates - keeping them task oriented (e.g COBOL source data set or Load module data set etc.) is natural. The default values for those are actually published. Also, the newer versions of compilers would require the load modules to be in PDSE's and PDSE usage is very prevalent for new datasets - so having the templates taking care of that and not letting the user worry about what they need to specify for a PDS vs SEQ vs PDSE would be optimal. For example: The default for a COBOL source dataset can be LIBRARY (PDSE) but if a user so chooses can (e.g via a drop down) select SEQUENTIAL or SEQ. Net net - the 80% of use cases in this can be handled via the pre-canned templates.

tucker01:
Keep in mind, as @jellypuno mentioned, folks can/do use SMS to manage allocations. We can create templates for allocation attributes... but we'll almost always be missing critical site specific info like SMS management/data/storage classes or `volser`. In fact, allocating without this info would cause some serious confusion in our shop (~3 days later 😄).

jelaplan:
@venkatzhub, can you point us to a site that shows the published values? I think having that in hand would be valuable. 

@tucker01 What is SMS? Does this impact the user experience? When we create data sets today, I don't think people specify a storage class.

venkatzhub:
@tucker01 - absolutely, the templates won't solve all the use cases, so as long as users have the flexibility to specify the required attributes when/if needed - we should be good. Just that we don't want force everyone to worry about it all the time.

venkatzhub:
@jelaplan : The values are in the specific language programming guides. For COBOL you can find it here: http://publibfp.boulder.ibm.com/epubs/pdf/igy6pg30.pdf?_ga=2.45657508.1807165600.1603123493-132489827.1600989033

dkelosky:
The templates are nice, but in practice, perhaps not many people create brand new data sets without first stealing all of the attributes from an existing, similar data set, e.g. in ISPF, you view data set information and then head to `3.2`.  

"Allocate Like" is awesome.  If we add the option (as Venkat said) to view & edit attributes while doing "Allocate Like" perhaps we could just leave the existing templates be.

jelaplan:
I don't think we should leave the existing templates be. Note that the impetus for revisiting the templates came from the survey that Zowe (ZLC + Squads) conducted for 2H2020. These come from customers / real users. A few quotes:
* worst choices for a dataset creation, makes me wonder who the target audience is.
* I don't know what you mean by Data Set Binary, or Data Set C etc. It would be nice to have a dialog where you could see the data set attributes and edit them if required
* I don't understand the difference between "C" and "Classic" - it's a PDS (with directory blocks), or a PDSE (LIBRARY). Is "Binary" a PDS or a PDSE with RECFM=U?
* Templates for all the above and PDS/E would be helpful!
* VSAM ??

dkelosky:
I don't know what 2H2020 is.  Did we have "allocate like" when that survey was done?

venkatzhub:
I agree with @jelaplan that based on the feedback we got, we need to do a better job with templates, I anticipate we will have both users - users using Allocate Like and Templates .. so making those two paths easy is the right thing to do. 
@dkelosky - Allocate Like was not there when the survey was done, and it would have impacted the survey results for sure.

jellypuno:
> @tucker01 What is SMS? Does this impact the user experience? When we create data sets today, I don't think people specify a storage class.

@jelaplan Just to answer the question, in all the companies that I've worked with, normally we specify the data class and that provides the default values that I could use. This is defined and set-up by Storage Team. 

Since I could see that dataclass , storage class etc are part of the fields that I could edit then I think that's okay.

jellypuno:
> WRT a survey, the additions to this issue was the start of expanding the conversation. I think a survey that reaches beyond our echo chamber is an excellent idea. I'm wondering how to frame the conversation and who to ask.

We could target the COBOL Programming Course group. Paul Newton is a wonderful SysAdmin that has a lot of experience in this. I bet he could provide great insights.

Issue #904, opened 2020-06-25
Implement the "like" data set capability available from zOSMF API


zdmullen:
**Is your feature request related to a problem? Please describe.**

I would like to be able to do "allocate like" inside of Zowe Explorer. This should be available now that zOSMF has a REST parameter for allocate like.

This should be implemented in Zowe CLI so that Zowe Explorer can implement it. 
Please refer to https://github.com/zowe/zowe-cli/issues/754 for SDK implementation details.

**Describe the solution you'd like**

Create a new data set and in the drop down of choices, be able to select "allocate like" at the bottom of the list.

Or

Right + click on a particular data set and select "Allocate like". 

Further refinement with regard to the UX will be needed, of course. 

**Additional context**

IBM documentation about the new REST parameter:

https://www.ibm.com/support/knowledgecenter/SSLTBW_2.4.0/com.ibm.zos.v2r4.izua700/IZUHPINFO_API_CreateDataSet.htm

jelaplan:
An idea here is to implement copying of data sets. And then have two paste options. One to paste the a copy of the dataset and one to allocate like. 

I'm wondering how fixed we are on the term allocate like. I guess that  is a mainstay of ISPF MF users but it is not self-revealing meaning wise. Pasting the file properties might be closer.

katelynienaber:
Working on getting this implemented in CLI [here](https://github.com/zowe/zowe-cli/pull/766).

jelaplan:
I have some concerns with the menu label Allocate Like. It is likely familiar to experienced mainframers. But for the modern developer experience, that name is obscure and hard to interpret. 

I looked at some [IBM doc around Eclipse and Z Open Development](https://www.ibm.com/support/knowledgecenter/SSUFAU_2.0.0/com.ibm.etools.zide.cmn.doc/topics/allocateLike.html) and read the following snippet "click Copy characteristics from an existing data set." That phrase is more descriptive of the feature. I wonder if we might append the menu label. Some ideas:
- Allocate Like (Copy File Type)
- Allocate Like (Copy File Format)
- Allocate Like (New File with Same File Format)
- Allocate Like (Copy DS Characteristics)
- Allocate Like (Copy File Details)
- Allocate Like (Copy File Attributes)
- Allocate Like (New File with  Same Attributes)
- Allocate Like (New Dataset)
- Create New Dataset Modelled on This
- New Dataset with the Same Format
- Create New Dataset Like This One

[...]

PR #959, opened 2020-07-28
Issue #904: Implement Allocate Like in Explorer

katelynienaber:
## Proposed changes

[Issue #904](https://github.com/zowe/vscode-extension-for-zowe/issues/904): Added Allocate Like command to data set context menus & command palette.

If the data set specified by the user to copy from doesn't exist, an error from the API will be shown.

allocateLike

## Release Notes

Milestone: 1.9
Changelog: Added Allocate Like command to data set context menus & command palette

## Types of changes

## Checklist


katelynienaber:
Failures are due to some dependency issues...

zdmullen:
So one question with this-- what if users don't have the proper release of zOSMF? From the documentation I can see that allocate like is available "V2R3 and V2R4 at 2020/4 with APAR PH22030." 

As this is the case, if a user has an older version of zOSMF, will the function be grayed out? Is there even a check for that? 

From what I can imagine, we should determine what the error message is when the Rest call is not available, and document this particular scenario. Thoughts?

katelynienaber:
> So one question with this-- what if users don't have the proper release of zOSMF? From the documentation I can see that allocate like is available "V2R3 and V2R4 at 2020/4 with APAR PH22030."
> 
> As this is the case, if a user has an older version of zOSMF, will the function be grayed out? Is there even a check for that?
> 
> From what I can imagine, we should determine what the error message is when the Rest call is not available, and document this particular scenario. Thoughts?

Unfortunately it seems that if we try to send a REST request with an invalid field name, it just gives us error 500:
image

I tried this with the CLI system tests as well as with Postman.

I did NOT try to send a request with more than just the faulty "like" field. I didn't, because if you are allocating like, this is the only parameter you will use. All other parameters are related to data set properties. Please let me know if you think this is a faulty assumption and I will try again :)

Alex mentioned that: "a sysprog would check the zosmf logs and see exactly to what is related..." but if the user is not a sysprog then they might not know what is going on.

Maybe we could add a note to the error? Like if allocateLike returns 500, we say something like:
**Note: This command requires zOSMF v2.4.0 or higher**

Alexandru-Dumitru:
@katelynienaber Your assumption was good. I think we should not go down the rabbit hole of interpreting the errors on behalf of the user. Looking in the documentation for that specific RC and Reason, I got this (page 573):
```
rc: 4 
reason: 13
message: Incorrect parameter was specified.
description: The request contained one or more parameters that are not valid.
```
If they do this basic root cause analysis they should be fine, in my opinion.
Info extracted from: https://www-01.ibm.com/servers/resourcelink/svc00100.nsf/pages/zOSV2R4sc278420/$file/izua700_v2r4.pdf

jelaplan:
Wrt to the 500 error, I'm not sure if you will add some nuance or take Alex advise to not go down the rabbit hole.  I'm a fan of trying to improve because the error messages are so unhelpful. We are trying to modernize the MF experience so 1960s messages don't fit.  

A simple improvement is to state that the command is not available on your Mainframe system and contact your system team to update your version of zOSMF TO V2.4.0+.  That way they could stop trying to make it work.

katelynienaber:
Hi John,

I will probably go with Alex's method, for 2 reasons.

First, the 500 error is very general and just means the server was unable
to complete the request. The user could get this error for some other
reason, and our clarifying message would maybe tell the user the wrong way
to fix it.

Also, we would have to keep updating our error messages whenever IBM added
something new. For example, if the user IS able to use the allocateLike
command, but has some other problem with their request/connection to zOSMF,
we would still tell them that they need a different version of zOSMF to use
the command.

So ideally IBM should update their own error messages so they make sense to
newbies :)

Thanks,
Katelyn Nienaber
Software Engineer | Mainframe Division
CA Technologies, A Broadcom Company
Phone: +420 226 207 737 ext. 7737
CA CZ, s.r.o. | V Parku 2316/12 Chodov | Praha 4, 148 00 Czech Republic
Katelyn.Nienaber@broadcom.com | broadcom.com
Identification No. 25694073, registered in the Commercial Register
maintained by the Municipal Court in Prague in Section C, Insert No. 61808


Dne po 17. 8. 2020 16:11 uživatel John laPlante <notifications@github.com>
napsal:

katelynienaber:
> I'm not sure I love the label in the right+click, Allocate Like (New File with Same Attributes), but it is the clearest. LGTM

I'm also not super pleased with how long it is...if you guys prefer something from @jelaplan's list, I can change it :)

- Allocate Like (Copy File Type)
- Allocate Like (Copy File Format)
- Allocate Like (New File with Same File Format)
- Allocate Like (Copy DS Characteristics)
- Allocate Like (Copy File Details)
- Allocate Like (Copy File Attributes)
- Allocate Like (New File with Same Attributes)
- Allocate Like (New Dataset)
- Create New Dataset Modelled on This
- New Dataset with the Same Format
- Create New Dataset Like This One

jellypuno:
@katelynienaber I'm trying to run build but I am having failures. Do I need to install something? I have the latest version of Zowe CLI.

image

crawr (commented):
Functionality works great for me for sequential and PDS, both in VSCode and in Theia.

However, I see that the method is also implemented in the context menu of data set members. I think this should be removed since you don't really allocate members, you simply create a new one under an allocated PDS.

jellypuno:
> Functionality works great for me for sequential and PDS, both in VSCode and in Theia.
> 
> However, I see that the method is also implemented in the context menu of data set members. I think this should be removed since you don't really allocate members, you simply create a new one under an allocated PDS.

I agree. An error message appears when you try to allocate like under members

jellypuno:
I found several issues:
1. Delete datasets (PD and PDS) are not working. It is working in master
2. Errors in Integration tests as well connected to delete.

katelynienaber:
> I found several issues:
> 
> 1. Delete datasets (PD and PDS) are not working. It is working in master
> 2. Errors in Integration tests as well connected to delete.

This should be fixed now..somehow master got merged into this branch but without the fix for delete.

katelynienaber:
> Functionality works great for me for sequential and PDS, both in VSCode and in Theia.
> 
> However, I see that the method is also implemented in the context menu of data set members. I think this should be removed since you don't really allocate members, you simply create a new one under an allocated PDS.

2nd push removes the Allocate Like option from the member context menu :)

jellypuno:
Delete is working as expected. Thanks for fixing. I found 1 more.

- Allocating through favorites doesn't work

AllocLike

katelynienaber:
> Delete is working as expected. Thanks for fixing. I found 1 more.
> 
> * Allocating through favorites doesn't work

Fixed!

AndrewTwydell (approved):
LGTM 👍

crawr (approved):
Thanks for addressing my comment re: members. Allocating under favorites worked for me.

PR #978, opened 2020-08-12
Fix issue 962

### LINKED ISSUE #962: Rename.dataset() doesn't convert dataset name to uppercase
**Describe the bug**

When renaming a dataset (including favourites), the rename function doesn't convert any lowercase characters in the dataset name to uppercase unlike create.dataset() where each lowercase character is made uppercase.

**To Reproduce**

1. Create a dataset.
2. Click on 'Rename'.
3. Rename dataset using a mixture of uppercase and lowercase characters.
4. See error (Dataset displays with the characters inputted in step 3).

**Expected behavior**

All dataset name characters should be displayed as uppercase after a rename where the user uses lowercase characters.

**Screenshots**

Screenshot 2020-07-29 at 16 05 33

-   OS: MacOS
-   Zowe Explorer Version: 1.7.1
-   (Optional) Zowe CLI Version: 6.19.1
-   (Optional) Are you using Secure Credential Store? No

**Additional context**

N/A

AndrewTwydell:
## Proposed changes

Resolves #962. This fix ensures datasets and dataset members are displayed in uppercase when renamed by the user.
Not sure which milestone this falls under.

## Release Notes

Milestone:

Changelog: Ensures datasets and dataset members are displayed in uppercase when renamed by the user.

## Types of changes

## Checklist

### DIFF
--- a/src/dataset/DatasetTree.ts
+++ b/src/dataset/DatasetTree.ts
@@ -632,19 +632,18 @@ export class DatasetTree extends ZoweTreeProvider implements IZoweTree<IZoweData
         } else {
             dataSetName = node.getParent().getLabel();
         }
-        const afterMemberName = (await vscode.window.showInputBox({ value: beforeMemberName })).toUpperCase();
+        const afterMemberName = await vscode.window.showInputBox({ value: beforeMemberName });
         const beforeFullPath = getDocumentFilePath(`${node.getParent().getLabel()}(${node.getLabel()})`, node);
         const closedOpenedInstance = await closeOpenedTextFile(beforeFullPath);
 
-        if (!afterMemberName || afterMemberName == beforeMemberName) {
-            throw new Error("New member name empty or unchanged.");
-        }
-
         this.log.debug(localize("renameDataSet.log.debug", "Renaming data set ") + afterMemberName);
-        if (afterMemberName) {
+        if (afterMemberName && afterMemberName.toUpperCase() !== beforeMemberName) {
             try {
-                await ZoweExplorerApiRegister.getMvsApi(node.getProfile()).renameDataSetMember(dataSetName, beforeMemberName, afterMemberName);
-                node.label = afterMemberName;
+                await ZoweExplorerApiRegister.getMvsApi(node.getProfile()).renameDataSetMember(
+                    dataSetName,
+                    beforeMemberName,
+                    afterMemberName.toUpperCase());
+                node.label = afterMemberName.toUpperCase();
             } catch (err) {
                 this.log.error(localize("renameDataSet.log.error", "Error encountered when renaming data set! ") + JSON.stringify(err));
                 await errorHandling(err, profileLabel, localize("renameDataSet.error", "Unable to rename data set: ") + err.message);
@@ -659,7 +658,7 @@ export class DatasetTree extends ZoweTreeProvider implements IZoweTree<IZoweData
             if (otherParent) {
                 const otherMember = otherParent.children.find((child) => child.label === beforeMemberName);
                 if (otherMember) {
-                    otherMember.label = afterMemberName;
+                    otherMember.label = afterMemberName.toUpperCase();
                     this.refreshElement(otherMember);
                 }
             }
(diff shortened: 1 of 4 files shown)


src/dataset/DatasetTree.ts
```diff
@@ -632,15 +632,18 @@ export class DatasetTree extends ZoweTreeProvider implements IZoweTree<IZoweData
         } else {
             dataSetName = node.getParent().getLabel();
         }
-        const afterMemberName = await vscode.window.showInputBox({value: beforeMemberName});
+        const afterMemberName = await vscode.window.showInputBox({ value: beforeMemberName });
```
lauren-li, line 635: Could this PR be made simpler by just doing `toUpperCase()` at the point when `afterMemberName` is assigned in this function?

src/dataset/DatasetTree.ts
```diff
@@ -685,31 +688,31 @@ export class DatasetTree extends ZoweTreeProvider implements IZoweTree<IZoweData
             favPrefix = node.label.substring(0, node.label.indexOf(":") + 2);
             beforeDataSetName = node.label.substring(node.label.indexOf(":") + 2);
         }
-        const afterDataSetName = await vscode.window.showInputBox({value: beforeDataSetName});
+        const afterDataSetName = await vscode.window.showInputBox({ value: beforeDataSetName });
```
lauren-li, line 691: Similar to my comment for `renameDataSetMember` - are there any downsides to just doing`toUpperCase()` at the point when `afterDataSetName` is assigned (rather than throughout the function)?

src/dataset/DatasetTree.ts
AndrewTwydell: Hi Lauren, thanks for reviewing my PR. 
Initially, we did just put the `toUpperCase()` method on the line that `afterMemberName` was assigned. However, when testing this, the following error occurred when cancelling the rename by pressing escape.

Screenshot 2020-08-13 at 10 15 59

This error occurred because the input box was returning nothing, hence why we changed it.

Do you think this error is significant enough to justify this approach?

src/dataset/DatasetTree.ts
AndrewTwydell: Hi Lauren, this is the same issue as the comment above. Please refer to that response 😄

src/dataset/DatasetTree.ts
lauren-li: @AndrewTwydell Thanks for pointing out the case where the user cancelled via Escape!

Could we avoid running into this error by doing a check for `afterDataSetName` after it's assigned, and returning early out of the rename function if `afterDataSetName` is undefined? Then, if it's not undefined, we could convert `afterDataSetName` to uppercase and continue on with the rename function. For instance, something like:

```
let afterDataSetName = await vscode.window.showInputBox({value: beforeDataSetName});
if (!afterDataSetName) {
    vscode.window.showInformationMessage(localize("renameDataSet.cancelled", "Rename operation cancelled."));
    return;
}
afterDataSetName = afterDataSetName.toUpperCase();
... (the rest of the function...)
```

Would this work?

AndrewTwydell: Another great suggestion @lauren-li, thank you 😃 

However one of the many different implementations tried was very similar to this. The issue I found was that `afterDataSetName` was declared as constant, so could not then be changed to uppercase with `afterDataSetName = afterDataSetName.toUpperCase();` after I made sure it was populated.

Above, you have declared it using `let` as opposed to `const` which would solve this problem.

Do you recommend changing `afterDataSetName` to be a variable rather than constant?

Pranay154: Hey @lauren-li , thanks for the suggestion, I tried this and it works! I will push the changes shortly 😃

lauren-li: Thanks @Pranay154 !

@AndrewTwydell I think it should be okay to use `let` instead of `const` in this case. That gives us the flexibility to reassign `afterDataSetName` to the uppercase version if it's not undefined.

If there is a concern with using `let afterDataSetName`, you could use `const afterDataSetName` (as you did before), and then assign the value `afterDataSetName.toUpperCase();` to a differently-named constant that you could then use for the remainder of the rename function. However, I didn't see the case for doing that, so I went ahead and used `let afterDataSetName`.

Pranay154:
@zFernand0 @lauren-li  Thanks for taking the time to review!

Is there a solution to the `npm audit` failing in the pipeline?

zFernand0:
> @zFernand0 @lauren-li Thanks for taking the time to review!
> 
> Is there a solution to the `npm audit` failing in the pipeline?

See #980 :yum:

PR #1030, opened 2020-09-30
Base profile billie

JillieBeanSim:

### DIFF
--- a/__tests__/__unit__/Profiles.unit.test.ts
+++ b/__tests__/__unit__/Profiles.unit.test.ts
@@ -29,6 +29,7 @@ import { createUSSSessionNode, createUSSTree } from "../../__mocks__/mockCreator
 import { createJobsTree, createIJobObject, createJobSessionNode } from "../../__mocks__/mockCreators/jobs";
 import { DefaultProfileManager } from "../../src/profiles/DefaultProfileManager";
 import { IZoweNodeType } from "../../src/api/IZoweTreeNode";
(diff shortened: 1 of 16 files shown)


katelynienaber (changes requested):
There are some errors but no circular dependency error in unit tests so AWESOME 👍 . Anyway the errors I found were:

- After editing a profile to have no user/pass (no base profile existing) I got this error, and whenever I tried to use the profile the same error occurs, instead of prompting. It also prevented having the new details stored to the yaml file. Same error occurs when I try to just store user but not password:
image
- When editing a profile to add user/pass, I get this error...it doesn't prevent the profile info being saved, and it doesn't occur if I create a new profile with all details:
image
- I created a new profile, and I used as the host: https://hosturl.com/. I got an error and when trying to use the profile I was then prompted for all information stored in the profile (not just the host). Maybe we could specify the format that should be entered here, or strip the https:// out of it?
- When trying to see jobs under a profile (base profile existing and should be used), I get this error (however favorites work fine):
image

Let me know how I can help! I could maybe work on unit tests at the same time as you are doing the code? But I don't want to touch your branch so lmk.

jellypuno:
@katelynienaber If you merge this branch to your branch then we are good to go on commit changes. I am also planning to commit some.

katelynienaber:
> @katelynienaber If you merge this branch to your branch then we are good to go on commit changes. I am also planning to commit some.

@jellypuno justmerged it

PR #1032, opened 2020-10-01
BREAKING CHANGE: Make ZE secure by default and add SCS setting

### LINKED ISSUE #574: Store credentials securely by default
The current behavior (in Zowe Explorer 1.5) is as follows:
* If the Secure Credential Store plugin is installed, ZE detects this and always stores credentials securely.
* If the Secure Credential Store plugin is *not* installed, ZE always stores credentials in plain text.

I believe the desired behavior is as follows:
* Zowe Explorer should *always* store credentials securely. It should not matter:
  * Whether Secure Credential Store plugin is installed
  * Whether "Zowe Security: Credential Key" is set or blank
* In case users do not want this behavior, perhaps a new option should be added to ZE that disables the default behavior of storing credentials securely.

### LINKED ISSUE #820: Zowe Security: Credential Key is confusing
As a new user of ZOWE Explorer, I am confused by options to store credentials.

Screenshot 2020-05-28 at 10 15 39

Better formulation, explanation, or a link to documentation that explains the options would be helpful.

ghost:
## Proposed changes

Resolves #574 (Store credentials securely by default). Also resolves #820 (Zowe Security: Credential Key is confusing).

This PR changes the default behavior of ZE to make it always store credentials securely, unless a user explicitly disables this behavior and chooses to store credentials in plain text.

These are the new security settings:
image
The first setting already existed, and its description was updated to reflect that most users can leave it blank. The second setting was added to allow users to disable secure credential storage, and is enabled by default.

When one of the above settings is changed, the following message now appears:
image

## Release Notes

Milestone:

Changelog: Make Zowe Explorer secure by default and add security settings

## Types of changes

## Checklist

## Further comments

This solution was chosen based on the assumption that we want security by default.

This replaces the previous behavior which was confusing:
* ZE would detect the scenario where Zowe CLI was installed without the Secure Credential Store plugin, and in this case stored credentials in plain text. Otherwise it would store them securely.
* The "Zowe Security: Credential Key" setting was documented incorrectly.


phaumer:
Ideally this should be done in the new api package in the monorepo that extensions will be able to reuse this for their profiles as well.

jellypuno:
@t1m0thyj Can we close this PR? Do you still want this to be implemented?

t1m0thyj:
@jellypuno I think we can close it. This PR would need to be rewritten to use the new API package in the monorepo, and I'm not sure whether the breaking change is desirable.

Perhaps we can revisit the related issues (#574 and #820) in the future when ZE adopts the new Profile Information API (https://github.com/zowe/imperative/issues/556)?

PR #1054, opened 2020-10-28
Update Keytar and Jest dev deps for Node 14

t1m0thyj:
This is a quick fix to make it possible to install and build ZE with Node.js 14 (the newest LTS version).


zFernand0 (approved):
LGTM!
I believe this is still applicable and we probably want to update these packages.
Thanks @t1m0thyj

t1m0thyj:
Updated this PR since it was out of date in several ways:
* Merged with master to use the new monorepo folder structure
* Removed package-lock.json and updated yarn.lock file
* Updated Keytar from v6 to v7 which adds support for Node.js 15

zFernand0 (approved):

Issue #914, opened 2020-06-26
Implement prompting for token/credentials (SSO for API ML)


zdmullen:
**Is your feature request related to a problem? Please describe.**

Part of SSO for API ML implementation

**Describe the solution you'd like**

With base profiles that I have created via Zowe CLI, if my token is no longer valid, I expect be to be prompted and able to log in again to get a token and begin my work. 

**Additional context**

This follows from the previous issue, #913. 

This will require some UX consideration.

zFernand0:
After discussions with the CLI squad, we can collaborate in ways to expose a callback function for providing credentials from the Zowe Explorer. See https://github.com/zowe/imperative/issues/422 for more details.

Issue #1000, opened 2020-09-03
Refactor Code for Profiles.ts


jellypuno:
**Is your feature request related to a problem? Please describe.**

I would like to re-factor the overall solution of Profiles.ts Maybe create a folder structure similar to datasets, uss or jobs

**Describe the solution you'd like**

Discuss with the squad on how to do efficiently do this.

**Describe alternatives you've considered**

**Additional context**

jellypuno:
Minor Issues from #979 that we can probably address

jellypuno:
TODO: Move addSession and addSingleSession to abstract/ZoweTreeProvider. Pattern it to editSession

zFernand0:
With the latest improvements to the ZE API NPM package, I believe this issue can be marked as resolved.

PR #1062, opened 2020-11-09
Issue 914: Add login and logout functions for base profiles

crawr:

## Proposed changes

Issue #914 

**This PR adds the following functions for base profiles:**

1. **Log in to Authentication Service.** The user will be prompted for username and password beforehand, same as in the CLI. The token will be stored in the default base profile yaml file.

log in to auth

2. **Log out from Authentication Service**

drawing
drawing

3. **Log in prompt.** This pop-up will appear when the base profile token is invalid or expired.

drawing

4. For regular service profiles, the log in and log out will be skipped

drawing
drawing

## Release Notes

Milestone: 2.0 (?)

Changelog: Added login and logout options for base profiles

## Types of changes

## Checklist

## Further comments

- The scope of this PR includes only base profiles already created via the CLI
- The login and logout functions will only be available for profiles relying on the base profile. It will be skipped for all other service profiles.


zdmullen:
I will test this out later today, hopefully. Please synchronize with @IgorCATech about documentation updates for this. Thanks!

jelaplan:
Correct me if I'm wrong, but I think we should use Login and Logout without the space. That is how it is in the CLI (zowe auth login apiml). 
image

jelaplan:
The CLI team chose to have syntax that supported logging in through other services. Presently, there is only apiml (i.e., zowe auth login apiml) but there could be others (i.e., zowe auth login otherService). There is only one service today but if we were to add another then we might be adding addditional menu items to login to x and logout of x. In that case, we might want to give users the choice in the quick pick. 

I think we should spell out the term mediation layer rather than ML because people may not be familiar with the abbreviation. 
Login to API Mediation Layer
Logout to API Mediation Layer

jelaplan:
Up until now, if a user wanted to use a different user name and password, they would click the pencil icon to update profile. Now we are adding a seperate way to do something similiar. 
image

I'm wondering whether you implemented the prompt "Would you like to login to apiml?"
image

jelaplan:
> 2. Log out from API ML
It also adds a re-prompt option when the base profile token is invalid:
image

WRT to your point 2., is this a situation where the base profile contains a token but it has expired or is invalid for some reason? I'm guessing you do some sort of validation. I think we should avoid referring to tokens and base profiles. Here is an rewritten message. 
Restore Connection



image

zdmullen:
At @jelaplan, so just to clarify several things. This flow that Richelle has introduced is for those people who are using a base profile that they have created from the CLI (and logged in with) and use in Zowe Explorer. If their token expires, they need a way to easily log in again. That's what this all provides. 

For the commands, I believe that in the context menu it should still read "log in" since it is a verb in this case. In the message that pops up, however, it should read "Login to the API ML was successful." 

Also, on your second message suggestion, I think I agree with the more general message, though the user will know that they are using that profile with a token. In this message, it should say "Please log in..." by the way.

crawr:
Thanks, @jelaplan @zdmullen for looking into this!

- I updated the PR description with the new wordings you suggested. Just to note that in the CLI, the description of the command is also two words. 
- I would keep the options to just API Mediation Layer for now, but it's good to note for the future to build that quick pick for the authentication service.
- The "edit session" icon will go away based on another open PR so everything will be in the context menu
- The "Would you like to login to apiml?" text has been replaced with the "Please log in" prompt

Question: What do you think of the "Log in skipped" text?

JillieBeanSim:
@crawr I have a question about this, does this look for the profile type "base" to decide or does it look at the meta data for the field option to have a tokenValue and tokenType?  I ask this since other profile types can have the tokenValue and tokenType fields in their meta data to authenticate with their server via a JWT token.  It would be nice to have this login and logout option for those profile types as well, and with this the wording would need to not include API ML but possibly have the wording to include "login for authentication" or "logout for authentication"

packages/zowe-explorer/src/Profiles.ts
```diff
@@ -1103,6 +1121,154 @@ export class Profiles extends ProfilesCache {
...
+                const updSession = new Session({
+                    hostname: combinedProfile.profile.host,
+                    port: combinedProfile.profile.port,
+                    user: newUser,
+                    password: newPass,
+                    rejectUnauthorized: combinedProfile.profile.rejectUnauthorized,
+                    tokenType: combinedProfile.profile.tokenType,
+                    type: combinedProfile.profile.type,
+                });
+                const loginToken = await zowe.Login.apimlLogin(updSession);
```
phaumer, line 1186: This needs to be added to the Common API as we cannot call zOSMF specific implementations in Zowe Explorer code.

phaumer (dismissed):
z/OSMF only implementation needs to be refactored to use an abstract API.

packages/zowe-explorer/src/Profiles.ts
```diff
@@ -1103,6 +1121,154 @@ export class Profiles extends ProfilesCache {
...
+            baseProfile &&
+            serviceProfile.profile.host &&
+            serviceProfile.profile.port &&
+            ((baseProfile.profile.host !== serviceProfile.profile.host &&
+                baseProfile.profile.port !== serviceProfile.profile.port) ||
+                (baseProfile.profile.host === serviceProfile.profile.host &&
+                    baseProfile.profile.port !== serviceProfile.profile.port))
+        ) {
+            vscode.window.showInformationMessage(
+                localize("ssoLogout.noBase", "Log out skipped. This profile does not support base profiles.")
```
JillieBeanSim, line 1246: this info message too.  maybe just "This profile does not support Log out"

packages/zowe-explorer/src/Profiles.ts
```diff
@@ -1103,6 +1121,154 @@ export class Profiles extends ProfilesCache {
...
+            const combinedProfile = await Profiles.getInstance().getCombinedProfile(serviceProfile, baseProfile);
+            const updSession = new Session({
+                hostname: combinedProfile.profile.host,
+                port: combinedProfile.profile.port,
+                rejectUnauthorized: combinedProfile.profile.rejectUnauthorized,
+                tokenType: combinedProfile.profile.tokenType,
+                type: combinedProfile.profile.type,
+                tokenValue: combinedProfile.profile.tokenValue,
+            });
+            await zowe.Logout.apimlLogout(updSession);
```
JillieBeanSim, line 1261: can't have specific zowe calls outside of the api file, the api file in the common section could have new login and logout methods that could be utilized

packages/zowe-explorer/src/utils/ProfilesUtils.ts
```diff
@@ -39,6 +39,10 @@ export function errorHandling(errorDetails: any, label?: string, moreInfo?: stri
         "Invalid Credentials. Please ensure the username and password for {0} are valid or this may lead to a lock-out.",
         label
     );
+    const errToken = localize(
+        "errorHandling.invalid.token",
+        "Your connection is no longer active. Please log in to the API Mediation Layer to restore the connection."
```
JillieBeanSim, line 44: this error could be more generalized to include alternate profile types whose token has expired, "Please log in to restore the connection."

packages/zowe-explorer/src/utils/ProfilesUtils.ts
```diff
@@ -51,6 +55,24 @@ export function errorHandling(errorDetails: any, label?: string, moreInfo?: stri
...
 
+            if (errorDetails.mDetails.additionalDetails) {
+                const tokenError: string = errorDetails.mDetails.additionalDetails;
+                if (tokenError.includes("Token is not valid or expired.")) {
+                    if (isTheia()) {
+                        vscode.window.showErrorMessage(errToken).then(async () => {
+                            await Profiles.getInstance().ssoLogin(null, label);
+                        });
+                    } else {
+                        vscode.window.showErrorMessage(errToken, "Log in to API Mediation Layer").then(async (selection) => {
```
JillieBeanSim, line 66: same here with error being more generalized to work with other profiles that don't authenticate with the API Mediation Layer.

JillieBeanSim (changes requested):

crawr:
Thank you @JillieBeanSim for reviewing. 

- I moved the login and logout call to the common API so that extenders can use their own functions
- I also updated the wordings as you suggested to make the authentication service generic
- This PR only handles log in and log out via base profiles. From the combined profile, there is no way for me to know whether the token is stored in the service profile or the base profile. So my assumption is that, if there is a token, it must come from the base profile and that is the profile I need to update with the new token from the login.
- Perhaps tokens stored in service profiles could be handled under a different issue and a different PR. I also have no profiles where I can test this use case.

JillieBeanSim (dismissed):
@crawr I have a question, for logout and the token is expired in the base profile I am getting this error below.  Shouldn't the logout also prompt for creds as I see the login does or give an error that states that it's expired to login like the prompt shown above?  Also, I didn't get the login prompt with the token being expired
image


I did a login via the cli and got a new token, logout after that was successful and login was successful via Zowe Explorer.  After obtaining a token the profile that matches with base gives me an error in Zowe Explorer but works via cli, even refreshing the view and vs code I still get the error like it's not recognizing the token in the base profile

image


image

Also if the logout has been issued and the profile doesn't have a valid token (since logout with Zowe Explorer doesn't remove token from yaml of base like it does with cli) shouldn't I get either a notice to login or a prompt to supply user/pass for authentication?  All I get is the error about profile being inactive.

jellypuno (approved):
So far everything looks good. I tested all functionalities with PR #1064 merged

crawr:
Hi @JillieBeanSim. Yes, the logout will not prompt for credentials. This behavior is also consistent with the CLI auth logout function.

I believe the other issues you saw are related to the refresh. This is being handled by Jelly's PR as it's a separate issue. So I think that PR should be merged first before this PR.

JillieBeanSim:
@crawr & @jellypuno if both PRs are needing to be merged to test should this have been one PR?

Also after merging locally and testing I ran into some issues
1. on login I got this error in ZE but was able to do the `auth login apiml` command via cli with no issues using these profiles below (base on the left and zosmf profile on the right)
image
image

2. after obtaining a token via cli I still get this error when trying to do a filter search for MVS (I am through the firewall)
image

3. after doing the right-click logout option I get a successful information message for logout but when I check the YAML the value has not been removed the tokenType and tokenValue as the cli logout command does

lauren-li:
@crawr One polish item for this PR would be if we can move the **Log in**/**Log out** context menu items higher up in the section they're in. Usually, the **Delete** action is at the very bottom of the context menu per VS Code's conventions, and it would be good to be consistent with their UI. (Actually, **Enable/Disable Validation** is taking that bottom place now, but perhaps we can discuss fixing that in a separate PR.)

crawr:
Hello @JillieBeanSim, I believe they should not be one PR since they are separate issues, but the problem arises when testing this PR.

1. This is a good use case as some base profiles may not contain any token type. I moved the setting of the token type to the zosmf specific API function, making use of the defined APIML constant in imperative. 
2. After logging in from the CLI, the ZE tree has to be refreshed for the token to be picked up by the loaded profile. 
3. Initially, I did not want to remove the token from the profile and the logout function just invalidates the token's session. But to be consistent with the CLI behavior, I addressed this as well and followed how imperative handles it. Now, the logout process removes both tokenType and tokenValue from the default base profile yaml.

lauren-li:
With @JillieBeanSim's help, I was able to log into API ML on Zowe CLI! Thanks Billie!

@crawr Should both tokenType and tokenValue be added to the base profile upon clicking "Log in" and entering credentials? Mine only shows the tokenValue getting added, but when I log in using Zowe CLI, I get both tokenType and tokenValue.

After logging in and getting the token, if I did not put credentials in my profile, I get the message that my profile is inactive, which keeps me from using that profile to filter data sets or do anything else:
Screen Shot 2020-11-24 at 3 20 44 PM

Here is what my base profile and z/OSMF service profile look like after logging in, respectively. For this screenshot, I left in the empty strings for username/password that occur when creating the profile in Zowe Explorer, but I get the same "inactive profile" error when removing them, as well as when also removing the lines for `encoding` and `responseTimeout`:
Screen Shot 2020-11-24 at 3 25 06 PM

jellypuno:
@lauren-li Can you please try to click RefreshAll after logging in through Zowe CLI? Please let me know of the results.

Also, please remove the username and password. In ZE, this profile is considered as an Optional Profile and will not go through the token process. (This is a precedence issue that is not handled in this PR and needs more UX )

lauren-li:
> @lauren-li Can you please try to click RefreshAll after logging in through Zowe CLI? Please let me know of the results.
> 
> Also, please remove the username and password. In ZE, this profile is considered as an Optional Profile and will not go through the token process. (This is a precedence issue that is not handled in this PR and needs more UX )

@jellypuno I got the same "inactive profile" message when the username/password lines are removed from my profile. (I also removed the lines for `encoding` and `responseTimeout` and got the same result.) Clicking the Refresh button and trying to filter data sets in Zowe Explorer after successfully doing the API ML login and retrieving data sets via Zowe CLI also results in the same "inactive profile" message.

jellypuno:
@lauren-li Can you please confirm if these are the steps that you have done:

1. Login using `zowe auth login apiml`
2. Create a profile in Zowe Explorer
3.  Refresh the tree
4. Filter a dataset
5. Profile inactive happens

is this correct?

jellypuno:
@lauren-li So I tested this one and the error is connected to refresh. After logging out and logging in again, the API for validating the session is not identifying my profile as a profile that uses a token. No idea why.

Untitled

Richelle and I will discuss on how we could handle this.

crawr:
- Refresh issue has been fixed
- Login process updates default base profile with both tokenType and tokenValue
- Logout process removes both tokenType and tokenValue from default base profile

lauren-li:
> @lauren-li Can you please confirm if these are the steps that you have done:
> 
>     1. Login using `zowe auth login apiml`
>     2. Create a profile in Zowe Explorer
>     3. Refresh the tree
>     4. Filter a dataset
>     5. Profile inactive happens
> 
> is this correct? Also, the programs are correctly compiled?

@jelly The PR compiles without any issues for me. However, my steps are more like:

1. Use Zowe CLI (`zowe auth login apiml --ru false`) to create a base profile with hostname, port, and `rejectUnauthorized: false`.
1. Create a profile without credentials in Zowe Explorer or Zowe CLI. (But if creating via Zowe Explorer, manually remove the lines with empty string or `0` values from the profile's YAML.)
1. Right-click on the profile in Zowe Explorer and select "Log in to Authentication Service".
1. When prompted, enter username and password.
    - I get the error message: `Unable to log in. Must have user & password OR base64 encoded credentials`.

I then clicked the (Data Sets) view's "Reset" button and tested with logging in via the CLI:
1. Have a Zowe base profile created, as well as a Zowe z/OSMF profile without credentials. (Remove lines with empty strings or `0` values in the profile's YAML.)
1. Use Zowe CLI command `zowe auth login apiml --ru false` and follow the prompts to create a base profile named "default".
1. Refresh the Zowe Explorer tree view by clicking the view's "Refresh" button.
1. Right-click on the profile in Zowe Explorer and select "Log in to Authentication Service". Follow the ZE prompts to input your username and password.
    - This step appears to succeed, because the `tokenType` and `tokenValue` fields are added to the base profile's YAML (or updated if they were already there, and I get the message: "Login to authentication service was successful."
1. Click on the magnifying glass for the profile to try to run a search filter.
    - This results in the error message: `Error: Profile Name noCreds is inactive. Please check if your Zowe server is active or if the URL and port in your profile is correct.` 
        - Looking at the error message that is passed to the log (via line 1050 of `Profiles.ts`), it says `Error: self signed certificate in certificate chain`.
        - Despite the error message above for Zowe Explorer, I am still able to use this profile to successfully retrieve data sets via Zowe CLI.

@crawr An additional polish item to consider:
- Do we want to add error handling for if the user clicks "Log in to Authentication Service" or "Log out from Authentication Service" while having no base profile? I noticed that if I have no base profile, when I click "Log in...", nothing happens. If I click "Log out...", then I get the following error message: `Unable to log out. Cannot read property 'profile' of undefined`. Perhaps we can direct users on how to create a base profile if none is found when they click on these commands.

crawr:
Hi Lauren, thanks for the detailed explanation of the process you performed. 
Because you pointed it out, I tested it on master branch and I was able to replicate the issue. It stems from the fact that during **creation of new profiles**, the combination of base and service profiles in one session is not performed. I can open a separate issue for this.
As a workaround, if you try to hide the profile and add it back, the session works as expected, with the base and service combined. It's just the creation of new profiles that is affected.

Yes, I can add the checking for no base profiles and have the message similar: `This profile does not support login.`

lauren-li:
@crawr Thanks for suggesting the workaround! Unfortunately, even with it, I still get the same errors about the `user & password OR base64 encoded credentials` or `self-signed certificate chain` when trying to log in/run a search filter on my combination of service profile + base profile. (I also ran into some strange behavior where I was occasionally unable to actually hide/re-add the profile, but could not figure out how to recreate that issue reliably.)

However, I also checked on the `master` branch, and got these same errors there. So perhaps my issue is not specific to this PR.

JillieBeanSim:
@crawr I retested your changes and I am running into the same issues as @lauren-li.  It's like going through the API ML the rejectUnauthorized = false is not being recognized and I get the error about self-signed cert when trying to access.  The strange thing is I have a second profile that is using the zosmf default port directly instead of the API ML port with base path and I am now getting the same error for that profile.  The merge of profiles (base & service) is still happening even though the port is different and adding the token type & value to the profile that is passed so I don't get prompted for the user/pass since it's not in the profile.  With this now none of my profiles for that server is working.  This is a breaking issue for a user that has self-signed certs and wants to use a base profile.  This is not the case using the base and zosmf profile via the CLI, with the CLI the ru value is recognized and no errors are recieved.
ze-ru-not-recognized

zFernand0:
would that be related to https://github.com/zowe/vscode-extension-for-zowe/issues/1073
`--ru` not used when going through the APIML?

crawr:
Thanks for linking that issue, @zFernand0. So it seems this is happening in our master branch as well? 

@JillieBeanSim I tried to replicate it on our system
1. Login to apiml via CLI `zowe auth login apiml --ru false`
```
host: hostname.com
port: 443
rejectUnauthorized: false
tokenType: apimlAuthenticationToken
tokenValue: validtoken
```
2. Create zosmf service profile
```
rejectUnauthorized: false
basePath:           api/v1
```
3. I am able to use this profile both in CLI and ZE (master and my branch)

Can you confirm this happens in master branch for you? It might be a problem in the APIML setup, too as I understand from the linked issue.

And can you confirm what your base profile and service profile contains so I can try to replicate it as close as possible?

JillieBeanSim:
@crawr I had this initially testing with default being the base profile, but I tried changing the profiles to look similar to what you just showed and still get the same error.  I looked at the issue @zFernand0 linked but didn't see a solution posted just that the customer was straight and Dan closed it still having the issue.
image

dkelosky:
@JillieBeanSim - In what you pasted, is `default.yaml` representing an API ML profile and `tivlp02.yaml` representing a z/OSMF profile?

JillieBeanSim:
> @JillieBeanSim - In what you pasted, is `default.yaml` representing an API ML profile and `tivlp02.yaml` representing a z/OSMF profile?

yes

crawr:
@JillieBeanSim This is not happening for me and I think the error that you are experiencing is already out of scope for this PR. But I do agree that this is a breaking issue and should be dealt immediately. So my solution is to create an issue, investigate and open a hot fix.

What do you think?

dkelosky:
I was also testing and could not reproduce the certificate errors mentioned here.  I did run into other problems attempting to test this:  

- It seems like residual user name and password can be loaded from the credential manager (even if `user` and `password` are removed from profile yaml files) which caused me confusing results
- I cannot seem to store `user` and `password` in a base profile and share with several z/OSMF profiles in Zowe Explorer, validation fails

I'll record those in another issue at some point.

crawr:
Issue #1094 opened for tracking. @lauren-li @JillieBeanSim kindly check if I captured the scenario correctly and please add more details if necessary. Thanks!

katelynienaber (dismissed):
A few things I noticed...

Larger issues, I think these should be fixed:
- The username/password boxes say `Optional`, but if I leave them blank I don't get logged in.
- I can't login. Service profile contains user/pass/host/port/basePath. Base profile contains host/port. I get error `Profile does not support login`.
- Related to last point: I still can't log in. I removed the password from the serviceProfile inside ZE, and refreshed the tree. Now it takes me through the login dialog. But in the end I get error `Unable to log in. Must have user & password OR base64 encoded credentials`.
- Unrelated issue: I logged in via CLI, then typed some random letters into the token, to simulate expiry. I tried to logon twice (with wrong password). Without refreshing workspace after trying to login twice, I logged out, then tried to log in again. This time I got ZE-specific error: `Unable to log in. Must have user & password OR base64 encoded credentials`.
- Related to last point: My ZE won't init. After logging out, I tried refreshing my workspace. I still get ZE-specific error: `Unable to log in. Must have user & password OR base64 encoded credentials`. And the profile doesn't show up in the sidebar.

Minor issues, could probably be ignored:
- If I add a base path to my profile later, I still see `Does not support login` message until I manually click the Refresh button.
- If my service profile contains the username/password, I am still prompted to enter it in the box

Maybe need new issue opened:
- When I try to edit my profile to contain base path, it isn't stored in base OR service profile. So I have to go to edit the yaml files directly to fix this.
- When I edit my profile to not contain password (ie. in yaml looks like `password: ""`) ZE won't activate. I get: `Error encountered while activating and initializing logger! Error: Must have user & password OR base64 encoded credentials`...maybe related to the init issues we were going to fix...somewhere idk I remember us talking about it on a call like a month ago...

jellypuno:
@zdmullen @zFernand0 I think my PR should be merged now so the testing for this PR will be smooth. what do you think?

jellypuno:
@katelynienaber There are related bugs that I've already fixed in PR #1063 . Unfortunately it is not merged yet so you still experienced them. 

Just a few points:

> The username/password boxes say Optional, but if I leave them blank I don't get logged in.

The function is re-using the existing prompting of username and password. maybe it would be better to re-factor that prompting and remove the word Optional. I will add that to issue #1000 

> I can't login. Service profile contains user/pass/host/port/basePath. Base profile contains host/port. I get error Profile does not support login.

Login function will not be called is user and pass exists in your service profile. This is the precedence that we have implemented. The Login function will be called is user and pass does not exist in your profile.

> If I add a base path to my profile later, I still see Does not support login message until I manually click the Refresh button.
> If my service profile contains the username/password, I am still prompted to enter it in the box

This is similar to the 2nd point. The precedence that we implemented is:
1. Service Profile that contains user and pass
2. Optional Profiles
3. Profiles with token

Other than these points, all issues that you have experienced should be handled in my PR.

katelynienaber:
If I add a base path to my profile later, I still see Does not support login message until I manually click the Refresh button.
If my service profile contains the username/password, I am still prompted to enter it in the box

This is similar to the 2nd point. The precedence that we implemented is:

Service Profile that contains user and pass
Optional Profiles
Profiles with token

I think these 2 can be ignored for now...I forgot this precedence thing.

zdmullen:
Jelly's PR is now merged. Update stuff, @katelynienaber and give this another try

jellypuno:
> Jelly's PR is now merged. Update stuff, @katelynienaber and give this another try

Thank you @zdmullen

katelynienaber:
@crawr Now I can login.

When I am already logged in, I can't edit the profile. I can go through the whole dialog (I am only trying to add the password) and when I get done, it says: `Error running command zowe.editSession: Cannot read property 'ISession' of undefined.`

Also logging in/out is now "not supported", after I tried to edit the profile (even though the yaml wasn't updated to include the password).

jellypuno:
@katelynienaber This out of scope. Unfortunately for base profiles / SSO in general we are not handling the maintenance of the profile. UX is still in discussion. The scope of this PR is only to allow login and logout. Any maintenance of a tokenized profile should be done through CLI. @zdmullen Can you please confirm this?

zdmullen:
@jellypuno, that is correct.

jellypuno (approved):
I have tested this and it is working for me. The errors that I am experiencing are already out of scope of this PR and I've opened github issues for those as well.

zdmullen (approved):
I've also gone through the original scope again and everything is fine for that. I feel that we've covered the other issues in subsequent issues, so we should be good to go. LGTM!

Issue #1063, opened 2020-11-09
Initialize Error when using base profiles


jellypuno:
**Describe the bug**

Initialize Error when using base Profiles

**To Reproduce**

1. Issue zowe auth logout apiml
2. Make sure that your session that uses base profile is loaded in ZE
3. Re-load VSCode

**Expected behavior**

This should be handled

**Screenshots**

![image](https://user-images.githubusercontent.com/45042840/98531618-e189e100-2280-11eb-9c14-d337049cf9a1.png)

**Desktop (please complete the following information):**

- OS: Windows
- Zowe Explorer Version: 10.1
- (Optional) Zowe CLI Version: latest
- (Optional) Are you using Secure Credential Store? Yes

**Additional context**

jellypuno:
Changes were merged into master

PR #1064, opened 2020-11-09
Fix: Initialize Error when using base profiles

jellypuno:
Signed-off-by: Jessielaine Punongbayan <jessielaine.punongbayan@broadcom.com>

## Proposed changes

This is a fix for issue #1063 - Initialize Error when using base profiles. When you are logged out in apiml and you want to use Zowe Explorer an initialization error would occur because tokenType and TokenValue are not found in your base profile.

## Release Notes

Milestone:

Changelog:

## Types of changes

## Checklist

- [ ] `npm run vscode:prepublish` has been executed

## Further comments


jellypuno:
There is an access error in the Theia build and I don't know where to find the other CI builds

packages/zowe-explorer/src/Profiles.ts
```diff
@@ -1079,11 +1079,25 @@ export class Profiles extends ProfilesCache {
...
+                cmdArgs.tokenValue = serviceProfile.profile.tokenValue
+                    ? serviceProfile.profile.tokenValue
+                    : baseProfile.profile.tokenValue;
             }
             if (commonApi.getSessionFromCommandArgument) {
-                session = await commonApi.getSessionFromCommandArgument(cmdArgs);
+                if (cmdArgs.tokenType === undefined || cmdArgs.tokenValue === undefined) {
+                    vscode.window.showInformationMessage(
+                        "Base Profile " +
+                            baseProfile.name +
```
JillieBeanSim, line 1093: since the check for tokenType and tokenValue has been updated for alternate profile types as well, this information message shouldn't be based on only base profiles but more general

packages/zowe-explorer/src/Profiles.ts
```diff
@@ -1079,11 +1079,25 @@ export class Profiles extends ProfilesCache {
...
+                            " is currently logged out. Please login and reload Zowe Explorer to be able to use it."
+                    );
+                    session = baseProfile.profile;
```
JillieBeanSim, line 1096: same here with only looking to base profile, there should be a check as to where the token value came from that that profile should be the session

JillieBeanSim (changes requested):

packages/zowe-explorer/src/Profiles.ts
jellypuno: @JillieBeanSim I'm wondering what would be the best way to make this more generic? "Your profile does not have a token. Please login and try again" 😄 

@zdmullen Do you have a suggestion for this?

jellypuno: Maybe I can change it to "Profile is not authorized. Please check the connection and try again"

jellypuno: I'm not really sure which profile name should I display in the error message. If I make it generic then it is confusing if I have 5 profiles in my tree. I need help here.

packages/zowe-explorer/src/Profiles.ts
jellypuno: So I tested this. I move the base profile because I needed the info if there is a base profile. If ever, they have the info in the service profile then it will be moved in line 1110.

packages/zowe-explorer/src/Profiles.ts
zdmullen: Hi. Your most recent question refers to the situation in which someone has multiple profiles that use a token, correct? But they only get the error message when they click on the profile, or is it at the activation of the extension? 

If I had to suggest an error message now, I would go with "Your profile no longer has a valid token. Please log in and try again."

jellypuno: This appears during initialization. And if I have multiple profiles in the tree, I'm not sure which one has no valid token. Unless I  just remove the error message here and let the initialization continue then there will still be an error message once the profile is used.

JillieBeanSim: I like that, with the error when the profile is used vs during initialization

jellypuno: ok so I'll just remove this error message and add it to the logs for record keeping.

packages/zowe-explorer/src/shared/utils.ts
```diff
@@ -85,19 +85,21 @@ export function labelRefresh(node: vscode.TreeItem): void {
...
-                sessNode.getSession().ISession.base64EncodedAuth = SessionProfile.base64EncodedAuth;
-                sessNode.getSession().ISession.hostname = SessionProfile.hostname;
-                sessNode.getSession().ISession.port = SessionProfile.port;
-                sessNode.getSession().ISession.rejectUnauthorized = SessionProfile.rejectUnauthorized;
+            const combinedSessionProfile = (await Profiles.getInstance().getCombinedProfile(profNode, baseProf))
+                .profile;
+            const sessionNode = sessNode.getSession();
+            for (const prop of Object.keys(combinedSessionProfile)) {
+                if (prop === "host") {
+                    sessionNode.ISession.hostname = combinedSessionProfile[prop];
```
zFernand0, line 99: This goes back to https://github.com/zowe/zowe-cli/issues/855#issuecomment-724778568
I really wish we didn't have to do this, but this is coming out of Imperative.
@dkelosky, thoughts on including this in `Imperative@next` and `CLI@next`?

dkelosky: That sounds good.  Do you want `hostname` gone entirely or both to exist, e.g. make `hostname` take precedence over `host` if it's there?

zFernand0: Personally, I don't mind either way, but we (in ZE) mostly use `host` since that's what profiles contain.

However, checkout this out:
CLI:
image

Imperative:
image

🤣

dkelosky: if we could be backwards compatible - I think that would be the preferred route to take

jellypuno: @dkelosky @zFernand0 The reason for this is because the APIs are using IProfileLoaded/IProfile format when a request is made. But when creating a session, the response that we get is in Session/ISession format. So we need to transfer the ISession to IProfile so that we could fire a request. This transfer is not a problem because they are the same except for the hostname.

packages/zowe-explorer/src/Profiles.ts
```diff
@@ -1079,11 +1079,30 @@ export class Profiles extends ProfilesCache {
...
+                    this.log.debug(
+                        localize("getCombinedProfile.noToken1", "Profile ") +
+                            baseProfile.name +
+                            " " +
+                            serviceProfile.name +
+                            localize(
+                                "getCombinedProfile.noToken2",
+                                " is not authorized. Please check the connection and try again"
+                            )
+                    );
```
lauren-li, line 1100: Just a note, you should be able to write the localization as:
```
this.log.debug(
    localize(
        "getCombinedProfile.noToken",
        "Profile {0} {1} is not authorized. Please check the connection and try again.",
         baseProfile.name,
         serviceProfile.name
    )
);
```

Keeping the message in a single localization string will make it easier for translators to understand what we are trying to say (in addition to looking neater).
- (Remember to re-run `yarn run vscode:prepublish` to update the localization strings if you decide to use this suggestion1)

jellypuno: Thanks for the review @lauren-li. I pushed new changes to apply what you said

JillieBeanSim (changes requested):
@jellypuno when I merged this PR and @crawr 's PR locally to test them I initially started testing with a base profile without the token fields mentioned in the description, but I never got a message about not found in your base profile, only the inactive profile message. 
Can see other issues I ran into while testing these 2 PRs together here https://github.com/zowe/vscode-extension-for-zowe/pull/1062#issuecomment-732498070

zFernand0:
After https://github.com/zowe/vscode-extension-for-zowe/pull/1093, all checks should be passing. :yum:
Also, this PR LGTM! 👍

zdmullen (approved):

zFernand0 (approved):
LGTM

zFernand0:
merging as this PR will technically be retested in 1062

Issue #1031, opened 2020-10-01
Ability to create a data set and edit the attributes


zdmullen:
**Is your feature request related to a problem? Please describe.**

I would like to be able to create a new data set and during the creation (allocation) process be able to edit the attributes as applicable. 

**Describe the solution you'd like**

![image](https://user-images.githubusercontent.com/38981509/94805761-2d797880-03ed-11eb-838c-b8ae6516cf2f.png)

**Describe alternatives you've considered**

I know that you can go into the settings.json, copy the attributes of a given data set, paste them into the settings.json and change them, but this is not the most straightforward approach (nor is it formally documented). 

**Additional context**

This is connected to what @katelynienaber has added in her draft PR #903 .

PR #1092, opened 2020-11-30
Issue #1031: Enable editing of DS attributes on allocation

katelynienaber:
## Proposed changes

This is an enhancement, linked to [issue #1031](https://github.com/zowe/vscode-extension-for-zowe/issues/1031).

It allows you to edit a data set's attributes before allocation, following the scheme presented in that issue and discussed with @jelaplan and @VitGottwald:

changeParamsDS

## Release Notes

Milestone: 2.0 (?)
Changelog: Enabled editing of DS attributes on allocation

## Types of changes

## Checklist


zdmullen:
I've just gone through this once and one major thing I noticed is that  "Directory Blocks" is in the list of attributes twice, and this pushes the "+Allocate Data Set" down so that I have to scroll to see it. What that issue got me thinking about is whether a user may just want to edit one attribute and then be able to "+ Allocate Data Set". Was this part of the original discussion, that we should allow them to allocate after editing just one attribute so they don't have to go back to the full list and select the "+ Allocate..." part?

katelynienaber:
> I've just gone through this once and one major thing I noticed is that "Directory Blocks" is in the list of attributes twice, and this pushes the "+Allocate Data Set" down so that I have to scroll to see it. What that issue got me thinking about is whether a user may just want to edit one attribute and then be able to "+ Allocate Data Set". Was this part of the original discussion, that we should allow them to allocate after editing just one attribute so they don't have to go back to the full list and select the "+ Allocate..." part?

So doing it this way would give 2 possible flows (I guess):
- User selects an item to edit...user edits...user hits enter to allocate
- User selects an item to edit...user edits...user arrows down to "Back" option, then hits enter to go back to full list

This would make allocation _more_ inconvenient for people editing many attributes. Because if you want to edit more attributes, you will have to arrow down and go "Back" for every option you edit.

Idk maybe John or @jellypuno or @VitGottwald can put in here...does anyone know if people most frequently edit one attribute or more than one?

_Also I fixed that double item in the list @zdmullen, and changed the item order in the big list. Now if a user enters an attribute value and hits Enter, they can hit Enter one more time and it will allocate automatically._

JillieBeanSim (changes requested):
@katelynienaber this is really cool, I do have a few questions about what is presented to the users in the parameters list:
1. During create should we offer the option to change the node label since the data set name was just entered before hand?
2. If we do decide to keep the option to rename during this, would we want to use more familiar labels like `Data Set Name` instead of `Node Type`?
3. Another thing with labels, `Node Type` and `DSN type` are presented in the list where `Node Type` is holding the value, but shouldn't `DSN Type` hold the value and possibly have the long name `Data Set Type`,  and `Node Type` not be in the list as it isn't a traditional label when allocating a data set

Functionally this works well, just requesting a few polish items to follow traditional labels for the attributes.

One other thing I notice when running integration tests is that the tests seem to hang on the Creating Data Sets and Members tests (tried multiple times with same result).  I wonder if this hang is caused due to changes in the UI for this new functionality.  I know these tests are inconsistent across users that run them, so if someone else could run them and not get the hang I experienced it could just be me. 
image

katelynienaber:
> 1. During create should we offer the option to change the node label since the data set name was just entered before hand?

I would still leave it, in case people make a typo or something

> 2. If we do decide to keep the option to rename during this, would we want to use more familiar labels like `Data Set Name` instead of `Node Label`?

I did change this :) It makes sense.

> 3. Another thing with labels, `Node Type` and `DSN type` are presented in the list where `Node Type` is holding the value, but shouldn't `DSN Type` hold the value and possibly have the long name `Data Set Type`,  and `Node Type` not be in the list as it isn't a traditional label when allocating a data set

I am getting these attributes from the object `zowe.ICreateDataSetOptions`, so I would like to include both. But I made some changes to the names to add clarity, let me know if this is adequate:
image

> One other thing I notice when running integration tests is that the tests seem to hang on the Creating Data Sets and Members tests (tried multiple times with same result).

This was an issue related to the new code...they are running okay for me now, please confirm!

packages/zowe-explorer/__tests__/__integration__/extension.integration.test.ts
```diff
@@ -152,18 +156,28 @@ describe("Extension Integration Tests", () => {
...
-            quickPickStub.returns("Data Set Sequential");
-
+            // 1st step: User names DS
             const testFileName = pattern + ".EXT.CREATE.DATASET.TEST";
             const inputBoxStub = sandbox.stub(vscode.window, "showInputBox");
             inputBoxStub.returns(testFileName);
 
+            // 2nd step: User selects DS type
+            const quickPickStub = sandbox.stub(vscode.window, "showQuickPick");
+            quickPickStub.onCall(0).returns("Data Set Sequential");
```
jellypuno, line 166: If possible can you change this to a foreach statement? The last PR that I opened has magic numbers error. I can see that yours doesn't have one but it would be great if we could prevent it in the future. I did it like this:

```
const returnVals = ["testUser", "testPass", "basePath", "encoding", "timeOut"];
        returnVals.forEach((value, index) => {
            showInputStub.onCall(index).returns(value);
        });
```

packages/zowe-explorer/src/dataset/actions.ts
```diff
@@ -297,31 +298,38 @@ export async function openPS(
...
             createOptions = vscode.workspace.getConfiguration("Zowe-Default-Datasets-Classic");
             break;
         case localize("createFile.dataSetPartitioned", "Data Set Partitioned"):
             typeEnum = zowe.CreateDataSetTypeEnum.DATA_SET_PARTITIONED;
+            cliDefaultsKey = "PARTITIONED";
             createOptions = vscode.workspace.getConfiguration("Zowe-Default-Datasets-PDS");
             break;
         case localize("createFile.dataSetSequential", "Data Set Sequential"):
             typeEnum = zowe.CreateDataSetTypeEnum.DATA_SET_SEQUENTIAL;
+            cliDefaultsKey = "SEQUENTIAL";
```
jellypuno, line 326: Do you think it is good to put this in globals? Maybe we can re-use it in future enhancements?

packages/zowe-explorer/src/globals.ts
```diff
@@ -54,6 +54,116 @@ export const ICON_STATE_OPEN = "open";
 export const ICON_STATE_CLOSED = "closed";
 export const VSCODE_APPNAME: string[] = ["Visual Studio Code", "VSCodium"];
 export const ROOTPATH = path.join(__dirname, "..", "..");
+export const DATA_SET_PROPERTIES = [
```
jellypuno, line 57: Can you import the descriptions here: C:\Users\jp669971\Documents\GitHub\Community\vscode-extension-for-zowe\node_modules\@zowe\cli\lib\zosfiles\-strings-\en.d.ts

jellypuno: And also, I think it needs to be localize for webpack

katelynienaber: How do I do this?

jellypuno: Maybe we should talk to CLI team about it but they did it in this way : https://github.com/zowe/zowe-cli/blob/a9b40537e432935fda7b03e27467bb854cf3db2e/packages/cli/src/zosfiles/hMigrate/HMigrate.definition.ts#L18

jellypuno: Or maybe @lauren-li and @JillieBeanSim can give us some hints on how to localize these properties

packages/zowe-explorer/src/dataset/actions.ts
katelynienaber: I am not sure if this is useful...the `cliDefaultsKey` is just the `CreateDataSetTypeEnum`, but without `DATA_SET_`

ie. if the `CreateDataSetTypeEnum` is `DATA_SET_BINARY`, then the `cliDefaultsKey` is `BINARY`...

So it feels a little pointless to me to make a whole new enum just to remove the `DATA_SET_` part of `CreateDataSetTypeEnum`

The CLI is irritating me a bit with all their different names for the same things. `CreateDataSetTypeEnum` should just match the keys used in the CLI to identify the default settings (keys used in `zowe.CreateDefaults` file). We are having the same issue with sending profile values to Imperative as CLI commands. I think they need to standardize their naming procedures, and we shouldn't be making a bunch of different data structures that only have slight differences in their values to accommodate all these slightly different names across CLI and Imperative.

Rant over. Maybe I will bring this up in standup.

jellypuno: ok. I understand. Do you think trim would be useful since you know that you want to remove DATA_SET_? like this:

const session = node.label.trim().substring(node.label.trim().indexOf("[") + 1, node.label.trim().indexOf("]"));

jellypuno: There's an example in C:\Users\jp669971\Documents\GitHub\Community\vscode-extension-for-zowe\packages\zowe-explorer\src\dataset\actions.ts. Line 989

packages/zowe-explorer/src/globals.ts
katelynienaber: I have done the localization but I am not sure how to import them into that file...so yeah it would be cool to ask someone about how to use that file :)

packages/zowe-explorer/src/dataset/actions.ts
katelynienaber: The problem is that I can't get the key from the enum. Ex. I don't have DATA_SET_SEQUENTIAL as a string, I have just the number: 4.

If you look here, since it is a `const enum`, the key is removed completely on compilation...so it doesn't even exist there for me to get ahold of: [link to docs](https://www.typescriptlang.org/docs/handbook/enums.html#const-enums)

Maybe it makes sense to add it as an enum for this reason, but I still think it's a bit overkill P:

katelynienaber: I could add an enum like this:
```
CreateDataSetTypeKeysEnum {
    0 = DATA_SET_BINARY,
    1 = DATA_SET_C,
    2 = DATA_SET_CLASSIC,
    3 = DATA_SET_PARTITIONED,
    4 = DATA_SET_SEQUENTIAL
}

```
And that would at least be more useful, in that we can take the `CreateDataSetTypeEnum` and reverse it to get the keys when we need them...but it is sort of a hack to get around their `const enum` definition, and it makes me wonder: why did they define it as `const enum` in the first place? Maybe they can remove it? Maybe we shouldn't go around it?

jellypuno: I think that "hack" is the approach that I'm gonna go for. But only if you find it very useful. 

> why did they define it as const enum in the first place? 
That's a good question. We can open an enhancement issue for this. I pretty sure they didn't expect anybody to get the description of the dataset. So this might be the first time.

jellypuno:
For the newly created datasets, make sure it is deleted in the posttest:integration script

katelynienaber:
> For the newly created datasets, make sure it is deleted in the posttest:integration script

Done

jellypuno:
This is not an issue in this PR but make sure that the dataset name is not too long. It has a limit of 44 chars only. For more reference: https://www.ibm.com/support/knowledgecenter/zosbasics/com.ibm.zos.zjcl/zjclt_dstypesNnames.htm

katelynienaber:
> This is not an issue in this PR but make sure that the dataset name is not too long. It has a limit of 44 chars only. For more reference: https://www.ibm.com/support/knowledgecenter/zosbasics/com.ibm.zos.zjcl/zjclt_dstypesNnames.htm

I am working on this in another branch...see ds-attribute-validation...basically over there the framework for adding better validation for any of the attributes is set up. [Here's a new discussion thread](https://github.com/zowe/vscode-extension-for-zowe/discussions/1141) where you can add anything else you think of (but I can add this suggestion you just added to that branch, you don't have to post again)

jellypuno:
I'm having issues with integration test with the latest commit. It was stopping here:

```
            newDSProperties.forEach((property) => {
                Object.keys(propertiesFromDsType).forEach((typeProperty) => {
                    if (typeProperty === property.key) {
                        property.value = propertiesFromDsType[typeProperty].toString();
                        property.placeHolder = propertiesFromDsType[typeProperty];
                    }
                });
            });
```

image

packages/zowe-explorer/src/globals.ts
jellypuno: The first part is right, I would like to import that file and use it. And I'm wondering if we could localize it without re-writing it in our codebase or can we use their own localization? (Maybe not 😢 )

lauren-li: @jellypuno Assuming you mean that we should import descriptions from `@zowe\cli\lib\zosfiles-strings-\en.js` and display them localized in Zowe Explorer (as @JillieBeanSim described), I took a look at [this file](https://github.com/zowe/zowe-cli/blob/a9b40537e432935fda7b03e27467bb854cf3db2e/packages/cli/src/zosfiles/-strings-/en.ts ) in the Zowe CLI repo, and it seems we will need to do some JSON processing, especially since the keys there don't match the ones we need.

If we process the JSON and store their values in variables in Zowe Explorer, then I believe we should just be able to localize the values of those variables per our usual localization method.

JillieBeanSim: so we would need to create a schema that could be used to process the JSON and could set a way to localize all Description fields?

lauren-li: That might be one way to do it, although perhaps in a separate PR.

jellypuno: I agree. Thank you for your guidance Lauren and Billie. I appreciate it 😊

@katelynienaber Just use the file for now and then the localization will be handled on a different PR. 😉

lauren-li: I just had the thought that if we are using the imported strings file from `@zowe`, we'll be able to pull in the latest updated strings, but that means we probably won't be able to localize the strings ourselves. 🤔 But perhaps we could set some variable to update the path to the import based on the user's VS Code language setting, and use whatever localization is provided by `@zowe` (although currently, this only appears to be English).

katelynienaber:
@jellypuno The integration tests should be fixed, and localization has been added for the texts in the list.

@JillieBeanSim I was thinking about adding the MF parameter names to all the items in the dropdown list, to help people search...do you think this is overkill?

Ie. all the items in the list would have the MF param name next to the text in parenthesis:
image

jellypuno:
I created a Sequential Dataset with the default values and I encountered this error:

image

katelynienaber:
> I created a Sequential Dataset with the default values and I encountered this error:
> 
> image
> 
> I think there's a conflict with the DIRBLK. It should be 0 instead of 5
> 
> I also notice that this happened after I created a partition dataset. Somehow the DIRBLK is transferred to my next allocation

Hey @jellypuno I am not getting this error. Could you tell me how you are going about making your DS?

I tried it by selecting Sequential Data Set in the list & allocating directly after, and also I tried going to the edit params list and allocating from there. It is getting the defaults from this file:
image
And there is no default DIRBLK value set for Sequential DS in that file. I can confirm no setting is being passed:
image

JillieBeanSim:
instead of timing out I am now getting this error for those same integration tests
image

JillieBeanSim:
when I try to change the spacu/alcunit to BLOCK I get these errors but also get a success message even though the data set isn't created.  I tried the value `blk`, `BLK`, `BLOCKS`, and `blks`, according to what I see in code it should be `blk` but still get the errors
image

katelynienaber:
> instead of timing out I am now getting this error for those same integration tests
> image

I think that's old, it's because I merged in master and it had a wrong variable in there. It should be fixed

JillieBeanSim:
I get strange login errors when refreshing the Data Set View
image

I can't change the spacu/alcunit to `trk` either with this error
image

but when using `TRK` the data set gets created but I get this error popping up about `name not defined`.  User input should be case insensitive and not sure what is causing the error about name undefined and with the data set getting created I didn't get the info pop up about it being created
image

jellypuno:
@katelynienaber This is how I did it:

dsAttributes

katelynienaber:
> @katelynienaber This is how I did it:
> 
> dsAttributes

@jellypuno @JillieBeanSim Please pull again to prevent this error, it was caused by merging with master. I am checking the other issues...

katelynienaber:
> when I try to change the spacu/alcunit to BLOCK I get these errors but also get a success message even though the data set isn't created. I tried the value `blk`, `BLK`, `BLOCKS`, and `blks`, according to what I see in code it should be `blk` but still get the errors
> image

Allocating new data set isn't a success message, it's more of an "attempting to" message. I think we should add a progress spinner here but would prefer to do so in another PR. We could change the message to something more clear...maybe something like "Attempting to allocate data set..." ?

Also in the `zowe.Create` function, [you can't pass BLOCK as an alcunit value](https://github.com/zowe/zowe-cli/blob/a9b40537e432935fda7b03e27467bb854cf3db2e/packages/zosfiles/src/methods/create/Create.ts#L178)...I don't know why XD Asking @jellypuno now...

katelynienaber:
> I get strange login errors when refreshing the Data Set View, these only happen if the creation fails and I refresh. If the creation was successful and I refresh view I don't see the error

I'm not getting this...I think maybe this is coming from that name error...please pull again XD

> I can't change the spacu/alcunit to `trk` either with this error

Sorry not getting this either >.<

> but when using `TRK` the data set gets created but I get this error popping up about `name not defined`. User input should be case insensitive and not sure what is causing the error about name undefined and with the data set getting created I didn't get the info pop up about it being created

This is still that error caused by merging with master, it should be fixed already.

crawr (commented):
I was testing this in Theia and I found an issue when allocating a data sets and trying to edit an attribute.

create classic

As you can see in the gif, I received an error but the data set was allocated successfully.


It is not a showstopper for me since the existing allocation still works, I just cannot edit the attributes. Perhaps the Theia specific issue can be investigated and handled in a separate PR.

katelynienaber:
@crawr I have opened this in an issue here: #1167

jellypuno:
@katelynienaber I tried to allocate a binary using default values

allocerr

katelynienaber:
@jellypuno Looks like I missed a `name` after that merge :V

Try again please. Maybe this will fix tests as well XD

jellypuno:
@katelynienaber Did you add a fix for the sequential dataset as well? I'm still having it even though I cloned a new copy again.

jellypuno:
The "allocating new dataset" message, is this just an informational message or is it a withProgress message? I'm expecting that it will be removed after the dataset was allocated.

katelynienaber:
> @katelynienaber Did you add a fix for the sequential dataset as well? I'm still having it even though I cloned a new copy again.

I'm not getting that error, and I thinkk I am doing the same thing as you to allocate...
errorGif

katelynienaber:
> The "allocating new dataset" message, is this just an informational message or is it a withProgress message? I'm expecting that it will be removed after the dataset was allocated.

It's an info message, like I said to Billie, I think it should be withProgress but I would add that later and instead change the message now if it's not clear enough

jellypuno:
> > The "allocating new dataset" message, is this just an informational message or is it a withProgress message? I'm expecting that it will be removed after the dataset was allocated.
> 
> It's an info message, like I said to Billie, I think it should be withProgress but I would add that later and instead change the message now if it's not clear enough

I would suggest to just remove it. Because there is a Get list progress that is displayed as well. So having that allocate can be removed for now

katelynienaber:
@jellypuno OK I removed it!

jellypuno:
Error message is too vague. I executed a file creation CLI and did the same thing in Zowe Explorer. The message in the CLI is better. If we can display that instead of what we have now that would be helpful.

image

jellypuno:
I am not able to allocate a dataset in ZE but I can do it in CLI. I have the same attributes:

image

jellypuno:
Also, Show Attributes is a CLI tag to display the attributes in the Command Line after the allocation. We don't need it and you can remove it in dialog box.

katelynienaber:
> Error message is too vague. I executed a file creation CLI and did the same thing in Zowe Explorer. The message in the CLI is better. If we can display that instead of what we have now that would be helpful.

CLI doesn't send us any more info. @jellypuno is talking to @t1m0thyj about this.

> I am not able to allocate a dataset in ZE but I can do it in CLI. I have the same attributes:

Fixed!

> Also, Show Attributes is a CLI tag to display the attributes in the Command Line after the allocation. We don't need it and you can remove it in dialog box.

I have removed this field as well.

jellypuno:
@katelynienaber thank you! don't forget to push your changes 😉

katelynienaber:
@jellypuno should be there now! Was just running tests

jellypuno (approved):
I think the overall functionality works for me and it looks great. 

However, I see 2 issues under this PR.
1. To get a more descriptive error message. This can be opened under CLI team.
2. To have a condition that will not display the EDIT Attributes in Theia environment.

We can open new issues for these and merge this PR in master today

JillieBeanSim:
Functionality looks good to me and all unit tests are passing.

> For the newly created datasets, make sure it is deleted in the posttest:integration script

not sure it this is getting deleted else where...

@katelynienaber @jellypuno I see an error here running posttest and then pretest in this PR where posttest gives this
image
which stops the script so on pretest this is the error dealing with USS since the posttest stopped and didn't continue with USS actions
image

I end up having to go in and delete USS dir manually then run pretest to have a clean testing environment for integration tests

I have this failing integration test that I haven't seen yet while testing various branches
image

katelynienaber:
@JillieBeanSim I fixed the integration test, and yes those DS are deleted after each run, so I removed the line deleting them in the posttest.

JillieBeanSim (approved):

JillieBeanSim:
Thanks for all of your hard work on this @katelynienaber! all expected tests and checks are passing now and the scripts are working as they should be.  Approved!

jellypuno (approved):

Issue #1065, opened 2020-11-10
handle new schema fields for zosmf profile


jellypuno:
**Describe the bug**

In the latest CLI, there 2 new fields added in the profile schema. We are not currently handling it so if I choose to cancel the profile creation on those fields I encounter this error:

![image](https://user-images.githubusercontent.com/45042840/98732153-0ffdde00-239f-11eb-87a8-8bb9bd4ff19c.png)

**To Reproduce**

**Expected behavior**

Handle the error

**Screenshots**

**Desktop (please complete the following information):**

- OS:
- Zowe Explorer Version:
- (Optional) Zowe CLI Version:
- (Optional) Are you using Secure Credential Store?

**Additional context**

PR #1100, opened 2020-12-08
Quick Fix: Small issues in profiles

jellypuno:
## Proposed changes

These are small changes and are not high priority.

Changes Include:
- Remove the word Optional in Username and Password
- Display the error message when Profile Inactive is encountered
- Handle the additional fields when creating and editing a profile when the user cancelled it.

## Release Notes

Milestone:

Changelog:

## Types of changes

## Checklist

## Further comments

If the validation is on and the profile that you are using doesn't have a username and password or token, you will encounter this error:
image

I have opened an issue in CLI for this:
https://github.com/zowe/zowe-cli/issues/875


JillieBeanSim (commented):
@jellypuno this is looking good but I did get this error when updating a profile and going all the way through the update process.  It fully updated the profile, yaml included but got the error on top of successful message
image

jellypuno:
> @jellypuno this is looking good but I did get this error when updating a profile and going all the way through the update process. It fully updated the profile, yaml included but got the error on top of successful message
> image

Thank you for testing @JillieBeanSim . I will take a look at this tomorrow.

packages/zowe-explorer/src/Profiles.ts
```diff
@@ -482,10 +482,24 @@ export class Profiles extends ProfilesCache {
                             if (!Number.isNaN(Number(updValue))) {
                                 updSchemaValues[value] = Number(updValue);
```
zFernand0, line 483: If input is an empty string (i.e. the user hit enter), `Number("") === 0`
This results in the profile still being saved with `responseTimeout: 0`.

We could check for `updValue !== "" && !Number.isNaN(Number(updValue)` but that leads to another problem. (see next comment)

This comments applies to both Creating a new Connection (`createNewConnection`) and Editing a session (`editSession`)

packages/zowe-explorer/src/Profiles.ts
```diff
@@ -482,10 +482,24 @@ export class Profiles extends ProfilesCache {
...
-                                } else {
-                                    break;
+                                switch (true) {
+                                    case schema[value].optionDefinition.hasOwnProperty("defaultValue"):
+                                        if (schema[value].optionDefinition.defaultValue === null) {
+                                            vscode.window.showInformationMessage(
+                                                localize("editConnection.rejectUnauthorize", "Operation Cancelled")
+                                            );
+                                            return undefined;
+                                        }
```
zFernand0, line 492: If the user didn't enter a value for a numeric property, and the schema contains a null value, we should still create the profile without this property.

For comparison, here are the schema definitions encoding and responseTimeout
```
{
  "name": "encoding",
  "aliases": ["ec"], 
  "description": "The encoding for download and upload of z/OS data set and USS files. The default encoding if not specified is 1047.",
  "type":"number"
}
```

```
{
  "name": "responseTimeout",
  "aliases": ["rto"],
  "description": "The maximum amount of time in seconds the z/OSMF Files TSO servlet should run before returning a response. Any request exceeding this amount of time will be terminated and return an error. Allowed values: 5 - 600",
  "type": "number",
  "defaultValue": null,
  "numericValueRange": [5,600]
}
```

zFernand0 (commented):
Added a couple of comments 😋

jellypuno:
@zFernand0 Just a clarification, are you saying that if the user enters a blank in the property (e.g. responseTimeOut) we should create the profile with a blank value instead of zero? or are saying that we should not add the property into the profile?

jellypuno:
Since this is a generic checking of number, If I try to allow the creation of profile that is equal to null, we are having this error.

image

In my opinion, if it is defined as a number, the default number should not be null. I opened an issue for that: https://github.com/zowe/zowe-cli/issues/871

zFernand0:
> are you saying that if the user enters a blank in the property (e.g. responseTimeOut) we should create the profile with a blank value instead of zero? or are saying that we should not add the property into the profile?

The latter. We should not store properties that: 
  - Do not have a default value set by the schema
  - The user didn't specify a value

jellypuno:
@zFernand0 I plan to add this in the future. The goal of this PR is just to handle those additional fields. Because right now, if I cancelled the creation / editing with those additional fields it will display an error and not "Operation Cancelled"

katelynienaber:
@jellypuno 2 things:

1: Does this fix [this issue (#1065)](https://github.com/zowe/vscode-extension-for-zowe/issues/1065)?
2: I think it might be better to store empty strings than totally remove the fields. The reason is that Imperative is using `deepmerge` to merge old and new profiles when editing them, and if the new profile doesn't have all the same fields as the old one, `deepmerge` retains the old fields (and their values) when merging. See [this PR](https://github.com/zowe/vscode-extension-for-zowe/pull/1120) for some affected fields...maybe there is a better way to deal with this, I am not sure.

jellypuno:
> @jellypuno 2 things:
> 
> 1: Does this fix [this issue (#1065)](https://github.com/zowe/vscode-extension-for-zowe/issues/1065)?
> 2: I think it might be better to store empty strings than totally remove the fields. The reason is that Imperative is using `deepmerge` library to merge a profile's old and new values when updating the profile. If the new profile doesn't have all the same fields as the old one, `deepmerge` retains the old fields (and their values) when merging. See [this PR](https://github.com/zowe/vscode-extension-for-zowe/pull/1120) for some fields affected by `deepmerge`. Maybe there is a better way to deal with this than storing empty strings...what do you think?

I don't think I'm removing it. I'm storing it as spaces or default value. The handling that I added is that if you cancelled the Profile Creation on those new fields the error message is better. But maybe I need to double check

katelynienaber:
> > @jellypuno 2 things:
> > 1: Does this fix [this issue (#1065)](https://github.com/zowe/vscode-extension-for-zowe/issues/1065)?
> > 2: I think it might be better to store empty strings than totally remove the fields. The reason is that Imperative is using `deepmerge` library to merge a profile's old and new values when updating the profile. If the new profile doesn't have all the same fields as the old one, `deepmerge` retains the old fields (and their values) when merging. See [this PR](https://github.com/zowe/vscode-extension-for-zowe/pull/1120) for some fields affected by `deepmerge`. Maybe there is a better way to deal with this than storing empty strings...what do you think?
> 
> I don't think I'm removing it. I'm storing it as spaces or default value. The handling that I added is that if you cancelled the Profile Creation on those new fields the error message is better. But maybe I need to double check

I was referring to your conversation with @zFernand0 ...so I will tag him in that comment :)

katelynienaber (changes requested):
Looks good and works as expected. The basePath is still not updated when the profile is edited...but I think that's out of scope and is already fixed in [another PR](https://github.com/zowe/vscode-extension-for-zowe/pull/1120).

IMHO though we should maybe leave the `optional` placeholder when creating/editing a profile, but remove it when prompted for username/password.

jellypuno:
> Looks good and works as expected. The basePath is still not updated when the profile is edited...but I think that's out of scope and is already fixed in [another PR](https://github.com/zowe/vscode-extension-for-zowe/pull/1120).
> 
> IMHO though we should maybe leave the `optional` placeholder when creating/editing a profile, but remove it when prompted for username/password.

I removed the optional word because it is used by ssoLogin and it is not supposed to optional.

katelynienaber:
> > Looks good and works as expected. The basePath is still not updated when the profile is edited...but I think that's out of scope and is already fixed in [another PR](https://github.com/zowe/vscode-extension-for-zowe/pull/1120).
> > IMHO though we should maybe leave the `optional` placeholder when creating/editing a profile, but remove it when prompted for username/password.
> 
> I removed the optional word because it is used by ssoLogin and it is not supposed to optional. It didn't make sense to create another text just for ssoLogin so I just removed the optional because we intend to remove the Optional Credentials in the future

Ah okay, SO GLAD we will remove optional credentials! I didn't know that was the plan so I will approve now.

katelynienaber (approved):
LGTM!

JillieBeanSim (commented):
Hey @jellypuno when I was creating and editing a zosmf profile, I hit the escape at the encoding value.  I was expecting the `operation cancelled` message to pop up then but instead it went to the next textbox for response timeout then I clicked escape again to get the `operation cancelled` message.
Is this expected, as I was not expecting to get prompted with another textbox after hitting escape the first time?

jellypuno:
> Hey @jellypuno when I was creating and editing a zosmf profile, I hit the escape at the encoding value. I was expecting the `operation cancelled` message to pop up then but instead it went to the next textbox for response timeout then I clicked escape again to get the `operation cancelled` message.
> Is this expected, as I was not expecting to get prompted with another textbox after hitting escape the first time?

Did you get the latest commit? It is not happening for me.

operCancel

JillieBeanSim (dismissed):
@jellypuno I must have been missing a commit.  This LGTM, functionality is working as expected but I do get a few failing integration tests that deal with creating a profile now that I didn't get on Master branch. 

image

jellypuno:
@JillieBeanSim I fixed the integration test. Thank you for testing and reviewing this PR.

katelynienaber (approved):
LGTM :)

JillieBeanSim (approved):
This LGTM, all expected tests and checks are passing and functionality works great! Thanks for all of your work on this @jellypuno, fixing these small issues make a huge positive impact for users.

Issue #1095, opened 2020-12-02
Enhance Optional Profiles


jellypuno:
Currently, Optional Profiles are creating a profile with User and Pass as spaces. And because of this feature, it creates a lot of inconsistencies and contradictions with the recent Zowe SDK functionalities. I would like to enhance it and keep it consistent with the SDK so we can have a better profile experience.

katelynienaber:
Link to [issue #1000](https://github.com/zowe/vscode-extension-for-zowe/issues/1000)

jellypuno:
Closing issue. Fix is merged in v1.12.x

Issue #1090, opened 2020-11-25
Use the --dd Disable defaults parameter when creating a profile


zdmullen:
**Is your feature request related to a problem? Please describe.**

This a question/request. Should we pass the -dd (disable defaults) parameter from CLI when a user creates a new profile in Zowe Explorer? This would go along with the ability to have all parameters optional and support the use case with base profiles (that I need to create a zosmf or service profile that only has the base-path for the mediation layer. 

**Describe the solution you'd like**

A discussion of whether we want to pass this parameter and then an implementation (or not).

This in part is linked to our issues connected to base profiles (such as #1073 or #1044).

jellypuno:
@katelynienaber Is it okay to close this issue?

katelynienaber:
@jellypuno yes I think so.

I am not sure if #1044 can be closed. I found [this additional issue](https://github.com/zowe/vscode-extension-for-zowe/issues/1203) which is somewhat related...

jellypuno:
Closing this issue. Fix is available in v1.12.1

Issue #1038, opened 2020-10-09
Update our codebase to call createBasicZosmfSessionFromArguments


jellypuno:
Change our Profiles codebase to  call createBasicZosmfSessionFromArguments. 

This can be handled in issue #1000

jellypuno:
The issue that we are having is that we are not allowed to create a session with incomplete information. Maybe if we update it then it would not encounter any errors.

jellypuno:
Note to self: createSessCfgFromArgs is returning ISession rather than Session. CLI team suggested to use addPropsOrPrompt instead so that a Session will be returned

JillieBeanSim:
@zFernand0 I believe this is done with the next branch.  Is that correct? if so we can close this.

Issue #1016, opened 2020-09-15
Default responseTimeout=0 Causes Zowe CLI Error


dkelosky:
**Is your feature request related to a problem? Please describe.**

If I create a profile in Zowe Explorer and bypass this prompt by hitting enter:
![image](https://user-images.githubusercontent.com/14940561/93247397-3ebf5580-f75c-11ea-854d-4eb5467bcd74.png)

My profile is created  with:
`responseTimeout: 0`

Then if I run a command like `zowe files download ds "some.data.set(member)" --zosmf-p ze-profile` where `ze-profile` is a profile I created with Zowe Explorer, I get the following error on the CLI:
```
Invalid numeric value specified for option:
--responseTimeout

You specified:
0

Value must be between 5 and 600 (inclusive)
```

**Describe the solution you'd like**

Perhaps Zowe Explorer should create the profile with a default value.  

**Describe alternatives you've considered**

**Additional context**

zFernand0:
> Perhaps Zowe Explorer should create the profile with a default value.

or better yet, ZE could avoid adding optional fields to the profile if the user didn't specify a value for it. 😋

dastagg:
Just so you know:

I can confirm that I have seen (a few) Profiles created through the Zowe Explorer Interface that did have:
responseTimeout: 0
as a parameter in their Profile.

But today( Oct 1, 2020) , using Zowe Explorer 1.9.0:
I created a profile.
It did ask for:
The base path for your API mediation layer instance
The encoding for download and upload of files

But it did not ask for:
The maximum amount of time.

By not entering any information and hitting "Enter" for each of the two last options it created this profile:
```
cat Test3.yaml 
host: 192.168.2.5
port: 10443
user: Test3
password: test3
rejectUnauthorized: false
encoding: 0
```
By not entering any information and hitting "Esc" for each of the two last options it created this profile:
```
$ cat Test4.yaml 
host: 192.168.2.6
port: 10043
user: Test4
password: test4
rejectUnauthorized: false
```

(yeah, I don't have zowe secure installed on this computer.)

jellypuno:
Closing issue. Fix is merged in v1.12.x

PR #1120, opened 2021-01-04
Multiple issues: Fixed errors with updating optional profile fields

katelynienaber:
## Proposed changes

This is one change that fixes several issues.

1. **Fix for issue #989:** There has been a bugfix! 😃 Which means in Zowe Explorer we can now pass an `IUpdateProfile` object to `CliProfileManager.update()`...and we don't have to change the profile object key names to resemble CLI command params. This was the cause of #989.

2. **Fix for issue #1095:** I set `merge: false` in the profile, so only the new values will be kept, not the old (ie. now we don't have to store an empty string as a placeholder for an optional field).

3. **Fix for issue #1016:** There are two new conditions in our `updateProfile()` function, which will remove the `responseTimeout` field if it was set to 0, and any other field if it was set to an empty string. And because of fix 2 (above), #1016 should not reoccur.

Linked issues (bugs):
[#989](https://github.com/zowe/vscode-extension-for-zowe/issues/989)
[#1016](https://github.com/zowe/vscode-extension-for-zowe/issues/1016)
[#1095](https://github.com/zowe/vscode-extension-for-zowe/issues/1095)

Other linked issues (to-dos that are no longer needed):
[#1038](https://github.com/zowe/vscode-extension-for-zowe/issues/1038)
[#1090](https://github.com/zowe/vscode-extension-for-zowe/issues/1090)

## Release Notes

Milestone: 1.12
Changelog: Fixed errors with updating optional profile fields

## Types of changes

## Checklist

### DIFF
--- a/packages/zowe-explorer/src/abstract/ZoweTreeProvider.ts
+++ b/packages/zowe-explorer/src/abstract/ZoweTreeProvider.ts
@@ -193,10 +193,8 @@ export class ZoweTreeProvider {
                 node.contextValue.toLowerCase().includes("session") ||
                 node.contextValue.toLowerCase().includes("server")
             ) {
-                // change contextValue only if the word inactive is not there
-                if (node.contextValue.toLowerCase().indexOf("inactive") === -1) {
-                    node.contextValue = node.contextValue + globals.INACTIVE_CONTEXT;
-                }
+                node.contextValue = node.contextValue.replace(/(?<=.*)(_Active|_Inactive|_Unverified)$/, "");
+                node.contextValue = node.contextValue + globals.INACTIVE_CONTEXT;
                 const inactiveIcon = getIconById(IconId.sessionInactive);
                 if (inactiveIcon) {
                     node.iconPath = inactiveIcon.path;
@@ -224,10 +222,8 @@ export class ZoweTreeProvider {
                 node.contextValue.toLowerCase().includes("session") ||
                 node.contextValue.toLowerCase().includes("server")
             ) {
-                // change contextValue only if the word active is not there
-                if (node.contextValue.toLowerCase().indexOf("active") === -1) {
-                    node.contextValue = node.contextValue + globals.ACTIVE_CONTEXT;
-                }
+                node.contextValue = node.contextValue.replace(/(?<=.*)(_Active|_Inactive|_Unverified)$/, "");
+                node.contextValue = node.contextValue + globals.ACTIVE_CONTEXT;
                 const activeIcon = getIconById(IconId.sessionActive);
                 if (activeIcon) {
                     node.iconPath = activeIcon.path;
@@ -238,10 +234,8 @@ export class ZoweTreeProvider {
                 node.contextValue.toLowerCase().includes("session") ||
                 node.contextValue.toLowerCase().includes("server")
             ) {
-                // change contextValue only if the word unverified is not there
-                if (node.contextValue.toLowerCase().indexOf("unverified") === -1) {
-                    node.contextValue = node.contextValue + globals.UNVERIFIED_CONTEXT;
-                }
+                node.contextValue = node.contextValue.replace(/(?<=.*)(_Active|_Inactive|_Unverified)$/, "");
+                node.contextValue = node.contextValue + globals.UNVERIFIED_CONTEXT;
             }
         }
         await this.refresh();
--- a/packages/zowe-explorer/src/abstract/ZoweTreeProvider.ts
+++ b/packages/zowe-explorer/src/abstract/ZoweTreeProvider.ts
@@ -177,9 +177,6 @@ export class ZoweTreeProvider {
         try {
             // Remove the edited profile from profilesForValidation since it should be revalidated
             Profiles.getInstance().profilesForValidation.forEach((checkProfile, index) => {
-                if (index === 0) {
-                    Profiles.getInstance().profilesForValidation = [];
-                }
                 if (checkProfile.name === profileName) {
                     Profiles.getInstance().profilesForValidation.splice(index, 1);
                 }
--- a/packages/zowe-explorer/src/abstract/ZoweTreeProvider.ts
+++ b/packages/zowe-explorer/src/abstract/ZoweTreeProvider.ts
@@ -168,7 +168,7 @@ export class ZoweTreeProvider {
             if (await node.getSession()) {
                 await setSession(node, EditSession as ISession);
             } else {
-                this.deleteSessionByLabel(node.label);
+                zoweFileProvider.deleteSession(node.getSessionNode());
                 this.mHistory.addSession(node.label);
                 zoweFileProvider.addSession(node.getProfileName());
             }
--- a/packages/zowe-explorer/src/abstract/ZoweTreeProvider.ts
+++ b/packages/zowe-explorer/src/abstract/ZoweTreeProvider.ts
@@ -156,7 +156,7 @@ export class ZoweTreeProvider {
         return undefined;
     }
 
-    public async editSession(node: IZoweTreeNode) {
+    public async editSession(node: IZoweTreeNode, zoweFileProvider: IZoweTree<IZoweNodeType>) {
         const profile = node.getProfile();
         const profileName = node.getProfileName();
         // Check what happens is inactive
@@ -165,11 +165,16 @@ export class ZoweTreeProvider {
         if (EditSession) {
             node.getProfile().profile = EditSession as IProfile;
             await setProfile(node, EditSession as IProfile);
-            await setSession(node, EditSession as ISession);
+            if (await node.getSession()) {
+                await setSession(node, EditSession as ISession);
+            } else {
+                this.deleteSessionByLabel(node.label);
+                zoweFileProvider.addSession(node.getProfileName());
+            }
             this.refresh();
         }
         try {
-            // refresh profilesForValidation to check the profile status again
+            // Remove the edited profile from profilesForValidation since it should be revalidated
             Profiles.getInstance().profilesForValidation.forEach((checkProfile, index) => {
                 if (index === 0) {
                     Profiles.getInstance().profilesForValidation = [];
@@ -178,8 +183,6 @@ export class ZoweTreeProvider {
                     Profiles.getInstance().profilesForValidation.splice(index, 1);
                 }
             });
-
-            await this.checkCurrentProfile(node);
         } catch (error) {
             await errorHandling(error);
         }
--- a/packages/zowe-explorer/src/abstract/ZoweTreeProvider.ts
+++ b/packages/zowe-explorer/src/abstract/ZoweTreeProvider.ts
@@ -169,6 +169,7 @@ export class ZoweTreeProvider {
                 await setSession(node, EditSession as ISession);
             } else {
                 this.deleteSessionByLabel(node.label);
+                this.mHistory.addSession(node.label);
                 zoweFileProvider.addSession(node.getProfileName());
             }
             this.refresh();
--- a/packages/zowe-explorer-api/src/profiles/ProfilesCache.ts
+++ b/packages/zowe-explorer-api/src/profiles/ProfilesCache.ts
@@ -224,6 +224,7 @@ export class ProfilesCache {
             profile: profileInfo,
             name: profileName,
             type: profileType,
+            overwrite: true,
         });
         return newProfile.profile;
     }
--- a/packages/zowe-explorer-api/src/tree/IZoweTree.ts
+++ b/packages/zowe-explorer-api/src/tree/IZoweTree.ts
@@ -47,7 +47,7 @@ export interface IZoweTree<T> extends vscode.TreeDataProvider<T> {
      * Edit a session to the container
      * @param node This parameter identifies the node that needs to be called
      */
-    editSession(node: IZoweNodeType): Promise<void>;
+    editSession(node: IZoweNodeType, zoweFileProvider: IZoweTree<IZoweNodeType>): Promise<void>;
 
     /**
      * Add a new session to the container
--- a/packages/zowe-explorer/__tests__/__unit__/KeytarCredentialManager.unit.test.ts
+++ b/packages/zowe-explorer/__tests__/__unit__/KeytarCredentialManager.unit.test.ts
@@ -120,8 +120,7 @@ describe("KeytarCredentialManager Unit Tests", () => {
             error = err;
         }
         expect(error).toBeDefined();
-        expect(error.additionalDetails).toContain("Service = Zowe-Plugin");
-        expect(error.additionalDetails).toContain("Awesome-Service\n  Account = user5");
+        expect(error.additionalDetails).toContain("Could not find an entry in the credential vault");
     });
 
     it("Test saving passwords to credential store", async () => {
--- a/packages/zowe-explorer/__tests__/__unit__/Profiles.unit.test.ts
+++ b/packages/zowe-explorer/__tests__/__unit__/Profiles.unit.test.ts
@@ -1015,7 +1015,7 @@ describe("Profiles Unit Tests - Function updateProfile", () => {
         };
         newMocks.profiles = await Profiles.createInstance(newMocks.log);
         newMocks.profileInstance = createInstanceOfProfile(newMocks.profiles);
-        newMocks.changedImperativeProfile.profile = { user: "test2", password: "test2" };
+        newMocks.changedImperativeProfile.profile = { user: "test2", password: "test2", rejectUnauthorize: true };
         newMocks.profileInfo = newMocks.changedImperativeProfile;
         Object.defineProperty(globalMocks.mockCliProfileManager, "load", {
             value: jest.fn(() => {
--- a/packages/zowe-explorer/__tests__/__unit__/Profiles.unit.test.ts
+++ b/packages/zowe-explorer/__tests__/__unit__/Profiles.unit.test.ts
@@ -648,7 +648,7 @@ describe("Profiles Unit Tests - Function createNewConnection", () => {
         globalMocks.mockShowQuickPick.mockResolvedValueOnce("False");
 
         await blockMocks.profiles.createNewConnection("alternate");
-        expect(globalMocks.mockShowInformationMessage.mock.calls[0][0]).toBe("Operation Cancelled");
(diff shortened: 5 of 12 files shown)


JillieBeanSim:
Hey @katelynienaber I was testing this PR, while creating a profile I still get `""` for optional values left blank along with the `0` value for encoding and response timeout that where left blank/entered passed.  
image

Then I edited the same profile leaving those same values blank and still got `0` in the yaml for them.
image

JillieBeanSim (changes requested):
see comment above for issues I ran into.

the changes requested are:
Should we use this same way of thinking during create?
Also, would like to see the default value for encoding used or leave it out of the yaml 
Finally, shouldn't it leave response timeout out of the yaml if value = 0?

katelynienaber:
> see comment above for issues I ran into.
> 
> the changes requested are:
> Should we use this same way of thinking during create?
> Also, would like to see the default value for encoding used or leave it out of the yaml
> Finally, shouldn't it leave response timeout out of the yaml if value = 0?

Hey @JillieBeanSim, give it a shot now.

**Should we use this same way of thinking during create?**
I think so. Create is kind of hacky right now because the `CliProfileManager.save` function does some validation, and won't let you pass in non-string values for user & password. So I'm calling our `updateProfile` function on newly-created profiles, just after the profile manager is refreshed, to remove any blank fields.

**Also, would like to see the default value for encoding used or leave it out of the yaml**
This one should be fixed. I rearranged the conditions in that loop I added.

**Finally, shouldn't it leave response timeout out of the yaml if value = 0?**
This should also be good now.

jellypuno:
> Should we use this same way of thinking during create?
> I think so (and now Create should work in the new way, not storing empty values). Create is kind of hacky right now because the CliProfileManager.save function does some validation, and won't let you pass in non-string values for user & password. So I'm calling our updateProfile function on newly-created profiles, just after the profile manager is refreshed, to remove any blank fields.

@katelynienaber Let's think of another way and not implement a hack. We can check zowe/cli code as to how they implmented this.

@JillieBeanSim When you say "same way of thinking during create", Are you thinking of re-factoring that process? Code-wise or UX-wise or both? If yes, do you a design that we can look into?

katelynienaber:
@JillieBeanSim Ok, now we have a not-hacky solution :) @jellypuno pointed out that [this option](https://github.com/zowe/imperative/blob/741f24f7246028879317dbd9590b91d7efbdef39/packages/cmd/__tests__/profiles/CliProfileManager.test.ts#L280) could allow passing optional parameters, and it works (at least for me)

Please check again when you have time ^^"

JillieBeanSim:
> 
> @JillieBeanSim When you say "same way of thinking during create", Are you thinking of re-factoring that process? Code-wise or UX-wise or both? If yes, do you a design that we can look into?

@jellypuno I was only thinking of the way it was saved to the yaml, ie. no empty strings and `0` values, not a UI change.

Thanks for looking into my questions and concerns @katelynienaber & @jellypuno I will retest the PR today.

JillieBeanSim:
@katelynienaber the functionality works well now and as expected for create and update of profiles without optional user/pass and for the empty encoding and response timeout values.  I do see this message pop up sometimes after editing a profile that doesn't have the values mentioned.  Not sure but I believe this error is being handled in @jellypuno PR https://github.com/zowe/vscode-extension-for-zowe/pull/1100 I know I ran into it there as well. If this is correct @jellypuno let me know and I will approve this PR

zdmullen:
Merge this after Jelly's PR #1100!

katelynienaber:
@JillieBeanSim I think this issue was fixed in @jellypuno's PR, could you check it out again? I don't see it in mine.

lauren-li (dismissed):
This PR is looking good to me. All unit and expected integration tests are passing. I am able to create and edit profiles without specifying username, password, encoding, and/or timeout, and the yaml is created/updated without empty fields, which is really great. I do sometimes see the error that @JillieBeanSim mentioned in her [comment](https://github.com/zowe/vscode-extension-for-zowe/pull/1120#issuecomment-758738277) above, ... but I can also get it to happen in v1.11.1 the same way (but not reliably). We may want to look into it as a separate issue.

Another issue I encountered while testing, which I also see in 1.11.1 (and thus not a blocker for this PR) is that I get prompted for credentials immediately upon finishing creation or editing of a profile without username/password. I suppose this behavior has been around for a while, but somehow I only just now noticed it, and it feels unexpected since I am not trying to use the profile to access anything yet; I'm just udpating/creating it. Again, this could be investigated as a separate issue. I just wanted to mention it here so others are aware that it isn't a new behavior with this PR.

Thanks @katelynienaber for the fixes in this PR! It was such a pain to deal with empty profile fields when troubleshooting/testing other profile-related PRs, and I am happy to see that getting resolved.

jellypuno:
Maybe its just me but after I created my profile, I edited it and I encountered this error:

image

This is my profile:
```
host: 192.86.32.250
port: 10443
rejectUnauthorized: false
```

lauren-li:
@JillieBeanSim and I have seen an issue in `master` branch, which is not a regression from this PR. However, the fact that this PR works like it's supposed to makes the existing issue much easier for users to encounter. It seems that when clicking a view's "RefreshAll" button, profiles without a user/password field are not refreshed:

ze-refreshAll

In the gif above, the profiles `d`, `ncz`, and `tivNC` all had in common that they were missing user/password fields. Just to be sure, I also tested profiles that had user/password filled in, but lacked encoding/timeout fields (both with good hosts and bad hosts to test active vs inactive), and those profiles were also able to refresh fine.

I also get this message in the debug console when the issue occurs: `TypeError: Cannot read property 'ISession' of undefined at Object.refreshTree`.

We should be sure we can fix this issue in 1.12 before merging this PR in. Again, this PR does not seem to be causing the issue, but it just happens to expose the user to it very easily. (I am still very excited that we are getting rid of empty strings in the profile to mirror what Zowe CLI does, though.)

katelynienaber:
> Another issue I encountered while testing, which I also see in 1.11.1 (and thus not a blocker for this PR) is that I get prompted for credentials immediately upon finishing creation or editing of a profile without username/password.

@lauren-li I will look into this as part of PR #1150

katelynienaber:
@lauren-li I disagree that we should wait to merge this. I'm not convinced we can get another PR finished and reviewed twice by tomorrow. And therefore we would be delaying a bugfix for 3 issues affecting 5 profile fields until the next release, because it reveals 1 issue affecting 2 profile fields.

jellypuno:
Issues that needs to be addressed:

1. Cannot create a profile if I am using SCS
image

2. This appears if I edit a newly created profile: 
image

    the yaml file was updated though.

    image

3. Similar to  @lauren-li comment, the status of the profile is not refreshing.

jellypuno:
For the SCS issue, I think there's something that we need to update in our own SCS. I am not sure what though. I'm hoping @t1m0thyj can help us.... 😊

katelynienaber:
> 1. Cannot create a profile if I am using SCS

Blargh SCS. This still needs fixing...

> 2. This appears if I edit a newly created profile:
>    image

I've fixed this, just copied it out of the old base profiles PR

> 3. Similar to @lauren-li comment, the status of the profile is not refreshing.

@lauren-li @jellypuno I went ahead and fixed this here. Please please can we merge it like this it is such a small change, another PR for this would be overkill 8|

Original code:
`if (sessNode.getSession().ISession !== SessionProfile) {`
New fix:
`if (sessNode.getSession() && sessNode.getSession().ISession !== SessionProfile) {`
Now, it will not try to set the session at all, if the node doesn't already contain one. This is during refresh _only_. It will still update the session if you ex. edit the profile, or enter your creds at a prompt.

> 4. After step# 2 I clicked the magnifying glass and this error appears:
>    image
>    I've enabled validation and I am sure that the hostname is active. I expected that I will be prompted to enter my credentials

I am guessing you have created a profile with user/pass, and then are trying to immediately edit it and remove the user/pass...at least that's what I fixed. I just removed validation in the end of the `editSession` function. When you try to use the profile it will validate you as usual.

jellypuno:
@katelynienaber Once you push your commit, I'll test it again.

katelynienaber:
@jellypuno lol totally forgot to do this

lauren-li:
> @lauren-li I went ahead and fixed this here. Please please can we merge it like this it is such a small change, another PR for this would be overkill 8|

I think it's fine if it is fixed in this PR (which it now appears to be). My intention here wasn't to say that the fix _had_ to be in a separate PR, but that it could be if you didn't want to make any additional changes to this PR, since it was looking pretty good to me at the time I wrote my comment. However, since you've also made updates to address issues @jellypuno found, I will test the PR for those as well now. Thanks @katelynienaber for your hard work on this! I will let you know soon how my testing goes.

t1m0thyj:
> I think issue is [here](https://github.com/zowe/imperative/blob/741f24f7246028879317dbd9590b91d7efbdef39/packages/cmd/src/profiles/CliProfileManager.ts#L408)...imperative's save function is trying to delete the field from an existing profile, if you pass in a profile where the field is missing. Since we are making a new profile, the field doesn't exist and can't be deleted. The causeError I am getting is `Error: Unable to delete credentials. at KeytarCredentialManager.`

@katelynienaber This same issue used to happen in the SCS plugin. It was fixed by updating CredentialManager class to fail silently when it attempts to delete a non-existent secure field. This ensures that the field gets deleted if it exists, and eliminates the unwanted failure.

https://github.com/zowe/zowe-cli-scs-plugin/pull/7/files#diff-7fd952b958aa47583d7349b4404413e8918ccac03052333f1dfcb8990b5af5cfL105-L114

I'd suggest implementing the same fix for ZE's KeytarCredentialManager. If you agree it seems safe, you can remove the error being thrown in `deleteCredentialsHelper` 🙂 
https://github.com/zowe/vscode-extension-for-zowe/blob/f0f93e90e01eb0f725c9c630a2d406d61ae7b51d/packages/zowe-explorer/src/KeytarCredentialManager.ts#L73-L80

katelynienaber:
@t1m0thyj Thank you very much! Merging this fix now. I'm logging the error but it won't fail completely.

lauren-li:
Referencing @jellypuno's comment: https://github.com/zowe/vscode-extension-for-zowe/pull/1120#issuecomment-766713463

> 1. Cannot create a profile if I am using SCS

I don't see this issue when testing now.

> 2. This appears if I edit a newly created profile:

I don't see the error message @jellypuno posted; however, I do see this one when trying to edit a profile that has no username/password (whether newly-created or from a previous session):
Screen Shot 2021-01-25 at 1 17 09 PM

> 3. Similar to @lauren-li comment, the status of the profile is not refreshing.

This appears to have been fixed!

> 4. After step# 2 I clicked the magnifying glass and this error appears:

- I still get these errors in MVS and Jobs when trying to run a search using a profile without credentials. (I also have validation enabled, and was expecting to be prompted for username/password.)
- In USS, when I try to run a search using a profile without credentials, I get a different error:
    Screen Shot 2021-01-25 at 1 11 04 PM

~

Some other odd behavior I saw that I wasn't able to produce on `master` branch:
- After deletion, profiles show up again in the dropdown list of available profiles when clicking the **+** button, even though their corresponding `yaml` file was deleted. (This doesn't happen 100% of the time, but I was able to recreate it fairly frequently, particularly for profiles without credentials.)
    ze-1120-deletep

- Editing a profile (especially one without creds/encoding/timeout) causes it to be removed from the `sessions` list in the `settings.json` file (and therefore, the profile does not load automatically upon session restart). I also was not able to recreate this with 100% consistency, but I was still able to recreate it fairly frequently:
    ze-1120-editp

katelynienaber:
> I don't see the error message @jellypuno posted; however, I do see this one when trying to edit a profile that has no username/password (whether newly-created or from a previous session):

This error is present in master and is related to [this line](https://github.com/zowe/vscode-extension-for-zowe/blob/06a5fcc2bc087c80d00f4ee704616c078274f12b/packages/zowe-explorer/src/abstract/ZoweTreeProvider.ts#L163). It isn't related to the original problem of empty fields being stored improperly and removing this line might cause adverse affects, so I won't fix it here. It doesn't stop you from editing the profile in any case.

> > 1. After step# 2 I clicked the magnifying glass and this error appears:
> 
> * I still get these errors in MVS and Jobs when trying to run a search using a profile without credentials. (I also have validation enabled, and was expecting to be prompted for username/password.)

I am inclined not to fix this one here. It's kind of a big deal XD And it's already going to happen in master if you have any invalid profile and no base profile, it's not caused by the changes in this PR. It occurs because we get to [this point](https://github.com/zowe/vscode-extension-for-zowe/blob/06a5fcc2bc087c80d00f4ee704616c078274f12b/packages/zowe-explorer-api/src/profiles/ZoweExplorerZosmfApi.ts#L82) without ever having prompted for creds. In the original base profiles PR we did [this](https://github.com/zowe/vscode-extension-for-zowe/blob/f29a955740b6c4db74189f501cb994ae3011ebd2/src/api/ZoweExplorerZosmfApi.ts#L147), but to implement something like that would be a huge change that would require lots of testing.

> * In USS, when I try to run a search using a profile without credentials, I get a different error:

I rearranged the search function, so that it tries to retrieve the session from the node _after_ the node's profile is validated. **If you don't have user/pass on the profile, the error above will still occur.** But like I said that error is a biggie and should be dealt with separately. The workaround is to disable validation.

> * After deletion, profiles show up again in the dropdown list of available profiles when clicking the **+** button, even though their corresponding `yaml` file was deleted. (This doesn't happen 100% of the time, but I was able to recreate it fairly frequently, particularly for profiles without credentials.)

This should be fixed.

> * Editing a profile (especially one without creds/encoding/timeout) causes it to be removed from the `sessions` list in the `settings.json` file (and therefore, the profile does not load automatically upon session restart). I also was not able to recreate this with 100% consistency, but I was still able to recreate it fairly frequently:

This should also be fixed.

jellypuno:
@katelynienaber I am still having this error:

image


Steps to reproduce:
1. Create a new profile with invalid URL and without User and Password
2. Edit the said profile and enter the correct URL
3. Click the magnifying glass to search for a dataset

Note: My base profile contains host, port and ru. Same with my service profile. And the URL for both profiles are different

jellypuno:
I can create and edit profiles with token. thank you for that. I tried to submit a job and I encountered this error.

image

katelynienaber:
@jellypuno Can you give it another shot please?

packages/zowe-explorer/src/abstract/ZoweTreeProvider.ts
```diff
@@ -169,6 +169,7 @@ export class ZoweTreeProvider {
                 await setSession(node, EditSession as ISession);
             } else {
                 this.deleteSessionByLabel(node.label);
+                this.mHistory.addSession(node.label);
```
lauren-li: Adding L172 ` this.mHistory.addSession(node.label);` seems to have fixed the issue where the profile was getting removed out of the `settings.json` file after editing, but I'm a little confused what lines 171-172 are meant to be doing now. It looks like L171 just removes the session from `settings.json` and then L172 just adds it back in and cancels L171 out. Could these two lines just be removed?

jellypuno:
Small issue, I think we can just filter this out and stop displaying it.

This appears when I right-click > Update Profile

image

jellypuno:
Error when I tried to add a newly created profile in USS favorites:

image

katelynienaber:
@jellypuno Fixed this...there is still the "profile inactive" error popping up, but the profile is added to favorites. I think this inactive error should be dealt with in another PR, because it occurs all over ZE, not just in profiles operations.

lauren-li:
@katelynienaber What is the expected result if the user has made profile ABC without credentials, does not have a base profile, and uses profile ABC to run a search? I get the set of errors below. I was thinking we should be prompting users for username/password, but I don't get a chance to do that with my new profile:

Screen Shot 2021-01-28 at 12 49 18 PM

katelynienaber:
@lauren-li If you can turn off validation for the profile and get a prompt thereafter...see my post above XD Basically the flow is messed up for validation and it's too big of an unrelated issue to be addressed here.

If you turn off validation and it still won't prompt, let me know, I will try again.

> I am inclined not to fix this one here. It's kind of a big deal XD And it's already going to happen in master if you have any invalid profile and no base profile, it's not caused by the changes in this PR. It occurs because we get to this point without ever having prompted for creds. In the original base profiles PR we did this, but to implement something like that would be a huge change that would require lots of testing.

JillieBeanSim:
@katelynienaber @jellypuno @lauren-li if the user has to do a bunch of work arounds to have this PR in and their profiles to show up and work then I personally don't think this should go out in this release.  I agree the inactive error should be dealt with in another PR but I also believe that this PR and the inactive one will need to go out together possibly as bug fixes in 1.12.1.  This is just my opinion so we don't break users and have to point them to workarounds like disabling validation or requiring them to have a base profile.

lauren-li:
@katelynienaber I'm inclined to agree with @JillieBeanSim on this. This PR is in pretty good shape, I really like it, and like you said, fixing the root cause of the error prompts is a bigger issue. But introducing this feature without addressing those prompts has the potential to degrade the user experience, since they would now be getting errors for something they were able to do without any problems before.

JillieBeanSim:
I may be able to even pitch in with the inactive error/prompting for creds if the field is missing PR as to have these ready for 1.12.1, I was fiddling with it a couple of weeks ago before I jumped in with the filter icon PR

katelynienaber:
@lauren-li @JillieBeanSim It's fine with me if you want to wait on this one, it wasn't promised in the PI goals :D We should bring it up with @zdmullen and @jellypuno as well.

JillieBeanSim (approved):
This PR works great now with https://github.com/zowe/vscode-extension-for-zowe/pull/1178 added in.  I don't see any of the issues we were seeing before.  Thanks @katelynienaber for doing this work.  All expected tests and checks are passing

lauren-li (approved):
This looks good to me now! All unit and expected integration tests are passing. Omitting credentials when creating and updating a profile in Zowe Explorer now omits the respective fields instead of creating empty strings. With the merging-in of PR #1178, this works much more seamlessly than before (i.e. no new "error" pop-ups). This PR also appears to be "backwards compatible" with the existing credential-less profiles, in that those profiles are still able to be used, and updating them with empty credentials will just completely remove the blank `user` and `password` fields.

Thanks @katelynienaber for your hard work on this PR!

PR #1243, opened 2021-03-18
Add connection buttons

VitGottwald:
Add-Connection

## Proposed changes

Adds welcome content with big buttons if no profiles are defined in dataset/uss/jobs trees.

## Release Notes

When no profiles are defined, the dataset, uss, and jobs tree views now display that information along with a big button to add a connection to the tree.

Milestone: N/A

Changelog:

## Types of changes

- [x ] New feature (non-breaking change which adds functionality)

## Checklist

- [x gif or screenshot is included if visual changes are made


VitGottwald:
Unit tests are obviously failing, because the tree content has been changed. I would like to get input whether this POC of welcome content is heading in the right direction before putting energy into updating tests.

@jelaplan, @tomofbroadcom, @jellypuno

VitGottwald:
One question I was asked when showing this POC was whether it would make more sense to use

- Add New Dataset Connection
- Add New USS Connection
- Add New Jobs Connection

instead of 'Add New Connection' everywhere?

lauren-li:
> One question I was asked when showing this POC was whether it would make more sense to use
> 
>     * Add New Dataset Connection
> 
>     * Add New USS Connection
> 
>     * Add New Jobs Connection
> 
> 
> instead of 'Add New Connection' everywhere?

Since the connections can be reused between different trees, I think a generic "Add New Profile" is appropriate for the buttons. (I also picked the word "Profile" instead of "Connection" since it seems to have become our favored terminology.)

I really like the concept of using these welcome views when the user has no profiles. Something I want to test (but haven't gotten to yet) is if these welcome views will activate if a user only has extender profiles but no z/OSMF profiles. For example, if a user is just using FTP to connect and only has zFTP profiles, would these welcome screens be activated before the zFTP profiles have a chance to load into the trees?

VitGottwald:
Thank you for looking at this @lauren-li! The _welcome view_ will be displayed if the tree has no nodes. So if there are any profiles, they will be shown instead of the welcome view.

Regarding the button names. Currently when we hover over the `+` button to add new profile the tooltips is called "Add New Connection". So I sided with it. I also think it is a bit more friendly to new users who most likely do not have an understanding what a "profile" is.

I do not mind changing it to `Add New Profile` if that would be preferable.

packages/zowe-explorer/src/dataset/DatasetTree.ts
```diff
@@ -473,6 +472,10 @@ export class DatasetTree extends ZoweTreeProvider implements IZoweTree<IZoweData
      */
     public deleteSession(node: IZoweDatasetTreeNode) {
         this.mSessionNodes = this.mSessionNodes.filter((tempNode) => tempNode.label.trim() !== node.label.trim());
+        // if we only have favorites node, remove it so that viewsWelcome button shows up
+        if (this.mSessionNodes.length === 1) {
```
nickImbirev, line 476: Soo, why we are deleting the 'Favorite' session node? 

I mean, in case I have have something in 'Favorites' - I don't want to use the proposed welcome screen, but it still appears.

Maybe we can keep 'Favorite' node separately inside tree provider,  instead of mixing all nodes in `mSessionNodes` variable? In that case, we would have to verify, that 'Favorites' is empty and show the welcome view, that we designed. Does it make sense?

lauren-li:
> Thank you for looking at this @lauren-li! The _welcome view_ will be displayed if the tree has no nodes. So if there are any profiles, they will be shown instead of the welcome view.

This is true, and I have tested that this also works if the user only has extender (non-z/OSMF) profiles as well. However, the welcome view does show for a couple seconds while Zowe Explorer is activating (even with z/OSMF profiles). We might want to look into seeing if we can get rid of that, or at least minimize it. The good thing is that it's not too disruptive to the user experience, in that if they do click on the "Add new connection" button while Zowe Explorer is activating, it doesn't cause errors or crash the extension or anything like that when the profiles load into the UI.

> Regarding the button names. Currently when we hover over the `+` button to add new profile the tooltips is called "Add New Connection". So I sided with it. I also think it is a bit more friendly to new users who most likely do not have an understanding what a "profile" is.
> 
> I do not mind changing it to `Add New Profile` if that would be preferable.

@VitGottwald You make a good point about new users maybe not understanding what a "profile" is. Maybe we can use "Add New Connection Profile"? Or maybe we can get some input from other squad members (or even better, users) on the labeling for this button.
Quick note: In my PR #1308 (Clean up commands in Command Palette), the hover text for the "+" button was changed from "Add Profile"/"Add Session" (in v1.14.0 release) to "Add Profile to x View" (where "x" is Data Sets, USS, or Jobs). This is because the hover text is shared with the command's label in the Command Palette, and if users are accessing the "Add Profile" command through the Command Palette, they should be able to distinguish which view they are adding a profile to.

packages/zowe-explorer/src/dataset/DatasetTree.ts
VitGottwald: Without profiles, there is no way to connect to a z/OS LPAR. So there is no way to make use of the favourites. So I believe it is correct that we show the welcome view.

I like the idea of separating the nodes rather than having magic numbers. But I wanted to limit the scope of changes just to adding the new welcome view.

nickImbirev: Hmm, but I have my profiles in the file system, they also available via 'Favorites' and I can expand them and access LPAR, even if they were removed from the tree (except 'Favorites' node).

Will we remove this ability?

VitGottwald: Interesting, I just tried it out and you are right @nickImbirev. Thanks for pointing that out! I guess I do not understand the favourites section. It looks like it duplicates the profile tree, only being less powerful (individual data sets instead of filters).

VitGottwald: Ok, I think I get it now. The search filter will not survive vscode restarts, but the favourites do. Thank you @jellypuno. 

So I believe we should not display the welcome view if favourites exist as you suggested @nickImbirev.

Issue #650, opened 2020-03-31
USS Tree collapses after folder rename


lauren-li:
When renaming a USS folder, the entire tree will collapse (to refresh the contents) after the rename. This means the user has to subsequently go look for the renamed folder in order to continue working with it. It would be a friendlier UX if the tree was able to stay open after the rename. (Perhaps we could achieve this by refreshing at the folder level, rather than refreshing the whole USS tree.)

![ussRename2](https://user-images.githubusercontent.com/45975633/78074371-a598b000-7370-11ea-96c6-17b116edd60c.gif)

In addition: 

USSTree collapses when a new directory or file is created. It is inconsistent with the way Dataset works. The list should just refresh and not collapse
- Note: This just seems to happen at the top level, and not with creation under subfolders.

zFernand0:
Crazy idea: 
- Recursively goo back (or `node->parent`) until the root doing refreshElements (which should keep the parent elements open) :yum:

zdmullen:
Revisit after the pull request for issue #563  is merged.

katelynienaber:
Note: **only renaming a folder** collapses the tree.

Creating a folder/file OR renaming a file does not.

JillieBeanSim:
Thanks @katelynienaber we can revisit this after the PR is merged in

Issue #973, opened 2020-08-06
Rename causes VS Code to flip through editor tabs


lauren-li:
**Describe the bug**
Renaming data set and USS items via Right click > Rename flips between active editor tabs while completing the rename.

**To Reproduce**
1. Have several editor tabs open (at least 3), and take note of which one is currently active
1. Right-click a data set or USS item
1. Click on 'Rename'
1. Type a new name in the dialog box, then press Enter
1. See VS Code flip through a couple editor tabs before ending on a different active editor tab than in Step 1.
1. Rename is completed

**Screenshot**
(Note: This gif demonstrates the issue with local Zowe Explorer files open, but the issue occurs just the same when anything is open in the editor tabs, including files/data sets opened through Zowe Explorer.)

![Zowe Explorer rename issue - med](https://user-images.githubusercontent.com/45975633/89584078-76f08180-d809-11ea-9262-741bf935358d.gif)

**Expected behavior**
Right click > Rename should rename data set and USS items without changing the active editor tab.

**Desktop**
-   OS: Mac OS
-   Zowe Explorer Version: 1.8.0
-   (Optional) Zowe CLI Version: 6.13.0

**Additional context**
This issue is due to the `closeOpenedTextFile()` function that was added 2 months ago: https://github.com/zowe/vscode-extension-for-zowe/blob/master/src/dataset/DatasetTree.ts#L690
Thanks @jellypuno for this tip!

zFernand0:
It seems that an unexpected error can pop-up due to server instability after a rename action occurs.
(Happened to me earlier today)
Steps?:
- Someone has a few datasets and USS files already opened
- They rename something
- Some dataset fails to be retrieved (I tried reproducing by deleting the dataset outside of Zowe Explorer, but no luck)

lauren-li:
Branch with preliminary work to address this: https://github.com/zowe/vscode-extension-for-zowe/tree/rename-tabs

traeok:
This no longer occurs in v3

PR #1259, opened 2021-04-01
Issue #650 partial fix: USS tree collapses after folder rename

lauren-li:
## Proposed changes

This PR:
- Partially resolves issue #650 (USS tree collapses after folder rename). Folders underneath the renamed directory still collapse, but at least the USS tree stays open down to the level of the renamed directory.
- Removes the `renameUSSNode` function from `packages/zowe-explorer/src/uss/actions.ts`, as it appears to be a near-duplicate code of the `rename` function in `packages/zowe-explorer/src/uss/USSTree.ts` and was only referenced by its unit tests (which I also removed).
- Adds a check to prevent the user from trying to rename an opened, unsaved file, or a folder containing an opened, unsaved file. (Currently, Zowe Explorer doesn't handle this renaming situation properly, so we probably want to prevent it from trying.) See gif below:

  ze-renameOpened

**Note:** The behavior in the gif with the opened file closing during the rename is also present in the current Zowe Explorer release (v1.13.1). A separate issue has been filed to address this: #973 (Rename causes VS Code to flip through editor tabs).

Other USS rename changes made in this PR:
- Fix for rename path issue. (See comments [here](https://github.com/zowe/vscode-extension-for-zowe/pull/1259#discussion_r613920261) and [here](https://github.com/zowe/vscode-extension-for-zowe/pull/1259#discussion_r613924332).)
- Fix rename of corresponding node in Sessions section when rename is initiated from a favorited subfolder. (See comment [here](https://github.com/zowe/vscode-extension-for-zowe/pull/1259#issuecomment-823181886).)

## Release Notes

Milestone: 1.14.0

Changelog: Keep USS tree open to the level of the renamed directory.

## Types of changes

## Checklist


lauren-li:
All unit tests are passing for me locally. I'm not quite sure why they aren't passing on GitHub. If reviewers could please let me know if the unit tests pass or fail for you locally so we can investigate more, I would really appreciate it. Thank you!

jellypuno:
@lauren-li I am experiencing this error when running tests.

image

I executed `yarn` and then `compile` before running tests. Maybe that would work for you?

lauren-li:
@jellypuno I still get all unit tests passing locally when I run `yarn` and then `compile` before running the unit tests. However, I looked into this some more, and it turns out the test is failing because Windows uses `\\` in its path names, and the paths in the unit tests' mock text documents use `/` (which works great on my local Mac). Thank you for checking this out, and thanks to @zFernand0 for tipping me off that it might be a Windows/Mac thing! I am working on a fix for this now.

zFernand0 (approved):
LGTM!

jellypuno:
@lauren-li Can you please check the theia tests?

packages/zowe-explorer/src/uss/USSTree.ts
```diff
@@ -94,10 +94,34 @@ export class USSTree extends ZoweTreeProvider implements IZoweTree<IZoweUSSTreeN
         const oldFavorite: IZoweUSSTreeNode = contextually.isFavorite(originalNode)
```
lauren-li, line 94: ```const parentPath = originalNode.fullPath.substr(0, originalNode.fullPath.indexOf(originalNode.label));```

This line (92) causes a pre-existing issue where if the user is renaming a USS item whose label happens to be a substring of a folder name further up its parent path, the renamed item will be moved up the folder hierarchy to the same level as the folder its label was originally a substring of.

For example, if you wanted to rename `/u/user1/abc/a` to `/u/user1/abc/z`, the result would actually be `/u/user1/z` because `a` is a substring of `abc`.

One way we could fix this is by using the path library to determine the parent path instead:
`const parentPath = path.dirname(originalNode.fullPath);`

Questions for reviewers: Do you see any issues with the proposed fix? Do we want to include this in this PR since it involves USS rename?

packages/zowe-explorer/src/uss/USSTree.ts
```diff
@@ -109,20 +133,29 @@ export class USSTree extends ZoweTreeProvider implements IZoweTree<IZoweUSSTreeN
                 newNamePath = newNamePath.replace(/\\/g, "/"); // Added to cover Windows backslash issue
```
lauren-li, line 133:
```
let newNamePath = path.join(parentPath + newName);
newNamePath = newNamePath.replace(/\\/g, "/"); // Added to cover Windows backslash issue
```

Can the code to get `newNamePath` be replaced with :
`const newNamePath = path.join(parentPath, newName);`

@katelynienaber was there any special reason we used the `+` in the `path.join` argument and then did the manual replace afterwards?

lauren-li:
I made a branch based off of this PR that has the file path suggestions I made in earlier comments. If we want to include those for 1.14.0, I could just pull those changes into this PR so reviewers can look at all the USS rename changes in one PR: https://github.com/zowe/vscode-extension-for-zowe/tree/rename-paths

packages/zowe-explorer/src/uss/USSTree.ts
lauren-li: Addressed by pulling in commit 4a052cb after scrum discussion.

packages/zowe-explorer/src/uss/USSTree.ts
lauren-li: Addressed by pulling in commit 4a052cb after scrum discussion.

lauren-li:
> I made a branch based off of this PR that has the file path suggestions I made in earlier comments. If we want to include those for 1.14.0, I could just pull those changes into this PR so reviewers can look at all the USS rename changes in one PR: https://github.com/zowe/vscode-extension-for-zowe/tree/rename-paths

Per scrum discussion, I have merged the contents of the `rename-paths` branch into this PR. The changes from the `rename-paths` branch can be found in commit 4a052cb.

zFernand0:
> > I made a branch based off of this PR that has the file path suggestions I made in earlier comments. If we want to include those for 1.14.0, I could just pull those changes into this PR so reviewers can look at all the USS rename changes in one PR: https://github.com/zowe/vscode-extension-for-zowe/tree/rename-paths
> 
> Per scrum discussion, I have merged the contents of the `rename-paths` branch into this PR. The changes from the `rename-paths` branch can be found in commit [4a052cb](https://github.com/zowe/vscode-extension-for-zowe/commit/4a052cb8dbbe95eed7eb74f1103d81c3dd8c22ac).

LGTM! : )

zFernand0 (approved):

jellypuno:
@lauren-li Maybe I am missing something but this is not working for me. I am having this error when I try to rename a directory:

image


Then if I try and FTP Profile, this is the error:
image

I am not sure why in FTP it detects it as a dataset.

JillieBeanSim (changes requested):
@lauren I am seeing these errors with a zosmf profile
image

my rse profile worked fine, but I will install ftp cli and extension and test that as well

lauren-li:
It seems that `path.join()` was not actually returning the correct paths on Windows systems, so it has been replaced with `path.posix.join()` in the latest commit (60ca99c). Please give this a try and see if resolves the error messages.
Thank you @jellypuno for finding this issue, and thank you @JillieBeanSim for helping me debug it!

JillieBeanSim:
Thanks @lauren-li for getting the path name fix in, I am not seeing those errors anymore with zOSMF or zFTP profiles.  I do see some strange behavior, not sure if this is known since the PR states that it's a partial fix.
1.  When hiding profiles the `get USS File list command submitted` info message pops up multiple times, this isn't consistent though.
2.  When renaming a favorite with the session and file path to the favorite is open as well the rename doesn't show up automatically in the session tree.  Where if the file is open and closed then reopened for the rename of favorite it does reflect in the session when the tree is repainted.

ZE-rename-USS

lauren-li:
> Thanks @lauren-li for getting the path name fix in, I am not seeing those errors anymore with zOSMF or zFTP profiles. I do see some strange behavior, not sure if this is known since the PR states that it's a partial fix.
> 
> 1. When hiding profiles the `get USS File list command submitted` info message pops up multiple times, this isn't consistent though.
> 
> 1. When renaming a favorite with the session and file path to the original of the favorited item is expanded in the tree as well the rename doesn't show up automatically in the session tree.  Where if the file is open and closed then reopened for the rename of favorite it does reflect in the session when the tree is repainted.
> 
> 1. Also, not shown in the gif.  When dealing with number 2 and try to click on the filename that wasn't yet changed within the session I get an error back but at the same time of the `Not Found` error the file is being renamed in the USS tree so if clicked on after receiving the message and the rename it will then open.

Thanks @JillieBeanSim for checking over the path fixes! Regarding the additional issues you saw:

- For 1, I believe this is pre-existing behavior in the current release and unrelated to this PR. It just depends how many folders/subfolders you have open when you hide profiles. I think it is running `getChildren` as part of the hide, which results in “get USS File list...” messages for the open folders that are getting refreshed after the profile is hidden.

- For 2 and 3 (which are tied to the same problem), thank you for catching this case! It appears this was happening because you tested by favoriting a subdirectory (not a direct child of the session node). In our current (`master` branch) code, when we search to see if there is an equivalent non-favorite node (i.e. `findNonFavoritedNode` in `USSTree.ts`), we were only looking in the session node's first level of children. (This means this issue actually also exists in 1.13.1.)  However, this use case seems like it should be fairly common, and USS rename does feel broken without it working, so I pushed a fix for it (at the risk of scope creep for this PR 😅). I also added/updated unit tests for the revised code.

chladnefazole:
Hello, so the functionality works great! Does exactly what it says on the tin.

I tried renaming a file and a folder quickly after each other...it seems to create a lot of popups. Doesn't break anything but is a bit excessive. See the gif...also when you just rename one file/folder, it creates the popup like 3 times, but it's really obvious if you do it twice.

renameDoesntCollapse

**All of those popups are from renaming only 2 documents.**

(In the video, it looks like I renamed 3 documents in succession. But I only renamed 2, and then didn't do the 3rd, because after 2 the effect was already clear enough.)

lauren-li:
@chladnefazole Thanks for looking at this PR! I agree there a lot of pop-ups that result from this. Some of them are directly related to the refresh implementation (and are needed for renaming to sync between Favorites and session nodes), but I think some of them have to do with the opening of the USS file. I've noticed in our current 1.13.1 release, opening a USS file will also trigger a series of “get USS File list...” messages. It seems that the more folders the user has expanded when a USS file is opened, the more “get USS File list...” messages pop up.

Below is an example with 1.13.1. At first I open the file `abc` with only its parent folder expanded. Then, I close the file, expand a couple of additional folders, and then try to open `abc` again. A significant additional number of “get USS File list...” messages pop up the second time `abc` is opened. This might be something we want to look at separately with the operation to open USS files:
ze-openRefresh

katelynienaber (approved):
LGTM :)

JillieBeanSim:
Thanks @lauren-li for fixing prior issues, I do see one error pop up with renaming a favorited folder while open.  You don't see it if the folder is closed or if renaming the same folder within the session tree open or closed.  It looks like getChildren is being called and old dir name is probably being sent to get the 404 Not Found error.
ZE-rename-USS-2

JillieBeanSim (approved):
@lauren-li this looks good now. Thanks for your quick response on the errors/issues I ran into and your hard work on this PR.

lauren-li:
Thanks @katelynienaber for creating issue #1272 to address the excessive pop up messages!

Thanks @JillieBeanSim for catching the case with renaming an open Favorited folder. I have implemented your suggested fix to add an await allowing the favorite rename to finish before refreshElement is called on it. I really appreciate your troubleshooting help on this one!

katelynienaber (approved):
LGTM :)

PR #1271, opened 2021-04-20
Add connection buttons test fixes

katelynienaber:
@VitGottwald I just went back to your 3rd commit (8214407) and fixed the tests from there.

Because I don't know what you fixed in the code later but this is only changing test files which were affected by the original feature.

Pls lmk if I should change anything!


VitGottwald:
Thank you for looking into this @katelynienaber !

VitGottwald (approved):
Thank you @katelynienaber !

Issue #1124, opened 2021-01-08
When creating a new profile in ZE, I am prompted for port twice


MikeBauerCA:
When creating a new profile in ZE, I receive both of these prompts:
![image](https://user-images.githubusercontent.com/35504792/104058685-24edc680-51c2-11eb-8626-1b2eb08a6cfa.png)
![image](https://user-images.githubusercontent.com/35504792/104058704-2b7c3e00-51c2-11eb-9c16-8e90202bc4b7.png)

ZE should only prompt for port one time.

OS: Windows 10 & Ubuntu 16.04
Zowe Explorer Version: 1.11.1

zdmullen:
If you enter a url with a host and port, you won't get the subsequent port dialog. That said, this seems to be a carryover from mirroring the behavior in the CLI. We'll talk in the squad about whether to remove the second one or not, because we would have to change the validation for the field I suppose (though perhaps that isn't much of an issue).

jellypuno:
Adding to what @zdmullen said, the port dialog box is prompted just in case you forgot  to put the port in the URL dialog box. It still gives you that chance to put the port instead of going to edit. In my opinion, we shouldn't remove this port dialog box because it only appears if you failed to put a port. It makes sense to be prompted again because your profile will be incomplete.

MikeBauerCA:
> If you enter a url with a host and port, you won't get the subsequent port dialog.

@zdmullen, that is not how it is working for me. 

I enter port 443 with the host and still get prompted for port again during initial profile creation (update doesn't share this issue; during update I only see it as a host:port combo).

I created the gif below to summarize my profile creation/update experience with Zowe Explorer:
![ZoweExplorerProfileCreationAndUpdate](https://user-images.githubusercontent.com/35504792/104194989-0ec54d80-53f0-11eb-93f3-9aa230c038df.gif)

jellypuno:
now, I understand. It is because you are using port 443. Since we are using NodeJS URL function to handle the URL information, it converts port 443 to 0 that's why the port dialog box is displayed again. I cannot explain it but for some reason NodeJS is having a lot of validation/issues with 443. 

Reference: https://nodejs.org/api/url.html#url_url_port
```
// Default ports are automatically transformed to the empty string
// (HTTPS protocol's default port is 443)
myURL.port = '443';
console.log(myURL.port);
// Prints the empty string
console.log(myURL.href);
// Prints https://example.org/
```

[...]

PR #1298, opened 2021-05-03
Issue #1124: Remove port prompt

katelynienaber:
## Proposed changes

This is a fix for the double port prompting mentioned [here](https://github.com/zowe/vscode-extension-for-zowe/issues/1124).

Now, there is no prompt for port alone. You should enter your port at the end of the hostname, in the prior input box.

## Release Notes

Removed second prompt for port during profile creation & editing.

Milestone: Backlog

Changelog: Removed second prompt for port during profile creation & editing.

## Types of changes

## Checklist

## Further comments


packages/zowe-explorer/src/Profiles.ts
```diff
@@ -582,20 +567,7 @@ export class Profiles extends ProfilesCache {
                     }
                     break;
                 case "port":
-                    if (schemaValues[value] === undefined) {
```
nickImbirev, line 585: Maybe we change our dialog validation instead?

Now, I can specify only a hostname in the dialogs, without any port and everything will be fine, until the whole profile will be created. And I can get an error during profile validation (which is not really clear in some cases).

We have `validateAndParseUrl` method, which we can use to cover the bug (or a feature) inside the Node js URL constructor. Maybe we can add an additional check there? Like, if the protocol is equal to `https`, and the port is `0`, than we can assume, the port is `443` - default one. We can cover it by some unit tests and it will not be a surprise.

What do you think?

nickImbirev: There is also a super related issue (about `443` port and `https` protocol combination). I think, proper validation will fix it as well.

How to reproduce: once I specified a `443` port and a `https` protocol during profile creation, I do not see the correct profile value during profile **updating**.

katelynienaber: Hey, is there an issue already open for this 443 port/https issue? If so could you post the link here? Then we could close both with this PR possibly.

packages/zowe-explorer/src/Profiles.ts
```diff
@@ -401,20 +400,7 @@ export class Profiles extends ProfilesCache {
...
-                                    "editConnection.undefined.port",
-                                    "Invalid Port number provided or operation was cancelled"
-                                )
-                            );
-                            return undefined;
-                        }
-                        updSchemaValues[value] = updPort;
-                        break;
-                    }
+                    // Do nothing, the user should put in the port at the end of the hostname
```
jellypuno, line 403: please bring the prompt for the port back. The user should have the ability to put the port if they didn't put it in the host

packages/zowe-explorer/src/Profiles.ts
```diff
@@ -582,20 +567,7 @@ export class Profiles extends ProfilesCache {
...
-                                    "createNewConnection.undefined.port",
-                                    "Invalid Port number provided or operation was cancelled"
-                                )
-                            );
-                            return undefined;
-                        }
-                        schemaValues[value] = newPort;
-                        break;
-                    }
+                    // Do nothing, the user should put in the port at the end of the hostname
```
jellypuno, line 570: same here

packages/zowe-explorer/src/Profiles.ts
```diff
@@ -1392,30 +1364,6 @@ export class Profiles extends ProfilesCache {
         });
     }
 
-    private async portInfo(input: string, schema: {}) {
```
jellypuno, line 1395: also here.

jellypuno (changes requested):
@katelynienaber The request is to remove the double prompt and not remove the port prompt completely. Please put it back. 

You can reference this code to fix the double prompt: https://github.com/zowe/vscode-extension-for-zowe/blob/master/packages/zowe-explorer/src/Profiles.ts#L399

This only happens if the port is 443

packages/zowe-explorer/src/Profiles.ts
katelynienaber: I disagree with this. Showing the user that they can include the port in the URL, and then asking them for the port specifically, amounts to asking them twice for the port. Putting the port at the end of the URL seems like basic knowledge to me, everyone will understand how to do this, it's not like we are being unclear somehow about how to specify your port.

I would prefer to ask the team before putting it back the way it was.

packages/zowe-explorer/src/Profiles.ts
nickImbirev: Nope, I just played with it and identified :)

packages/zowe-explorer/src/Profiles.ts
nickImbirev: Maybe I can propose to add some validation of user value as a compromise? 

Like to force user to include port in the first dialog and disallow them to continue profile creation without it? Like a runtime validation?

packages/zowe-explorer-api/src/profiles/ProfilesCache.ts
```diff
@@ -134,7 +134,12 @@ export class ProfilesCache {
             return validationResult;
         }
 
-        validationResult.port = Number(url.port);
+        if (url.protocol.includes("https") && url.port === "") {
```
nickImbirev, line 137: I would propose to add some unit tests for this function to cover this 'weird' behaviour.

What do you think about it? The tests will be pretty easy to write, since this function is pure.

And second point: maybe you can extract these magic numbers and strings into some understandable constants? Because, now, I understand, why these constants are here, but eventually, we will forget about it and why we need such a weird `443` and `https` here.
Does it make sense?

nickImbirev: Maybe we can check for the port value equal to 0?

Because now you are checking number for equality with empty string... Why?...

I think unit tests may cover such things..

packages/zowe-explorer/src/Profiles.ts
katelynienaber: What if they don't want to add the port? It seems weird to force them to put in a zero, we should be able to tell this if they simply don't add anything. And if they decide later they want to add the port they can just go through the edit dialog again.

packages/zowe-explorer-api/src/profiles/ProfilesCache.ts
katelynienaber: I will add some unit tests. I am checking against an empty string because that's what is stored in url.port if the user doesn't enter anything. I'll check what happens if they enter 0.

packages/zowe-explorer/src/Profiles.ts
nickImbirev: Hmm, but they have to specify it, right?

And we have this updated validation function now, that can handle default port values, so even if they will not specify anything, but protocol will be `https` - everything will be fine.

packages/zowe-explorer-api/src/profiles/ProfilesCache.ts
nickImbirev: OMG, they store empty string in number field?.... I mean, port can be only number, as far as I know...

packages/zowe-explorer/src/Profiles.ts
jellypuno: @katelynienaber let's talk offline

katelynienaber: With the updated validation function, they don't have to specify the port, we will use 443 as default if it isn't specified.

So, since they don't have to specify it, if they don't we should just deal with it, not give them some validation error imho

packages/zowe-explorer-api/src/profiles/ProfilesCache.ts
JillieBeanSim: wouldn't it be better to just check for `!url.port` which would cover null and undefined?

packages/zowe-explorer/src/Profiles.ts
JillieBeanSim: Why would you force the user to enter the port or the protocol in the hostname textbox?  We changed this before to allow the user to have both options.  If it's entered as a whole URL with protocol, hostname, and port this entry should be split up into appropriate pieces and if a piece was missing a following prompt would show up to get the missing information.  The point of this issue is to not re-prompt if it was already entered.

JillieBeanSim: I just tested PR and only entered hostname and was never prompted for port while creating an RSE profile and it automatically assigned 443 which would be incorrect
image

JillieBeanSim:
@katelynienaber @jellypuno I have a super simple fix if the issue was just with the port being 443 here on this branch https://github.com/zowe/vscode-extension-for-zowe/tree/no-reprompt-port-work if you want to look at it

jellypuno:
@katelynienaber Thank you for fixes. Works well for me. Although there are 2 things that I've noticed:
1. It didn't ask me to input my port if my hostname ends in 443. It assumes that my port is 443.
url443

2. If I just put 443 and didn't put any hostname then it will automatically provide me a hostname. Although I don't think this is our problem.
image

packages/zowe-explorer/src/Profiles.ts
katelynienaber: @nickImbirev @jellypuno @JillieBeanSim Please check this again. I wasn't aware it had been discussed previously.

When creating a new profile or editing an old one, it should now...

Prompt you for port specifically if you enter after the URL:
- Nothing
- 0
- Something that is NaN

It will not prompt you for port if you enter:
- 443
- Any other number

katelynienaber:
@jellypuno lol, first issue. This one should be fixed.

For the 2nd one, what would the expected behavior be here?

jellypuno:
Thank you for adding the fix. 

For the 2nd one, no fix.  I don't think it is an issue. I just noticed it. So let's ignore it for now.

Lastly, please execute pre-publish. There are updates in the i18n file that was not included

packages/zowe-explorer-api/src/profiles/ProfilesCache.ts
```diff
@@ -137,7 +138,12 @@ export class ProfilesCache {
             return validationResult;
         }
 
-        validationResult.port = Number(url.port);
+        if (newUrl.includes(":443")) {
```
nickImbirev, line 141: Maybe we can reuse our new `DEFAULT_PORT` constant here?

I mean, it is super weird string here right now, I'm not sure, that the meaning of it will be remembered...

nickImbirev: Can we add a unit test(s) for such a behaviour?

nickImbirev (approved):

jellypuno (approved):
Functionality works well for me

nickImbirev (approved):

JillieBeanSim (approved):
LGTM @katelynienaber thank you for your work on this and for addressing the issues that came up.

katelynienaber:
@crawr can we merge this? :)

Issue #1266, opened 2021-04-09
Issue STOP command  fails with Cannot read property 'trim' of undefined TypeError: Cannot read property 'trim' of undefined


VitekVlcek-Broadcom:
**Describe the bug**

When using `Issue STOP command` in zowe vscode extension 
I get `Cannot read property 'trim' of undefined TypeError: Cannot read property 'trim' of undefined`

**To Reproduce**

1. Submit job
2. show it in exporters and click `Issue STOP command` in right click menu

**Expected behavior**

Job stops

**Desktop (please complete the following information):**

- OS:
- Zowe Explorer Version:  v1.13.1
- (Optional) Zowe CLI Version: ??
- (Optional) Are you using Secure Credential Store?  I don't know

**Additional context**

@crawr was able to recreate issue

PR #1304, opened 2021-05-07
Issue #1266: stop job command error

nickImbirev:
## Proposed changes

Issue description and steps to reproduce: https://github.com/zowe/vscode-extension-for-zowe/issues/1266
Issue cause: **incorrect refresh session behaviour**.

Detailed explanation: 
AS IS:
Session is updated manually during refresh, via `Object[prop]` solution. As a result, session  was not built correctly after refresh (some default values were not used).

TO BE (implemented in the PR):
Session is updated via provided API: `ZoweExplorerApiRegister.get###Api(profileValue).getSession()`. Now, session object is correct and contains default values and applied strategies.

Additional improvements in the PR:
Util refresh function refactoring and jsdoc update accordingly.

## Types of changes

## Checklist

- [x] All checks have passed (DCO, Jenkins and Code Coverage) - excluding Theia tests


packages/zowe-explorer/src/utils/ProfilesUtils.ts
```diff
@@ -106,25 +107,34 @@ export function isTheia(): boolean {
...
-export async function refreshTree(sessNode: IZoweTreeNode) {
-    const profileType = sessNode.getProfile().type;
-    const allProf = Profiles.getInstance().getProfiles(profileType);
-    const baseProf = await Profiles.getInstance().getBaseProfile();
-    for (const profNode of allProf) {
-        if (sessNode.getProfileName() === profNode.name) {
-            setProfile(sessNode, profNode.profile);
-            const combinedSessionProfile = (await Profiles.getInstance().getCombinedProfile(profNode, baseProf))
-                .profile;
-            setSession(sessNode, combinedSessionProfile);
```
nickImbirev, line 123: The problem was here: `setSession(sessNode, combinedSessionProfile)`, where we manually updated session values, ignoring provided API to build session properly.

packages/zowe-explorer/src/uss/ZoweUSSNode.ts
```diff
@@ -171,7 +171,9 @@ export class ZoweUSSNode extends ZoweTreeNode implements IZoweUSSTreeNode {
                 this.label,
                 localize("getChildren.error.response", "Retrieving response from ") + `uss-file-list`
             );
-            await refreshTree(sessNode);
+            await syncSessionNode(Profiles.getInstance())((profileValue) =>
+                ZoweExplorerApiRegister.getUssApi(profileValue).getSession()
```
katelynienaber, line 175: I am leery of passing `getSession` in as a param. We have some circular dependency problem in ZE weith the API functions. The API imports functions from throughout ZE, so if ZE has to import API functions, we might get random test failures (this happened when we added base profile support awhile back).

This seems like a structural issue that we should really deal with at some point. Not in this PR or anything. But you've reminded me of it. I'll bring it up at the next standup probably...

katelynienaber (dismissed):
LGTM! Thanks for changing function name :)

packages/zowe-explorer/src/uss/ZoweUSSNode.ts
nickImbirev: I do not see any circular dependency issues here, because we are passing function as an argument, it is not a class or a module, it has no context.

But you are right, with passing classes and objects, there might be such a problem :(

packages/zowe-explorer/__tests__/__unit__/shared/utils.unit.test.ts
```diff
@@ -104,25 +104,100 @@ describe("Shared Utils Unit Tests - Function node.labelRefresh()", () => {
...
+        undefined,
+        undefined,
+        undefined,
+        {
+            profile: serviceProfile,
+            type: serviceProfileType,
+            message: "",
+            failNotFound: true,
+        }
+    );
```
crawr, line 129: I remember moving away from defining test items in this way and using the global mocks instead for more consistency. 
I'm fine with it personally in this unit test since it is a small function, but I would love to hear @katelynienaber 's thoughts on it.

crawr (commented):
Functionality works well for me. I tested it with SSO, optional credentials, and regular zosmf and zftp profile types. Thanks @nickImbirev 

I saw some audit issues but we'll need to discuss them as a squad. Perhaps we can create a separate PR to fix them in master so all the other PRs can just merge it in.

packages/zowe-explorer/__tests__/__unit__/shared/utils.unit.test.ts
katelynienaber: Yes, thank you for noticing this :D

The goal is to remove these big, repetitive mock declarations all over the place. You can find some mock creation functions in folder `zowe-explorer/__mocks__/mockCreators`, ex. I would use this one here: https://github.com/zowe/vscode-extension-for-zowe/blob/00f2ec21e95713bb2bfa2311fb23c6e4c4d3f904/packages/zowe-explorer/__mocks__/mockCreators/datasets.ts#L19

You can ofc change individual properties on your node after you use the mock creation function, but using it will make the declaration a one-liner.

Also you can check in the `createGlobalMocks` or `createBlockMocks` functions, to see if some session node has already been created there for you to use. The mocks created in those functions are recreated after every it() block, so you can be sure you won't affect later tests by changing the properties of any nodes declared there.

nickImbirev: Thank you for this point @katelynienaber , I will take a look and try to reuse already defined mock nodes here.

crawr (dismissed):
LGTM! Thanks for addressing my comment

katelynienaber (approved):
LGTMT
looks good to me too
legitimate
:D

jellypuno:
@nickImbirev Can you check the Theia errors, please?

image

nickImbirev:
> @nickImbirev Can you check the Theia errors, please?
> 
> image

Yep, you are right. I was not prepared for such a rule, even for it to be applied during Theia tests only... I fixed this, thank you for pointing it!

crawr (approved):

jellypuno (approved):

Issue #1061, opened 2020-11-06
Trailing slash in USS filter causes 500 error code


ChrisBoehmCA:
**Describe the bug**

Entering a USS filter with a trailing slash fails -- USS files in that directory can't be listed. 

The error looks like: 
Retrieving response from uss-file-list Error: z/OSMF REST API Error: Rest API failure with HTTP(S) status 500 category: 1 rc: 4 reason: 6 message: Path name is not valid

**To Reproduce**

Go to the USS section of the Zowe explorer menu. Select an active profile and click the filter (magnifying glass) button. 

Enter a valid, existing USS path without the trailing slash. For example, `/u/users/ibmuser`. It should display the list of files correctly (it doesn't matter if you do this first, it's just for comparison). Now enter the same path with a trailing slash at the end -- for example, `/u/users/ibmuser/`. All of the z/OSMF instances I have tried have had the same issue, so it doesn't seem system-specific.

**Expected behavior**

If the trailing slash causes an error in z/OSMF for some reason, maybe it should be stripped off before the API is called. I have a vague memory that there used to be code to handle this, but it might have been lost in a refactor. 

**Desktop (please complete the following information):**

- OS: Windows 10 
- Zowe Explorer Version: 1.10.1
- (Optional) Zowe CLI Version: 6.24.2
- (Optional) Are you using Secure Credential Store? No

PR #1313, opened 2021-05-12
Fix trailing slash issue in USS search

katelynienaber:
## Proposed changes

Quick fix for this issue: https://github.com/zowe/vscode-extension-for-zowe/issues/1061

## Release Notes

Milestone: 1.15.0 (?)

Changelog: Fixed USS search issue with trailing slash.

## Types of changes

## Checklist


nickImbirev (dismissed):

nickImbirev:
Maybe we can even remove 'duplicated' filters from already existed filters?

I mean, now, once I tried two practically the same filters: `/u/users` and `/u/users/`, they both appears in my quickPick list.

Screenshot 2021-05-12 at 14 11 43

I see, that we do the same thing for filters with a lot of trailing slashes, so for example, if I will use filter: `/u/users////` and `/u/users`, there will be only one filter saved and showed in quickPick. Maybe we can do the same here?

What do you think?

packages/zowe-explorer/src/uss/USSTree.ts
```diff
@@ -622,7 +622,7 @@ export class USSTree extends ZoweTreeProvider implements IZoweTree<IZoweUSSTreeN
                 }
             }
             // Sanitization: Replace multiple preceding forward slashes with just one forward slash
-            const sanitizedPath = remotepath.replace(/\/\/+/, "/");
+            const sanitizedPath = remotepath.replace(/\/\/+/, "/").replace(/\/+$/, "");
```
nickImbirev, line 625: Maybe we can even remove 'duplicated' filters from already existed filters?

I mean, now, once I tried two practically the same filters: `/u/users` and `/u/users/`, they both appears in my quickPick list.

Screenshot 2021-05-12 at 14 11 43

I see, that we do the same thing for filters with a lot of trailing slashes, so for example, if I will use filter: `/u/users////` and `/u/users`, there will be only one filter saved and showed in quickPick. Maybe we can do the same here?

What do you think?

jellypuno: @katelynienaber Was this captured or will this be a different issue that we should open and implement in the future?

katelynienaber: I haven't fixed it yet but I will throw it in there because it seems small.

katelynienaber: @nickImbirev @jellypuno Please check this again...

Now, any duplicate slashes in the path are reduced to 1 (except at the end of the path, where they are removed).

I didn't write any code to remove previously-used filters, if they are duplicates in the list except for the number of slashes. There is a max limit on filters in the list, so as people use more, the old ones will be removed.

But I also made a global variable for the default max number of filters and whatnot. So now you can check those numbers in the globals file.

packages/zowe-explorer/src/PersistentFilters.ts
```diff
@@ -10,7 +10,7 @@
  */
 
 import * as vscode from "vscode";
-import * as zowe from "@zowe/cli";
```
jellypuno, line 13: Why is Zowe CLI removed? Is it not needed anymore?

packages/zowe-explorer/src/globals.ts
```diff
@@ -30,6 +30,8 @@ export let DS_DIR;
 export let ISTHEIA: boolean = false; // set during activate
 export let LOG: Logger;
 export const COMMAND_COUNT = 93;
+export const MAX_SEARCH_HISTORY = 5;
```
nickImbirev, line 33: Do we need to specify, that these constants are used only in 'uss' tree? Or maybe we can rename/replace them somewhere?

packages/zowe-explorer/src/uss/USSTree.ts
```diff
@@ -622,7 +622,7 @@ export class USSTree extends ZoweTreeProvider implements IZoweTree<IZoweUSSTreeN
                 }
             }
             // Sanitization: Replace multiple preceding forward slashes with just one forward slash
```
nickImbirev, line 624: I would update the comment here, because now it is outdated and may be confusing. 

I mean, now we are basically removing all the slashes from the path end, right?

nickImbirev (approved):

packages/zowe-explorer/src/globals.ts
katelynienaber: They aren't just used in USS tree, they are used for any list of previous searches (also DS and jobs trees have search capability)

And MAX_FILE_HISTORY is also in use for DS tree, you can see recently-opened data sets & members.

nickImbirev: Hmm, this is weird, I thought you just add these constants.. okay :)

nickImbirev (approved):

JillieBeanSim (approved):
LGTM, all expected tests and checks are passing as well.  Thanks for your work on this @katelynienaber

Issue #1263, opened 2021-04-07
Standardize Zowe Explorer configuration property names to make them consistent with VS Code's format


lauren-li:
**Is your feature request related to a problem? Please describe.**
Currently, Zowe Explorer's configuration property names use a variety of formats, such as kebab-case (e.g. `"Zowe-DS-Persistent"`), camel case (e.g. `"searchHistory"`), and title case (e.g. `"Zowe Security: Credential Key"`). This provides an inconsistent settings file interface for Zowe Explorer users, and also does not conform to how VS Code formats its own configuration property names (camel case). Although Zowe Explorer settings display like other extension settings in VS Code, they appear oddly when Zowe Explorer is run in Theia, like a long list of separate categories:

<img width="293" alt="Screen Shot 2021-04-07 at 5 49 07 PM" src="https://user-images.githubusercontent.com/45975633/113939939-a4436400-97ca-11eb-9aba-312407ef1e5e.png">

**Describe the solution you'd like**

VS Code formats its configuration property names using a combination of periods (to denote hierarchical categories) and camel case as documented in the Properties subsection of the [Configuration Schema section of their API docs](https://code.visualstudio.com/api/references/contribution-points#Configuration-schema).

For example:
<img width="440" alt="Screen Shot 2021-04-07 at 5 34 29 PM" src="https://user-images.githubusercontent.com/45975633/113939014-61cd5780-97c9-11eb-81d0-3a62ae1b8f47.png">

Proposal:
- Review all Zowe Explorer configuration property names and make them consistent with VS Code's settings format for Zowe Explorer 2.0.
- Provide a way to convert Zowe Explorer settings that use the current format into the new standardized format (in order to make the user's transition smoother).
- Add standardization of configuration property names to the conformance criteria.
- Store the `"Zowe-Temp-Folder-Location"` setting as a string rather than an object.

lauren-li:
Possible impact on/from issue #820 (Zowe Security: Credential Key is confusing).

rudyflores:
Hi everyone I wanted to provide my proposal for the standardized configuration names that will be applied with my upcoming PR for issue #1263. I would like to hear everyone's input on my current proposed names so I can make any changes if someone finds a more optimal way of keeping the new names as consistent as possible.

- zowe.ds.default.binary
- zowe.ds.default.c
- zowe.ds.default.classic
- zowe.ds.default.pds
- zowe.ds.default.ps
- zowe.ds.history
[...]

PR #1347, opened 2021-06-07
Support multiple VS Code windows for files opened via Zowe Explorer

JillieBeanSim:
## Proposed changes

Note to testers - run integration clean up on master branch before moving to this branch and run integration setup on this branch.  I changed AllocLike integration test to fix other broken tests on master so it will remove it's data set after test vs cleanup script doing it.

Note # 2 This PR will need to be built, packaged, and installed for `Open New Window` to be tested.  May need to clean local cache of old install.

What Changed:
Removed the call to deactivate during activation and have deactivate check for the new Setting `zowe.files.temporaryDownloadsFolderCleanup` whose default is true.  When true, deactivate will clean up temp folder as ususal.  When false, deactivate will not remove temp folder or any of it's files.

image

tempFolder-setting-defTrue-final

## Release Notes

Milestone: 1.17.0

Changelog: Zowe Explorer cleans up the folders with file contents downloaded from z/OS during deactivation to not leave source code behind. If you open multiple VS Code windows to edit the same files side-by-side and then close one of these windows it would mean that file content gets deleted even if the file is still being edited in the other window. To prevent this kind of clean-up enable this setting. If you want to ensure that no source is left behind on a development machine then keep this setting disabled and only use one VS Code window at a time.

## Types of changes

## Checklist

## Further comments


packages/zowe-explorer/src/extension.ts
```diff
@@ -66,16 +66,22 @@ export async function activate(context: vscode.ExtensionContext): Promise<ZoweEx
...
-    // this is to handle if the application crashed on a previous execution and
-    // VSC didn't get a chance to call our deactivate to cleanup.
-    await deactivate();
-
     try {
-        fs.mkdirSync(globals.ZOWETEMPFOLDER);
-        fs.mkdirSync(globals.ZOWE_TMP_FOLDER);
-        fs.mkdirSync(globals.USS_DIR);
-        fs.mkdirSync(globals.DS_DIR);
+        if (!fs.existsSync(globals.ZOWETEMPFOLDER)) {
```
nickImbirev, line 70: Just out of curiosity: why we are using the synchronous API to work with the file system?

Just a suggestion: Maybe we can use the [async API](https://code.visualstudio.com/api/references/vscode-api#FileSystem), which is built in VSCode?

The benefits will be: better user experience, better resources consumption and of course, this VSCode API will work fine in Theia, in Eclipse Che or in some browser VSCode versions with ZE installed. It will also handle potential problems between the operating systems, because it is basically another layer of abstraction above the usual file system API.

JillieBeanSim:
All integration tests are fixed here except the Delete failure which is being handled in a different PR

packages/zowe-explorer/src/extension.ts
```diff
@@ -66,16 +66,13 @@ export async function activate(context: vscode.ExtensionContext): Promise<ZoweEx
     // Determine the runtime framework to support special behavior for Theia
     globals.defineGlobals(preferencesTempPath);
 
-    // Call cleanTempDir before continuing
-    // this is to handle if the application crashed on a previous execution and
-    // VSC didn't get a chance to call our deactivate to cleanup.
-    await deactivate();
```
phaumer, line 72: Oh wow, that was a bad idea. Why not call cleanup directly? What if others would have added more code to the deactiviate function that would have broken activation?

phaumer (changes requested):
We can merge this with the requested changes, but I am wondering if our users would expect this to work automatically using reference counters?

packages/zowe-explorer/src/extension.ts
JillieBeanSim: @phaumer should I add back a call to cleanup?  calling cleanup during activation is what was causing the deletion of a file right when new window was opened.  With this PR that is not done now but if the new setting is set to false (default) once one of the new windows closes `deleted` will show in other open windows if a ZE file is opened.  If set to true, that doesn't happen since no cleanup is done on temp folder.

packages/zowe-explorer/package.nls.json
```diff
@@ -73,6 +73,7 @@
   "Zowe-Default-Commands-Persistent-Favorites": "Toggle if Commands persist locally",
   "Zowe-Environment": "Environment where the extension is running, default is VSCode",
   "Zowe-Automatic-Validation": "Allow automatic validation of profiles.",
+  "Zowe-Disable-TempFolder-Cleanup": "Disable the temporary folder cleanup during deactivation.",
```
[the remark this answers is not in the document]
lauren-li: @phaumer @JillieBeanSim I think if we want to go with the sub-groups, it's fine to start doing that in this PR for the setting you are adding. We can go ahead and use `zowe.files.disableTemporaryDownloadsFolderCleanup` for this PR.

For other existing settings that would need to be migrated, I would leave those for @rudyflores' WIP that will handle the settings migration.

packages/zowe-explorer/package.json
```diff
@@ -1652,6 +1652,12 @@
           },
           "description": "%Zowe-Environment%",
           "scope": "window"
+        },
+        "zowe.disableTemporaryDownloadFolderCleanup": {
+          "type": "boolean",
+          "default": false,
+          "description": "%zowe.disableTemporaryDownloadFolderCleanup%",
```
lauren-li, line 1659: @JillieBeanSim Would it be less confusing for users if we labeled this setting in the positive? Saying "Disable Temporary Downloads Folder Cleanup" forces the user to think in double negatives when setting this to `false`, i.e. "I am not disabling `x`, so therefore I am enabling `x`".

Perhaps we could use one of the following:
- "Enable Temporary Downloads Folder Cleanup" (true/false)
- "Temporary Downloads Folder Cleanup"
  - Settings description: "Enable temporary downloads folder cleanup" (true/false).
  
I remember we had a similar discussion about double negatives when we added the global enable/disable profile validation setting, and this is what we ended up going with:

Screen Shot 2021-06-29 at 9 34 06 AM

JillieBeanSim: @lauren-li so we don't want to include enable or disable in the label, just an open ended label, and add Allow to the description like with validation?

JillieBeanSim: `zowe.files.temporaryDownloadsFolderCleanup`: `Allow deletion of downloaded file content at editor deactivation (closing a Window) to support using separate VS Code windows editing the same files.`

JillieBeanSim: image

JillieBeanSim: Sorry @lauren-li , the above examples are all wrong.

What we will need to do if we follow the validation example is change to this with default true and allow the user to disable it by unchecking the box.
image

lauren-li: @JillieBeanSim I would rephrase the setting description to:
```
Allow deletion of downloaded file content during deactivation (closing a window). Disable this setting to support using separate VS Code windows editing the same files.
```
Do we want to add a note that disabling the setting means the temporary files will persist after the window is closed?

Actually, would it be clearer if we renamed the setting to `Persist temporary downloads`? (The default would have to be set to `false` again, though.)

packages/zowe-explorer/src/extension.ts
phaumer: No, sorry, I was just surprised that we had that code. Should have not commented.

JillieBeanSim:
as discussed, we will follow the validation example and change to this with default true and allow the user to disable it by unchecking the box.
image

I will update the PR description to match these changes

packages/zowe-explorer/src/utils/TempFolder.ts
```diff
@@ -104,12 +105,16 @@ export function cleanDir(directory) {
  * @export
  */
 export async function cleanTempDir() {
+    // Get temp folder location from settings
```
lauren-li, line 108: @JillieBeanSim The comment in this line doesn’t seem to describe what is actually happening in the block below it. Should it maybe say `// Get temp folder cleanup preference from settings`?

lauren-li (dismissed):
I think this PR looks good. All unit and expected integration tests are passing. The new setting description makes clear the benefit of enabling versus disabling the temp folder cleanup, and toggling the setting on/off works as expected in manual testing.

The only (minor) question I had was about a comment (see below), but I don't think it's a blocker to this PR. I don't mind re-approving if edits are made to the comment.

Thanks @JillieBeanSim for adding this feature!

lauren-li (dismissed):
This looks good to me now. Thanks @JillieBeanSim for addressing my comment!

phaumer:
Should this comment still be addressed? https://github.com/zowe/vscode-extension-for-zowe/pull/1347#discussion_r647182685

JillieBeanSim:
> Should this comment still be addressed? [#1347 (comment)](https://github.com/zowe/vscode-extension-for-zowe/pull/1347#discussion_r647182685)

@phaumer @nickImbirev when I tested with the VS Code APIs there was some misbehaviors, it may be why we are using the synchronous API to work with the file system.  Could be something to look further into as technical debt.

lauren-li:
@JillieBeanSim @crawr @phaumer This PR refers to the temp folder as `TemporaryDownloadsFolder`, whereas PR #1373 refers to it as `TempFolder`. We should pick one way of referring to the temp folder and be consistent with using it.

This will also affect how @rudyflores renames the `Zowe-Temp-Folder-Location` setting for his upcoming PR.

rudyflores:
After posting in the issue thread for #1263 the proposal for standardized names, it was decided to change the current `zowe.files.temporaryDownloadsFolderCleanup` to the option proposed by @VitGottwald: `zowe.files.temporaryDownloadsFolder.cleanup` in order to keep the configuration names consistent.

Could you change the current name to the one that Vit proposed @JillieBeanSim ?

jellypuno (dismissed):
This works really well! Thank  you @JillieBeanSim

JillieBeanSim:
Please re-review @lauren-li & @jellypuno.  I addressed comment https://github.com/zowe/vscode-extension-for-zowe/pull/1347#issuecomment-876724996

JillieBeanSim:
This is what the settings look like now with the changes
image

jellypuno (approved):

lauren-li (approved):
This PR looks good to me. All unit and expected integration tests are passing, and manual testing still works well with the renamed setting. Thanks @JillieBeanSim for addressing all the comments!

Issue #791, opened 2020-05-12
Provide better error message for deleted profile in favorites


crawr:
> The only thing I am not sure is this part of the error message:
"To resolve this, you can create a profile with this name, or remove the favorites ..."
>
> If the user creates another profile with the same name but points to a different place, it can possibly bring confusion. Wouldn't it be just safer to leave out this part of the message and keep just the advice to drop the missing favorite item...?

_Originally posted by @zimlu02 in https://github.com/zowe/vscode-extension-for-zowe/issues/455#issuecomment-624783184_

zFernand0:
@jelaplan any thoughts?

jelaplan:
@zFernand0 I'm not sure I fully understand what is happening. I do realize people are trying to use a favorite that relies on a delete profile and they get an error message. The suggestion to just advise to delete the favorite without telling them to repair their connection profiles is ok. But the reason is that it is simple to understand and doesn't present complicated advise about repairing things.  It seems farfetched that someone would create a connection with the same name and a different LPAR. A radical idea would be to delete any dependent favorites when deleting a profile and get out of the business of creating elaborate error messages :)

lauren-li:
Now that the UX flow for removing favorited profiles has been updated, this error message can also be updated to just advise the user to use Zowe Explorer's UI to remove the deleted profile from the Favorites section.

(Full current error message, for reference):
<img width="455" alt="Screen Shot 2021-01-13 at 4 46 18 PM" src="https://user-images.githubusercontent.com/45975633/104514482-53ebaa00-55bf-11eb-875b-ee6c485735f9.png">

PR #1349, opened 2021-06-07
Error string fix for profile not found in favorites

rudyflores:

## Proposed changes

Corrections for error message string displayed upon triggering an error when trying to access a profile on the favorites folder for datasets, USS, and jobs. #791 

Screen Shot 2021-06-08 at 1 49 46 PM

## Release Notes

Milestone: Release v1.16.0
Changelog: correction to error message string when accessing a profile in the favorites section.

## Types of changes

## Checklist


packages/zowe-explorer/__tests__/__unit__/dataset/DatasetTree.unit.test.ts
```diff
@@ -497,6 +497,7 @@ describe("Dataset Tree Unit Tests - Function loadProfilesForFavorites", () => {
         expect(resultFavProfileNode).toEqual(expectedFavProfileNode);
     });
     it("Checking that error is handled if profile not successfully loaded for profile grouping node in Favorites", async () => {
+        jest.restoreAllMocks();
```
lauren-li, line 500: Since we are just concerned with clearing the old data from `showErrorMessageSpy` here, a less heavy-handed way to do this might be to just use [Jest's `mock.mockClear()` function](https://jestjs.io/docs/mock-function-api#mockfnmockclear) immediately after the previous time the spy was used. You can try adding `showErrorMessageSpy.mockClear();` after the `expect` statement on [line 254](https://github.com/zowe/vscode-extension-for-zowe/blob/master/packages/zowe-explorer/__tests__/__unit__/dataset/DatasetTree.unit.test.ts#L254), and removing the `restoreAllMocks` at line 500.

lauren-li (changes requested):
@rudyflores This is looking good so far! The Yes/No buttons for removing the non-existent profile's favorites seem to generally function as expected. However, I did notice that the profile's favorites will still get listed (orange box) when the error appears. This makes it possible for the user to ignore the error message and try to continue interacting with the non-existent profile's favorites, which will lead to more error messages. For example, when I try to expand the favorited PDS `ZOWELKL.ABC`, I get the error in the pink box:

Screen Shot 2021-06-08 at 8 15 17 AM

To avoid this behavior, you might want to add a `return` in the catch statement with your error message.

After you have gotten the behavior working as expected for the Data Sets view, you might also want to update the error messages for the USS and Jobs views, too. :)

packages/zowe-explorer/__tests__/__unit__/job/ZosJobsProvider.unit.test.ts
```diff
@@ -238,11 +238,10 @@ describe("ZosJobsProvider unit tests - Function getChildren", () => {
...
         testTree.mFavorites.push(favProfileNode);
 
-        const loadProfilesForFavoritesSpy = jest.spyOn(testTree, "loadProfilesForFavorites");
-
+        mocked(vscode.window.showErrorMessage).mockResolvedValueOnce({ title: "Yes" });
         await testTree.getChildren(favProfileNode);
-
-        expect(loadProfilesForFavoritesSpy).toHaveBeenCalledWith(log, favProfileNode);
+        expect(showErrorMessageSpy).toBeCalledTimes(1);
+        showErrorMessageSpy.mockClear();
```
lauren-li, line 244: This particular unit test is supposed to check that `getChildren` will call `loadProfilesForFavorites` with the correct parameters given a certain context. It is not supposed to test the error message you added in this PR, and the `loadProfilesForFavorites` function doesn't actually need to be executed. (The fact that it is being executed here has more to do with how the test was originally written (probably by me...) as opposed anything you did.) To avoid unnecessarily running `loadProfilesForFavorites` and hitting the error condition, we can just mock the implementation of `loadProfilesForFavorites` instead of changing what the unit test is checking.

Try putting this test back to the way it was before, and then modifying the `loadProfilesForFavoritesSpy` as follows:

```
       const loadProfilesForFavoritesSpy = jest
            .spyOn(testTree, "loadProfilesForFavorites")
            .mockImplementationOnce(() => Promise.resolve([]));
```

This mock will just pretend that `loadProfilesForFavorites` ran and returned an empty array, so none of the actual code inside `loadProfilesForFavorites` will be executed.

lauren-li (changes requested):
@rudyflores This PR is looking much better now! Thank you for addressing the comments for it. In manual testing, it is working great for all 3 views. I have a couple more suggestions to refine the unit tests and strings (just capitalization and localization).

(Friendly reminder: After updating the strings, remember to run the following command to keep the generated files up-to-date!:
`yarn workspace vscode-extension-for-zowe vscode:prepublish`)

lauren-li (approved):
This PR looks good to me now. All unit and expected integration tests are passing. In manual testing, the text is updated and the added buttons behave as expected.

Thanks @rudyflores for your hard work on this PR!

JillieBeanSim (approved):
LGTM thanks @rudyflores for your work on this.  Functionality looks good and all expected tests and checks are passing.

PR #1359, opened 2021-06-15
Get account and other information from tso profile, if present.

fswarbrick:
Signed-off-by: Frank Swarbrick <fswarbrick@gmail.com>

## Proposed changes

Currently only the account number is pulled from the tso profile, if there is one.  Since there is (currently) no interface asking the user to enter other parameters (regionSize et al), there is no way for said parameter(s) to be passed as part of the TSO command request.

This change will pull all of the TSO profile parameters, if they exist, and pass them as part of the TSO command request.

## Release Notes

Milestone:

Changelog:

When issuing a TSO command, pass all parameters from TSO profile (if one exists), not just the account parameter.

## Types of changes

## Checklist

### DIFF
--- a/packages/zowe-explorer-api/src/profiles/ZoweExplorerApi.ts
+++ b/packages/zowe-explorer-api/src/profiles/ZoweExplorerApi.ts
@@ -440,7 +440,11 @@ export namespace ZoweExplorerApi {
          * @returns {zowe.IIssueResponse>}
          * @memberof ICommand
          */
-        issueTsoCommand?(command: string, acctNum?: string): Promise<zowe.IIssueResponse>;
+        issueTsoCommand?(
+            command: string,
+            acctNum?: string,
+            startParams?: zowe.IStartTsoParms
+        ): Promise<zowe.IIssueResponse>;
 
         /**
          * Issues a MVS Command and returns a Console Command API response.
--- a/packages/zowe-explorer-api/src/profiles/ZoweExplorerZosmfApi.ts
+++ b/packages/zowe-explorer-api/src/profiles/ZoweExplorerZosmfApi.ts
@@ -329,8 +329,12 @@ export class ZosmfJesApi extends ZosmfApiCommon implements ZoweExplorerApi.IJes 
  */
 // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access
 export class ZosmfCommandApi extends ZosmfApiCommon implements ZoweExplorerApi.ICommand {
-    public async issueTsoCommand(command: string, acctNum: string): Promise<zowe.IIssueResponse> {
-        return await zowe.IssueTso.issueTsoCommand(this.getSession(), acctNum, command);
+    public async issueTsoCommand(
+        command: string,
+        acctNum: string,
+        startParams: zowe.IStartTsoParms
+    ): Promise<zowe.IIssueResponse> {
+        return await zowe.IssueTso.issueTsoCommand(this.getSession(), acctNum, command, startParams);
     }
 
     public async issueMvsCommand(command: string): Promise<zowe.IConsoleResponse> {
--- a/packages/zowe-explorer/src/command/TsoCommandHandler.ts
+++ b/packages/zowe-explorer/src/command/TsoCommandHandler.ts
@@ -19,6 +19,7 @@ import { Profiles } from "../Profiles";
 import { ZoweExplorerApiRegister } from "../ZoweExplorerApiRegister";
 import { errorHandling, FilterDescriptor, FilterItem, resolveQuickPickHelper } from "../utils/ProfilesUtils";
 import { ZoweCommandProvider } from "../abstract/ZoweCommandProvider";
+import { IStartTsoParms } from "@zowe/cli";
 
 // Set up localization
 nls.config({
@@ -128,9 +129,9 @@ export class TsoCommandHandler extends ZoweCommandProvider {
             if (Profiles.getInstance().validProfile !== ValidProfileEnum.INVALID) {
                 const commandApi = ZoweExplorerApiRegister.getInstance().getCommandApi(profile);
                 if (commandApi) {
-                    let acctNum: string;
+                    let tsoParams: IStartTsoParms;
                     if (profile.type === "zosmf") {
-                        acctNum = await this.getAccountNumber();
+                        tsoParams = await this.getTsoParams();
                     }
                     let command1: string = command;
                     if (!command) {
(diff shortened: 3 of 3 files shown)


fswarbrick:
This PR does not pass all unit tests, but works in practice.  Please advise.

packages/zowe-explorer-api/src/profiles/ZoweExplorerApi.ts
```diff
@@ -440,7 +440,11 @@ export namespace ZoweExplorerApi {
          * @returns {zowe.IIssueResponse>}
          * @memberof ICommand
          */
-        issueTsoCommand?(command: string, acctNum?: string): Promise<zowe.IIssueResponse>;
+        issueTsoCommand?(
+            command: string,
+            acctNum?: string,
+            startParams?: zowe.IStartTsoParms
+        ): Promise<zowe.IIssueResponse>;
```
JillieBeanSim, line 448: @fswarbrick since account number is included in the startParams I would think just changing `acctNum` to `startParams` then see message in ZoweExplorerZosmfApi.ts

packages/zowe-explorer-api/src/profiles/ZoweExplorerZosmfApi.ts
```diff
@@ -329,8 +329,12 @@ export class ZosmfJesApi extends ZosmfApiCommon implements ZoweExplorerApi.IJes
...
 // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access
 export class ZosmfCommandApi extends ZosmfApiCommon implements ZoweExplorerApi.ICommand {
-    public async issueTsoCommand(command: string, acctNum: string): Promise<zowe.IIssueResponse> {
-        return await zowe.IssueTso.issueTsoCommand(this.getSession(), acctNum, command);
+    public async issueTsoCommand(
+        command: string,
+        acctNum: string,
+        startParams: zowe.IStartTsoParms
+    ): Promise<zowe.IIssueResponse> {
+        return await zowe.IssueTso.issueTsoCommand(this.getSession(), acctNum, command, startParams);
```
JillieBeanSim, line 337: here, with passing only startParams we could make the call 
`return await zowe.IssueTso.issueTsoCommand(this.getSession(), startParams.account, command, startParams);`

packages/zowe-explorer/src/command/TsoCommandHandler.ts
```diff
@@ -247,7 +249,11 @@ export class TsoCommandHandler extends ZoweCommandProvider {
                         title: localize("issueTsoCommand.command.submitted", "TSO command submitted."),
                     },
                     () => {
-                        return ZoweExplorerApiRegister.getCommandApi(profile).issueTsoCommand(command, acctNum);
+                        return ZoweExplorerApiRegister.getCommandApi(profile).issueTsoCommand(
+                            command,
+                            startParams.account,
+                            startParams
+                        );
```
JillieBeanSim, line 256: and this call just passing the command and startParams

JillieBeanSim:
> This PR does not pass all unit tests, but works in practice. Please advise.

@fswarbrick I will take a look at the unit tests

packages/zowe-explorer-api/src/profiles/ZoweExplorerApi.ts
fswarbrick: Seems like six of one, half dozen of the other.  I wanted to minimize the number of modules I changed.  But I can do so if you'd prefer.

JillieBeanSim: @fswarbrick if those changes are made then only 1 unit test is failing and that can be fixed this way
image

I will build and package your branch and test it with extenders to make sure nothing is broken in that case

fswarbrick: @JillieBeanSim:  Thanks for your help.  Can you point me to any documentation that explains what these tests are doing?  Apparently they count the number of parameters passed?

Also, isn't it wrong that the "getTsoParams" test is (still) returning a string, instead of IStartTsoParams, but there is no test that catches that?

JillieBeanSim: @fswarbrick we have documentation on unit tests in the wiki https://github.com/zowe/vscode-extension-for-zowe/wiki/Best-Practices:-Unit-Tests, mostly testing functions and methods with mock returns for some calls.  The Object.defineProperty is mocking the call to getTsoParams returning an object with only account within in the image above, so any method in this set of unit tests that call getTsoParams will get this object back.

JillieBeanSim:
@fswarbrick changing the api will break extenders so what we will need to do is create a new api and set the old as deprecated
image

then there will have to be a check in the issueCommand method in TsoCommandHandler similar to this https://github.com/zowe/vscode-extension-for-zowe/blob/3a463bc10d709120b42ef6e8e88fa8a1c78c86b5/packages/zowe-explorer/src/shared/utils.ts#L216 to check for new api call with the updated parameter passed if it exists use that else use the deprecated api.  For the ZoweExplorerZosmfApi.ts you can update the tso api to use the new one.

fswarbrick:
> @fswarbrick changing the api will break extenders so what we will need to do is create a new api call in ZoweExplorerApi.ts and set the old as deprecated
> 
> then there will have to be a check in the issueCommand method in TsoCommandHandler similar to this
> 
> https://github.com/zowe/vscode-extension-for-zowe/blob/3a463bc10d709120b42ef6e8e88fa8a1c78c86b5/packages/zowe-explorer/src/shared/utils.ts#L216
> 
> to check for new api call with the updated parameter passed if it exists use that else use the deprecated api. For the ZoweExplorerZosmfApi.ts you can update the tso api to use the new one.

@JillieBeanSim  I now remember why I added a new parameter instead of replacing it in the first place.  Wouldn't doing so keep the API essentially "unchanged" (any calls without IStartTsoParms would continue with the existing behavior), and this not require deprecation?

JillieBeanSim:
> @JillieBeanSim I now remember why I added a new parameter instead of replacing it in the first place. Wouldn't doing so keep the API essentially "unchanged" (any calls without IStartTsoParms would continue with the existing behavior), and this not require deprecation?

@fswarbrick even adding an additional parameter would break the api for extenders, there api file is set up to match the ZoweExplorerApi.ts like the ZoweExplorerZosmfApi.ts.  Changing the api call in any way will break it for extenders, adding a new optional api call like shown above allows the extenders to have time to change to the new call before the deprecated call is removed completely.

fswarbrick:
I replaced this PR with https://github.com/zowe/vscode-extension-for-zowe/pull/1378, as I couldn't figure out how to update the PR successfully.

JillieBeanSim:
@fswarbrick should we close this one in favor of the other?

fswarbrick:
Closed.  Replaced by https://github.com/zowe/vscode-extension-for-zowe/pull/1378.

PR #1367, opened 2021-06-18
Add patching for Eclipse Che environment

crawr:

## Proposed changes

Add patching scripts to allow for ZE to be built as an Eclipse Che plugin:
1. Specify the Zowe-Temp-Folder-Location
2. Remove `yarn lint` to allow yarn package to work
3. Set VS Code workspace configuration target

The result of running the new `che:package` script is a new `vscode-extension-for-zowe-<version>-che.vsix`  in the `/dist` folder

## Release Notes

Milestone: 1.17

Changelog: Added patching scripts to allow for ZE to be built as an Eclipse Che plugin

## Types of changes

## Checklist

## Further comments

### DIFF
--- a/packages/zowe-explorer/package.json
+++ b/packages/zowe-explorer/package.json
@@ -1629,6 +1629,7 @@
     "pretest:integration": "ts-node ../../scripts/create-env.ts",
     "posttest:integration": "ts-node ../../scripts/clean-env.ts",
     "updateStrings": "node ../../scripts/stringUpdateScript.js && prettier --write  --loglevel warn ./i18n package.nls*",
+    "che:package": "node ../../scripts/che-patch.js && vsce package --yarn && node ../../scripts/che-pack.js vscode-extension-for-zowe vsix",
     "package": "vsce package --yarn && node ../../scripts/mv-pack.js vscode-extension-for-zowe vsix",
     "license": "node ../../scripts/license.js",
     "watch": "yarn createTestProfileData && gulp build && webpack --mode development --watch --info-verbosity verbose",
--- a/packages/zowe-explorer/package.json
+++ b/packages/zowe-explorer/package.json
@@ -1681,6 +1681,7 @@
     "jest-stare": "^2.2.0",
     "jsdoc": "^3.6.3",
     "jsdom": "^16.0.0",
+    "json-diff": "0.5.4",
     "keytar": "^7.2.0",
     "log4js": "^3.0.5",
     "markdownlint-cli": "^0.16.0",
--- a/scripts/che-pack.js
+++ b/scripts/che-pack.js
@@ -1,0 +1,19 @@
+/*
+ * This program and the accompanying materials are made available under the terms of the *
+ * Eclipse Public License v2.0 which accompanies this distribution, and is available at *
+ * https://www.eclipse.org/legal/epl-v20.html                                      *
+ *                                                                                 *
+ * SPDX-License-Identifier: EPL-2.0                                                *
+ *                                                                                 *
+ * Copyright Contributors to the Zowe Project.                                     *
+ *                                                                                 *
+ */
+
+const fs = require("fs");
+const path = require("path");
+const packageName = process.argv[2];
+const extension = process.argv[3];
+const fullPackageName = `${packageName}-${process.env.npm_package_version}.${extension}`;
+const targetPath = path.join("..", "..", "dist", `${packageName}-${process.env.npm_package_version}-che.${extension}`);
+fs.renameSync(fullPackageName, targetPath);
+console.log(`Published package to ${targetPath}.`);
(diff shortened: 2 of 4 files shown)


phaumer (changes requested):
How will the script be used? I vote against such a solution as it looks very brittle. Patching source code seems a bad practice making the code harder to maintain working around our change process and will break when we make changes in the code assuming that no other developer will know about this or forget about it. Using feature flags within the code would be a better solution. 

However, I would rather investigate into better ways for dynamically finding a temp location as part of the VS Code extension. Assuming it will be /projects is not an ideal assumption either as the folder will appear in the user's workspace and Che file explorer. 

Users can also configure a different folder than /project for their workspace location as well mount other storage areas. A Che user might want to specify all these things in their devfile instead. The latest additions to devfiles allow exactly that. See an example here: https://github.com/IBM/zopeneditor-about/blob/59afdf5e935d77cb284201a90d820d34c3599556/che/ibm-z-devfile.yaml#L23

filipkroupa:
@phaumer This is not about dynamically setting the temp folder location, but about fixing it's location under /projects. The reason is interoperability with other extensions. There exists a limitation in Eclipse Che environment, which allows enablement of LSP features only on files under /project. Since Zowe Explorer puts dynamically downloaded files into it's temp folder, we want this temp folder to be under /project so LSP extensions (Cobol LSP, HLASM LSP) would work with files downloaded/browsed by Zowe Explorer.
Part of the full patch should also have contribution into workspace configuration, hiding the temp folder under /projects.

VitGottwald:
Thank you @crawr for opening this PR to start the discussion!

VitGottwald:
This is something we've inherited as a quick fix to get Zowe Explorer running under Che. Richelle actually cleaned it up quite a bit and it is now clear what it does. Thank you @crawr!

However, I agree @phaumer that it cannot stay like this. Thank you for pointing us to the Che plugin `preferences` option, it looks pro missing. One thing I do not see, however, is how you achieve hiding of the `temp` folder so that it does not show up in `/projects`. The devfile you pointed to does not seem to mention it at all.

phaumer:
@VitGottwald I would like to see it as a user, but you can just add a "files.exclude" setting to the devfile as well, couldn't you?

@filipkroupa I understand and agree with the problem, but not the proposed solution. Hacking source code during build would lead to a maintenance nightmare. We already have Theia-specific code in our codebase. Testing for Che (e.g. check for the existence of a Che env vars such as CHE_WORKSPACE_ID) and providing Che-specific code in our main codebase, i.e. code that would set different default values for these two settings, could also be a valid solution I would approve.

phaumer:
`/projects` is also just the defaults value and a user can change that, so Zowe Explorer should query the the `CHE_PROJECTS_ROOT` env variable for the actual location.  (see <https://www.eclipse.org/che/docs/che-7/end-user-guide/authoring-devfiles-version-1/>)

VitGottwald:
> `/projects` is also just the defaults value and a user can change that, so Zowe Explorer should query the the `CHE_PROJECTS_ROOT` env variable for the actual location. (see https://www.eclipse.org/che/docs/che-7/end-user-guide/authoring-devfiles-version-1/)

That is a very good point, thank you @phaumer .

phaumer:
@VitGottwald I tried it as well and you are right. This might be a limitation that there are no preferences for the cheEditor component type. Let's see if there is someone responding to your Mattermost question. 

If not I propose the following for this PR:

- In our main codebase during activation test for Che and if the temp setting is not provided by the user. If true for both then assign the temp to be the value of `${CHE_PROJECTS_ROOT}/.zowe-temp` (not temp as that is too generic; what if the user has a temp folder in their project?)
- Provide a new Zowe Explorer setting called hideZoweTempFolderInWorkspace (default can be true) that will then programmatically update the files.exclude value in the user settings with a new entry "**/.zowe-temp". If the Zowe temp value was provided as part of the user settings and the path is below ${CHE_PROJECTS_ROOT} then modify with that resulting value. 
 
There are several other issues filed against Zowe Temp which we can address independently, but only by supporting the Che cases in our main code we will be able to address them all consistently.

VitGottwald:
Thank you @phaumer . I was thinking along the identical lines. Since no one has responded to me yet in the Mattermost chat I think the extra setting for hiding temp folder is our best choice for now.

I am however not sure whether we should honor `Zowe-Temp-Folder-Location` even if `CHE_PROJECTS_ROOT` is defined or whether we should rather append `Zowe-Temp-Folder-Location` to `CHE_PROJECTS_ROOT`, because `.zowe-temp` just another "default" like "tmp" was.

Perhaps we could make it _templatable_ and allow for the insertion of environment variables instead like `'${CHE_PROJECTS_ROOT}/.zowe-temp'`?

crawr:
Thanks for all the helpful discussion. I am closing this PR in favor of #1373

PR #1373, opened 2021-06-24
Add new setting to hide temp folder from settings

crawr:

## Proposed changes

Add a new setting to hide the Zowe Temp Folder from the workspace
setting

## Release Notes

Milestone: 1.17.0

Changelog: Added a new setting to hide the Zowe Temp Folder from the workspace

## Types of changes

## Checklist

## Further comments

Changes here wer proposed in the comments from PR #1367

### DIFF
--- a/packages/zowe-explorer/package.json
+++ b/packages/zowe-explorer/package.json
@@ -1562,6 +1562,12 @@
           "description": "Zowe-Temp-Folder-Loacation",
           "scope": "window"
         },
+        "Zowe-Hide-Temp-Folder": {
+          "type": "boolean",
+          "default": true,
+          "description": "%Zowe-Hide-Temp-Folder%",
+          "scope": "window"
+        },
         "Zowe Security: Credential Key": {
           "type": "string",
           "default": "",
--- a/packages/zowe-explorer/package.nls.json
+++ b/packages/zowe-explorer/package.nls.json
@@ -67,6 +67,7 @@
   "Zowe-Default-Datasets-PS": "Default values of Sequential data set creation",
   "Zowe-Default-Datasets-Favorites": "Toggle if favorite files persist locally",
   "Zowe-Temp-Folder-Loacation": "Path to temporary folder location",
+  "Zowe-Hide-Temp-Folder": "Hide the Zowe temporary folder from the workspace",
   "Zowe-Builtin-Security": "Built-in Security. Specify a service of either 'Zowe-Plugin' when using the Secure Credentials Store plugin or @zowe/cli to describe a keytar based solution. The Default 'blank' represents no security. User names and passwords are stored in the clear.",
   "Zowe-Default-USS-Persistent-Favorites": "Toggle if USS favorite files persist locally",
   "Zowe-Default-Jobs-Persistent-Favorites": "Toggle if Jobs favorite files persist locally",
--- a/packages/zowe-explorer/src/extension.ts
+++ b/packages/zowe-explorer/src/extension.ts
@@ -41,6 +41,7 @@ import SpoolProvider from "./SpoolProvider";
 import * as nls from "vscode-nls";
 import { TsoCommandHandler } from "./command/TsoCommandHandler";
 import { cleanTempDir, moveTempFolder } from "./utils/TempFolder";
+import { PersistentFilters } from "./PersistentFilters";
 
 // Set up localization
 nls.config({
@@ -65,6 +66,16 @@ export async function activate(context: vscode.ExtensionContext): Promise<ZoweEx
 
     // Determine the runtime framework to support special behavior for Theia
     globals.defineGlobals(preferencesTempPath);
+
+    if (PersistentFilters.getDirectValue("Zowe-Hide-Temp-Folder") as boolean) {
+        vscode.workspace
+            .getConfiguration("files")
+            .update(
+                "exclude",
+                { [getZoweDir()]: true, [globals.ZOWETEMPFOLDER]: true },
+                vscode.ConfigurationTarget.Workspace
+            );
+    }
 
     // Call cleanTempDir before continuing
     // this is to handle if the application crashed on a previous execution and
--- a/packages/zowe-explorer/i18n/sample/package.i18n.json
+++ b/packages/zowe-explorer/i18n/sample/package.i18n.json
@@ -67,6 +67,7 @@
   "Zowe-Default-Datasets-PS": "Default values of Sequential data set creation",
   "Zowe-Default-Datasets-Favorites": "Toggle if favorite files persist locally",
   "Zowe-Temp-Folder-Loacation": "Path to temporary folder location",
+  "Zowe-Hide-Temp-Folder": "Hide the Zowe temporary folder from the workspace",
   "Zowe-Builtin-Security": "Built-in Security. Specify a service of either 'Zowe-Plugin' when using the Secure Credentials Store plugin or @zowe/cli to describe a keytar based solution. The Default 'blank' represents no security. User names and passwords are stored in the clear.",
   "Zowe-Default-USS-Persistent-Favorites": "Toggle if USS favorite files persist locally",
   "Zowe-Default-Jobs-Persistent-Favorites": "Toggle if Jobs favorite files persist locally",


VitGottwald (dismissed):
LGTM

VitGottwald (approved):
LGTM

packages/zowe-explorer/package.json
```diff
@@ -1567,6 +1567,12 @@
           "description": "Zowe-Temp-Folder-Loacation",
           "scope": "window"
         },
+        "Zowe-Hide-Temp-Folder": {
```
lauren-li, line 1570: @crawr Do we want to go ahead and try to use the [standardized VS Code formatting](https://code.visualstudio.com/api/references/contribution-points#Configuration-schema) for this setting? Something like `zowe.hideTempFolder`?

phaumer: Agree. Otherwise, user will start using it and then have to migrate again.

phaumer (changes requested):
This writes an update to the workspace settings of each and every one of our users on every platform when they start up. This means an SCM changes for everyone using Git. This violates design principles of minimum impact of changes and it would get everyone very confused about why this is necessary for VS Code users. 

As this was intended for the small percentage of Eclipse Che users out there we should either (1) add an Eclipse Che test as previously discussed or (b) make the default of zowe.hideTempFolder false. Then we could set it to true in devFiles, which would then still hide for Che users when they start up.

The other thing I am not sure about is if the files.exclude needs to be written into the workspace settings. If the devfile sets the zowe.hideTempFolder in the workspace, the actual files.exclude could still be done in the user settings, couldn't it? What if different users in the team have different opinions about it? Perhaps they want each team member decided and not have in the devfile. (I personally would never want to use it.) For those it needs to be a user setting.

packages/zowe-explorer/package.json
crawr: Yes, this makes sense to me too. I've changed it in the latest commit

crawr:
Thanks for the review. I changed the default to `false` and set the configuration target to `global` as suggested

packages/zowe-explorer/src/extension.ts
```diff
@@ -66,6 +67,16 @@ export async function activate(context: vscode.ExtensionContext): Promise<ZoweEx
     // Determine the runtime framework to support special behavior for Theia
     globals.defineGlobals(preferencesTempPath);
 
+    if (PersistentFilters.getDirectValue("zowe.hideTempFolder") as boolean) {
+        vscode.workspace
+            .getConfiguration("files")
+            .update(
+                "exclude",
+                { [getZoweDir()]: true, [globals.ZOWETEMPFOLDER]: true },
```
phaumer, line 75: Testing this with VS Code I see that it adds these settings for me

```json
"files.exclude": {
    "/Users/phaumer/.zowe": true,
    "/Users/phaumer/Dev/Zowe/vscode-extension-for-zowe/packages/zowe-explorer/resources/temp": true
  }
```

However, these settings do not overlap with the workspace path I am using, so they are really unnecessary. Would it make more sense to test if the path of zowe dir and zowe temp actually overlap with the current workspace path and then only if it does to add the setting?

packages/zowe-explorer/src/extension.ts
```diff
@@ -66,6 +67,16 @@ export async function activate(context: vscode.ExtensionContext): Promise<ZoweEx
     // Determine the runtime framework to support special behavior for Theia
     globals.defineGlobals(preferencesTempPath);
 
+    if (PersistentFilters.getDirectValue("zowe.hideTempFolder") as boolean) {
```
phaumer, line 70: Adding this code to the activation means that the other has to restart the editor when they change the hideTempFolder. If you would rather do via an `onDidChangeConfiguration()` event handler then the files.excludes settings gets added as soon as the user saves the hideTempFolder settings change without a restart. I think it works something like this (also in activation):

```ts
context.subscriptions.push(
        vscode.workspace.onDidChangeConfiguration(async event => {
            if (event.affectsConfiguration("zowe.hideTempFolder") &&
                PersistentFilters.getDirectValue("zowe.hideTempFolder") as boolean) {
...
```

packages/zowe-explorer/package.nls.json
```diff
@@ -67,6 +67,7 @@
   "Zowe-Default-Datasets-PS": "Default values of Sequential data set creation",
   "Zowe-Default-Datasets-Favorites": "Toggle if favorite files persist locally",
   "Zowe-Temp-Folder-Loacation": "Path to temporary folder location",
+  "zowe.hideTempFolder": "Hide the Zowe temporary folder from the workspace",
```
phaumer, line 70: @rudyflores @lauren-li do you have an updates list of settings names? Would it make sense to use the "files" group? I.e. `zowe.files.hideTempFolder` and `zowe.file.tempFolderPath` ?

phaumer (changes requested):
This works now. I am just making some proposals of minor changes would should discuss. We also should agree on the settings name as I see this PR using `zowe.files.<something>`: https://github.com/zowe/vscode-extension-for-zowe/pull/1347/files

packages/zowe-explorer/package.nls.json
lauren-li: @crawr @phaumer The format being proposed for @JillieBeanSim's PR #1347 (Support multiple VS Code windows) is `zowe.files.temporaryDownloadsFolderCleanup`. I agree with the proposal in that PR, and I also think it makes sense to use the `files` group here in this PR. Below are the settings we could put in the `files` group thus far via @rudyflores' work:

- `zowe.files.temporaryDownloadsFolderCleanup`
- `zowe.files.tempFolderPath`(or `zowe.files.temporaryDownloadsFolderPath` as proposed in [this PR 1347 comment](https://github.com/zowe/vscode-extension-for-zowe/pull/1347#discussion_r660206692))
- `zowe.files.hideTempFolder`

We probably want to standardize on if we want to refer the temp folder as `tempFolder` or `temporaryDownloadsFolder`.

crawr: I think to be consistent, we can go with `zowe.files.hideTemporaryDownloadsFolder` for this PR

VitGottwald:
Thank you for the comments and for testing this out @phaumer !

packages/zowe-explorer/src/extension.ts
VitGottwald: While I agree with all the other comments, I do not agree with this one. Zowe Explorer works independently of the vscode workspace. This suggests that we make the code behaviour workspace dependent. I think that would be worse than the current behaviour (which is not ideal either).

This way the behaviour predictable and easy to understand for the users. So I would suggest that we keep it as is.

VitGottwald: This touches the subject of how we should we properly handle workspace vs. user settings which is a much larger discussion that we will have to have.

phaumer: @VitGottwald can you clarify what you mean by "workspace dependent"? The "files.exclude" setting's purpose is to modify how the workspace is being presented. It has to be specific to the current users workspace. If a user does not use a workspace then then there is no problem. I rather think that we should only fill in values that represent appropriate workspace locations otherwise it would be confusing to the end user or they might think this is a bug.

packages/zowe-explorer/src/extension.ts
crawr: I also added it to the event handler as suggested. Thanks!

rudyflores:
After posting in the issue thread for #1263 the proposal for standardized names, it was decided to change the current `zowe.files.hideTemporaryDownloadsFolder` to the option proposed by @VitGottwald: `zowe.files.temporaryDownloadsFolder.hide` in order to keep the configuration names consistent.

Could you change the current name to the one that Vit proposed @crawr ?

phaumer (approved):
I am approving this PR now as it is usable for Che and VS Code, but there are two items that I strongly still encourage to do perhaps in a separate PR, which @lauren-li volunteered doing while I am out on vacation:

- The description of the new setting is not detailed enough. A user will not be able to understand what it is for and how it should be used with VS Code or Che. Perhaps @rudyflores can revise it together with his other work of revising all settings and descriptions of these settings.
- The discussion around what to add to file.excludes was not concluded and I really still believe that adding folder here that do not overlap with the current workspace folder will be really confusing and just cause unnecessary settings file changes. There should be a test that check if the path is actually within the current workspace path and only then add it. As this code is executed during every activation, i.e. workspace change, it will ensure that only valid workspace locations are listed here and not just some random set of folders as it is now.

jellypuno (dismissed):

VitGottwald (dismissed):

PR #1483, opened 2021-09-09
Fix Multi-Delete Bug 

JillieBeanSim:

## Proposed changes

After investigation the delete data sets was no longer passing a right-clicked node that isn't fully selected in blue resulting in the data set that was opened/selected being deleted vs the member that was right-clicked for deletion.

This PR fixes that passing the right-clicked node again, then it also checks against the selected list.  If the node is not in the selected list only that node is deleted like vsc does with their tree. 

Also, a new warning message is presented to the user for confirmation including the list of items to be deleted similar to vsc's deletion confirmation.  I will create a new issue to unify the other trees with this new deletion confirmation.

image

## Release Notes

Milestone: 1.19.0

Changelog: Bug fix to address multi-select deletion of entire data set when a member is right-clicked for deletion, with a new confirmation message for delete of data sets and members.

## Types of changes

## Checklist

## Reminder

## Further comments


packages/zowe-explorer/src/dataset/actions.ts
```diff
@@ -239,28 +280,21 @@ export async function deleteDatasetPrompt(datasetProvider: IZoweTree<IZoweDatase
...
-    ) {
-        globals.LOG.debug(
-            localize("deleteDatasetPrompt.showQuickPick.log.debug", "Dataset deletion action was canceled.")
-        );
-        return;
-    }
+    const deleteButton = localize("deleteDatasetPrompt.confirmation.delete", "Delete");
+    const message = localize(
+        "deleteDatasetPrompt.confirmation.message",
+        "Are you sure you want to delete these items?\nThis will permanently remove these data sets and/or members from your system.\n\n{0}",
```
lauren-li, line 286: @JillieBeanSim What would you think about adding an item count in this message like how VS Code does theirs? E.g. "Are you sure you want to delete these 30 items?"
Adding the count of items to be deleted can help the user quickly assess if they have accidentally selected more (or less) than the expected number of items for deletion. 
I think it shouldn't be too hard to add; you might be able to get the count just by doing `deletedNodes.length`:

```
    const message = localize(
        "deleteDatasetPrompt.confirmation.message",
        "Are you sure you want to delete these {0} items?\nThis will permanently remove these data sets and/or members from your system.\n\n{1}",
        deletedNodes.length,
        deletedNodes.toString().replace(/(,)/g, "\n")
    );
```

In fact, VS Code is a bit fancier with generating different messages based on if you have a file(s), folder(s), or mix selected, and if you have one versus multiple selected. So we can consider additional refinements like those in future PRs.

JillieBeanSim: @lauren-li `these 1 items` will look funny, maybe it should be `the following items`?

lauren-li: @JillieBeanSim Mostly I wanted to have the item count in the message. We could write: `Are you sure you want to delete the following {0} item(s)?"

phaumer (changes requested):
I selected four members and get a long list with redundant entires. Why even use a List of this should be a Set?

packages/zowe-explorer/src/dataset/actions.ts
```diff
@@ -198,19 +198,60 @@ export async function uploadFile(node: ZoweDatasetNode, doc: vscode.TextDocument
...
+                selectedNode.getParent() &&
+                !contextually.isFavorite(selectedNode) &&
+                !contextually.isFavorite(selectedNode.getParent()) &&
+                !contextually.isSession(selectedNode) &&
+                !contextually.isInformation(selectedNode)
+        );
+        // Check that child and parent aren't both in array
+        const newNodeList: IZoweDatasetTreeNode[] = [];
+        for (const item of nodes) {
+            if (contextually.isDsMember(item)) {
```
phaumer, line 229: I the problem is here missing the `!`: Shouldn't it be `if (!contextually.isDsMember(item)) {`

phaumer (commented):
Cancel does not work. If I select four and see that the first gets deleted and then click cancel the other three still get deleted:

phaumer:
Sorry for messing up all the comments and review comments. To solve the cancel problem I suggest to not use the event handler, but a simple if test in the deletion for loop. In line 316 we could add this to check the token reference before each delete an exit out when the cancel button was clicked.

```ts
      if (token.isCancellationRequested) {
            vscode.window.showInformationMessage(
                localize("deleteDatasetPrompt.deleteCancelled", "Delete action was cancelled.")
            );
            return;
        }
```

phaumer:
@JillieBeanSim I pushed a branch in which I fixed the problems mentioned above, which I am using for the Theia fix. 
https://github.com/zowe/vscode-extension-for-zowe/commits/delete-on-theia

If you agree with my changes feel free to cherry pick these commits: a2a4357ff426b8bad33942ca4f798accaae4294b, 9890784dec25f5ddd4c79ab7339cd92b5f73c274, 48058eab95d51d2e8624e51271cc1a406c736ad9

phaumer:
I created a draft pull request that fixes the issues I found during my review as well as deleting on Theia/Che. 
However, I need some mentoring on the failing tests. For some reasons I cannot set break points in tests anymore (@lauren-li and @rudyflores see the same issue); plus I have some questions about these tests as they do not seem to mock the withProgress VS Code API as other tests do.

packages/zowe-explorer/src/dataset/actions.ts
JillieBeanSim: @lauren-li it's in now

phaumer (dismissed):

lauren-li:
There is a regression in the current VS Code Marketplace release of Zowe Explorer (1.18.1), where selecting to delete a favorited data set or favorited member from the Favorites section now silently fails. (See issue https://github.com/zowe/vscode-extension-for-zowe/issues/1488.)
Although fixing this regression may be out of scope for this PR, this PR does make it a bit more obvious. Selecting to delete a favorited PDS gives the below message:

Screen Shot 2021-09-13 at 5 50 51 PM

Trying to delete a favorited member or PS will give a generic VS Code error:

`Error running command zowe.ds.deleteMember: Running the contributed command: 'zowe.ds.deleteMember' failed.. This is likely caused by the extension that contributes zowe.ds.deleteMember.`

We might want to make these errors/messages a bit more user friendly until we can fix the regression.

JillieBeanSim:
This is a temp fix for release, the full bug/regression fix will be handled by https://github.com/zowe/vscode-extension-for-zowe/issues/1488

delete-favs-message

with multi-select of all favorites the information message will pop up in lower right corner, but if favorites and session items selected for multi-select delete only the session items will show in the confirmation message
image

lauren-li (approved):
This PR looks good to me! All unit and expected integration tests are passing. In manual testing, I found that selecting a member to delete no longer deletes its parent PDS, and single/multi-select deletion of PDS, PS, and members now appears to be working as expected. I appreciate the addition of the item numbers in the confirmation messages, as well as the added info message for our currently-defunct favorited data set/member delete menu item (pre-exisiting issue*).

*I filed issues for the following pre-existing bugs I found:
- #1488
- #1489

Thanks @JillieBeanSim for your hard work and patience on this PR!

katelynienaber (approved):
Works for me!

PR #1497, opened 2021-09-16
1.19.0 to next

katelynienaber:
## Proposed changes

Merge 1.19.0 changes into next branch for testing purposes.


JillieBeanSim:
@katelynienaber the theia failures where just timeout errors, I re-ran the job and all worked

.github/workflows/deployment.yml
```diff
@@ -11,7 +11,7 @@ on:
     inputs:
       version:
         description: Version to be published
-        default: v1.18.0
+        default: v1.18.1
```
jellypuno, line 14: We might need to wait for the doc and the version change before merging this to `next`

lauren-li (commented):
@katelynienaber Are you able to bring the remaining changes from the 1.19.0 release into this PR?

One thing we might want to add in this PR is to update the `refresh()` function's `apiRegister` parameter to be required in `ProfilesCache.ts` (Zowe Explorer API).
 - I noticed that on the [`master` branch](https://github.com/zowe/vscode-extension-for-zowe/blob/master/packages/zowe-explorer-api/src/profiles/ProfilesCache.ts#L84), `apiRegister` is required, but on the [`next` branch](https://github.com/zowe/vscode-extension-for-zowe/blob/next/packages/zowe-explorer-api/src/profiles/ProfilesCache.ts#L98), `apiRegister` is currently optional. It seems the update to change `apiRegister` from optional to required somehow missed being pulled into the `next` branch. This `refresh()` function does not look like it was written to work without `apiRegister` passed in. Additionally, I did a reference check, and Zowe Explorer does not make any calls to the ProfilesCache `refresh()` without passing in an `apiRegister`, so I do not believe making this update would break any code/functionality.

lauren-li:
@katelynienaber Just to confirm, was this closed in favor of PR #1509?

katelynienaber:
@lauren-li Yep!

Issue #1485, opened 2021-09-10
Standardize Confirmation Popups/QuickPicks


JillieBeanSim:
Would like to standardize confirmation pop-ups and quick picks with VS Code standard for:
  - Delete
  - ~Save Credentials~ handled in https://github.com/zowe/vscode-extension-for-zowe/pull/1486 with `Check Credentials` popup
  - any others we may find that have a yes/no option or buttons that require user interaction

This is done in PR #1483 for delete data sets/members 

![image](https://user-images.githubusercontent.com/49491949/132859583-f9d573f2-69b9-477e-bf28-034bbef220ed.png)

The code below is what was used to accomplish that in the linked PR

```
const deleteButton = localize("deleteDatasetPrompt.confirmation.delete", "Delete");
const message = localize(
        "deleteDatasetPrompt.confirmation.message",
        "Are you sure you want to delete these items?\nThis will permanently remove these data sets and/or members from your 
        system.\n\n{0}",
        deletedNodes.toString().replace(/(,)/g, "\n")
);
await vscode.window.showWarningMessage(message, { modal: true }, ...[deleteButton]).then((selection) => {
        if (!selection || selection === "Cancel") {
            globals.LOG.debug(
                localize("deleteDatasetPrompt.confirmation.cancel.log.debug", "Dataset deletion action was canceled.")
            );
            return;
        }
});
```

JillieBeanSim:
more pop ups with buttons mentioned in a bug #1349

JillieBeanSim:
**loadProfilesForFavorites()**
- DatasetTree.ts line 402 
- ZosJobsProvider.ts line 368
- USSTree.ts line 815

JillieBeanSim:
Once both of the above attached PRs are merged, this issue can be closed.

PR #1506, opened 2021-09-27
Update to error confirmation msg for removing favorites

JillieBeanSim:

## Proposed changes

This is part of #1485 
The error message pop up with `Remove` and `Cancel` buttons are now using the VSC standard confirmation message in Data Sets, USS, and Job trees.

image

## Release Notes

Milestone: 1.20.0

Changelog: Add Visual Studio Code's standard confirmation prompt for the Data Sets, USS, and Job trees when clicking on a Favorited profile that doesn't exist.

## Types of changes

## Checklist

## Reminder

## Further comments

There is a secondary confirmation on this action that uses the quickpick for confirmation in each tree views `removeFavProfile()`, I wonder if this should either (1) also be a VSC standard confirmation or (2) be removed completely?


JillieBeanSim:
About the **Further Comments** section question `There is a secondary confirmation on this action that uses the quickpick for confirmation in each tree views removeFavProfile(), I wonder if this should either (1) also be a VSC standard confirmation or (2) be removed completely?` I personally don't think a second confirmation is needed.  If this is agreed I can remove it in this PR

lauren-li:
> About the **Further Comments** section question `There is a secondary confirmation on this action that uses the quickpick for confirmation in each tree views removeFavProfile(), I wonder if this should either (1) also be a VSC standard confirmation or (2) be removed completely?` I personally don't think a second confirmation is needed. If this is agreed I can remove it in this PR

@JillieBeanSim I believe the confirmation for `removeFavProfile()` occurs when the user tries to remove an entire profile's folder from the Favorites section. Since this action will (irreversibly) remove all favorites associated with that profile, I would keep that message to allow the user to confirm if they really do want to potentially remove multiple favorites all at once. That should be its own UX flow, and not secondary the non-existent-CLI-profile message in this PR. (If you are seeing the `removeFavProfile()` confirmation appear secondary to the one this PR, then that would be a bug that we should address.)

JillieBeanSim:
Another idea is:
I think removing the boolean value that is passed to that method removeFavProfile() would easily resolve the double confirmation, or we just move the VSC dialog into that method and have it pop up on delete so the caught error will call it and delete on the favorite will call it. removing the double confirmation and standardizing the delete of favorites all together

lauren-li:
> Another idea is:
> I think we just move the VSC dialog into the `removeFavProfile()` method and have it pop up on delete so the caught error will call it and delete on the favorite will call it. removing the double confirmation and standardizing the delete of favorites all together

I think I will need to see this code to understand this proposal better. Generally speaking, I think we should display the `removeFavProfile.confirm` dialog when the user manually right-clicks a favorited profile and selects to remove it. But I do not think it is necessary to display it in this PR's use case where the user has selected an invalid favorited profile, as this would mean all favorites affected by the removal action are also associated with the invalid profile. Presumably, if the user confirms they want to remove the invalid profile in this PR's dialog, they would also want to remove it to resolve the other "favorites-with-invalid-profile" instances, as well.

JillieBeanSim:
OK, since the invalid fave profile is a different message than the delete fave items I am thinking of just passing false if non-existent fave profile removal is selected to avoid the double confirmation that happens with true being passed.  I am also going to standardize the delete confirmation of profiles vs them being in the quick-pick that way all delete actions in Zowe Explorer will be the same format.

JillieBeanSim:
what the dialog looks like in theia
image

packages/zowe-explorer/src/dataset/DatasetTree.ts
```diff
@@ -683,25 +681,24 @@ export class DatasetTree extends ZoweTreeProvider implements IZoweTree<IZoweData
...
-            // If user did not select "Continue", do nothing.
-            if (
-                (await vscode.window.showQuickPick([continueRemove, cancelRemove], quickPickOptions)) !== continueRemove
-            ) {
-                return;
-            }
+            await vscode.window
+                .showWarningMessage(checkConfirmation, { modal: true }, ...[continueRemove])
+                .then((selection) => {
+                    if (!selection || selection === "Cancel") {
```
crawr, line 695: At first I was worried about removing the localization from the selection, but it looks like this condition handles other languages. I tried it with spanish language and when I clicked "Cancelar" it returned undefined so it still counted as cancelled.

crawr (approved):
Great work on this, @JillieBeanSim ! And congrats on fixing the Theia tests 👍 
I tested it locally on Theia 1.15 and it works well.

packages/zowe-explorer/src/dataset/DatasetTree.ts
JillieBeanSim: @crawr I am wondering if the `Cancel` button will always return undefined since it's not a passed button into the confirmation, the modal:true confirmation boxes have a built in Cancel button.

jellypuno (approved):

PR #1472, opened 2021-08-31
[next] Add initialization of zowe config json file

crawr:
## Proposed changes

This PR adds the initialization of zowe config json file in Zowe Explorer using the new SDK APIs

teamconfiginit

## Release Notes

Milestone:

Changelog:

## Types of changes

What types of changes does your code introduce to Zowe Explorer?
_Put an `x` in the boxes that apply_

- [ ] Bugfix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Updates to Documentation or Tests (if none of the other choices apply)

## Checklist

_Put an `x` in the boxes that apply. You can also fill these out after creating the PR. If you're unsure about any of them, don't hesitate to ask. We're here to help! This checklist will be used as reference for both the contributor and the reviewer_

- [x] I have read the [CONTRIBUTOR GUIDANCE](https://github.com/zowe/vscode-extension-for-zowe/wiki/Best-Practices:-Contributor-Guidance) wiki
- [x] PR title follows [Conventional Commits Guidelines](https://www.conventionalcommits.org/en/v1.0.0-beta.2/)
- [x] PR Description is included
- [x] gif or screenshot is included if visual changes are made
- [x] `yarn workspace vscode-extension-for-zowe vscode:prepublish` has been executed
- [ ] All checks have passed (DCO, Jenkins and Code Coverage)
- [ ] I have added unit test and it is passing
- [ ] I have added integration test and it is passing
- [ ] There is coverage for the code that I have added
- [ ] I have tested it manually and there are no regressions found
- [ ] I have added necessary documentation (if appropriate)
- [ ] Any PR dependencies have been merged and published (if appropriate)

## Reminder

After a PR is merged into the `master` branch, create a PR from `master` to the `next` branch resolving any conflicts.

## Further comments

To be merged after #1471

PR #1509, opened 2021-09-30
1.19 to next (2nd go)

katelynienaber:
## Proposed changes

Merge 1.19.0 changes into next branch for testing purposes.


katelynienaber:
@JillieBeanSim Would you have any idea how to fix the failing tests here? It has to do with an NPM version it seems...

JillieBeanSim:
> @JillieBeanSim Would you have any idea how to fix the failing tests here? It has to do with an NPM version it seems...

@katelynienaber it looks like with the bringing in of master branch you are also bringing in the node 12 tests which are removed in the next branch as Node 10 is not supported anymore.

JillieBeanSim:
removed the node 10 from GHA tests as well as node 16 like done in #1472 https://github.com/zowe/vscode-extension-for-zowe/blob/e9fb19d9e864494177dbcfa0b939ea7152ac4a13/.github/workflows/zowe-explorer-ci.yml#L19

JillieBeanSim (commented):
~most new functionality is working well, I do see that the FTP profile is not validating though. This is with old style profiles and if config file is created via cli and includes FTP. @lauren-li @tiantn~
UPDATE: My bad, I forgot to package and install the FTP extension from this branch, all is working!

Another thing I see that could be a separate issue/PR would be that `Issue TSO Command` isn't recognizing the TSO profile information like acct number from the config file.

JillieBeanSim (approved):
Thanks @katelynienaber for putting this together to update next with the most recent changes.  I will create the issue for the TSO stuff I noticed.

crawr (approved):
LGTM. Thanks @katelynienaber 

I did find an issue in the `next` branch where the Job ID link does not work unless a search has already been performed on the profile. But this also exists outside of this PR so I will file a separate issue.

PR #1579, opened 2021-11-15
Expression over statements

VitGottwald:
Signed-off-by: Vit Gottwald <vit.gottwald@gmail.com>


crawr:
@VitGottwald can you please fill out the PR template so we know the description better?

jellypuno:
@VitGottwald was this change tested? Can you also add a gif so that we could see what we need to test and see if there are no functional change. Thanks.

VitGottwald:
There is no functional change. It is only a 3 line code cleanup to make the code more readable.

jellypuno (approved):
tested it and it works

crawr (approved):
Works well for me. I tested it in a Theia environment too 👍 
Thanks @VitGottwald

crawr:
Theia tests and audit issues will be fixed in a different PR

Issue #845, opened 2020-06-08
Invalid Job Spool Display


jellypuno:
**Describe the bug**

The user executed a JCL with a Job card that started with: 

```
//SET1 SET PN=EXER3
//&PN JOB
```

This resulted to a JCL Error and Zowe Explorer displayed it like this:

![Capture](https://user-images.githubusercontent.com/45042840/84059586-26040e80-a9bb-11ea-812f-a7642a6874be.PNG)

**To Reproduce**

1. Run a JCL with a Job Card similar to the one above
2. Submit it
3. Retrieve the result in Zowe Explorer

**Expected behavior**

A better error handling

**Screenshots**

Error Message:
![Capture2](https://user-images.githubusercontent.com/45042840/84060416-6f089280-a9bc-11ea-9466-decac02d42ed.PNG)

From Zowe CLI:
![Capture3](https://user-images.githubusercontent.com/45042840/84060428-74fe7380-a9bc-11ea-9a4c-d0b06261c4c4.PNG)

**Desktop (please complete the following information):**

-   OS: Windows 10
-   Zowe Explorer Version: 1.6.0
-   (Optional) Zowe CLI Version: 6.13.0
-   (Optional) Are you using Secure Credential Store? No

**Additional context**

Follow the conversation: https://openmainframeproject.slack.com/archives/C011NE32Z1T/p1591632786034100

zFernand0:
Is this a problem of us not URI-encoding something?

P.S. I believe the conversation is lost on slack.

zdmullen:
Possible solution here:
We should only show the name of the JCL and "JCL error" next to it, without any entries for JES (since there not any produced). This would keep things clean and prevent users from selecting something that provides large error messages.

zdmullen:
This is dependent on type of security being used, RACF vs ACF2 etc., because they may offer more in the DD.

jellypuno:
This is still happening for v1.18.1

roman-kupriyanov:
[...]

PR #1584, opened 2021-11-19
[Issue #845] Do not show undefined spool files for a job

roman-kupriyanov:
## Proposed changes

Resolves an issue #845 by skipping all the objects which are returned from `GetSpoolFiles` API call and do not correspond to the proper `zowe.IJobFile` type. As a result there will be no more `undefined:undefined(undefined)` items in the spool files tree.

## Types of changes

## Checklist

### DIFF
--- a/packages/zowe-explorer/i18n/sample/src/job/ZoweJobNode.i18n.json
+++ b/packages/zowe-explorer/i18n/sample/src/job/ZoweJobNode.i18n.json
@@ -1,6 +1,7 @@
 {
   "getChildren.search": "Use the search button to display jobs",
   "ZoweJobNode.getJobs.spoolfiles": "Get Job Spool files command submitted.",
+  "getChildren.noSpoolFiles": "There are no JES spool messages to display",
   "ZoweJobNode.getJobs.jobs": "Get Jobs command submitted.",
   "getChildren.error.response": "Retrieving response from "
 }
--- a/packages/zowe-explorer/src/job/ZoweJobNode.ts
+++ b/packages/zowe-explorer/src/job/ZoweJobNode.ts
@@ -102,39 +102,52 @@ export class Job extends ZoweTreeNode implements IZoweJobTreeNode {
                         );
                     }
                 );
-                const refreshTimestamp = Date.now();
-                spools
+                spools = spools
                     // filter out all the objects which do not seem to be correct Job File Document types
                     // see an issue #845 for the details
                     .filter(
                         (item) => !(item.id === undefined && item.ddname === undefined && item.stepname === undefined)
-                    )
-                    .forEach((spool) => {
-                        let prefix = spool.stepname;
-                        if (prefix === undefined) {
-                            prefix = spool.procstep;
-                        }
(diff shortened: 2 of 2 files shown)


jellypuno (changes requested):
As per discussion, please add the text " There are no JES spool messages to display" when I click on the arrow. Thank you and well done on your first PR! 🎉

roman-kupriyanov:
> As per discussion, please add the text " There are no JES spool messages to display" when I click on the arrow. Thank you and well done on your first PR! 🎉

Thank you, Jelly! 😊 

That would be a great addition for the UX in general. Nice idea! Will add it..

uzust01:
Looks good!

nickImbirev:
I'm not sure, that the proposed behaviour will solve the problem. In the end, the user will still try to submit the job with the '&' in the job name (in the beginning) and they will not be able to see any spool files for it. I would consider this as a bug :(

But I do understand, that there is additional contribution required from the SDK authors. Maybe we can at least submit a GH issue for them? And make additional issue in our repo to fix the `?`, `&` and other problematic symbols in our URI serialisation process?

jellypuno:
`And make additional issue in our repo to fix the ?, & and other problematic symbols in our URI serialisation process`

@nickImbirev I do not understand what do you mean by this. If mainframe is accepting `?`, `&` in their JCLs, we are going to do accept that as well.

I agree that we should open an issue against the SDK or the CLI team because if you fire the APIs the records are available. I am not sure why they are giving us an empty response.

image

FIles:
image

Records:
image

jellypuno:
@roman-kupriyanov I am going to accept this as a solution for the issue but I agree with @nickImbirev that we should open a GH issue about the response that we are getting from the SDK. As seen in the screenshots, The z/OSMF response contains the records that we need. We need investigate why the SDK is not giving that to us. Can you please open an issue in the zowe/cli repo?

roman-kupriyanov:
> Can you please open an issue in the zowe/cli repo?

Yes, sure, I was about to propose the same today.

A root cause for that is in the z/OSMF request URL, where `&,?` are the special symbols to organize the query part of that URL and it is constructed inside `@zowe/cli` API's for `GetSpoolFiles` call, out of our control on on the ZE level at the moment. From z/OSMF perspective though, alternatively, you can use **a Job correlator** value to do the same call and it does not contain any special characters, which breaks URL. Also, it is specified in those response fields `files-url` and `records-url`, which is why those links are working perfectly. `@zowe/cli` does not provide "the Job correlator way" of making calls to z/OSMF, except using `jobname/jobid`, which will be broken each time they contain `&,?` or any other special symbols:
https://github.com/zowe/zowe-cli/blob/0cc2ed46de784876172d79d16c1548e2b3fb23c1/packages/zosjobs/src/GetJobs.ts#L253

In sum, I assume, an issue about adding the alternative way of calls for `@zowe/cli` is what may help us with this issue in general. So I will create it 😉

jellypuno:
@roman-kupriyanov That is a good point. Yes. suggest that alternative way 👍

PR #1645, opened 2022-01-27
[Issue #1560] Fix input cancelling messaging

### LINKED ISSUE #1560: TSO error message doesnt appear
If escape when prompted for a command `operation cancelled` info message doesn't pop up...[possible solution here](https://github.com/zowe/vscode-extension-for-zowe/pull/1479#issuecomment-919110641)

roman-kupriyanov:
## Proposed changes

This one fixes #1560. 
_(The issue itself is opened for TSO command execution feature initially, but this bugfix affects a helper function, which is used in many places, and as a result it also fixes the cancelling cases for other commands (e.g. Issue MVS command, filter selections and etc.)_

Added an `onDidHide` handler for `resolveQuickPickHelper`, which always return `undefined` result. 

## Release Notes

Milestone:

Changelog:

## Types of changes

## Checklist

## Reminder

## Further comments

The API doc says:
> There are several reasons why this UI might have to be hidden and the extension will be notified through QuickInput.onDidHide. (Examples include: an explicit call to QuickInput.hide, the user pressing Esc, some other input UI opening, etc.)

This means the informational message will be shown in the cases above, including if some other input will appear in the middle of TSO command issuing.

Also, a current implementation already contains `No selection made.` message, which appears when a choice from the picker is falsy (`undefined` in a case of cancelling through `Esc`). The issue however specifies that `Operation cancelled.` is expected.
Should the message be changed or even, should the cancelling event itself be distinguished from others somehow to show operation cancelled message exactly in this case?

### DIFF
--- a/packages/zowe-explorer/__tests__/__unit__/command/tsoCommandHandler.unit.test.ts
+++ b/packages/zowe-explorer/__tests__/__unit__/command/tsoCommandHandler.unit.test.ts
@@ -323,7 +323,7 @@ describe("TsoCommandHandler unit testing", () => {
             placeHolder: "Select the Profile to use to submit the TSO command",
         });
         expect(showInformationMessage.mock.calls.length).toBe(1);
-        expect(showInformationMessage.mock.calls[0][0]).toEqual("No selection made.");
+        expect(showInformationMessage.mock.calls[0][0]).toEqual("No selection made. Operation cancelled.");
     });
 
     it("tests the issueTsoCommand function user escapes the command box", async () => {
--- a/packages/zowe-explorer/i18n/sample/src/command/TsoCommandHandler.i18n.json
+++ b/packages/zowe-explorer/i18n/sample/src/command/TsoCommandHandler.i18n.json
@@ -7,7 +7,7 @@
   "issueTsoCommand.apiNonExisting": "Not implemented yet for profile of type: ",
   "issueTsoCommand.command.hostname": "Select a TSO command to run immediately against ",
   "issueTsoCommand.command.edit": " (An option to edit will follow)",
-  "issueTsoCommand.options.noselection": "No selection made.",
+  "issueTsoCommand.options.noselection": "No selection made. Operation cancelled.",
   "issueTsoCommand.command.hostnameAlt": "Select a TSO command to run against ",
   "issueTsoCommand.command": "Enter or update the TSO command",
   "issueTsoCommand.enter.command": "No command entered.",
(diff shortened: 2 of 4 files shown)


JillieBeanSim:
@roman-kupriyanov I think just changing the info message to `Operation cancelled` or extending the existing one to `No selection made. Operation cancelled.` in the TsoCommandHandler.ts line 211 would suffice for the issue to be closed.  Pressing escape on this instance does cancel the operation so we are good on that. Thanks for working on this.

roman-kupriyanov:
Hey @JillieBeanSim, thanks a lot for looking at this 😊 

> just changing the info message to `Operation cancelled` or extending the existing one to `No selection made. Operation cancelled.` in the TsoCommandHandler.ts line 211 would suffice for the issue to be closed.

What about other cases, when the `resolveQuickPickHelper` is used? I suppose, changing that commonly used helper function is fixing not only the TSO issue command, but other calls too, where `No selection made.` was not shown as well. For example:

- _MvsCommandHandler.ts_: a similar message is used:
  https://github.com/zowe/vscode-extension-for-zowe/blob/f215d6dc1d79f57abda110007a4f9e9e5cc1a0df/packages/zowe-explorer/src/command/MvsCommandHandler.ts#L207
- _DatasetTree.ts_: same message for filter selection:
  https://github.com/zowe/vscode-extension-for-zowe/blob/f215d6dc1d79f57abda110007a4f9e9e5cc1a0df/packages/zowe-explorer/src/dataset/DatasetTree.ts#L914
- And many more.. There is actually an entire list of them:
  
  image

Should I change the message for them too? Or maybe to create another issue to cover all other commands similarly, but in another PR?

JillieBeanSim:
@roman-kupriyanov I would create a new issue and capture what you have in the comment here so it can be looked at and see what is being presented to the user and what we want presented to the user in each case.  What do you think of that idea?

roman-kupriyanov:
> I would create a new issue and capture what you have in the comment here so it can be looked at and see what is being presented to the user and what we want presented to the user in each case. What do you think of that idea?

@JillieBeanSim sounds good to me as well. I will open a new one, thanks 😉

JillieBeanSim (approved):
LGTM @roman-kupriyanov! thanks for working on this and creating the follow up issue.  If you wouldn't mind, could you add a link in this PR for that once it's created? Thanks.

rudyflores (approved):
LGTM! I have tested issuing a TSO command and escaping in order to view to cancellation message pop-up. The unit/integration tests seem to all be passing as well! Thanks for your work on this @roman-kupriyanov !

JillieBeanSim:
@roman-kupriyanov you are good to go to create the follow up PR to bring this into the `next` branch now :slightly_smiling_face:

PR #1673, opened 2022-02-15
Another approach for refreshing after config file creation/deletion

JillieBeanSim:
Silently reload Zowe Explorer to reset which treeProviders are used after types of profile handling are changed, old style => config, config -> old sytle, etc.  User is presented with a botton in the Popup Message after config file creation to `Refresh Zowe Explorer` other options for this action are the Command Palette `Zowe Explorer: Refresh Zowe Explorer` and quick keys `alt+z`

refresh-zowe-explorer

### DIFF
--- a/packages/zowe-explorer/__tests__/__unit__/extension.unit.test.ts
+++ b/packages/zowe-explorer/__tests__/__unit__/extension.unit.test.ts
@@ -122,6 +122,7 @@ async function createGlobalMocks() {
         mockExtension: null,
         appName: vscode.env.appName,
         expectedCommands: [
+            "zowe.extRefresh",
             "zowe.all.config.init",
             "zowe.ds.addSession",
             "zowe.ds.addFavorite",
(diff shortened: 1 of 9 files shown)


JillieBeanSim:
I would like to get the `Refresh Zowe Explorer` icon in the Zowe title bar or in the option in the `...` of the title bar but haven't figured out how to do it yet.

crawr:
Thanks @JillieBeanSim . The code makes sense to me as it is similar to our current UX but the silent refresh approach makes it better.

The config file was created and I'm able to use it. I'm getting the following errors, though, after clicking the refresh button in the popup:

crawr:
Yep, if I just refresh it without updating the file I get both messages. The top one was confusing since the config file was created successfully.
But if I update the host, everything goes smoothly. 👍

JillieBeanSim:
> Yep, if I just refresh it without updating the file I get both messages. The top one was confusing since the config file was created successfully. But if I update the host, everything goes smoothly. 👍

ok, cool. this is pre-existing even with old method of restarting VSC after the config is created.  It will be good to catch this error and tell the user to update the file with the host.

PR #1737, opened 2022-04-07
Refactor ProfilesCache

JillieBeanSim:
Signed-off-by: Billie Simmons <49491949+JillieBeanSim@users.noreply.github.com>

refactored the ProfilesCache

## Release Notes

Milestone:

Changelog:

## Types of changes

What types of changes does your code introduce to Zowe Explorer?
_Put an `x` in the boxes that apply_

- [ ] Bugfix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Updates to Documentation or Tests (if none of the other choices apply)

## Checklist

_Put an `x` in the boxes that apply. You can also fill these out after creating the PR. If you're unsure about any of them, don't hesitate to ask. We're here to help! This checklist will be used as reference for both the contributor and the reviewer_

- [ ] I have read the [CONTRIBUTOR GUIDANCE](https://github.com/zowe/vscode-extension-for-zowe/wiki/Best-Practices:-Contributor-Guidance) wiki
- [ ] PR title follows [Conventional Commits Guidelines](https://www.conventionalcommits.org/en/v1.0.0-beta.2/)
- [ ] PR Description is included
- [ ] gif or screenshot is included if visual changes are made
- [ ] `yarn workspace vscode-extension-for-zowe vscode:prepublish` has been executed
- [ ] All checks have passed (DCO, Jenkins and Code Coverage)
- [ ] I have added unit test and it is passing
- [ ] I have added integration test and it is passing
- [ ] There is coverage for the code that I have added
- [ ] I have tested it manually and there are no regressions found
- [ ] I have added necessary documentation (if appropriate)
- [ ] Any PR dependencies have been merged and published (if appropriate)

## Reminder

After a PR is merged into the `master` branch, create a PR from `master` to the `next` branch resolving any conflicts.

## Further comments

If this is a relatively large or complex change, kick off the discussion by explaining why you chose the solution you did and what alternatives you considered, etc...

Issue #1663, opened 2022-02-09
[next] Custom secure properties not loaded


zFernand0:
**Describe the bug**

The ZE extension won't load a profile (i.e. fails to validate) when provided with a team config that has a profile required property (host, port) in the `secure` array.

**To Reproduce**

1. Copy the `zowe.config.json` (see below) into a new workspace
2. Open the Zowe Explorer on the above workspace
3. Profile gets automatically added, but it's practically unusable 😢 
```json
{
    "$schema": "./zowe.schema.json",
    "profiles": {
        "ibmzosmf": {
            "type": "zosmf",
            "properties": {
                "basePath": "/ibmzosmf/api/v1"
            }
        },
        "base": {
            "type": "base",
            "properties": {
                "port": 1234,
                "rejectUnauthorized": true,
                "tokenType": "apimlAuthenticationToken"
            },
            "secure": [
                "host",
                "tokenValue"
            ]
        }
    },
    "defaults": {
        "zosmf": "ibmzosmf",
        "base": "base"
    },
    "autoStore": true
}
```

**Expected behavior**

All properties are loaded regardless of whether or not they are stored securely

**Screenshots**

![image](https://user-images.githubusercontent.com/37381190/153213398-5cd92553-f449-478a-980a-89874069e1b9.png)

**Desktop (please complete the following information):**

- OS: Windows
- Zowe Explorer Version: #1637 
- (Optional) Zowe CLI Version:
- (Optional) Are you using Secure Credential Store?

**Additional context**

Using a VSIX from #1637

JillieBeanSim:
@zFernand0 without the host in the config you will receive the error shown for missing host. isn't that expected behavior?

zFernand0:
I guess I would expect Zowe Explorer to work the same way as the CLI here, and be able to look for the host in the vault.
And if it's not set (in the vault or in the config file) then it can ask for it, and store it back (depending on the config file)

JillieBeanSim:
That could be a great enhancement to add after v2 goes out! :grin:

zFernand0:
The prompting for missing properties I agree that it could be an enhancement after v2 (as a 2.1).
However, I do believe that storing port or host securely and not reading it from ZE could be a patch release (e.g. 2.0.1)

JillieBeanSim:
[...]

PR #1743, opened 2022-04-11
fix(settings): Remove Secure Credential Setting and more

### LINKED ISSUE #1713: API ML Token doesn't work across platforms
**Describe the bug**

if an API ML token is obtained with Zowe CLI that token isn't recognized by Zowe Explorer and vice versa

### LINKED ISSUE #1736: Preferred services should be the same as CLI (`Zowe`)
Wondering if we should default the setting to `SCS_ZOWE_CLI_V2`

Thanks to how the saveCredentials is implemented, ZE will eventually migrate all secure properties to the new preferred service.
 
https://github.com/zowe/vscode-extension-for-zowe/blob/6c89e532b2571fb2e99f1d380237dfee465b203f/packages/zowe-explorer-api/src/security/KeytarCredentialManager.ts#L198-L201

In addition to that, we look at all services for credentials, so that shouldn't fail either 😋

_Originally posted by @zFernand0 in https://github.com/zowe/vscode-extension-for-zowe/pull/1693#discussion_r829122245_

### LINKED ISSUE #1738: Issues after prompt credentials with v1 profiles
Having a v1 profile without creds and doing an action will prompt the user for the credentials as expected.  But, after the prompting errors are popping up and the credentials are automatically stored in the keystore.  Not sure if autostore for v1 is expected results and we definitely shouldn't get the errors.  Have to click the view refresh to update the profile so it is usable. 
Errors:
image

### LINKED ISSUE #1739: Remove `Security: Credential Plugin` setting
We previously had a discussion about the setting for Security: Credential Plugin, should we remove or change it to something more useful like an true/false for autoStore or something?

_Originally posted by @JillieBeanSim in https://github.com/zowe/vscode-extension-for-zowe/issues/1735#issuecomment-1090822764_

We should default to the `Zowe` service (see #1736)

zFernand0:
## Proposed changes
In this PR:
- Fix #1739
- Fix #1736 
- Fix #1713 
- Fix #1738

## Release Notes

Milestone: Zowe v2

Changelog: 
- Remove `Security: Credential Plugin` Setting
- Default credential storage service to `Zowe` (same as Zowe CLI)
- Fixed issue with Token (and other secure properties) being out of sync between CLI and ZE
- Handle port and host being in different profiles (e.g. host in base and port in service)
  - This may cause merge conflicts with #1737 
  - Will need to re-test

## Types of changes

## Checklist

## Further comments

- For the issue below, is it normal to still see the error (missing creds) if the operation was cancelled?
  - #1738
- I'm afraid I'm still seeing the issue below on my local testing. 
  - #1663 
  - If it's just me, I'm happy to move forward and just close the issue 😋


zFernand0:
Known Issues:
- Imperative: Secure array not updated on logout. (only the token gets removed from the vault)
- ZE: After login, an extension refresh is required to be able to use the profile (since the config might have gotten updated)
- ZE: After logout, the icon changes to a project profile icon (instead of the home - for global profiles)

JillieBeanSim (approved):
Approving for now to get RC1 but one thing I see weird and is at an imperative level with the allowance of being able to use global and project level configs, if base profile is the same name in both the project level takes precedence so the global profile that uses the global base is trying to use the project level base breaking the global level profile.  Then when I create 2 bases with different names it breaks everything.
Unfortunately that makes the global level profile unusable when cant get the secured creds from it's base.  These are show stoppers that may need an RC2 from the CLI team to fix this for them and us as well since we had to update the CLI version we have as a dependency for another fix we needed and this new functionality was included.

PR #1778, opened 2022-05-04
Fix log4js activation issue

rudyflores:

## Proposed changes

Fix previously discovered activation issue with log4js by bumping to latest version 6.4.6.

## Release Notes

Milestone:

Changelog: Bump log4js version to 6.4.6 to fix activation for Che.

## Types of changes

## Checklist

## Further comments

### DIFF
--- a/package.json
+++ b/package.json
@@ -31,7 +31,6 @@
     "eslint-config-prettier": "^8.5.0",
     "eslint-plugin-prettier": "^4.0.0",
     "mocha": "^4.0.0",
-    "log4js": "6.4.6",
     "prettier": "2.6.0",
     "rimraf": "^3.0.2",
     "typescript": "^4.5.3",
@@ -42,6 +41,7 @@
   },
   "resolutions": {
     "**/ansi-regex": "^5.0.1",
+    "**/log4js": "6.4.6",
     "**/minimist": "^1.2.6",
     "**/moment": "^2.29.2"
   },
--- a/package.json
+++ b/package.json
@@ -31,6 +31,7 @@
     "eslint-config-prettier": "^8.5.0",
     "eslint-plugin-prettier": "^4.0.0",
     "mocha": "^4.0.0",
+    "log4js": "6.4.6",
     "prettier": "2.6.0",
     "rimraf": "^3.0.2",
     "typescript": "^4.5.3",
@@ -41,7 +42,6 @@
   },
   "resolutions": {
     "**/ansi-regex": "^5.0.1",
-    "**/log4js": "6.3.0",
     "**/minimist": "^1.2.6",
     "**/moment": "^2.29.2"
   },
--- a/packages/zowe-explorer-api/src/profiles/ProfilesCache.ts
+++ b/packages/zowe-explorer-api/src/profiles/ProfilesCache.ts
@@ -59,7 +59,7 @@ export class ProfilesCache {
     protected defaultProfileByType = new Map<string, imperative.IProfileLoaded>();
     protected profileManagerByType = new Map<string, imperative.CliProfileManager>();
     public constructor(protected log: imperative.Logger, protected cwd?: string) {
-        this.cwd = cwd != null ? fs.realpathSync.native(cwd) : undefined;
+        this.cwd = cwd != null ? fs.realpathSync(cwd) : undefined;
     }
 
     public async getProfileInfo(): Promise<imperative.ProfileInfo> {
--- a/packages/zowe-explorer-api/src/profiles/ProfilesCache.ts
+++ b/packages/zowe-explorer-api/src/profiles/ProfilesCache.ts
@@ -49,6 +49,13 @@ export function getZoweDir(): string {
     return zowe.getZoweDir();
 }
 
+export function getFullPath(anyPath: string): string {
+    if (os.platform() === "win32") {
+        return fs.realpathSync.native(anyPath);
+    }
+    return fs.realpathSync(anyPath);
+}
+
 export class ProfilesCache {
     public profilesForValidation: IProfileValidation[] = [];
     public profilesValidationSetting: IValidationSetting[] = [];
@@ -59,7 +66,7 @@ export class ProfilesCache {
     protected defaultProfileByType = new Map<string, imperative.IProfileLoaded>();
     protected profileManagerByType = new Map<string, imperative.CliProfileManager>();
     public constructor(protected log: imperative.Logger, protected cwd?: string) {
-        this.cwd = cwd != null ? fs.realpathSync(cwd) : undefined;
+        this.cwd = cwd != null ? getFullPath(cwd) : undefined;
     }
 
     public async getProfileInfo(): Promise<imperative.ProfileInfo> {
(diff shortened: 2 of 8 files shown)


phaumer (changes requested):
@rudyflores let's keep a resolution to 6.4.6 as Imperative would still include 6.4.0 otherwise.
@zFernand0 fyi.

package.json
```diff
@@ -41,7 +41,7 @@
   },
   "resolutions": {
     "**/ansi-regex": "^5.0.1",
-    "**/log4js": "6.3.0",
+    "**/log4js": "6.4.6",
```
zFernand0, line 44: Imperative is being updated to use log4js 6.4.6 as its dependency 9and the CLI will be updated shortly after)
Not sure if we want to keep the resolution here or not. 

I don't necessarily have a strong preference other than the resolutions list getting unnecessarily long.

packages/zowe-explorer-api/src/profiles/ProfilesCache.ts
```diff
@@ -59,7 +59,7 @@ export class ProfilesCache {
     protected defaultProfileByType = new Map<string, imperative.IProfileLoaded>();
     protected profileManagerByType = new Map<string, imperative.CliProfileManager>();
     public constructor(protected log: imperative.Logger, protected cwd?: string) {
-        this.cwd = cwd != null ? fs.realpathSync.native(cwd) : undefined;
+        this.cwd = cwd != null ? fs.realpathSync(cwd) : undefined;
```
zFernand0, line 62: Curious about this particular change.
I'm afraid that this could cause ZE and CLI to not be able to retrieve credentials from the vauls as the path referencing the vault property will contain a lower-case letter for the drive
`c:/path/to/config`

zFernand0 (changes requested):
Most of the changes look ok to me.
I just have one comment/question before I feel comfortable approving this PR.

packages/zowe-explorer-api/src/profiles/ProfilesCache.ts
rudyflores: Perhaps we could convert the path string to become all uppercase?

zFernand0: That won't work as path are case sensitive on *nix environments 😢

packages/zowe-explorer-api/src/profiles/ProfilesCache.ts
```diff
@@ -59,7 +66,7 @@ export class ProfilesCache {
     protected defaultProfileByType = new Map<string, imperative.IProfileLoaded>();
     protected profileManagerByType = new Map<string, imperative.CliProfileManager>();
     public constructor(protected log: imperative.Logger, protected cwd?: string) {
-        this.cwd = cwd != null ? fs.realpathSync.native(cwd) : undefined;
+        this.cwd = cwd != null ? getFullPath(cwd) : undefined;
```
phaumer, line 69: @zFernand0 I wrapped the realpathSync so we could support both platforms.

zFernand0 (approved):

zFernand0:
Thanks for the quick workaround.
I really wish there was a solution that apply to all platforms (and I thought the `.native(...)` call worked on all three (from my testing) but I guess I keep forgetting Che 😅

packages/zowe-explorer-api/src/profiles/ProfilesCache.ts
```diff
@@ -49,6 +49,13 @@ export function getZoweDir(): string {
     return zowe.getZoweDir();
 }
 
+export function getFullPath(anyPath: string): string {
+    if (os.platform() === "win32") {
+        return fs.realpathSync.native(anyPath);
+    }
+    return fs.realpathSync(anyPath);
+}
+
```
rudyflores, line 58: My suggestion on this function is to add a try/catch block to catch any possible errors this might through @phaumer

phaumer: So far we never saw this one crashing. This problem was so far only that the method itself was not found.

phaumer:
@zFernand0 I am still not sure what the root cause for this is. It feels like something that happened when we switched to node 14.

zFernand0:
> I am still not sure what the root cause for this is. It feels like something that happened when we switched to node 14.

It's in the TS Doc for the fsPath property
```typescript
		/**
		 * The string representing the corresponding file system path of this Uri.
		 *
		 * Will handle UNC paths and normalize windows drive letters to lower-case. Also
		 * uses the platform specific path separator.
		 *
		 * * Will *not* validate the path for invalid characters and semantics.
		 * * Will *not* look at the scheme of this Uri.
		 * * The resulting string shall *not* be used for display purposes but
		 * for disk operations, like `readFile` et al.
		 *
		 * The *difference* to the {@linkcode Uri.path path}-property is the use of the platform specific
		 * path separator and the handling of UNC paths. The sample below outlines the difference:
		 * ```ts
		 * const u = URI.parse('file://server/c$/folder/file.txt')
		 * u.authority === 'server'
		 * u.path === '/shares/c$/file.txt'
		 * u.fsPath === '\\server\c$\folder\file.txt'
		 * ```
		 */
		readonly fsPath: string;
```
Source: https://github.com/microsoft/vscode/blob/a40b90e9a49f0eb3033fb852bd7616993dd377e9/src/vscode-dts/vscode.d.ts#L1358-L1378

phaumer (dismissed):

packages/zowe-explorer/src/utils/ProfilesUtils.ts
```diff
@@ -14,8 +14,8 @@
 import * as vscode from "vscode";
 import * as fs from "fs";
 import * as globals from "../globals";
-import { Session, IProfile, IProfileLoaded, ProfileInfo, IConfigLayer } from "@zowe/imperative";
```
zFernand0, line 17: Seems like `IConfigLayer` is still needed in this file 😋 
Otherwise, builds are failing 😅

zFernand0 (changes requested):

zFernand0:
IConfigLayer added back.
Will merge if all builds are green 😋

zFernand0 (approved):

package.json
JillieBeanSim: @zFernand0 when was the cli/imperative scheduled to have this update in a release? I know this was mentioned 5 days ago, is that release out yet?

zFernand0: It is included in Imperative 5.2.2, CLI 7.1.3
Those versions should likely make their way to public NPM tomorrow 😋 
We can open another PR to include them throughout the week 😋

packages/zowe-explorer/src/utils/ProfilesUtils.ts
```diff
@@ -14,7 +14,7 @@
 import * as vscode from "vscode";
 import * as fs from "fs";
```
phaumer, line 15: This one still shows as unused for me @zFernand0, but I am approving to get this in. We can clean up when we add eslint perhaps.

phaumer (approved):

packages/zowe-explorer/src/utils/ProfilesUtils.ts
zFernand0: the `IConfigLayer` really ? 🤔 
https://github.com/zowe/vscode-extension-for-zowe/blob/450db8fa9a362ee35618a9682ccce7e3741b514d/packages/zowe-explorer/src/utils/ProfilesUtils.ts#L256

phaumer: No, the fs import

Issue #1870, opened 2022-07-13
ZE API registerCustomProfileType throws an error when team config file is in use.


uzuko001:
**Describe the bug**

Using the ZE API ProfilesCache().registerCustomProfileType() throws an error when team config file is in use since the folder `os.homedir()}/.zowe/profiles/${profileTypeName}` does not exist.
This means that the external profile type is not added and that makes them unreadable.

Trying to `initForZowe()` first does not help either, since it checks if team config file is in use before creating the folder specified above.

**To Reproduce**
1. Convert existing v1 profiles using zowe-cli (renames existing os.homedir()}/.zowe/profiles folder to profiles-old)
2. Try using rofilesCache().registerCustomProfileType() 

- OS: Windows 10
- Zowe Explorer Version: 2.1.0

**Additional context**
As extenders going for v2 conformance in our next release this is a major issue.

nickImbirev:
I upvote for the issue, this is a show stopper for our extension to adopt the ZE APIs in the upcoming release and to be conformant as well.

PR #1885, opened 2022-07-25
fix v1 profile not exists v2 exists case

KutluOzel-b:
Signed-off-by: KutluOzel-b <kutlu.ozel@broadcom.com>

## Proposed changes

Fix registerCustomProfileType for extenders throwing error while v1 profile not in use but v2 in use. 

Checking if v1 or v2 is used if so then registers profile.

## Release Notes

Milestone:

https://github.com/zowe/vscode-extension-for-zowe/issues/1870

Changelog:

## Types of changes

## Checklist

- [x ] PR title follows [Conventional Commits Guidelines](https://www.conventionalcommits.org/en/v1.0.0-beta.2/)
- [ x] There is coverage for the code that I have added


packages/zowe-explorer-api/src/profiles/ProfilesCache.ts
```diff
@@ -123,8 +123,7 @@ export class ProfilesCache {
      * @param {string} type Type of Profile, defaults to "zosmf" if nothing passed.
      *
      * @returns {IProfileLoaded[]}
-     */
-    public getProfiles(type = "zosmf"): zowe.imperative.IProfileLoaded[] {
+     */ public getProfiles(type = "zosmf"): zowe.imperative.IProfileLoaded[] {
```
roman-kupriyanov, line 126: Probably, unnecessary change here.

packages/zowe-explorer-api/src/profiles/ProfilesCache.ts
```diff
@@ -136,14 +135,17 @@ export class ProfilesCache {
...
      */
-    public registerCustomProfilesType(profileTypeName: string): void {
-        const exists = fs.existsSync(path.posix.join(`${os.homedir()}/.zowe/profiles/${profileTypeName}`));
-        if (!exists) {
-            throw new Error(
-                `Zowe Explorer Profiles Cache error: Tried to register a custom profile type named: ${profileTypeName} that does not yet exist. Extenders must call "zoweExplorerApi.getExplorerExtenderApi().initForZowe()" first.`
-            );
-        }
-        this.allExternalTypes.add(profileTypeName);
+    public registerCustomProfilesType(profileTypeName: string) {
```
roman-kupriyanov, line 138: According to the `eslint` rules the function return type is required. I would leave the `void` type in place.

packages/zowe-explorer-api/src/profiles/ProfilesCache.ts
```diff
@@ -136,14 +135,17 @@ export class ProfilesCache {
...
+        const existsV1Profile = fs.existsSync(path.posix.join(`${os.homedir()}/.zowe/profiles/${profileTypeName}`));
+        this.getProfileInfo().then((res) => {
```
roman-kupriyanov, line 140: This call requires a full `then..catch` block according to the `eslint` rules again. So I would suggest to add simple:
```
.then(...)
.catch((error) => { throw error; });
```
to fix the build.

roman-kupriyanov: An alternative solution might be to convert `registerCustomProfilesType` function into `async` function to fully use `await` operator. But since this is a public API, you may want to introduce a new `asyncRegisterCustomProfilesType` call and deprecate the existing one.

This approach would be preferable, because otherwise we are loosing error throwing from `registerCustomProfilesType` function itself (they will stuck in callbacks).

roman-kupriyanov (commented):
Hi @KutluOzel-b! Thank you very much for this change! 😃 

I tested it with our E4E extension which uses Zowe Extenders API very heavily and I can confirm that it fixed our issues 🥳 

Though I have some suggestions/improvements for the fixes, please consider them before releasing the fix.

packages/zowe-explorer-api/src/profiles/ProfilesCache.ts
KutluOzel-b: Thanks Roman your inputs much appreciated :)

packages/zowe-explorer-api/src/profiles/ProfilesCache.ts
```diff
@@ -131,7 +131,7 @@ export class ProfilesCache {
     /**
      * Used for extenders to register with Zowe Explorer that do not need their
      * profile type in the existing MVS, USS, and JES
-     *
+     * @deprecated Please use asyncRegisterCustomProfilesType
```
roman-kupriyanov, line 134:
```ts
    /**
     * @deprecated Please use {@link asyncRegisterCustomProfilesType}.
     * 
     * Used for extenders to register with Zowe Explorer that do not need their
     * profile type in the existing MVS, USS, and JES
     *
     * @param {string} profileTypeName Type of Profile, defaults to "zosmf" if nothing passed.
     *
     * @returns {void}
     */
```

Minor: a proposal to use a link for the new function and move `@deprecated` annotation to the top of the JSDoc.

roman-kupriyanov (approved):
Thanks for the updates @KutluOzel-b 👍 

I recheck once more with our extension, it still works and the errors are handled properly. Thank you again 😃 

I have a small suggestion for the deprecation message, please consider to apply.

roman-kupriyanov (dismissed):

JillieBeanSim:
Looks like we have a DCO issue on this branch, the fix is stated here https://github.com/zowe/vscode-extension-for-zowe/pull/1885/checks?check_run_id=7520732371

KutluOzel-b:
Rebase conflict, closing this pr and opening a new one.

PR #1887, opened 2022-07-27
(Fix) registerCustomProfileType for extenders throwing error while v1 profile not in use but v2 in use. 

KutluOzel-b:
Signed-off-by: KutluOzel-b <kutlu.ozel@broadcom.com>

## Proposed changes

Fix registerCustomProfileType for extenders throwing error while v1 profile not in use but v2 in use. 

Checking if v1 or v2 is used if so then registers profile.

## Release Notes

Milestone:

https://github.com/zowe/vscode-extension-for-zowe/issues/1870

Changelog:

## Types of changes

## Checklist

- [x ] PR title follows [Conventional Commits Guidelines](https://www.conventionalcommits.org/en/v1.0.0-beta.2/)
- [ x] There is coverage for the code that I have added

### DIFF
--- a/packages/zowe-explorer-api/src/profiles/ProfilesCache.ts
+++ b/packages/zowe-explorer-api/src/profiles/ProfilesCache.ts
@@ -497,13 +497,16 @@ export class ProfilesCache {
     /**
      * Used for extenders to register with Zowe Explorer that do not need their
      * profile type in the existing MVS, USS, and JES
-     * @param {string} profileTypeName Type of Profile, defaults to "zosmf" if nothing passed.
+     * @param {string} profileTypeName Type of Profile
      *
      * @returns {void}
      */
     public async asyncRegisterCustomProfilesType(profileTypeName: string): Promise<void> {
         const existsV1Profile = fs.existsSync(path.posix.join(`${os.homedir()}/.zowe/profiles/${profileTypeName}`));
-        const existsV2Profile = (await this.getProfileInfo()).usingTeamConfig;
+        const profileInfo = await this.getProfileInfo();
+        const existsV2Profile =
+            profileInfo?.usingTeamConfig &&
+            profileInfo.getAllProfiles().find((profile) => profile.profType === profileTypeName);
         if (!existsV1Profile && !existsV2Profile) {
             throw new Error(
                 `Zowe Explorer Profiles Cache error: Tried to register a custom profile type named: ${profileTypeName} that does not yet exist. Extenders must call "zoweExplorerApi.getExplorerExtenderApi().initForZowe()" first.`
--- a/packages/zowe-explorer-api/src/profiles/ProfilesCache.ts
+++ b/packages/zowe-explorer-api/src/profiles/ProfilesCache.ts
@@ -149,6 +149,7 @@ export class ProfilesCache {
     }
 
     /**
+     * @deprecated Please use {@link asyncRegisterCustomProfilesType}.
      * Used for extenders to register with Zowe Explorer that do not need their
      * profile type in the existing MVS, USS, and JES
      *
@@ -491,5 +492,23 @@ export class ProfilesCache {
             externalTypeArray.filter((exType) => registeredTypes.every((type) => type !== exType))
         );
         return allTypes;
+    }
+
+    /**
+     * Used for extenders to register with Zowe Explorer that do not need their
+     * profile type in the existing MVS, USS, and JES
+     * @param {string} profileTypeName Type of Profile, defaults to "zosmf" if nothing passed.
+     *
+     * @returns {void}
+     */
+    public async asyncRegisterCustomProfilesType(profileTypeName: string): Promise<void> {
+        const existsV1Profile = fs.existsSync(path.posix.join(`${os.homedir()}/.zowe/profiles/${profileTypeName}`));
+        const existsV2Profile = (await this.getProfileInfo()).usingTeamConfig;
+        if (!existsV1Profile && !existsV2Profile) {
+            throw new Error(
+                `Zowe Explorer Profiles Cache error: Tried to register a custom profile type named: ${profileTypeName} that does not yet exist. Extenders must call "zoweExplorerApi.getExplorerExtenderApi().initForZowe()" first.`
+            );
+        }
+        this.allExternalTypes.add(profileTypeName);
     }
 }
--- a/packages/zowe-explorer-api/src/profiles/ProfilesCache.ts
+++ b/packages/zowe-explorer-api/src/profiles/ProfilesCache.ts
@@ -149,7 +149,6 @@ export class ProfilesCache {
     }
 
     /**
-     * @deprecated Please use {@link asyncRegisterCustomProfilesType}.
      * Used for extenders to register with Zowe Explorer that do not need their
      * profile type in the existing MVS, USS, and JES
      *
@@ -158,12 +157,6 @@ export class ProfilesCache {
      * @returns {void}
      */
     public registerCustomProfilesType(profileTypeName: string): void {
-        const exists = fs.existsSync(path.posix.join(`${os.homedir()}/.zowe/profiles/${profileTypeName}`));
-        if (!exists) {
-            throw new Error(
-                `Zowe Explorer Profiles Cache error: Tried to register a custom profile type named: ${profileTypeName} that does not yet exist. Extenders must call "zoweExplorerApi.getExplorerExtenderApi().initForZowe()" first.`
-            );
-        }
         this.allExternalTypes.add(profileTypeName);
     }
 
@@ -492,26 +485,5 @@ export class ProfilesCache {
             externalTypeArray.filter((exType) => registeredTypes.every((type) => type !== exType))
         );
         return allTypes;
-    }
-
-    /**
-     * Used for extenders to register with Zowe Explorer that do not need their
-     * profile type in the existing MVS, USS, and JES
-     * @param {string} profileTypeName Type of Profile
-     *
-     * @returns {void}
-     */
-    public async asyncRegisterCustomProfilesType(profileTypeName: string): Promise<void> {
-        const existsV1Profile = fs.existsSync(path.posix.join(`${os.homedir()}/.zowe/profiles/${profileTypeName}`));
-        const profileInfo = await this.getProfileInfo();
-        const existsV2Profile =
-            profileInfo?.usingTeamConfig &&
-            profileInfo.getAllProfiles().find((profile) => profile.profType === profileTypeName);
-        if (!existsV1Profile && !existsV2Profile) {
-            throw new Error(
-                `Zowe Explorer Profiles Cache error: Tried to register a custom profile type named: ${profileTypeName} that does not yet exist. Extenders must call "zoweExplorerApi.getExplorerExtenderApi().initForZowe()" first.`
-            );
-        }
-        this.allExternalTypes.add(profileTypeName);
     }
 }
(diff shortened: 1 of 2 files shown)


JillieBeanSim:
I am not able to recreate the original issue, I had v1 style profiles and ran the cli convert command. Then had an extension installed that uses the registerCustomProfileType ZE API call and refreshed VSC.  The only error/warning that was thrown was the message from the ZE extender about the profiles I had set to use in their settings wasn't found.  All still initialized as they should.

JillieBeanSim:
Is there a public extension I can try that has seen this issue?

roman-kupriyanov:
Hey @JillieBeanSim 😃 Thank you a lot for checking it out!

This is very weird that the extension your mentioned is somehow passing that issue. Is it public? Could I take a look on how they use Zowe API calls? Maybe there is a way how to use it properly.

Unfortunately, our extension release containing ZE API usage is not published yet, but I would like to try to give you a glimpse on how we use the API and what is the issue for us.

We are using two custom profiles for our product. This is a part of code that does the initialization for them:
```ts
    const desiredProfileTypes = ['endevor', 'endevor-location'];
    const zoweExplorerExtenderApi:
      | ZoweExplorerApi.IApiExplorerExtender
      | undefined =
      ZoweVsCodeExtension.getZoweExplorerApi().getExplorerExtenderApi();
    if (!zoweExplorerExtenderApi) {
      return new Error(`Missing Zowe Explorer API`);
    }
    const cache = zoweExplorerExtenderApi.getProfilesCache();
    for (const profileType of desiredProfileTypes) {
      await zoweExplorerExtenderApi.initForZowe(profileType);
      cache.registerCustomProfilesType(profileType);
    }
```

Running this with the latest at the moment ZE version `v2.2.0` and just after `zowe config convert` command is issued, first of all, we are getting this error:
```
Cannot read properties of undefined (reading 'forEach').
```

This one is related to the fact that we are not providing any schemas for our custom profiles, because we are going to work in read-only mode with our profiles, and the schemas array is optional anyway for `initZoweApi()` call. Anyway, this bug is addressed by this PR #1884.

Let's say, to avoid this bug, we are now providing the list of schemas for the init call. So, the next issue we are getting right away is this:
```
Zowe Explorer Profiles Cache error: Tried to register a custom profile type named: endevor that does not yet exist. Extenders must call "zoweExplorerApi.getExplorerExtenderApi().initForZowe()" first..
```

This happens, because the conversion Zowe CLI command actually renames our v1 `profiles` folder to `profiles-old` as a backup and creates a Team Config for us instead, but the `registerCustomProfilesType()` call still expects that the folder with `.zowe/profiles/<profileType>` name is existing in `profiles` folder, even though we switched to Team Config now.

As far as I understood, `initZoweApi()` call is aware of the Team Config usage and also does not create any folders, even though we provided a list of schemas to it, because now probably they should be included in the `zowe.schema.json` instead.

So the actual problem here is that the init call is aware of the v2 profiles are used, but the register call is still not and expecting some v1 folders to exist even in the v2 profiles mode.

So, I am very curious to see how the extension with a custom profile you mentioned is avoiding this problem 😄

JillieBeanSim:
Hey @roman-kupriyanov, in the case of the extension I was testing with we do pass the meta info/schema to initForZowe() which actually recreates the profile directory with meta info during that call but only when no team config is in use (This is another issue raised that is being addressed).  This could be the root of the problem for what you are seeing with the error even when passing schema info.  The passing of the meta info allows the extensions in the VS Code space to be independent of the CLI and not require Zowe CLI or plugins to be installed.  Also without the meta info passed the new profile type won't be considered for the creation of the config file via ZE.  There is a disconnect between the CLI and VS Code where we don't recognize the CLI plugins and the CLI doesn't recognize the VSC ZE extensions yet.  We are working with the imperative team on this. 

Is there a reason your team doesn't want to pass schema info?

On another note, with just investigating the code changes, for v1 we are checking that the type exists in the profile folder.  With v2 shouldn't we check that the type exists in the config schema vs just checking for existence of config file? @KutluOzel-b

KutluOzel-b:
@JillieBeanSim Good catch with the profileTypeName check thanks !  Going to add it.

JillieBeanSim:
Thanks @KutluOzel-b.  @roman-kupriyanov would you mind re-testing and I believe squad testing can use the CICS extension to recreate and test updates since it's creates a new profile type not included in Zowe CLI core.

JillieBeanSim:
@KutluOzel-b DCO failed, instructions to fix are here https://github.com/zowe/vscode-extension-for-zowe/pull/1887/checks?check_run_id=7561853908

roman-kupriyanov:
Hey @JillieBeanSim! Thanks for continuing this investigation 😃 

> Is there a reason your team doesn't want to pass schema info?

We just thought this is not our primary case to provide the schemas and to init some profiles there. In the current state our VSCode extension will work only in the readonly mode with the custom Zowe profiles. There is actually a Zowe CLI plugin, which introduces and controls the schemas, which also provides full functionality to work with those profiles, to update them, etc. through Zowe CLI. Since we considered `initForZowe()` call has the list of schemas as an optional argument, we can continue without providing the schemas and will be able to read required profiles if they present. 

In other words we need a way to try to read the custom profiles from Zowe API and if there are some - show it to the user, if not - just move along.

> @roman-kupriyanov would you mind re-testing

I did test the latest state of the fix and it works, but only for the case when we have at least one profile of the provided type in the Team Config. If it contains no such profiles, we are getting the error below again, no matter if we are calling `initForZowe()` with or without the schemas:

```
Zowe Explorer Profiles Cache error: Tried to register a custom profile type named: endevor that does not yet exist. Extenders must call "zoweExplorerApi.getExplorerExtenderApi().initForZowe()" first..
```

KutluOzel-b:
> @KutluOzel-b DCO failed, instructions to fix are here https://github.com/zowe/vscode-extension-for-zowe/pull/1887/checks?check_run_id=7561853908

@JillieBeanSim  I think I'm facing the issue mentioned here : https://github.com/dcoapp/app/issues/121

In commits it appears as it is signed and verified but dco bot checking fails.

I'm sharing this just in case for someone else getting same error in future.

roman-kupriyanov:
Hey @JillieBeanSim, thank you very much for the explanation of the issue with the schemas 😃 I think this is pretty much clear now what are the implications of that. Am I right that in the current state if the user will try to write not only own custom profiles, but also any of the core profiles using ZE (zosmf, uss, etc.), it would overwrite the existing schemas?

> Hopefully a decision will be made soon on how meta info can be appended to an existing schema vs only having the option from imperative to overwrite it at this time.

Yeah, if I understood the issue correctly that seems a good way to approach it. Meanwhile, we will discuss it within our team.

> A way to handle testing for existence or just moving along could include a try/catch around the call to initForZowe() and registerCustomProfileTypes() and if that error is returned due to that type not being in the config schema or in the v1 profiles folder on top of the fix we are trying to supply here.

But how the users can create their very first profile if they cannot even register the custom type until at least one profile already exist? This would suggest that they always need to create their first profile outside of the VSC extension and ZE API. Is it true for the Team config case?

Also, `initForZowe()` cannot fix the case in any ways possible for this situation, but the Error message sort of implies it: `Extenders must call "zoweExplorerApi.getExplorerExtenderApi().initForZowe()" first..`, which is confusing. The v1 Profiles users, on the other hand, can fix it by calling init function first, since it creates the required directories to pass the check, but not the v2 users, unfortunately.

As the extenders we would expect it to work as something like this:

```ts
    // extension activation
    const profileType = 'my-profile';
    // ... get zoweExplorerExtenderApi ...
    const cache = zoweExplorerExtenderApi.getProfilesCache();
    await zoweExplorerExtenderApi.initForZowe(profileType);
    cache.registerCustomProfilesType(profileType);
    // later in the code we can use a cache object to do different operations with the custom profiles
    // for now we are going to ask for the list of them
    const profiles = cache.getProfiles(profileType);
    // profiles === [] if no profiles defined in the Team config
```

It allows us to init/register custom type ZE API at our extension activation and to use `ProfilesCache` object later, when some commands are called. But without proper registration first (because of try/catched `Zowe Explorer Profiles Cache error`), we need to interpret the `undefined` return value from `cache.getProfiles()` call as "no profiles" state for each of the calls performed, which in certain cases might be an indication of other issues during the init.

JillieBeanSim:
> > A way to handle testing for existence or just moving along could include a try/catch around the call to initForZowe() and registerCustomProfileTypes() and if that error is returned due to that type not being in the config schema or in the v1 profiles folder on top of the fix we are trying to supply here.
> 
> But how the users can create their very first profile if they cannot even register the custom type until at least one profile already exist? This would suggest that they always need to create their first profile outside of the VSC extension and ZE API. Is it true for the Team config case?
> 
To register the custom type for v2 instance I suggested our fix include checking the config schema for the particular types meta info is included.  This will show that the types' cli plugin is installed.  similar to our v1 check for the creation of the types folder within the profile's folder.

packages/zowe-explorer-api/src/profiles/ProfilesCache.ts
```diff
@@ -497,13 +497,16 @@ export class ProfilesCache {
...
      *
      * @returns {void}
      */
     public async asyncRegisterCustomProfilesType(profileTypeName: string): Promise<void> {
         const existsV1Profile = fs.existsSync(path.posix.join(`${os.homedir()}/.zowe/profiles/${profileTypeName}`));
-        const existsV2Profile = (await this.getProfileInfo()).usingTeamConfig;
+        const profileInfo = await this.getProfileInfo();
+        const existsV2Profile =
+            profileInfo?.usingTeamConfig &&
+            profileInfo.getAllProfiles().find((profile) => profile.profType === profileTypeName);
```
JillieBeanSim, line 509: @KutluOzel-b this will only work if a profile of that type already exists, it there a call we can make to imperative to check against the type's existence in the configs schema?

KutluOzel-b: @JillieBeanSim Isn't that we check for in the first place , that if the specified profile type is already exists or not ? I'm having lack of knowledge here. We can get loaded profile from cli too though but I'm can't tell if this is the right call.

JillieBeanSim: I am starting to think that we shouldn't require the check for v1 or v2 profiles types and just have this method use the final line `this.allExternalTypes.add(profileTypeName);` to register the type and leave any checks for existing profiles be done by extenders using this.  What do you think @KutluOzel-b

KutluOzel-b: @JillieBeanSim according to the need of extenders since profileTypeName is optional or it is irrelevant to have one more right to say, I agree with you that we don't need to check any existing profiles but to add  directly via `this.allExternalTypes.add(profileTypeName);`

roman-kupriyanov:
Hey @KutluOzel-b, @JillieBeanSim!

I retested the latest changes with our extension and it is initialize and register the custom profile types successfully 👍 Thank you so much!

The only obstacle left is that `cache.getProfiles()` call returns `undefined` if no profiles of requested type created yet. Is it ok for us to go and interpret the `undefined` return value as "no profiles" from now on? Can it be an indication of something else wrong in other cases?

JillieBeanSim:
@Rajpreetgill could you test this with the CICS ext?  you will have to build and package this then install the packaged Zowe Explorer into VSCode then point the CICS ext dependency of the ZE API to the branches packaged ZE API and test that activation is successful following the recreation steps in the linked issue.

Rajpreetgill (approved):
Looks great, it has been tested with the CICS extension. Thanks!

PR #1891, opened 2022-07-29
Zowe Explorer still activate when error reading config file

JillieBeanSim:

## Proposed changes

update returned error message to include location of file and issue with file, also opening the file for users to address issues.  allow views to be created.  profilesCache.refresh returns empty profile arrays if issues reading config files. initForZowe will continue if error calling getProfileInfo due to issue reading config file.

## Release Notes

Milestone: 2.2.1

Changelog: Address Zowe Explorer activation issues due to invalid Team Config file.

## Types of changes

## Checklist

https://user-images.githubusercontent.com/49491949/181829135-d585c5a2-e44a-4b11-b6d5-62dfd9feeeda.mov


packages/zowe-explorer/src/utils/ProfilesUtils.ts
```diff
@@ -243,31 +243,93 @@ export function getProfile(node: vscode.TreeItem) {
...
+                            : "Not available"
+                    } `
+            );
+            globals.LOG.debug(
+                "Summary of team configuration files considered for Zowe Explorer: %s",
+                JSON.stringify(layerSummary)
+            );
+        }
+    } catch (error) {
+        if (error.message.toString().includes("Error parsing JSON")) {
```
kristinochka, line 275: Could there be another profile error that does not result in this error message? Looks like if error message is different, Zowe Explorer will fail to activate. Checking for error message strings is not a robust approach.

kristinochka: I think if Imperative fails to read profile for any reason it should still activate, and open the file if it exists.

packages/zowe-explorer/src/ZoweExplorerExtender.ts
```diff
@@ -97,9 +97,16 @@ export class ZoweExplorerExtender implements ZoweExplorerApi.IApiExplorerExtende
...
-            mProfileInfo = await getProfileInfo(globals.ISTHEIA);
+        let mProfileInfo: imperative.ProfileInfo;
+        try {
+            mProfileInfo = await globals.PROFILESCACHE.getProfileInfo();
+            if (!mProfileInfo) {
+                mProfileInfo = await getProfileInfo(globals.ISTHEIA);
+            }
+        } catch (error) {
+            if (!error.toString().includes("Error parsing JSON")) {
+                throw new Error(error);
```
kristinochka, line 108: Same here regarding checking for error message string.

packages/zowe-explorer/src/shared/refresh.ts
```diff
@@ -29,22 +26,23 @@ import { createJobsTree } from "../job/ZosJobsProvider";
...
-            resetValidationSettings(sessNode, setting);
-            returnIconState(sessNode);
-            await syncSessionNode(Profiles.getInstance())((profileValue) =>
-                ZoweExplorerApiRegister.getCommonApi(profileValue).getSession()
-            )(sessNode);
-        }
-    });
-    treeProvider.refresh();
+    if (treeProvider.mSessionNodes.length > 0) {
+        treeProvider.mSessionNodes.forEach(async (sessNode) => {
```
kristinochka, line 31: Why does it need to check the length? forEach will not execute the loop if array is empty.

kristinochka:
One issue I noticed:
If I have a **valid** global profile and **invalid** project profile, Zowe Explorer fails to activate
I think Zowe Explorer should behave as though local config does not exist in this situation. 

image

JillieBeanSim:
> One issue I noticed: If I have a **valid** global profile and **invalid** project profile, Zowe Explorer fails to activate I think Zowe Explorer should behave as though local config does not exist in this situation.
> 
> image
> 
> I'm actually seeing ZE fails to activate with the same error message if any profile is invalid. I am running activation-fix branch and `yarn` command ran successfully. I did have a bunch of rse profiles added before making my JSON invalid.

those messages about the activation are errors created by Z Open Editor from a try/catch with other calls that will have to be addressed on that side and not ZE.  The only message coming from ZE is the error with the path to JSON file and location of the error within the file.  Also if the user has v1 and v2 profile management imperative will only recognize the v2 and if there is an error in the JSON file it will not fall back to v1 it will give the error and wait for the fix in v2 profile management.

packages/zowe-explorer/src/shared/refresh.ts
JillieBeanSim: with the return from ProfilesCache arrays possibly being empty it is good to have a check for length of the sessionNodes in tree before the forEach call

kristinochka: If treeProvider.mSessionNodes is an empty array, no need to check for length because forEach loop will not execute.

JillieBeanSim: addressed in commit https://github.com/zowe/vscode-extension-for-zowe/pull/1891/commits/42158f234ca77c0251a97aa22e2433b4f9c40b12

packages/zowe-explorer/src/utils/ProfilesUtils.ts
JillieBeanSim: this is already handled whether it is a parsing error or an fs error the error will be caught and message shown continuing with activation.  Only with parsing error the config file in question will be opened

JillieBeanSim:
https://user-images.githubusercontent.com/49491949/182421220-f94fb7e5-8bbb-49f4-8579-ed417a89712c.mov

JillieBeanSim:
Since the error is thrown from Imperative while reading information from the disk, if the user has both, global and project level config, and the error is in only one that error will need to be fixed before either of the config files can be used.

rudyflores (dismissed):
LGTM! Thanks for the fix @JillieBeanSim !

rudyflores (dismissed):

kristinochka:
Zowe Explorer fails to activate if user creates project config file without a hostname. Refreshing the Zowe Extension does not re-activate

JillieBeanSim:
> LGTM! 😋 Also, thanks for using the imperative instance that comes with the CLI package 😉

I figured we could go ahead and start to change files over we are working on :slightly_smiling_face:

rudyflores (approved):
LGTM!

Issue #1428, opened 2021-08-04
Change context menus for multi-select


katelynienaber:
If multiple nodes are selected, the context menus should only show commands which support multi-select.

For example, delete is available for multiple nodes at once, but cancellation of jobs is not available. So Cancel Job should be hidden when multiple jobs are selected in the tree.

zFernand0:
Please refer to 7f97adfeb2cdeecf7b93c994e56c9d01df45e6f2 for the latest code changes.
`git checkout 7f97adfeb2cdeecf7b93c994e56c9d01df45e6f2`

KutluOzel-b:
Currently we are not allowing users to open multiple editor tabs at the same time. Is it intentional ? Considering selection of multiple files , allowing edit / pull from mainframe etc.. does not seems logical if we have no intention of allowing users to edit multiple files at the same time. 

My question is : I can't see any harm of allowing users to edit multiple files at the same time . Do you see any pitfalls in here ?

zFernand0:
Personally, I don't see any harm in allowing such behavior. 😋

PR #1882, opened 2022-07-22
Feature/multi select/context menu/uss

KutluOzel-b:
## Proposed changes

Added multiple selection to USS view context menu 

-  pull from mainframe
- add to fav
- refresh directory
- toggle binary / toggle text 
- delete node 

Fix error in check fav files when deleting node

## Release Notes

Milestone:

https://github.com/zowe/vscode-extension-for-zowe/issues/1428

completed USS view part 

Changelog:

## Types of changes

What types of changes does your code introduce to Zowe Explorer?
_Put an `x` in the boxes that apply_

- [x ] Bugfix (non-breaking change which fixes an issue)
- [ x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Updates to Documentation or Tests (if none of the other choices apply)

## Checklist

- [x ] I have read the [CONTRIBUTOR GUIDANCE](https://github.com/zowe/vscode-extension-for-zowe/wiki/Best-Practices:-Contributor-Guidance) wiki
- [ x] PR title follows [Conventional Commits Guidelines](https://www.conventionalcommits.org/en/v1.0.0-beta.2/)
- [ x] PR Description is included
- [ ] gif or screenshot is included if visual changes are made
- [x ] `yarn workspace vscode-extension-for-zowe vscode:prepublish` has been executed
- [ ] All checks have passed (DCO, Jenkins and Code Coverage)
- [ ] I have added unit test and it is passing
- [ ] I have added integration test and it is passing
- [x ] There is coverage for the code that I have added
- [ x] I have tested it manually and there are no regressions found
- [ ] I have added necessary documentation (if appropriate)
- [ ] Any PR dependencies have been merged and published (if appropriate)

PR #1917, opened 2022-08-19
Feature/multi select/context menu/data sets

KutluOzel-b:
## Proposed changes

Added multiple selection to manage context menu of Datasets and Jobs views

**Datasets** 

- Removed Allocate Alike
- Removed Create New Member
- Removed Rename
- Add / Remove Favourites
- Add Delete 
- Add Migrate Datasets
- Add Pull From Mainframe
- Add Recall Datasets
- Add Refresh
- Add Show Datasets attributes
- Add Upload Member
- Add Edit

**Jobs**

- Removed Issue Modify Command
- Add / Remove Favourites
- Add delete
- Add Download Files
- Add Get JCL (multiple jcl in a single file ) 
- Add Issue Multiple Stops

## Release Notes

Milestone:

https://github.com/zowe/vscode-extension-for-zowe/issues/1428

Completed DS and Jobs Views parts

## Types of changes

- [ x] New feature (non-breaking change which adds functionality)
- [ x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist

- [x ] I have read the [CONTRIBUTOR GUIDANCE](https://github.com/zowe/vscode-extension-for-zowe/wiki/Best-Practices:-Contributor-Guidance) wiki
- [ x] PR title follows [Conventional Commits Guidelines](https://www.conventionalcommits.org/en/v1.0.0-beta.2/)
- [x ] All checks have passed (DCO, Jenkins and Code Coverage)
- [x ] There is coverage for the code that I have added
- [ x] I have tested it manually and there are no regressions found


zFernand0 (changes requested):
There are a few unexpected things with the add to favorites compared to what the `main` branch does 😢 

PR: uses the selected node
Main: uses the node clicked

Similar behavior happens on other trees
Here is a GIF to help explain what's happening (only includes the Jobs tree)
  my intention was to do it for all trees, but the gif got a bit too long
temp

KutluOzel-b:
Thanks @zFernand0 for reviewing it and good catch 👍 . That made me realize that we have the same issue on other views as well and leads some other bugs like selection of unintended files . So I changed the getting of selected nodes method and fixed it at https://github.com/zowe/vscode-extension-for-zowe/pull/1917/commits/bc294be97327f9e0c8af14993fe167ab3ca4b64e

The modifications also includes https://github.com/zowe/vscode-extension-for-zowe/pull/1882 this pr context. 

It seems it is fixed now but being skeptical for both jobs / uss / ds views would be beneficial while testing.

JillieBeanSim (changes requested):
DATA SETS

- multi-select upload member didn't upload into both datasets only the latest selected
- ds member pull from mainframe flips through files in editor for each selected with only a single tab showing at all times when file not already opened.  seems like strange behavior. maybe have it behave like edit does, that one works really well.
- migrate & recall still require the tree view to be refreshed to see changes causing the sessions to close, not relevant to this PR, I believe it's prior behavior but noting here to remember to create issue for this

JOBS

- same issue here as in DS view with pull from mainframe multi select
- get JCL only opened the first selected of multiple in text editor

KutluOzel-b:
@JillieBeanSim Thanks for the review and comments 👍 

### DATA SETS 
> multi-select upload member didn't upload into both datasets only the latest selected 

> >

Uploading multiple files(members) to a single dataset is what is available now. Still it shouldn't show up on the context menu though when multi selection happens.I thought It was unnecessary to add such behaviour. What do you think ? Is there a use case for it or would be nice to add ?

> ds member pull from mainframe flips through files in editor for each selected with only a single tab showing at all times when file not already opened. seems like strange behavior. maybe have it behave like edit does, that one works really well.
> >
Resolved in commit : https://github.com/zowe/vscode-extension-for-zowe/pull/1917/commits/a44b7384c7d69ebcaf0a5235b13891abb7c70bb7

### JOBS

> same issue here as in DS view with pull from mainframe multi select

>>

Resolved in commit : https://github.com/zowe/vscode-extension-for-zowe/pull/1917/commits/a44b7384c7d69ebcaf0a5235b13891abb7c70bb7

>get JCL only opened the first selected of multiple in text editor
>>
This is actually related with the approach . I'm adding JCL statements together into a single file. I thought it would be more elegant way to manage multiple JCL Statements. What do you think ? 
here what is it looks like ;

 
GetJCL

JillieBeanSim:
Hey @KutluOzel-b thanks for addressing my comments.

For the getJCL, I don't think it is a good idea to concatenate in to a single file.  We may in the future bring in a functionality to edit and re-submit and this could prove problematic.

For the DS upload, I don't know of a use case for uploading the same file into different DS's.  If the option can be removed from the right-click menu that would be good to avoid bugs/issues being raised if users try and it doesn't work.

zFernand0:
based on our quick discussion on standup, it seems like we still need to remove the getJCL and upload DS on multi-selection 😋 
Once those are removed, I'm happy to approve.

packages/zowe-explorer/src/job/actions.ts
```diff
@@ -406,16 +408,6 @@ async function deleteMultipleJobs(
...
-    for await (const jobContent of jobContentList) {
-        docContent += jobContent + "\n";
-    }
-    const jclDoc = await vscode.workspace.openTextDocument({ language: "jcl", content: docContent });
-    await vscode.window.showTextDocument(jclDoc);
-}
-
 // class Spool extends Job {
 //     constructor(
 //         label: string,
```
zFernand0, line 413: curious if this commented-out class be removed as well ?

zFernand0 (approved):
LGTM! 😋 
Thanks for addressing the feedback so quickly 👍

packages/zowe-explorer/src/job/actions.ts
KutluOzel-b: Thanks I didn't changed the lines but also don't see any reason to leave them there. Hence removing them ..

JillieBeanSim (approved):
LGTM! thanks @KutluOzel-b for the fixes and updates

zFernand0 (approved):
LGTM 😋

PR #1969, opened 2022-10-12
Add option for non secure profile creation

rudyflores:

## Proposed changes

Create non-secure Zowe v2 profiles when the `zowe.security.enableSecureCredentials` VS Code setting is set to false. The default value of this new VS Code setting is set tot `true`.

## Release Notes

Milestone: 2.4.0

Changelog: Added option for creating non-secure Zowe v2 profiles when setting `zowe.security.enableSecureCredentials` to false.

## Types of changes

## Checklist


packages/zowe-explorer/i18n/sample/package.i18n.json
```diff
@@ -77,6 +77,7 @@
   "zowe.commands.history": "Toggle if Commands persist locally",
   "zowe.automaticProfileValidation": "Allow automatic validation of profiles.",
   "zowe.settings.version": "Track if migration has been processed",
+  "zowe.security.enableSecureCredentials": "Store username and password for z/OS as clear text in Zowe team config profiles if disabled",
```
kristinochka, line 80: I think better naming is `zowe.security.disableSecureCredentials`

because it's enabled by default and nothing needs to be done. Option enableSecureCredentials makes it seems like secure credentials need to be enabled, like it's an optional/experimental feature. In reality it's the default option, and if user wants to disable it, they should be tweaking `disableSecureCredentials` setting.

kristinochka (changes requested):

packages/zowe-explorer/i18n/sample/package.i18n.json
JillieBeanSim: unchecking the setting box the value changes to false in the settings.json so this request would not make much sense.  possibly have it as `secureCredentialsEnabled` so when unticked that will = false

kristinochka: I updated to secureCredentialsEnabled

JillieBeanSim, line 80: This name doesn't match the name used in package.json `zowe.security.secureCredentialsEnabled`

JillieBeanSim (changes requested):

packages/zowe-explorer/i18n/sample/package.i18n.json
kristinochka: Done.

kristinochka:
It's detecting duplicated code in the unit test file that was there already.

rudyflores:
> It's detecting duplicated code in the unit test file that was there already.

The code seems to have existed before this PR, should we create an issue and address this in a separate PR instead? @kristinochka @JillieBeanSim

JillieBeanSim:
> > It's detecting duplicated code in the unit test file that was there already.
> 
> The code seems to have existed before this PR, should we create an issue and address this in a separate PR instead? @kristinochka @JillieBeanSim

I don't think we need to worry about the dup code in test files, I am sure there is a lot of dup code in test files.  We can add the to-do to the test coverage story or as a part of that work in a sep story.

phaumer:
@rudyflores @kristinochka @JillieBeanSim I tested it together with the changes from https://github.com/zowe/vscode-extension-for-zowe/pull/1979. 

I works really well for the happy day scenario if the user specifies a correct team config file. It is definitely is a great improvement and could be delivered as is.

However, if the user makes a mistake they are still broken. If I have not read the docs and use code completion to add secure fields back in and save then Zowe Explorer is broken. Zowe Explorer refuses to open the broken file back in the editor so I could fix the problem. Zowe Explorer + menus still appear, but when I click any option I just get this dialog and the editor does not open. We need to catch these exceptions and always open the editor back up.



The only way to fix it is use vi to edit the team config file in the terminal. 

Another solution is to add a json validator that (a) checks the file for correctness against the schema and (b) makes sure it does not contain secure fields if secure is disabled.

packages/zowe-explorer/src/extension.ts
```diff
@@ -121,7 +121,49 @@ export async function activate(context: vscode.ExtensionContext): Promise<ZoweEx
...
+        const isSecureCredsEnabled = vscode.workspace
+            .getConfiguration()
+            .get(globals.SETTINGS_SECURE_CREDENTIALS_ENABLED);
+
+        if (isImperativeSettingEnabled !== isSecureCredsEnabled) {
+            const yesButton = localize("zoweSecurity.disable.button", "Yes");
+            const keepCurrentSettingButton = localize("zoweSecurity.keepEnabled.button", "Keep current setting");
+            const infoMsg = localize(
+                "zoweSecurity.setting.infoMessage",
+                `Imperative.json setting appears to not match VS Code setting zowe.security.secureCredentialsEnabled, do you want to allow Zowe Explorer to correct this setting?`
```
kristinochka, line 149: I think user may be confused by this message, especially if they're not familiar with these settings, like imperative.json. Unless you're developer I don't expect user to understand that file.

packages/zowe-explorer/i18n/sample/package.i18n.json
```diff
@@ -77,6 +77,7 @@
   "zowe.commands.history": "Toggle if Commands persist locally",
   "zowe.automaticProfileValidation": "Allow automatic validation of profiles.",
   "zowe.settings.version": "Track if migration has been processed",
+  "zowe.security.secureCredentialsEnabled": "Store username and password for z/OS as clear text in Zowe team config profiles if disabled",
```
JillieBeanSim, line 80: if we are setting autoStore to false during creation the info will not be stored in plain text.

packages/zowe-explorer/package.nls.json
```diff
@@ -77,6 +77,7 @@
   "zowe.commands.history": "Toggle if Commands persist locally",
   "zowe.automaticProfileValidation": "Allow automatic validation of profiles.",
   "zowe.settings.version": "Track if migration has been processed",
+  "zowe.security.secureCredentialsEnabled": "Store username and password for z/OS as clear text in Zowe team config profiles if disabled",
```
JillieBeanSim, line 80: same as before with setting autoStore: false during creation

JillieBeanSim (changes requested):
I tried changing the setting on an existing config file and didn't get prompted about the file not matching the setting or any messages about having secure arrays in file. I wasn't notified about the difference after creating a new file with toggled setting either. Do we want to have the actions within extension.ts for this setting/command in a separate method and add the setting to the `vscode.workspace.onDidChangeConfiguration()` call, line 210 extension.ts,  to watch for changes and make appropriate method call to check and update?

I get a lot of these errors when running debugger.
`rejected promise not handled within 1 second: Error: command 'zowe.toggleSecureCredentials' already exists`

packages/zowe-explorer/i18n/sample/package.i18n.json
phaumer: I like this, because we still want to provide the more secure solution of getting prompted every time by default. The user needs to explicitly make it less secure by enabling autoStore, which will store in clear text (on the encrypted drive).

phaumer (changes requested):
This works for the happy day scenario of users providing a correct .zowe/settings/imperative.json file as well as set the user setting. However, if I delete that file and reload I do not get prompted of the problem. I also see in the code that it would toggle the VS Code user setting, but that was not the requirement. The VS Code user setting should enforce that the file exists, recreate if not, and ensure it contains a false value for the CredentialsManager. 

The dialog should never change the VS Code setting. The user would have to do that in the VS Code settings itself, because before they can do that they would have to create a proper entry for the CredentialsManager. Remember, that in the future it needs to point to K8s Credentials Manager, which has to be user configured.

packages/zowe-explorer/i18n/sample/package.i18n.json
JillieBeanSim: I agree with the creation setup, this is just a comment about the description of new setting in the Settings.  Could just reword to mention `if autoStore: true credentials will be stored in plain text.`

phaumer (approved):
Great work!

phaumer (approved):

Issue #1549, opened 2021-10-25
Copy/Paste and Multiple USS Files


jellypuno:
As a ZE User, I would like to have the capability of copy and pasting files and directories inside another directory.

Goal:
1. Copy/Paste files and directories within 1 profile

Possible UX:
1. The user will select multiple files or directories
2. Right-Click > Copy
3. Right-Click on another directory > Paste
4. All selected items are copied to the new directory
    - This will only cover copy/paste. Move is a different issue
    - Can we use Ctrl + C and Ctrl + V?

Note: Same UX goes for the Stretch but you are just copying it to a different profile

jellypuno:
[Stretch] Copy/Paste files and directories across different profiles

- Should be a different issue. We should consider file conversion

jellypuno:
Zowe CLI is not able to move but it can copy. So we can remove move from our acceptance criteria

jellypuno:
Create a separate issue for:
- Copy/paste USS files (including multi-select) for different profiles
- Research/Implement on Move USS files

PR #2004, opened 2022-11-09
Feature/uss/copy

KutluOzel-b:
## Proposed changes

Adds multiselect copy / paste feature to uss view 

## Release Notes

https://github.com/zowe/vscode-extension-for-zowe/issues/1549

Milestone:

Changelog:

## Types of changes

- [ x] New feature (non-breaking change which adds functionality)

## Checklist

- [ x] PR title follows [Conventional Commits Guidelines](https://www.conventionalcommits.org/en/v1.0.0-beta.2/)
- [ x] I have added unit test and it is passing
- [ x] I have tested it manually and there are no regressions found


packages/zowe-explorer-api/CHANGELOG.md
```diff
@@ -11,6 +11,10 @@ All notable changes to the "zowe-explorer-api" extension will be documented in t
 
 ## `2.4.0`
 
+## TBD Release
```
JillieBeanSim, line 14: this should be above the 2.4.0 release

packages/zowe-explorer-api/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@zowe/zowe-explorer-api",
-  "version": "2.5.0-SNAPSHOT",
+  "version": "2.6.0-SNAPSHOT",
```
JillieBeanSim, line 3: this needs to be reverted back to 2.5.0-SNAPSHOT

packages/zowe-explorer-ftp-extension/package.json
```diff
@@ -47,7 +47,7 @@
     "vscode": "^1.53.2"
   },
   "dependencies": {
-    "@zowe/zowe-explorer-api": "2.5.0-SNAPSHOT",
+    "@zowe/zowe-explorer-api": "2.6.0-SNAPSHOT",
```
JillieBeanSim, line 50: same with revert to 2.5.0-SNAPSHOT

packages/zowe-explorer/CHANGELOG.md
```diff
@@ -16,6 +16,10 @@ All notable changes to the "vscode-extension-for-zowe" extension will be documen
 
 ## `2.4.0`
 
+## TBD Release
```
JillieBeanSim, line 19: move CHANGELOG above the 2.4.0 release tag

packages/zowe-explorer/package.json
```diff
@@ -1736,7 +1776,7 @@
     "@types/promise-queue": "^2.2.0",
     "@types/selenium-webdriver": "^3.0.4",
     "@types/yargs": "^11.0.0",
-    "eslint-plugin-zowe-explorer": "2.5.0-SNAPSHOT",
+    "eslint-plugin-zowe-explorer": "2.6.0-SNAPSHOT",
```
JillieBeanSim, line 1779: 2.5.0-SNAPSHOT

packages/zowe-explorer/package.json
```diff
@@ -1776,7 +1816,7 @@
     "webpack-cli": "^3.3.11"
   },
   "dependencies": {
-    "@zowe/zowe-explorer-api": "2.5.0-SNAPSHOT",
+    "@zowe/zowe-explorer-api": "2.6.0-SNAPSHOT",
```
JillieBeanSim, line 1819: 2.5.0-SNAPSHOT

JillieBeanSim (changes requested):
Hey @KutluOzel-b I commented on a couple things that need to be changed

KutluOzel-b:
@JillieBeanSim  Thanks for reviewing and feedbacks. Adressed all in : https://github.com/zowe/vscode-extension-for-zowe/pull/2004/commits/767a7038808cd1599b8f618f6e07520fd0edeb42

JillieBeanSim (changes requested):
thanks for the updates @KutluOzel-b I am seeing some issues with copy/paste.  
1. copy and paste are both always in the right click menu, DS copy/paste has only copy available until something is in the clipboard then the option changes to paste.
2. uss paste fails if user only right clicks on a node then selects paste, if the node is selected then the right click menu opened and paste selected it will succeed.  In DS tree this is not the case, it will succeed if user only right clicks the node without selecting the node first.
3. user isn't presented with an option to change the file name in uss like they are with DS.  this one I am not sure if expected behavior or if we should consider allowing the rename of the copied file.

traeok (changes requested):
I tried copying 3 files from my home directory to a newly-created folder: after selecting the files and clicking Copy File(s), I right-clicked on the new directory and clicked Paste. The following exception occurred after VSCode showed "Uploading copied files..." in the status bar, and it seemed to stop the rest of the logic from executing:

exception-uss-copy

I refreshed the USS panel after the exception occurred, but noticed that the directory/file structure is intact and the files were not copied into the directory.

KutluOzel-b:
> 

@JillieBeanSim  Thanks for reviewing it. here is the solutions / discussions on the feedbacks. 
1. **copy and paste are both always in the right click menu, DS copy/paste has only copy available until something is in the clipboard then the option changes to paste.**

- I'm not sure we have control over context menu dynamically. Made some research on it and on documented features of api there isn't listed such. 

2. **uss paste fails if user only right clicks on a node then selects paste, if the node is selected then the right click menu opened and paste selected it will succeed. In DS tree this is not the case, it will succeed if user only right clicks the node without selecting the node first.**

- good catch , adressed in : c99252cabfd323e821a497e0ea33a2fb497b1a25

3. **user isn't presented with an option to change the file name in uss like they are with DS. this one I am not sure if expected behavior or if we should consider allowing the rename of the copied file.**

- since it is not like datasets ( which has to have unique identifier for all files ) I thought adding (copy) to the end of filename is better UX if there is a file with the same name under the same directory. I'm also not sure if we should provide renaming while copying. I say we should add it only for single copy and if it is asked by users.

KutluOzel-b:
> I tried copying 3 files from my home directory to a newly-created folder: after selecting the files and clicking Copy File(s), I right-clicked on the new directory and clicked Paste. The following exception occurred after VSCode showed "Uploading copied files..." in the status bar, and it seemed to stop the rest of the logic from executing:
> 
> exception-uss-copy
> 
> I refreshed the USS panel after the exception occurred, but noticed that the directory/file structure is intact and the files were not copied into the directory.
> 
> As @JillieBeanSim suggested, I tried to paste with the directory node selected first, and also tested with just right-clicking on that node, but both methods failed to paste the files.
> 
> This _could_ be a z/OSMF-specific issue, so I will do some more testing within USS to confirm - but regardless we'll probably want to catch any exceptions that arise when trying to perform those actions.

Good catch @traeok thanks ! Added exception handler and resolved selection issue in c99252cabfd323e821a497e0ea33a2fb497b1a25

traeok:
@KutluOzel-b I tried to copy a file and paste it into the directory that I mentioned, but received the same error - this time the exception was handled and execution stopped as expected 👍 I will do some more debugging to confirm the bug is not something on my end. I verified that I am able to copy files to that folder within a USS shell.

KutluOzel-b:
> > 1. **copy and paste are both always in the right click menu, DS copy/paste has only copy available until something is in the clipboard then the option changes to paste.**
> > 
> > 
> > * I'm not sure we have control over context menu dynamically. Made some research on it and on documented features of api there isn't listed such.
> 
> For this I would look at how Data Sets tree is handling it, see gif attached
> 
> copy:paste [ copy:paste ](https://user-images.githubusercontent.com/49491949/206228701-189939f7-50f9-4e93-9000-d31df45abc23.gif) [ ](https://user-images.githubusercontent.com/49491949/206228701-189939f7-50f9-4e93-9000-d31df45abc23.gif)

@JillieBeanSim Thanks for the feedback. This doesn't also happens in datasets. In member view item it is always copy and in pds view Item it is always paste. Unfortunately it doesn't reacts according to clipboard contents 😞

packages/zowe-explorer/src/extension.ts
```diff
@@ -873,7 +886,7 @@ function initSubscribers(context: vscode.ExtensionContext, theProvider: IZoweTre
     }
 }
 
-function getSelectedNodeList(node: IZoweTreeNode, nodeList: IZoweTreeNode[]): IZoweTreeNode[] {
+export function getSelectedNodeList(node: IZoweTreeNode, nodeList: IZoweTreeNode[]): IZoweTreeNode[] {
```
JillieBeanSim, line 889: @KutluOzel-b this function seems to be a duplicate of getSelectedNodeList() in shared/utils?  We should only have one of these functions especially if they are exactly the same and I would suggest having it stay in the shared/utils file vs adding more to the extension.ts file.

JillieBeanSim (dismissed):
while testing, if I haven't copied anything and right-click on a directory and select paste the status bar mentions uploading but nothing is actually created.  If I paste into a directory with file name already existing that I copied it automatically overwrites without notifying the user and allowing them to rename the file/folder and since the option is there even if folder is closed, this would be a good check to have.

I would still like to research if we can find a way to check the vsc clipboard and only offer paste if something is in there.

packages/zowe-explorer/src/uss/ZoweUSSNode.ts
```diff
@@ -646,6 +646,59 @@ export class ZoweUSSNode extends ZoweTreeNode implements IZoweUSSTreeNode {
...
+            stageName: 0,
+        };
+        const options: IUploadOptions = {
+            task,
+        };
+        if (prof.profile.encoding) {
+            options.encoding = prof.profile.encoding;
+        }
+        const api = ZoweExplorerApiRegister.getUssApi(this.profile);
+        const apiResponse = await api.fileList(remotePath);
```
t1m0thyj, line 669: Should we wrap this API call in a try-catch block and handle any errors?

packages/zowe-explorer/src/uss/ZoweUSSNode.ts
```diff
@@ -646,6 +646,59 @@ export class ZoweUSSNode extends ZoweTreeNode implements IZoweUSSTreeNode {
...
+        const fileList = apiResponse.apiResponse?.items;
+
+        for (const localFile of localFileNames) {
+            try {
+                if (localFile.endsWith("/")) {
+                    const directoryPath = localFile.slice(0, -1);
+                    await api.uploadDirectory(
+                        directoryPath,
+                        remotePath + "/" + directoryPath.split("/").pop(),
```
t1m0thyj, line 678: This won't work on Windows - should we use `path.basename` instead?

packages/zowe-explorer/src/uss/ZoweUSSNode.ts
```diff
@@ -646,6 +646,59 @@ export class ZoweUSSNode extends ZoweTreeNode implements IZoweUSSTreeNode {
...
+                        options
+                    );
+                } else {
+                    let fname = localFile.split("/").pop();
```
t1m0thyj, line 682: This won't work on Windows - should we use `path.basename` instead?

t1m0thyj (changes requested):

packages/zowe-explorer/src/extension.ts
KutluOzel-b: @JillieBeanSim Thanks for the warning. This is caused by pulling from main branch. Keeping it in shared/utils

traeok:
I tested these changes and I'm able to copy files now on Windows 👍 I might have found an edge case - copying an empty directory into another directory. If a folder has files inside, I can copy it into another directory, but in the case of an empty folder nothing happens. No exceptions occur, but the directory structure remains the same.

It's unlikely that a user would want to copy an empty directory into another directory when they can just right-click on the parent -> "Create Directory" - but just wanted to mention it.

traeok (approved):

t1m0thyj (dismissed):
Thanks for fixing the issues with copying on Windows 🙂 

A few questions regarding the implementation:
* When you right-click on a USS directory, should the menu item say "Copy Directory" instead of "Copy Files"?
* If I have loaded directory `/a` in the USS tree that contains a subdirectory `b`, I don't see a way to create a copy of `/a/b` at the same level (`/a/b (copy)`). Should we add "Paste" to the right-click menu for the root USS node?

traeok (approved):

t1m0thyj (approved):

Issue #1974, opened 2022-10-14
ZE: `allocate-like` does not respect the `blksz` parameter


zFernand0:
**Describe the bug**

The `blksz` seems different in the new dataset (compared to the original).
This may not be expected by end users.

**To Reproduce**

1. Go to the ZE's Dataset tree
2. Right-click on a `zosmf` profile and search for an existing `data-set`
3. Right-click on the `data-set` and select `Show Data Set Attributes`
4. Right-click on the `data-set` and select `Allocate Like...`
5. Provide name for the new data set (e.g. `data-set-like`)
6. Right-click on the `data-set-like` and select `Show Data Set Attributes`
7. Compare the two `blksz` properties

**Expected behavior**

I believe the user would expect the `blksz` to be the same between `data-set` and `data-set-like`

**Screenshots**

![temp](https://user-images.githubusercontent.com/37381190/195883092-1fd6318f-94a1-4e62-a59c-eae6aa6ac83e.gif)

**Desktop (please complete the following information):**

- OS: Windows 10
- Zowe Explorer Version: 2.3.0
- (Optional) Zowe CLI Version: N/A
- (Optional) Are you using Secure Credential Store? N/A

**Additional context**

FYI, the GIF also showcases the following issue:
- #1973

traeok:
Seems like this issue is coming from z/OSMF's REST API Client - after fixing the anomaly bug in #1973, all attributes seem to be copied to the new dataset, as intended, w/ the exception of `blksz`.

Did some debugging on our calls to z/OSMF REST Client and the only attribute that we pass is an `options` object structured like so:

`{ like: 'USER.DATASET.ALLOC.LIKE' }`

Will soon report this through the necessary pipeline for z/OSMF issues.

traeok:
Going to reopen this as we will be addressing the API-specific issue with the z/OSMF team.

traeok:
We received a response back from the z/OSMF team confirming that the 'LIKE' API function used for the 'Allocate Like' feature does not support copying the blksz parameter. As the API does not support the required functionality to fix this issue, this issue will be closed.

https://www.ibm.com/docs/en/zos/2.4.0?topic=statement-like-parameter

Issue #1973, opened 2022-10-14
Anomaly message during an ALLOCATE LIKE


ROBIN-Olivier:
Each time I want to allocate a PDS, I get an error message:
Error while executing zowe.ds.allocateLike command: "Cannot read properties of undefined (reading 'pattern'). Probably due to the extension contributing zowe.ds.allocateLike.

However, the PDS seems well created...

**To reproduce:**
In the ZOWE EXPLORER view, right click on a PDS, then Allocate Like
Enter the new name, then Enter
The following message appears in the notifications:
![allocate like](https://user-images.githubusercontent.com/103997342/195869670-e8a171ba-83d9-429b-86d3-b016f8bb5f1c.png)

**Desktop (please complete the following information):**
OS: Windows
Zowe Explorer Version: v2.3.0
profile: zosmf

zFernand0:
I was able to reproduce the problem but it only happens the first time you use `Allocate-like` after the Zowe Explorer extension is loaded.

FWIW, I believe `allocate-like` does not respect the Block size option when creating the new dataset

ROBIN-Olivier:
Thank you

PR #2015, opened 2022-11-21
Fix: "Allocate Like" execution stops due to error

traeok:
## Proposed changes

- Fixes bug described in #1973 to allow `allocateLike` to continue execution
- **Update:** Originally #1974 was factored into this issue, but the block size problem appears to be an issue with z/OSMF REST client.
- Also updated logic in `removeFavorite` to skip `undefined` nodes

## Release Notes

Milestone: 2.5.0

Changelog: Fixed issue where "Allocate Like" feature fails to grab dataset node during execution

## Types of changes

## Checklist

### DIFF
--- a/packages/zowe-explorer/CHANGELOG.md
+++ b/packages/zowe-explorer/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 All notable changes to the "vscode-extension-for-zowe" extension will be documented in this file.
 
+## TBD Release
+
+### Bug fixes
+
+- Fixed "Allocate Like" feature failing to respect original dataset params and pattern (if specified). [#1974](https://github.com/zowe/vscode-extension-for-zowe/issues/1974)
+
 ## `2.4.1`
 
 ### Bug fixes
--- a/packages/zowe-explorer/CHANGELOG.md
+++ b/packages/zowe-explorer/CHANGELOG.md
@@ -7,7 +7,7 @@ All notable changes to the "vscode-extension-for-zowe" extension will be documen
 ### Bug fixes
 
 - Fixed missing localization for certain VScode error/info/warning messages. [#1722](https://github.com/zowe/vscode-extension-for-zowe/issues/1722)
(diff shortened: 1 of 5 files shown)


JillieBeanSim:
@traeok @zFernand0 should we get a couple unit tests for scenario to bring patch percentage up?

t1m0thyj (commented):
This PR fixes the error that was thrown by the Allocate Like command 👍 

It seems like blksize is still not respected. The CNTL2 data set on the right was created by right-clicking on the CNTL data set on the left and selecting "Allocate Like":
image

I believe this is not a problem with Zowe Explorer, but rather it is the way that z/OSMF behaves. The CLI command `zowe files create data-set` with the `--like` flag has the same behavior where blksize is not respected. Perhaps on #1974, we should leave a comment to document this as a known issue with z/OSMF?

My only request for changes on this PR would be to add some tests as @JillieBeanSim suggested 🙂 While increasing "project" coverage is more of a long term goal, we do want to make sure that individual PRs have adequate "patch" coverage.

traeok:
> I believe this is not a problem with Zowe Explorer, but rather it is the way that z/OSMF behaves. The CLI command `zowe files create data-set` with the `--like` flag has the same behavior where blksize is not respected. Perhaps on #1974, we should leave a comment to document this as a known issue with z/OSMF?

Was able to validate this as we only send one attribute to the z/OSMF REST Client for the "Attribute Like" command: `{ like: 'USERNAME.DATASET.TEST' }`. Renamed and reformatted this issue to reference the anomaly bug as that's what the fix technically solves - Allocate Like should function as intended once the block size issue is fixed within z/OSMF's REST API.

> My only request for changes on this PR would be to add some tests as @JillieBeanSim suggested 🙂 While increasing "project" coverage is more of a long term goal, we do want to make sure that individual PRs have adequate "patch" coverage.

Should be good to go now 👍

JillieBeanSim (approved):
LGTM! thanks @traeok for the fix

PR #2020, opened 2022-11-22
Fix: Cannot delete favorited PS, PDS, or member

traeok:
## Proposed changes

- Sync tree nodes in favorites list with their actual state under "Data Sets" panel

## Release Notes

Milestone: 2.5.0

Changelog: Fixed issue where Favorites section was not refreshing after changes were made to data sets or their members.

## Types of changes

## Checklist

## Further comments

- **Future Proposal:** Change method of storing favorites such that the nodes can be mapped 1:1; this should shorten lookup time for corresponding nodes and make logic more straightforward for adding/removing favorites. Since we are searching based on labels, we could assign the node labels as keys for a map-based structure.
- **For next standup:** The `findFavoritedNode` and `findNonFavoritedNode` functions are marked deprecated, but they're being used in the codebase. Reimplement (as suggested above) or keep the same data structure?

### DIFF
--- a/packages/zowe-explorer/src/dataset/actions.ts
+++ b/packages/zowe-explorer/src/dataset/actions.ts
@@ -247,35 +247,24 @@ export async function deleteDatasetPrompt(
     selectedNodes = selectedNodes.filter((val) => !childArray.includes(val));
 
     if (includedSelection || !node) {
-        // Filter out sessions, favorite nodes, or information messages
+        // Filter out sessions and information messages
         nodes = selectedNodes.filter(
             (selectedNode) =>
                 selectedNode.getParent() &&
-                !contextually.isFavorite(selectedNode) &&
-                !contextually.isFavorite(selectedNode.getParent()) &&
                 !contextually.isSession(selectedNode) &&
                 !contextually.isInformation(selectedNode)
         );
     } else {
-        if (
-            node.getParent() &&
-            !contextually.isFavorite(node) &&
-            !contextually.isFavorite(node.getParent()) &&
-            !contextually.isSession(node) &&
-            !contextually.isInformation(node)
-        ) {
+        if (node.getParent() && !contextually.isSession(node) && !contextually.isInformation(node)) {
             nodes = [];
             nodes.push(node);
         }
     }
 
-    // Check that there are items to be deleted, this can be caused by trying to delete favorites right now
+    // Check that there are items to be deleted
     if (!nodes || nodes.length === 0) {
         vscode.window.showInformationMessage(
-            localize(
-                "deleteDatasetPrompt.nodesToDelete.empty",
-                "Deleting data sets and members from the Favorites section is currently not supported."
-            )
+            localize("deleteDatasetPrompt.nodesToDelete.empty", "No data sets selected for deletion, cancelling...")
         );
         return;
     }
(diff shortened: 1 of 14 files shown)


JillieBeanSim (dismissed):
This works great! thanks @traeok for the fix

packages/zowe-explorer/src/dataset/actions.ts
```diff
@@ -438,6 +428,17 @@ export async function createMember(
             true,
             datasetProvider
         );
+
+        // Refresh corresponding tree parent to reflect addition
+        const findCorrespondingNode = contextually.isFavorite(parent)
+            ? datasetProvider.findNonFavoritedNode.bind(datasetProvider)
+            : datasetProvider.findFavoritedNode.bind(datasetProvider);
```
t1m0thyj, line 435: Can we call the `findEquivalentNode` method here to reduce duplication?

packages/zowe-explorer/src/dataset/actions.ts
```diff
@@ -1046,11 +1049,25 @@ export async function deleteDataset(node: api.IZoweTreeNode, datasetProvider: ap
...
+    datasetProvider.removeFavorite(node);
+
+    const isMember = contextually.isDsMember(node);
+
+    // If the node is a dataset member, go up a level in the node tree
+    // to find the relevant, matching node
+    const nodeOfInterest = isMember ? node.getParent() : node;
+    const parentNode = fav
+        ? datasetProvider.findNonFavoritedNode(nodeOfInterest)
+        : datasetProvider.findFavoritedNode(nodeOfInterest);
```
t1m0thyj, line 1064: Can we call the `findEquivalentNode` method here to reduce duplication (same question as above)?

packages/zowe-explorer/src/dataset/actions.ts
traeok: Yep! Not sure how I missed that, I added the wrapper function but didn't update the logic to use it 🤦‍♂️

t1m0thyj (approved):
Tested these changes and the favorites tree always seems to stay in sync with other data set trees now 👍

zFernand0 (dismissed):
TL;DR: LGTM! 😋 

At first I was afraid that `IZoweNodeType -> IZoweTreeNode` might be a breaking change to extenders, but after some trial and error, it seems that this should be fine as long as every type in `IZoweNodeType` extends `IZoweTreeNode` (which they are).

traeok:
Sorry @zFernand0 - had to re-request review as the rebase marked previous reviews as stale..

Is there a way to avoid this in the future or is it just an inevitable result of rebasing/updating branches to reflect main?

zFernand0:
> had to re-request review as the rebase marked previous reviews as stale. Is there a way to avoid this in the future or is it just an inevitable result of rebasing/updating branches to reflect main?

Thanks for sticking with a Rebase model 😋 
It seems like it's either dismiss reviews automatically after new commits or don't 😢 
image

zFernand0:
If we want to keep the history clean, we could do the same thing that the `zowe/docs-site` does.
- Disallow Merge Commits (in favor of squash merging)
image

zFernand0 (approved):
LGTM! 😋

t1m0thyj (approved):

Issue #1967, opened 2022-10-12
Refactor UI/UX code to stand alone and be reusable


JillieBeanSim:
Create a file for UI/UX methods that can be used across Zowe Explorer to promote easy maintainability

PR #2042, opened 2022-12-13
Use allowlist instead of blocklist in vscodeignore

t1m0thyj:
Remove extra files from the VSIX for both ZE and the FTP extension. Using an allowlist instead of a blocklist should make the `.vscodeignore` files easier to maintain in the future, as long as we remember to update the ignore list in any future PRs where new files should be included in the VSIX.

## Release Notes

Milestone: 2.6.0

Changelog: Not sure if we need one for this change? The only change users may notice is a smaller download size.

## Types of changes

What types of changes does your code introduce to Zowe Explorer?
_Put an `x` in the boxes that apply_

- [x] Bugfix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Updates to Documentation or Tests (if none of the other choices apply)

## Checklist

_Put an `x` in the boxes that apply. You can also fill these out after creating the PR. If you're unsure about any of them, don't hesitate to ask. We're here to help! This checklist will be used as reference for both the contributor and the reviewer_

- [ ] I have read the [CONTRIBUTOR GUIDANCE](https://github.com/zowe/vscode-extension-for-zowe/wiki/Best-Practices:-Contributor-Guidance) wiki
- [ ] PR title follows [Conventional Commits Guidelines](https://www.conventionalcommits.org/en/v1.0.0-beta.2/)
- [ ] PR Description is included
- [ ] gif or screenshot is included if visual changes are made
- [x] `yarn workspace vscode-extension-for-zowe vscode:prepublish` has been executed
- [ ] All checks have passed (DCO, Jenkins and Code Coverage)
- [ ] I have added unit test and it is passing
- [ ] I have added integration test and it is passing
- [ ] There is coverage for the code that I have added
- [ ] I have tested it manually and there are no regressions found
- [ ] I have added necessary documentation (if appropriate)
- [ ] Any PR dependencies have been merged and published (if appropriate)

PR #2047, opened 2022-12-14
feat(gui): UI refactor

traeok:
## Proposed changes

- Consolidates `vscode.window` function calls into one `Gui` module for standalone use: #1967 

## Release Notes

Milestone: 2.6.0

Changelog: 
- All `vscode.window` functions were consolidated into one `Gui` module for standalone use. #1967 

## Types of changes

## Checklist


packages/zowe-explorer-api/src/globals/Gui.ts
```diff
@@ -0,0 +1,276 @@
...
+        return vscode.window.showQuickPick(items, options, token);
+    }
+
+    /**
+     * Displays a message to the user, including any specified options.
+     * @param message The message to display
+     * @param options Any additional options for the message
+     * @returns A thenable containing the selected item (if items were specified), or `undefined`
+     */
+    export function showMessage(message: string, options?: GuiMessageOptions): Thenable<string> {
```
zFernand0, line 210: Should all exported functions have as few parameters as possible (following this one as an example)?

For instance, from `createWebviewPanel(type, title, showOptions, options)` to `createWebviewPanel(title, options: GuiWebviewPanelOptions)`

I guess I'm hoping that this will be more flexible if we need to add optional properties in the future 😋 

This may apply to other places, and maybe we can refactor some other interfaces and exports to have as few parameters as possible. 😋 

P.S. this is not a request, just a suggestion😋

packages/zowe-explorer-api/src/logger/IZoweLogger.ts
```diff
@@ -30,7 +30,7 @@ type Log4JsCfg = {
 
 const LOGGER_CONFIG: Log4JsCfg = loggerConfig;
 
-export enum MessageSeverityEnum {
+export enum MessageSeverity {
```
zFernand0, line 33: will this be a possible breaking change for anyone having the `@zowe/zowe-explorer-api` as a dependency?

One way to avoid the braking change is to add a type with the old name
```javascript
/**
 * @deprecated Please use MessageSeverity
 */
export type MessageSeverityEnum = MessageSeverity;
```

packages/zowe-explorer-api/src/vscode/ZoweVsCodeExtension.ts
```diff
@@ -45,24 +45,6 @@ export class ZoweVsCodeExtension {
         return undefined;
     }
 
-    /**
-     * Reveal an error in VSCode, and log the error to the Imperative log
-     *
-     */
-    public static showVsCodeMessage(message: string, severity: MessageSeverityEnum, logger: IZoweLogger): void {
```
zFernand0, line 52: instead of removing them, we could 
```typescript
/**
 * @deprecated Please use Gui.showMessage
 */
```

Then, make this a wrapper of `Gui.showMessage`

packages/zowe-explorer-api/src/vscode/ZoweVsCodeExtension.ts
```diff
@@ -136,20 +118,11 @@ export class ZoweVsCodeExtension {
         return undefined;
     }
 
-    public static async inputBox(inputBoxOptions: vscode.InputBoxOptions): Promise<string> {
```
zFernand0, line 139: same as the previous comment.

Just trying to prevent any possible disruptions before v3 (2024)

packages/zowe-explorer-api/src/logger/IZoweLogger.ts
```diff
@@ -76,25 +77,25 @@ export class IZoweLogger {
      * Log an error to the Imperative log
      *
      */
-    public logImperativeMessage(message: string, severity: MessageSeverityEnum): void {
```
zFernand0, line 79: wondering if we should
```typescript
public logImperativeMessage(message: string, severity: MessageSeverityEnum);
public logImperativeMessage(message: string, severity: MessageSeverity) { ... }
```

zFernand0 (changes requested):
Thanks for taking a look. 😋 
And sorry I didn't get t o review this sooner 😢 

This is looking great btw! 🥳

packages/zowe-explorer-api/src/logger/IZoweLogger.ts
traeok: Thinking that this will be covered by TypeScript as the deprecated type is now aliasing the renamed enum, but can do some local testing to confirm.

packages/zowe-explorer-api/src/vscode/ZoweVsCodeExtension.ts
```diff
@@ -46,21 +47,25 @@ export class ZoweVsCodeExtension {
...
+     * @param logger The IZoweLogger object for logging the message
      *
+     * @deprecated Please use `Gui.showMessage` instead
      */
-    public static showVsCodeMessage(message: string, severity: MessageSeverityEnum, logger: IZoweLogger): void {
-        logger.logImperativeMessage(message, severity);
-
-        const errorMessage = `${logger.getExtensionName()}: ${message}`;
+    public static showVsCodeMessage(message: string, severity: MessageSeverity, logger: IZoweLogger): void {
+        void Gui.showMessage(message, { severity: severity, logger: logger });
```
zFernand0, line 58: I'm still trying to understand the difference between using the `void` vs `await`.
I believe both may perform a void statement in practice.

I guess I don't see a problem with statements like these
```javascript
undefined; // same as `await promise-void`
1234; // same as `await promise-number`
true; // same as `await promise-boolean`
```

packages/zowe-explorer-api/src/vscode/ZoweVsCodeExtension.ts
```diff
@@ -136,20 +141,11 @@ export class ZoweVsCodeExtension {
...
-        }
-        return await vscode.window.showInputBox(inputBoxOptions);
-    }
-
     private static async saveCredentials(profile: imperative.IProfileLoaded): Promise<boolean> {
         let save = false;
         const saveButton = "Save Credentials";
         const message = `Save entered credentials in plain text for future use with profile ${profile.name}?\nSaving credentials will update the local information file.`;
-        await vscode.window.showInformationMessage(message, { modal: true }, ...[saveButton]).then((selection) => {
+        await Gui.showMessage(message, { items: [saveButton], vsCodeOpts: { modal: true } }).then((selection) => {
```
zFernand0, line 148: And I think this is why I'm asking. 😋 
Why not just use `void` here too if we don't plan to use the value (which should be `undefined`)?

I'm not suggesting we change the use of the `void` operator.
These questions are just to help me (and hopefully others) understand this a bit better.

packages/zowe-explorer-ftp-extension/src/ZoweExplorerAbstractFtpApi.ts
```diff
@@ -109,11 +107,11 @@ export abstract class AbstractFtpApi implements ZoweExplorerApi.ICommon {
                         "Invalid Credentials. Please ensure the username and password for " +
                         validateProfile?.name +
                         " are valid or this may lead to a lock-out.";
-                    ZoweVsCodeExtension.showVsCodeMessage(errMsg, MessageSeverityEnum.ERROR, ZoweLogger);
+                    await Gui.errorMessage(errMsg, { logger: ZoweLogger });
                     throw new Error();
                 } else {
                     // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access
-                    ZoweVsCodeExtension.showVsCodeMessage(e.message as string, MessageSeverityEnum.ERROR, ZoweLogger);
+                    await Gui.errorMessage(e.message as string, { logger: ZoweLogger });
```
zFernand0, line 114: For consistency, should these two `errorMessage` calls be `void`? 😋

zFernand0 (approved):
LGTM! 😋 
I promise my questions/comments were mostly for my own understanding.
I'm happy to dismiss the 2 code smells due to the `void` operator if we determine that's the way we want to go.

packages/zowe-explorer-ftp-extension/src/ZoweExplorerAbstractFtpApi.ts
```diff
@@ -60,10 +59,9 @@ export abstract class AbstractFtpApi implements ZoweExplorerApi.ICommon {
 
     public checkedProfile(): imperative.IProfileLoaded {
         if (!this.profile?.profile) {
-            ZoweVsCodeExtension.showVsCodeMessage(
+            void Gui.showMessage(
```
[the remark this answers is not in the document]
traeok: Currently `checkedProfile` is a synchronous function, by voiding the Promise we can call it without using `await` - we can update `checkedProfile` to be `async` but will likely cause a cascade of functions that need changed to be `async` in order to `await checkedProfile`.

In hindsight we can probably use `.then(() => ...)` in this scenario, in which we would throw the error, and avoid having to use `void` entirely here.

For context the original `ZoweVsCodeExtension.showVsCodeMessage` was taking the same approach but within the function itself:

image

packages/zowe-explorer-api/src/globals/Gui.ts
```diff
@@ -0,0 +1,275 @@
...
+        });
+    }
+
+    /**
+     * Wrapper function for `showMessage` to display an informational message.
+     * @param message The message to display
+     * @param options Additional options for the displayed message
+     * @returns A thenable containing the selected item (if items were specified), or `undefined`
+     */
+    export function infoMessage(message: string, options?: Omit<GuiMessageOptions, "severity">): Thenable<string> {
```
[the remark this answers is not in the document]
traeok: It's sort of odd to be honest, VSCode has their own `Thenable` which functions similarly to a `Promise` but can't be returned in an `async` function... However you can `await` these calls in a similar fashion to an `async` function call.

packages/zowe-explorer-api/src/vscode/ZoweVsCodeExtension.ts
traeok: This is essentially mimicking the original behavior of `showVsCodeMessage` (screenshot above), but we can definitely update the function to not use `void`.

packages/zowe-explorer-api/src/vscode/ZoweVsCodeExtension.ts
traeok: Primarily because `await` is exclusive to `async` functions, it is more of a shorthand here to dismiss the Promise instead of wrapping each call in a `.then()` and `.catch()` - honestly not keen about the use of `void` but was just adapting its current use in the codebase to the new `Gui` calls.

packages/zowe-explorer-ftp-extension/src/ZoweExplorerAbstractFtpApi.ts
traeok: `await` > `void` but sometimes `await` is needed for those `Thenable`s in synchronous functions, and would involve changing the function prototype to `async` and then `await`ing it everywhere it is used. 

Trying to a(void) `void` only because its not the ideal way of handling `Thenable`s/`async` calls, but is more of a convenience when a lot of synchronous functions reference the function that's using `void`, where usually that function needs to be changed to `async` for `await` to work.

JillieBeanSim:
@traeok the sonarCloud failed with a couple code smells and very low test coverage.  could this be addressed please.

zFernand0:
Re: `void` operator
There are instances in which we may need to ignore this specific code smell.
Otherwise we might be calling promises without an `await` operator OR we will need to make a bunch of functions `async` 😅

traeok:
> Re: `void` operator There are instances in which we may need to ignore this specific code smell. Otherwise we might be calling promises without an `await` operator OR we will need to make a bunch of functions `async` 😅

When removing `void` and replacing with `.then/.catch` blocks, this typescript-eslint error appears:

"Promises must be awaited, end with a call to .catch, end with a call to .then with a rejection handler or be explicitly marked as ignored with the `void` operator.eslint[@typescript-eslint/no-floating-promises](https://typescript-eslint.io/rules/no-floating-promises)"

Seems there are conflicts between the typescript-eslint plugin and SonarCloud code scanning. In these two cases, we are disregarding the response from the `async` function, so while `void` is recommended by the linter plugin, SonarCloud suggests that `void` is a code smell.

Will look over and address duplication 👍

traeok:
Have to fix a couple tests locally, should be able to have pipeline fixed tomorrow

packages/zowe-explorer-api/package.json
```diff
@@ -21,13 +21,68 @@
...
+  "jest-html-reporter": {
+    "pageTitle": "Zowe Extension Tests",
+    "outputPath": "results/unit/results.html"
+  },
+  "jest-stare": {
+    "resultDir": "results/unit/jest-stare",
+    "coverageLink": "../coverage/lcov-report/index.html"
+  },
+  "jest-junit": {
+    "outputDirectory": "<rootDir>/results/unit",
```
zFernand0, line 85: Will this overwrite the ZE of ZE-ftp test results ?
Maybe we should place the results folders inside their packages directory... and with the PR below, we don't have to add the ze or ze-ftp results folder to a blocklist
- #2042

JillieBeanSim (approved):
LGTM! thanks @traeok for the updates

zFernand0 (approved):
LGTM! 😋 
Thanks for resolving all merge conflicts

packages/zowe-explorer-api/package.json
traeok: I did some tests locally to confirm that `rootDir` is pointing to the root of the API package, so the `results` folder is created in the package directory as expected 👍

Issue #1550, opened 2021-10-25
Copy Paste Multiple Datasets


jellypuno:
As a ZE User, I would like to have the capability of copy and pasting multiple datasets and dataset members inside another dataset.

Goal:

- Copy/Paste multiple DS and members within 1 profile
- [Stretch] Copy/Paste multiple DS and members across different profiles
- Scenario:
     - PDS members to PDS members
     - IF there's a mix (Seq, PDS and PDS members) then that will not be allowed. An error message will be displayed or filter the selection to just PDS members and notify the user if other datasets are not copied.
     - Mimic the CLI

Possible UX:

1. The user will select multiple DS and members
2. Right-Click > Copy
3. Right-Click on another DS/PDS > Paste
4. All selected items are copied to the new DS
    - This will only cover copy. Move is not part of our acceptance criteria
    - Can we use Ctrl + C and Ctrl + V?

Note: Same UX goes for the Stretch but you are just copying it to a different profile

PR #2051, opened 2022-12-20
Feature copy datasets 

KutluOzel-b:
Please see this comment for context: https://github.com/zowe/vscode-extension-for-zowe/pull/2051#issuecomment-1431999564

---

## Proposed changes

add feature copy datasets ( pds, sequential, members across pds) with multi select capabilities.

## Release Notes

https://github.com/zowe/vscode-extension-for-zowe/issues/1550

Changelog:

## Types of changes

## Checklist

---

Please see this comment for context: https://github.com/zowe/vscode-extension-for-zowe/pull/2051#issuecomment-1431999564


traeok (changes requested):
Thanks for working on this feature @KutluOzel-b ! Just a couple things I've found while testing:

- Since the "Copy Data Set" option brings up a dialog for the new dataset(s) names and doesn't require the user to initiate the paste operation, should the feature be renamed to something like "Duplicate data set(s)" ? Also, would it be possible to add a title to the input box to describe its intent, e.g. "New name for copy of 'DATASET1'" ? This would help provide some clarity to the end-user when the input box appears.
- When right-clicking on data set members (PDS), the "Copy Member" option now says "Copy Data Set".
- Trying to copy one PS data set, and providing the name of an existing PS data set as the new name, results in a z/OSMF REST API error: 
image

In this situation we could probably disregard z/OSMF API errors, or just reject the operation before making the API request if the new name already matches one of the existing datasets.

KutluOzel-b:
> Thanks for working on this feature @KutluOzel-b ! Just a couple things I've found while testing:
> 
> * Since the "Copy Data Set" option brings up a dialog for the new dataset(s) names and doesn't require the user to initiate the paste operation, should the feature be renamed to something like "Duplicate data set(s)" ? Also, would it be possible to add a title to the input box to describe its intent, e.g. "New name for copy of 'DATASET1'" ? This would help provide some clarity to the end-user when the input box appears.
> * When right-clicking on data set members (PDS), the "Copy Member" option now says "Copy Data Set".
> * Trying to copy one PS data set, and providing the name of an existing PS data set as the new name, results in a z/OSMF REST API error:
>   image
> 
> In this situation we could probably disregard z/OSMF API errors, or just reject the operation before making the API request if the new name already matches one of the existing datasets.

Hey @traeok , Thanks for review . 

I had a purpose while huddling copy features together which was make it more simple and let user control over the operation he/she does. I'm thinking having "copy member", "copy dataset", "duplicate datasets", "copy sequential datasets" is a not a good way of achieving good ux. 
At this point it is maybe good to have more feedbacks from the community/team to continue.

Regarding the attempt of copying already existing a datasets ; I'm not sure there is an api for checking if a dataset with selected name is existing or not and dismissing api errors might be confusing considering operation is done or not.

traeok:
> I had a purpose while huddling copy features together which was make it more simple and let user control over the operation he/she does. I'm thinking having "copy member", "copy dataset", "duplicate datasets", "copy sequential datasets" is a not a good way of achieving good ux. At this point it is maybe good to have more feedbacks from the community/team to continue.

I agree that having too many different "Copy" operations would be redundant and confusing for the user - we could always stick with "Copy" here as the user is intentionally right-clicking on a dataset node anyway.

> Regarding the attempt of copying already existing a datasets ; I'm not sure there is an api for checking if a dataset with selected name is existing or not and dismissing api errors might be confusing considering operation is done or not.

In hindsight it is probably more complicated to ensure there isn't an overlap in existing/new data set names, so we could always leave that up to the user.

Going to do some quick tests to make sure it is working as intended and then I will update the review 👍

traeok (approved):

JillieBeanSim (changes requested):
Thanks @KutluOzel-b this is working good.  I do have a couple things I wanted to mention.
were we changing the `Copy Dataset` to just `Copy`?  I remember the discussion and see paste was changed to just `Paste`.  Also, I have to click on the data set (closing & reopening) to see the newly added files.  Is there a way we can have this automatically refresh after paste is finished?

KutluOzel-b:
> Thanks @KutluOzel-b this is working good. I do have a couple things I wanted to mention. were we changing the `Copy Dataset` to just `Copy`? I remember the discussion and see paste was changed to just `Paste`. Also, I have to click on the data set (closing & reopening) to see the newly added files. Is there a way we can have this automatically refresh after paste is finished?

Hey @JillieBeanSim, thanks for reviewing this. I changed copy datasets to just copy. Sorry that was my miss. 

Regarding refresh of ds contents ; I preserve the collapse state of partial datasets if the state is open then it could be seen new files are presenting. If closed yes user has to change the state by him/herself. Do you think we should open the expandable list in both cases ?

packages/zowe-explorer/src/dataset/actions.ts
```diff
@@ -1506,17 +1499,16 @@ export function downloadDs(node: ZoweDatasetNode) {
...
         default:
-            vscode.window.showErrorMessage(
-                localize("download.invalidNode", "downloadDataset() called with invalid node.")
-            );
+            vscode.window.showErrorMessage(localize("download.invalidNode", "downloadDataset() called with invalid node."));
             throw Error(localize("download.invalidNode", "downloadDataset() called with invalid node. "));
     }
     const filePath = getDocumentFilePath(lbl, node);
-    return ZoweExplorerApiRegister.getMvsApi(profile).getContents(lbl, {
+    let a = await ZoweExplorerApiRegister.getMvsApi(profile).getContents(lbl, {
```
traeok, line 1506: Instead of defining it and then returning this we can probably just `return ZoweExplorerApiRegister.getMvsApi...`

traeok (approved):
LGTM! Just left a comment about one of the return values

packages/zowe-explorer/src/dataset/actions.ts
KutluOzel-b: Thanks @traeok that was for debugging purposes my miss.

KutluOzel-b:
Hey @JillieBeanSim, Thanks for the feedback of refreshing ds after paste operation completes. 

https://user-images.githubusercontent.com/104970275/211808233-9c3fac74-bdbc-4ebb-b4b4-099c3d32316d.mov

Do you think should it behaves in above ss ? If not so, could you please provide an ss so I can take a look ? 

Thanks in advance

JillieBeanSim:
@KutluOzel-b I think the behavior in SS is perfect.  Are those changes in?

JillieBeanSim (approved):

JillieBeanSim:
Looks like the changes are in for the ds refreshing after paste.  Thanks @KutluOzel-b for the enhancement and for addressing all feedback

packages/zowe-explorer/package.json
```diff
@@ -101,6 +101,18 @@
...
         "when": "focusedView == zowe.uss.explorer || viewItem =~ /^(directory.*|ussSession.*)/ && !listSupportsMultiselect"
+      },
+      {
+        "command": "zowe.ds.copyDataSet",
+        "key": "ctrl+c",
+        "mac": "cmd+c",
+        "when": "focusedView == zowe.ds.explorer"
+      },
+      {
+        "command": "zowe.ds.pasteMember",
```
t1m0thyj, line 112: Should these commands be named consistently if they offer equivalent copy/paste behavior?
Either "copyDataSet" and "pasteDataSet", or "copyMember" and "pasteMember"

packages/zowe-explorer/src/dataset/actions.ts
```diff
@@ -1402,3 +1436,197 @@ export async function saveFile(doc: vscode.TextDocument, datasetProvider: api.IZ
...
+export async function copySequentialDatasets(nodes: ZoweDatasetNode[]) {
+    for (const node of nodes) {
+        try {
+            await downloadDs(node);
+            const lbl = node.getLabel().toString();
+            const inputBoxOptions: vscode.InputBoxOptions = {
+                value: lbl,
+                placeHolder: localize("pasteMember.inputBox.placeHolder", "Name of Data Set"),
+                validateInput: (text) => {
+                    return dsUtils.validateDataSetName(text) && (lbl !== text) === true ? null : "Enter valid dataset name";
```
t1m0thyj, line 1528: For the case where the destination DSN is identical to the original one (`lbl === text`), should we show a different error message?
Instead of "Enter valid dataset name", I think a more specific message like "Dataset name already exists" may be helpful.

packages/zowe-explorer/src/dataset/actions.ts
```diff
@@ -1402,3 +1436,197 @@ export async function saveFile(doc: vscode.TextDocument, datasetProvider: api.IZ
...
+ */
+export async function copyPartitionedDatasets(nodes: ZoweDatasetNode[]) {
+    for (const node of nodes) {
+        try {
+            const lbl = node.getLabel().toString();
+            const inputBoxOptions: vscode.InputBoxOptions = {
+                value: lbl,
+                placeHolder: localize("pasteMember.inputBox.placeHolder", "Name of Data Set"),
+                validateInput: (text) => {
+                    return dsUtils.validateDataSetName(text) && (lbl !== text) === true ? null : "Enter valid dataset name";
```
t1m0thyj, line 1586: Same question as above regarding having a more specific error message when "Dataset name already exists".

packages/zowe-explorer/src/dataset/actions.ts
KutluOzel-b: Thanks @t1m0thyj  for reviewing this. We need to validate new Input too and I think changing the text might be confusing. I hesitated to add "Dataset name already exists" because of this. Otherwise user would expect something like "datasets cannot start with : " or "maximum 8 chars allowed" .. and so on I think user should extract the logic what is missing/wrong the operation I'm doing at the moment if it this not extremely hidden or irrevelant to current operation.

packages/zowe-explorer/package.json
KutluOzel-b: Adressed in : a19a8de6821b088c5790d4b27484f48ff802a010

packages/zowe-explorer/src/dataset/actions.ts
KutluOzel-b: @t1m0thyj I got your concern. It was clear. My question was ; should we start to implement such messages ? If so I think then for consistency we need to add other ones as well. For example do we have it this one on rename function right now ?

traeok (approved):
LGTM, tested locally and it is still functioning as intended. 

Just a couple quick suggestions: 
- The `showInputBox` calls can be updated to use the new `Gui` module.
- We might want to pass a `title` value (same as `pasteMember` placeholder) to these input boxes, that will clarify the purpose of those dialogs to the end-user, e.g. "Enter new name for dataset {...}"

t1m0thyj (commented):
* When copying data sets and I am prompted for a new DSN, the input is rejected as invalid if the name is identical to the original. But when copying members, it accepts the original member name and doesn't require me to change it. Is this an inconsistency we should fix?
* When I try to copy data sets using an FTP profile, I get the following errors:
image
  Should it show the following error instead (that occurs when I try to copy members):
  image

traeok (changes requested):
Just saw @t1m0thyj 's comments regarding the FTP (thanks for testing that case), will approve once comments are addressed.

KutluOzel-b:
> LGTM, tested locally and it is still functioning as intended.
> 
> Just a couple quick suggestions:
> 
> * The `showInputBox` calls can be updated to use the new `Gui` module.
> * We might want to pass a `title` value (same as `pasteMember` placeholder) to these input boxes, that will clarify the purpose of those dialogs to the end-user, e.g. "Enter new name for dataset {...}"

Hey @traeok  thanks for the feedbacks. Adressed in : https://github.com/zowe/vscode-extension-for-zowe/pull/2051/commits/ea4cba23032e30d79ef7cfc8a9dba7f15687b29c

KutluOzel-b:
Hey @t1m0thyj  Thanks for the feedbacks. 

> * When copying data sets and I am prompted for a new DSN, the input is rejected as invalid if the name is identical to the original. But when copying members, it accepts the original member name and doesn't require me to change it. Is this an inconsistency we should fix?

**If you select multiple members to copy then we are not asking for every member( I think this is the correct behaviour otherwise it would be painful to copy multiple  ) name but if single is copied then user can change it. We could have implemented such feature that checks if member is already available for the current PDS , but this is costly since we are already doing that in api**

> * When I try to copy data sets using an FTP profile, I get the following errors:
>   image
>   Should it show the following error instead (that occurs when I try to copy members):
>   image

**Since we don't have an api to do this operation, new PDS is created with the process : download members, allocate new Pds , create new local files , upload to mainframe , pull again from mf. Each of the process which is not supported by API would reject due to various reasons .I think It doesn't seems good to ignore part of a process and inform users with "can't copy" since user might find solution of this wrt authorization and such .**

traeok:
@KutluOzel-b Thanks for addressing those comments 🙂 

I did find an API endpoint in `@zowe/zos-files-for-zowe-sdk`, and was wondering what your thoughts would be on using this instead of downloading and uploading? [Here](https://github.com/zowe/zowe-cli/blob/master/packages/zosfiles/src/methods/copy/Copy.ts#L40) is the method that I am referencing.

It seems like we should be able to do a copy operation on the mainframe, then pull the resultant dataset from that operation. Should help speed up runtime and would avoid having to download and re-upload files.

https://www.ibm.com/docs/en/zos/2.1.0?topic=interface-zos-data-set-member-utilities

KutluOzel-b:
> @KutluOzel-b Thanks for addressing those suggestions 🙂
> 
> I did find an API endpoint in `@zowe/zos-files-for-zowe-sdk`, and was wondering what your thoughts would be on using this instead of downloading and uploading? [Here](https://github.com/zowe/zowe-cli/blob/master/packages/zosfiles/src/methods/copy/Copy.ts#L40) is the method that I am referencing.
> 
> It seems like we should be able to do a copy operation on the mainframe, then pull the resultant dataset from that operation. Should help decrease runtime for that operation and would avoid having to make local copies first. I've added a link to the API below for context:
> 
> https://www.ibm.com/docs/en/zos/2.1.0?topic=interface-zos-data-set-member-utilities
> 
> I've also proposed a similar optimization for USS, here is that [PR](https://github.com/zowe/vscode-extension-for-zowe/pull/2082) for reference.

:exploding_head::exploding_head: I wish I knew this before. I thought we could only use what is already in cli/imperative/ftp api. 

If such api is available, then we should use it . At this point we might take this as an improvement or cancel this pr .

what do you think ?

traeok:
> 🤯🤯 I wish I knew this before. I thought we could only use what is already in cli/imperative/ftp api.
> 
> If such api is available, then we should use it . At this point we might take this as an improvement or cancel this pr .
> 
> what do you think ?

We might be able to update this PR to use that API instead, since there is a function within Zowe's z/OS files SDK for it. Either way, we can always implement that API into this feature as a future improvement, as the feature still functions as intended for this PR. Whatever works best for you 🙂

traeok (approved):
LGTM, thanks @KutluOzel-b 😀

packages/zowe-explorer-api/src/profiles/ZoweExplorerApi.ts
```diff
@@ -331,6 +331,17 @@ export namespace ZoweExplorerApi {
...
+        /**
+         * Copies a dataSet.
+         *
+         * @param {string} fromDataSetName
+         * @param {string} toDataSetName
+         * @param {string?} enq possible values : {SHR, SHRW, EXCLU}
+         * @param {boolean?} replace
+         * @returns {Promise<zowe.IZosFilesResponse>}
+         */
+        copyDataSet(fromDataSetName: string, toDataSetName: string, enq?: string, replace?: boolean): Promise<zowe.IZosFilesResponse>;
```
zFernand0, line 344: Don't we need to make this optional to avoid breaking extenders?

zFernand0 (changes requested):
left one comment about the ze-api new method

traeok (changes requested):
Sorry for the confusion, I've updated my review with a note about the APIs. I edited my original comment but I probably should have posted it as a follow-up comment instead.

@zFernand0 mentioned that we will have issues using that Copy API between LPARs. We can adapt this in the future to use the new endpoint, but until we add a check for pasting between sessions it will probably be best to use the original implementation.

packages/zowe-explorer-ftp-extension/CHANGELOG.md
```diff
@@ -4,7 +4,7 @@ All notable changes to the "vscode-extension-for-zowe" extension will be documen
 
 ### New features and enhancements
 
-- Updated UI/UX method calls to use standalone `Gui` module for better usability and maintainability. [#1967](https://github.com/zowe/vscode-extension-for-zowe/issues/1967)
+- Updated UI/UX method calls to use standalone `Gui` module for better usability and maintainability . [#1967](https://github.com/zowe/vscode-extension-for-zowe/issues/1967)
```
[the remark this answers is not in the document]
KutluOzel-b: I had to modify changelog to get it passed. Didn't want to add something like "added unsupported warning of copy datasets"

packages/zowe-explorer-api/src/profiles/ZoweExplorerZosmfApi.ts
```diff
@@ -281,6 +281,9 @@ export class ZosmfMvsApi extends ZosmfApiCommon implements ZoweExplorerApi.IMvs
     public dataSetsMatchingPattern(filter: string[], options?: zowe.IDsmListOptions): Promise<zowe.IZosFilesResponse> {
         return zowe.List.dataSetsMatchingPattern(this.getSession(), filter, options);
     }
+    public copyDataSet(fromDataSetName: string, toDataSetName: string, enq?: string, replace?: boolean): Promise<zowe.IZosFilesResponse> {
```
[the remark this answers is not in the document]
KutluOzel-b: addressed in : https://github.com/zowe/vscode-extension-for-zowe/pull/2051/commits/b62c2e4eed34167aa8faa83bb2baee2228ebc82d

packages/zowe-explorer-api/src/profiles/ZoweExplorerApi.ts
KutluOzel-b: Good catch ! adressed in : https://github.com/zowe/vscode-extension-for-zowe/pull/2051/commits/b62c2e4eed34167aa8faa83bb2baee2228ebc82d

KutluOzel-b:
> Sorry for the confusion, I've updated my review with a note about the APIs. I edited my original comment about this copy API, but I probably should have posted it as a follow-up comment instead.
> 
> @zFernand0 mentioned that we will have issues using that Copy API between LPARs. We can adapt this in the future to use the new endpoint, but until we add a check for pasting between sessions it will probably be best to use your original implementation. I noticed that `copyDataSet` will use this API regardless of the LPAR, so these changes might break copy/pasting between LPARs until cross-LPAR functionality is implemented for that API endpoint.

In current Implementation user doesn't asked to be choose LPAR while copying PDS and Sequential datasets and won't be a problem, but we can modify this in future once api is adjusted accordingly. 
Copying members of a PDS though, can be pasted into other LPAR and this gives the result " PDS is not available " from api.

packages/zowe-explorer/src/globals.ts
```diff
@@ -33,7 +33,7 @@ export let DS_DIR;
 export let CONFIG_PATH; // set during activate
 export let ISTHEIA: boolean = false; // set during activate
 export let LOG: imperative.Logger;
-export const COMMAND_COUNT = 103;
+export const COMMAND_COUNT = 102;
```
zFernand0, line 36: The fact that we have gone down in the number of commands scares me a little bit.

I know the FTP extension was updated as part of this PR, but can we teat this with another extender's extension that has not updated their implementation yet?

@rudyflores or @JillieBeanSim, can you help us here?

zFernand0 (commented):
Thanks for making the new API optional in the ZE-API interface :yum:
I'm still curious if other extenders may encounter small issues after this PR because of the removal of some commands :thinking:

packages/zowe-explorer/src/extension.ts
```diff
@@ -430,11 +430,24 @@ function initDatasetProvider(context: vscode.ExtensionContext) {
             }
         })
     );
+
     context.subscriptions.push(vscode.commands.registerCommand("zowe.ds.renameDataSet", (node) => datasetProvider.rename(node)));
-    context.subscriptions.push(vscode.commands.registerCommand("zowe.ds.copyMember", (node) => dsActions.copyDataSet(node)));
-    context.subscriptions.push(vscode.commands.registerCommand("zowe.ds.copyDataSet", (node) => dsActions.copyDataSet(node)));
-    context.subscriptions.push(vscode.commands.registerCommand("zowe.ds.pasteMember", (node) => dsActions.pasteMember(node, datasetProvider)));
```
zFernand0, line 436: Since we are in an LTS release, removing commands might be considered a breaking change.
We should implement this new behavior with new commands and deprecate the old ones.

Already in the middle of this 😋

zFernand0 (changes requested):
This could lead to a breaking change in the middle of an LTS release.
I'll suggest some changes in a commit later today

packages/zowe-explorer/src/extension.ts
zFernand0: 🤔 Removing the commands may not be a breaking change since they were never expose to the end user.
However, changing an exported function's signature is.
I'm happy to work on this in order to deprecate the copyDataSet method 😋

zFernand0:
Known issue:
While this PR successfully created the destination sequential dataset with the same attributes as the source sequential dataset...
The underlying z/OSMF API may prevent the operation due to the record format being used.
Tested with ZE and CLI and they both return a similar error regarding `recfm = U` 
image

zFernand0 (approved):

zFernand0 (approved):

zFernand0:
We believe that removing hidden (i.e. never exposed in the command palette) commands from the `context.subscriptions` is not a breaking change, thus we kept the removal of `copyDataSet`, `copyMember`, and `pasteMember`.

However, the change in behavior of an exported function is something we often try to avoid. For this reason, we decided the bring back the `copyDataSet` method. I took the liberty to rename the newly added methods (and their registered commands) to their plural forms (e.g. copyDataSets, pasteDataSets) since there the concept of method overloading in Typescript requires us to keep the same number of parameters (and we went from 1 to 3).

I'm happy to continue discussing these decision in a parking lot on our next call, or on our slack channel if needed 😋

zFernand0 (commented):
Not sure I should be approving this PR anymore
but, LGTM! 😋

traeok (dismissed):
LGTM! Thanks @KutluOzel-b and @zFernand0

packages/zowe-explorer-ftp-extension/__tests__/__unit__/Mvs/ZoweExplorerFtpMvsApi.unit.test.ts
```diff
@@ -195,6 +195,15 @@ describe("FtpMvsApi", () => {
         expect(DataSetUtils.deleteDataSet).toBeCalledTimes(1);
         expect(MvsApi.releaseConnection).toBeCalled();
     });
+    it("should give error copy datasets.", async () => {
+        try {
+            await MvsApi.copyDataSet(null, null);
+        } catch (error) {
+            // do nth
```
traeok, line 202: lines 204-205 can be placed in the catch block, but the result is the same in this situation either way

zFernand0 (changes requested):
Found a few showstoppers: 
- copy is not triggered for extenders
- there is no way to replace an existing dataset or member (even though the APIs support it)

zFernand0:
Test scenarios to try:
- [ ] Copy a PS dataset to a new PS dataset
- [ ] Copy a PS dataset to an existing dataset (Prompt: `replace dataset`)
- [ ] * Copy a PO dataset to a new PO dataset
- [ ] * Copy a PO dataset to an existing PO dataset (Prompt: `merge and replace`)
- [ ] Copy a member to a new member
- [ ] (NEW) Copy a member to an existing member (Prompt: `replace member`)

---
Scenarios not supported by ZE
- [ ] Copy a PS dataset to a new member in an existing PO dataset
- [ ] Copy a PS dataset to an existing member in an existing PO dataset
- [ ] Copy a PS dataset to a new member in a new PO dataset
- [ ] Copy a member to a new PS dataset
- [ ] Copy a member to an existing PS dataset

---
Notes:
- `*` not supported by the FTP plugin (https://github.com/zowe/zowe-cli-ftp-plugin/pull/131)
- All other scenarios are supported by the FTP plugin (https://github.com/zowe/zowe-cli-ftp-plugin/pull/131)

packages/zowe-explorer/src/shared/context.ts
```diff
@@ -386,10 +386,14 @@ export function asFavorite(node: TreeItem): string {
 }
 
 export function withProfile(node: IZoweTreeNode): string {
-    if (node?.contextValue?.includes(".profile=")) return node?.contextValue;
-    return node?.getParent()?.contextValue?.includes(".profile=")
-        ? node?.contextValue + ".profile=" + node?.getParent()?.contextValue?.split(".profile=")[1] ?? node?.contextValue
-        : node?.contextValue;
+    if (!node) return;
```
traeok, line 389: Looks like the linter wants braces for these one-line `if` statements 😅 (same w/ line 391)

packages/zowe-explorer/src/shared/context.ts
```diff
@@ -386,10 +386,14 @@ export function asFavorite(node: TreeItem): string {
...
+    const hasProfile = (n: IZoweTreeNode) => n?.contextValue?.includes(".profile=") ?? false;
+    if (hasProfile(node)) return node.contextValue;
+    if (hasProfile(node.getParent())) {
```
traeok, line 392: This *might* be optimized by JavaScript at some point, but I see there are two calls to `getParent` here. If this will increase code complexity in SonarCloud by declaring the parent, then this is good as-is 👍

packages/zowe-explorer/src/dataset/actions.ts
```diff
@@ -1618,6 +1589,77 @@ export async function copySequentialDatasets(nodes: ZoweDatasetNode[]) {
...
+            }
+            replace = stringReplace === (await api.Gui.showMessage(q, { items: [stringReplace, stringCancel] }));
+        }
+    }
+    // Sonar cloud code-smell :'(
+    const returnValueIfNotReplacing = q === null ? "notFound" : "cancel";
+    return replace ? "replace" : returnValueIfNotReplacing;
+}
+
+export async function _copyProcessor(nodes: ZoweDatasetNode[], type: replaceDstype, action: Function) {
```
traeok, line 1662: Should we define the type signature for the function parameter `action`? Just in case someone tries to call this function elsewhere (as it is exported)

traeok (approved):
LGTM! Just left a couple comments regarding lint warnings, etc

JillieBeanSim (dismissed):
this LGTM! thanks @zFernand0 & @KutluOzel-b 

We will definitely need to document the limitations as known issues or something

packages/zowe-explorer/src/shared/context.ts
zFernand0: I kind of forgot about this one 😢 
I don't mind storing that in a variable even though it's just a getter for the `mParent` property
```typescript
public getParent(): IZoweTreeNode {
   return this.mParent;
}
```

packages/zowe-explorer/src/dataset/actions.ts
```diff
@@ -1127,12 +1128,55 @@ export async function enterPattern(node: api.IZoweDatasetTreeNode, datasetProvid
...
+    let selectedNodes;
+    if (!(node || nodeList)) {
+        selectedNodes = datasetProvider.getTreeView().selection;
+    } else {
+        selectedNodes = getSelectedNodeList(node, nodeList);
+    }
+
+    const unique = [...new Set(selectedNodes.map((item) => item.contextValue))];
+    if (unique.length > 1) {
+        vscode.window.showErrorMessage(`${"Can't copy multiple dataset with different types"}: ${"Select same types to copy"}`);
```
t1m0thyj, line 1156: Should these strings be localized?

packages/zowe-explorer/src/dataset/actions.ts
```diff
@@ -1439,3 +1481,241 @@ export async function saveFile(doc: vscode.TextDocument, datasetProvider: api.IZ
...
+            }
+        );
+    });
+}
+
+/**
+ * Type of z/os dataset or member intended for replacement
+ * @export
+ */
+export type replaceDstype = "ps" | "po" | "da" | "mem";
```
t1m0thyj, line 1623: I'm not familiar with DA (Direct Access?) data sets. Has this PR been tested to verify we support them?

t1m0thyj (approved):

JillieBeanSim:
> Seems like the Theia tests broke after merging from main. 😢 Looking into it 👀

It could be from the Theia version update to 1.35.0

zFernand0:
> > Seems like the Theia tests broke after merging from main. 😢 Looking into it 👀
> 
> It could be from the Theia version update to 1.35.0

Seems like that's the case since replaying a successful run from `main`  now fails
https://github.com/zowe/vscode-extension-for-zowe/actions/runs/4235493965/jobs/7420836048

packages/zowe-explorer/src/dataset/actions.ts
zFernand0: I was trying to understand a bit more what Direct Access (DA) datasets were and, similar to VSAM, these Basic Direct Access Method (BDAM) datasets are a special type that we don't support for copy/paste operations.

I'll remove the DA specific considerations from this PR 😋

traeok (dismissed):
LGTM!

packages/zowe-explorer/src/dataset/actions.ts
```diff
@@ -1439,3 +1479,239 @@ export async function saveFile(doc: vscode.TextDocument, datasetProvider: api.IZ
...
+    await _copyProcessor(nodes, "ps", async (node: ZoweDatasetNode, dsname: string, replace: shouldReplace) => {
+        const lbl = node.getLabel().toString();
+
+        await api.Gui.withProgress(
+            {
+                location: vscode.ProgressLocation.Window,
+                title: localize("paste.dataSet.InPrg", "Copying File(s)"),
+            },
+            () => {
+                return ZoweExplorerApiRegister.getMvsApi(nodes[0].getProfile()).copyDataSet(lbl, dsname, null, replace === "replace");
```
JillieBeanSim, line 1572: If copyDataSet hasn't been picked up by an extender yet we should have a check for this otherwise user will see messages like below. 
We shouldn't expect extenders to add 'Not Supported' like I see in the FTP package here, it should be handled by the ZE extension like we have in the past where if the particular API didn't exist we either 
1. handle it the old way if this is a new way
2. return the error of this action not being supported if the new api doesn't exist

Screen Shot 2023-02-27 at 11 40 19 AM

JillieBeanSim (changes requested):

packages/zowe-explorer/i18n/sample/src/dataset/actions.i18n.json
```diff
@@ -91,6 +91,7 @@
   "paste.dataSet": "Invalid paste. Copy dataset(s) first",
   "paste.dataSet.InPrg": "Copying File(s)",
   "download.invalidNode": "downloadDataset() called with invalid node. ",
+  "copyDataSet.error.notSupported": "This operation is not supported at this time.",
```
traeok, line 94: Do we want to make this more specific to the "copy dataset" action? Something like "Copying datasets is not supported" - not a request to change, just wanted to get your thoughts on it

traeok (dismissed):
LGTM, check was added for the API function 👍 thanks @zFernand0

traeok (approved):

packages/zowe-explorer/src/dataset/actions.ts
zFernand0: And error message will appear indicating that `Copying datasets is not supported`
However, that leaves the question of whether we want to prevent the `allocateLike` from happening or not.

As it stands, I believe the new Copy/Paste experience is a little better than before.

Do we want to address the AllocateLike (and possibly other misbehaviors that we haven't found yet) in this PR?

(Personal opinion) 
I have considered going with a similar approach as in #2082 where we rely on the clipboard instead of performing a complex, multi-step, error-prone, user interaction for such functionality. 
However, some design decisions need to be discussed. For example, the USS file system may not require allocation of resources.

JillieBeanSim (approved):
thanks for the updates @zFernand0 LGTM

packages/zowe-explorer/src/dataset/actions.ts
JillieBeanSim: I think Allocate like can be a future PR

Issue #998, opened 2020-09-01
Confirmation before submitting a job


gcartier94:
**Is your feature request related to a problem? Please describe.**

Recently I was working with some of the JCL on my LPAR using the Zowe Explorer. When I decided to test it I've noticed that I've accidentally submitted the wrong job. Fortunately, I was working a controlled environment and that had no impact on my system. 

**Describe the solution you'd like**

It would be very helpful if the Zowe Explorer could prompt for a confirmation before submitting the currently selected JCL since is very easy to misclick and select the wrong member of the PDS instead.

zdmullen:
Thanks for opening the issue, @gcartier94. We've discussed it as a squad, and feel that there is further analysis needed because we should investigate confirmation messages more in general and the ability to potentially disable/enable them in settings. 

@jelaplan and Ilya, could you consider this from the UX point of view?

fbeyl:
This is an important improvement! Rightclick and click submit without seeing and confirming what will be submitted is easy and fast but potentially dangereous especially for highly authorized users like sysprogs.

IlyaKreynin:
If the error is easy to make, difficult to detect, and can have real consequences, a settings-toggleable confirmation message makes sense! The option to toggle confirmation messages off should also exist within the confirmation messages themselves.

To potentially reduce errors without confirmation messages, we could change the text of the action from "Submit Job" to "Submit *NAME OF JOB* As Job" or something to that effect. This gives users another chance to see what is actually being submitted without adding clicks. This change could be implemented with or without confirmation messages - should help either way!

zdmullen:
Note: this could also be an option that you could toggle in the settings.json.

zdmullen:
@jelaplan to reach out to guillerme!

jelaplan:
I will reach out to Cartier and talk to him about it. Submitting the wrong job is clearly a mistake to be avoided and the intention seems good but I suspect that a popup that asks are-you-sure is something to blow past so may offer little in the way of prevention. At the same time, an are-you-sure prompt is liable to be bothersome to some people. I like Ilya's idea to echo the job name in the menu command. But, is the job name sufficient and will people stop to read? The popular book "Don't Make Me Think" would indicate not. We could make the interaction more intentional by asking the user type in the job name. We could ask them to review the contents of the JCL before submitting the job. The idea of a setting.json option has potential too but do we want to add more complexity and things to maintain when few people would even realize it is there.

fbeyl:
[...]

PR #2054, opened 2022-12-21
feat(ds-jobs): Toggleable confirmation dialog for 'Submit Job' 

traeok:
Signed-off-by: Trae Yelovich <trae.yelovich@broadcom.com>

## Proposed changes

This is a feature implementation that follows up on issue #998. It adds a confirmation dialog for the "Submit Job" feature (right-click context menu for data set members), and is set to "All jobs" by default. This dialog can be configured from the extension settings - see the "Jobs: Confirm Submission" section.

| Proposed change | Screenshot |
| ------------------- | ------------ |
| Option in extension settings | image |
| Confirmation dialog | image |

## Release Notes

Milestone: 2.6.0

Changelog:
- Added confirmation message for "Submit Job" feature as an option in extension settings (set to "All jobs" by default) #998.

## Types of changes

## Checklist


zFernand0 (commented):
LGTM! 😋 
Before approving...
Should we:
1. open a docs-site issue to include this new configuration in [Configuring Zowe Explorer](https://docs.zowe.org/stable/user-guide/ze-install#configuring-zowe-explorer)
2. use a notification pop-up vs modal (new-window)? (how does this affect Theia)?

packages/zowe-explorer/package.json
```diff
@@ -1710,6 +1710,12 @@
           "description": "%zowe.jobs.history%",
           "scope": "application"
         },
+        "zowe.jobs.confirmSubmission": {
+          "type": "boolean",
+          "default": true,
```
zFernand0, line 1715: 🤔 Should the default be false? 
That way it's not a "breaking" change to the users

zFernand0 (changes requested):
Will this be a breaking change?

packages/zowe-explorer/package.json
traeok: Yes it should be `false`. Thanks for pointing that out, had it set to true for easy testing

traeok:
> LGTM! 😋 Before approving... Should we:
> 
> 1. open a docs-site issue to include this new configuration in [Configuring Zowe Explorer](https://docs.zowe.org/stable/user-guide/ze-install#configuring-zowe-explorer)?

Yep, works for me! 

> 2. use a notification pop-up vs modal (new-window)? (how does this affect Theia)?

Was debating this - not sure whether a modal will affect Theia behavior. Can do some testing to see if it will be affected, but always willing to switch to the cleaner notification-style dialogs.

traeok:
Looked into support for modal messages, and it seems that Theia will support the modal parameter:

```ts
/**
* Options to configure the message behavior.
*/
export interface MessageOptions {

  /*
   * Human-readable detail message that is rendered
   * less prominent. Note that detail is only shown
   * for modal messages.
   */
  detail?: string;

  /**
   * Indicates that this message should be modal.
   */
  modal?: boolean;
}
```

packages/zowe-explorer/src/dataset/actions.ts
```diff
@@ -974,10 +981,28 @@ export async function submitMember(node: api.IZoweTreeNode) {
...
         );
 
-        if (selection == null || selection?.title !== "Submit") {
-            return;
+        return selection != null && selection?.title === "Submit";
+    };
+
+    if (confirmationOption !== "Disabled") {
+        const ownsJob = datasetName.split(".")[0] === nodeProfile.profile?.user?.toUpperCase();
+        if (confirmationOption === "All jobs") {
```
zFernand0, line 989: Is is possible to use constants for these options?
and should they be localized?

packages/zowe-explorer/src/dataset/actions.ts
```diff
@@ -974,10 +981,28 @@ export async function submitMember(node: api.IZoweTreeNode) {
...
+            if (!(await showConfirmationDialog())) {
+                return;
+            }
+        } else {
+            if (!ownsJob && confirmationOption === "Other user jobs") {
+                if (!(await showConfirmationDialog())) {
+                    return;
+                }
+            } else if (ownsJob && confirmationOption === "Your jobs") {
```
zFernand0, line 998: Not suggesting to remove this, but I'm wondering why someone will want only confirmation dialogs for their own jobs and not other user's jobs. 🤔

packages/zowe-explorer/package.json
```diff
@@ -1711,8 +1711,20 @@
...
+            "Disabled",
+            "Your jobs",
+            "Other user jobs",
+            "All jobs"
+          ],
+          "enumDescriptions": [
+            "No confirmation dialogs will be shown when submitting a job.",
+            "Shows a confirmation dialog when submitting your jobs.",
+            "Shows a confirmation dialog when submitting jobs from another user.",
+            "Shows a confirmation dialog when submitting any job."
```
zFernand0, line 1726: Localization question also applies to these descriptions 😋

zFernand0 (approved):
LGTM! 😋 
Left a couple of questions. 😅

packages/zowe-explorer/src/dataset/actions.ts
traeok: Yep! Forgot about localization here 😅

packages/zowe-explorer/src/dataset/actions.ts
traeok: One situation I considered is if a user is working on a critical mainframe job that is not ready to be submitted yet. In that case the user might want to protect themselves from an accidental submission of their own job.

packages/zowe-explorer/package.json
traeok: Thanks for pointing that out - going to look into localizing the strings mentioned.

JillieBeanSim:
@traeok could the code smells be addressed please.

zFernand0:
Re: cognitive complexity
I finally understood why we have some `switch(true)`. All `case` statements are ignored towards the count.

General comment:
Sometimes trying to reduce cognitive complexity ends up decreasing the readability of a function 😅 

Happy to discuss during standup

JillieBeanSim:
> Re: cognitive complexity I finally understood why we have some `switch(true)`. All `case` statements are ignored towards the count.
> 
> General comment: Sometimes trying to reduce cognitive complexity ends up decreasing the readability of a function 😅
> 
> Happy to discuss during standup

I agree about the complexity reduction can lead to decreased readability sometimes but a lot of the time the complexity can be decreased easily by creating reusable methods or smaller private methods to handle a certain functional part of the code in the method flagged keeping it's readability.

zFernand0 (approved):

zFernand0 (approved):

JillieBeanSim (commented):
@traeok this works great and I like the drop-down settings.  should we also include a bit of info in the README under the [Submit Jobs section](https://github.com/zowe/vscode-extension-for-zowe/blob/main/packages/zowe-explorer/README.md#submit-a-jcl)

traeok:
> this works great and I like the drop-down settings. should we also include a bit of info in the README under the [Submit Jobs section](https://github.com/zowe/vscode-extension-for-zowe/blob/main/packages/zowe-explorer/README.md#submit-a-jcl)

@JillieBeanSim Thanks for the suggestion - I've updated the changelog to reflect the new drop-down options and added some documentation under that section. I can also replace the GIF w/ one showing the confirmation dialog, if desired.

packages/zowe-explorer/CHANGELOG.md
```diff
@@ -8,8 +8,8 @@ All notable changes to the "vscode-extension-for-zowe" extension will be documen
 
 - Added Job search prefix validator [1971](https://github.com/zowe/vscode-extension-for-zowe/issues/1971)
 - Added file association for `zowe.config.json` and `zowe.config.user.json` to automatically detect them as JSON with Comments. [#1997](https://github.com/zowe/vscode-extension-for-zowe/issues/1997)
-- Added confirmation message for "Submit Job" feature as an option in extension settings (disabled by default). [#998](https://github.com/zowe/vscode-extension-for-zowe/issues/998)
-- Added favorite job query to jobs view [#1947](https://github.com/zowe/vscode-extension-for-zowe/issues/1947)
+- Added favorite job query to jobs view. [#1947](https://github.com/zowe/vscode-extension-for-zowe/issues/1947)
+- Added confirmation message for "Submit Job" feature as an option in extension settings (set to "All jobs" by default). [#998](https://github.com/zowe/vscode-extension-for-zowe/issues/998)
```
JillieBeanSim, line 12: thanks for catching the change needed for default value

JillieBeanSim (approved):
LGTM! thanks @traeok for the new feature

zFernand0 (approved):

Issue #1986, opened 2022-10-26
Allow users to easily open config to address errors


JillieBeanSim:
Handling errors from readProfilesFromDisk() and allow users to easily open config to address errors

PR #2056, opened 2022-12-22
feat(config): Config error dialog w/ "Show Config" option

traeok:
**Update:** Going to improve patch coverage before merging.

## Proposed changes

- Updated the error dialog when encountering invalid Zowe profiles, with an option to "Show Config" for diagnostics. 

## Release Notes

Milestone: 2.6.0

Changelog:
- Updated the error dialog when encountering invalid Zowe profiles, with an option to "Show Config" for diagnostics. #1986  

## Types of changes

## Checklist


JillieBeanSim:
@traeok could we get the code smells addressed in this PR?

packages/zowe-explorer/src/ZoweExplorerExtender.ts
```diff
@@ -44,6 +44,68 @@ import { getProfileInfo, getProfile } from "./utils/ProfilesUtils";
...
+                    if (configPath == null) {
+                        return;
+                    }
+                }
+                vscode.window.showTextDocument(vscode.Uri.file(configPath));
+            });
+    }
+
+    /**
+     * Returns the location of a Zowe config (if it exists) relative to rootPath.
```
JillieBeanSim, line 88: @traeok I see this check for config in respect to the zowe home dir.  should we also include a check for config at the project level?

traeok: For `showZoweConfigError`, in most cases the directory is read from the error message. 

I just pushed an update to this - in edge cases where a directory cannot be parsed, I've added a check within that function to see if the error comes from a global Zowe config. Otherwise, it will assume that the culprit is within a project-level config and will check in the location for the active workspace within VS Code.

zFernand0:
I'm happy to re-approve once the new conflicts are resolved 😋

zFernand0 (approved):

t1m0thyj (commented):
When testing this PR I encountered an issue, not sure if its related or if we should file it separately. To reproduce:
* Introduce an error in global zowe.config.json
* Reload VS Code window
* Click "Show Config" and fix the error
* Notice that Zowe Explorer trees are still empty
* Run the "Refresh Zowe Explorer" command and config still fails to load until I reload VS Code

t1m0thyj (commented):
Thanks for updating the config file watcher so it doesn't reload the tree when config has errors - this addresses my earlier comment.

Will approve the PR once patch coverage is improved 🙂

zFernand0 (dismissed):
Some refactoring might be needed to use the new Gui methods,
But the changes LGTM! 😋

traeok:
> Some refactoring might be needed to use the new Gui methods, But the changes LGTM! 😋

@zFernand0 sorry, probably should have saved that refactoring so it didn't mark your review as stale, but I've updated the PR to change all `vscode.window` references to `Gui` (with the exception of `activeTextEditor` as it is a property and hasn't been implemented in `Gui`)

zFernand0 (approved):

t1m0thyj (approved):
LGTM, just needs to be updated with changes from main branch

zFernand0 (approved):

PR #2087, opened 2023-01-13
Take screenshots of failing Theia tests

t1m0thyj:
## Proposed changes

Take screenshots of failing Theia tests and save them in the "results/integration" folder which is archived by the GitHub workflow.

Hopefully this will help us to diagnose the problem causing several tests to occasionally have transient failures.

Example run with screenshots in the artifact: https://github.com/zowe/vscode-extension-for-zowe/actions/runs/3915784470

## Release Notes

N/A since only tests have been changed

### DIFF
--- a/packages/zowe-explorer/__tests__/__theia__/extension.theia.test.ts
+++ b/packages/zowe-explorer/__tests__/__theia__/extension.theia.test.ts
@@ -31,6 +31,12 @@ describe("Locate Tree Nodes", () => {
         await driverFirefox.clickOnZoweExplorer();
     });
 
+    afterEach(async function () {
+        if (this.currentTest.state === "failed") {
+            await driverFirefox.takeScreenshot(this.currentTest.fullTitle());
+        }
+    });
+
     it("should open Zowe Explorer and find the Favorites node", async () => {
         const favoriteLink = await driverFirefox.getFavouritesNode();
         expect(favoriteLink).to.equal("Favorites");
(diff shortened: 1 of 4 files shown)


t1m0thyj:
Verified this works locally, but the workflow skips uploading the artifact when tests fail. Marking as draft until this is fixed.

packages/zowe-explorer/__tests__/__theia__/extension.theia.test.ts
```diff
@@ -257,6 +280,8 @@ describe("Delete Profiles", () => {
         await driverChrome.clickOnZoweExplorer();
     });
 
+    afterEach(screenshotIfFailed(driverChrome));
```
traeok, line 283: Do we need `afterEach` multiple times in this `describe` block?

traeok (approved):
LGTM! Just left a question about one of the Jest functions

packages/zowe-explorer/__tests__/__theia__/extension.theia.test.ts
t1m0thyj: I believe there is only one `afterEach` per `describe` block in this PR, but the way the GitHub diff collapses the lines may make it tricky to review. This `afterEach` applies to the "Delete Profiles" `describe` on line 274.

traeok: Ah, sorry I should have taken a closer look. GitHub was hiding the `describe` blocks in between each `afterEach` 🤦‍♂️ thanks for clarifying!

zFernand0 (approved):

PR #2119, opened 2023-02-09
fix(jobs): Add `validateInput` for job owner entry 

### LINKED ISSUE #2112: Unable to change the owner in a job search
Hi, I met a problem when using the Create job search filter. I suppose the job owner, prefix and status all should be submit as a search query. But I can only change the prefix and confirm it to search. When I changed the owner or status I can't confirm it after press "Enter" key. It had not any response. See the gif file below:
job owner and job status can't confirm

It maybe not related to the issue directly. I am not sure if it is a bug or I missed something. Could you help to try it? Thank you very much!

_Originally posted by @tiantn in https://github.com/zowe/vscode-extension-for-zowe/issues/1971#issuecomment-1418539879_

traeok:
## Proposed changes

I've added a `validateInput` function to the job owner property, so that the input box correctly validates the input for the job owner. This fixes #2112 as a result.

## Release Notes

Milestone: 2.6.1

Changelog:

- Fixed issue where a user could not modify the job owner for an existing search query. #2112 

## Types of changes

## Checklist


packages/zowe-explorer/src/job/ZosJobsProvider.ts
```diff
@@ -878,21 +880,24 @@ export class ZosJobsProvider extends ZoweTreeProvider implements IZoweTree<IZowe
...
-                const options: vscode.InputBoxOptions = {
-                    value: jobProperties.find((prop) => prop.label === pattern).value,
-                    placeHolder: jobProperties.find((prop) => prop.label === pattern).placeHolder,
-                    validateInput: (text) => {
-                        return jobProperties.find((prop) => prop.label === pattern)?.validateInput(text);
-                    },
-                };
-                jobProperties.find((prop) => prop.label === pattern).value = await Gui.showInputBox(options);
+                let property = jobProperties.find((prop) => prop.label === pattern);
+                if (property != null) {
```
kristinochka, line 884: What if property is undefined?

traeok: `!= null` is not a strict type check, so since `null == undefined`, `!= null` will account for both

kristinochka: I think this will lead to confusion and linters recommend using strict checking to avoid this confusion  `===`

From JavaScript the good parts book: 
`JavaScript has two sets of equality operators: === and !==, and their evil twins == and !=. The good ones work the way you would expect. If the two operands are of the same type and have the same value, then === produces true and !== produces false. The evil twins do the right thing when the operands are of the same type, but if they are of different types, they attempt to coerce the values. the rules by which they do that are complicated and unmemorable.`

Looks like this function is checking if property is defined, then go head and assign values. Perhaps it would be better to check for truth-ish value `if (property)` or something like that.

kristinochka:
This solves the problem, now I can update owner string 👍

packages/zowe-explorer/src/job/ZosJobsProvider.ts
traeok: While I generally agree with the excerpt from the JS book, I'm unaware of an existent JavaScript runtime where `undefined` is coerced to a value that's not equal to `null`. If others agree that we should be using `===` here, then I will definitely update this check. For more context, we are already adopting this convention in other areas of the codebase, so I'm merely maintaining the same trend for consistency. 

For the suggestion on `if (property)`: AFAIK JavaScript will implicitly coerce the `property` value to `true/false` if used in this context, so type coercion would still be done here as well. The main difference would be an implicit coercion rather than explicit. However, I'm open to suggestions and willing to discuss this in further detail during stand-up if desired 🙂  

If we would want to change this to strict equality, we should probably update other areas of the codebase (where this check is being used) to maintain consistency.

Thoughts @zFernand0 @t1m0thyj ?

kristinochka: Yes, JavaScript will still coerce `if (property)`, however, this check will return false for any `falsy `condition: https://developer.mozilla.org/en-US/docs/Glossary/Falsy

What if property is `false` or an empty string? The logic inside the if statement should not execute if `property` is any falsy value. It's safer to check for any any falsy value.

traeok: > What if `property` is `false` or an empty string? 

It will never be `false` or an empty string though - in order to compile any calls to `handleEditingMultiJobParameters` w/ TypeScript, `jobProperties` must be of type `IJobPickerOption[]`:

```ts
interface IJobPickerOption {
    key: string;
    label: string;
    value: string;
    show: boolean;
    placeHolder: string;
    validateInput?;
}
```

If calls were made to this function that don't follow the type signature, I would argue that the function is being improperly used, as the type system in TypeScript is supposed to prevent incorrect calls from being compiled by design. However, I understand the concern and am willing to update this, just wanted to get some more perspectives on this topic before continuing.

kristinochka: Yes, using TypeScript will make it less likely that wrong type will be passed, however, in the future somebody could re-factor the code and eliminate the interface or bypass type checking or use the function improperly. If we have more robust check here it will lead to less bugs.

zFernand0: This is an interesting topic 😋 
I can see both side of the argument.

I'm curious why `eliminating an interface` won't be a red-flag for code-reviewers? 

I can understand that there may be places (mostly in tests) where we bypass certain types, but I'm sure that anyone reviewing a PR with an `any` in the codebase (not tests) may have some questions at least.

For what is worth, I believe the CLI squad has adopted the practice of using `!= null` and `!== <anything-else>`.

I understand that there are instances where it makes sense to check for falsy values. but I agree with Trae's point that this very specific case should not be called with a falsy value.

I'm personally biased since the language (JS/TS) allows us to type less with `!=`. However, I'm also ok if the squad decides to change their opinion and always using strict in-/equality going forward.

packages/zowe-explorer/src/job/ZosJobsProvider.ts
```diff
@@ -84,17 +84,18 @@ export class ZosJobsProvider extends ZoweTreeProvider implements IZoweTree<IZowe
         {
             key: `owner`,
             label: `Job Owner`,
-            value: undefined,
+            value: "",
```
zFernand0, line 87: I'm not opposed to this change, but is this the root cause of the discussion regarding `!=`? 😋

zFernand0 (approved):
LGTM! 😋 
We don't have to keep this PR open waiting for a decision on `!=` vs `!==`.
Happy to discuss the topic further in a parking log 😉

packages/zowe-explorer/src/job/ZosJobsProvider.ts
traeok: There are a few assignments in the `handleEditingMultiJobParameters` function that use this `.value` property - e.g. `jobProperties.find((prop) => prop.key === "owner").value`. This just ensures that `.value` will never be undefined for the owner value (the name of the job owner). Although they are both falsy values this will prevent setting one of the node properties, e.g. `node.owner` to something that isn't a string (undefined).

zFernand0: yeah, that makes sense 😋 
I thought that this `value: undefined` was one of the possible reasons 😋 
Thanks for explaining though 😅

t1m0thyj (approved):

Issue #1116, opened 2020-12-21
Submit start with a slash


moueza:
From my answer https://github.com/zowe/jobs/issues/38 :

Reproduced in openmainframeproject : on VS code Linux Ubuntu 20.10 cbl0002 (6. of 10.4 Lab of pdf1) : right click Submit job :
`Job submission failed
z/OSMF REST API Error:
Rest API failure with HTTP(S) status 400
rc: 4
reason: 17
category: 6
message: Submit input data does not start with a slash

Error: z/OSMF REST API Error:
Rest API failure with HTTP(S) status 400
rc: 4
reason: 17
category: 6
message: Submit input data does not start with a slash

`

Marjona6:
Are you submitting the COBOL source code or the JCL? I had this error when trying to submit the COBOL source.

moueza:
I had no JCL ( because I had not filtered for id.*) ... so I did from CBL.

You must do from JCL (by well filtering), and it is OK !

SOLVED

sravanisf8:
Command Error:
z/OSMF REST API Error:
Rest API failure with HTTP(S) status 400
rc:       4
reason:   17
category: 6
message:  Submit input data does not start with a slash

Error Details:
HTTP(S) error status "400" received.
Review request details (resource, base path, credentials, payload) and ensure correctness.

  
Can someone help me how to overcome the above error in ZOWE

zFernand0:
@LukeG-96 
Seems like this was related to an operation (filtering) done on the JCL before being submitted that caused the issue.
Could you verify that the JCL being submitted is properly formatted?

LukeG-96:
@zFernand0 
So what's weird is when I right click on the member name and click "Pull from mainframe" it adds 1 space in front of each line of the JCL. So I delete that line, save the member and try to re-run the job and it still gives me this same error. I am not sure what is causing this reformat to happen, but it basically makes the option to use zowe through vs code unusable...

zFernand0:
**TL;DR**: This could be related to a race condition in how we save and upload contents to the mainframe
- Details: https://github.com/zowe/vscode-extension-for-zowe/pull/2164

Workaround:
If you get the old contents (i.e. 1 space added in front of each line) after doing `Pull From Mainframe`, please make sure that the content is successfully uploaded after removing the extra space in front.

A quick way I personally work around this race condition is by triggering the upload request multiple times. For example:
1. adding a new line at the end of the file, 
2. saving the file, 
[...]

Issue #2099, opened 2023-01-25
Error not displayed in popup message for saving data


JillieBeanSim:
Per Bill Pereira:

```
Hi Team… in the past Zowe used to give feedback about data being saved, 
on mainframe side, or not with a popup, now it gives just a small message 
in the status bar… I had issues in the past of data not being reflected on zOS 
side when after long time using VSCode… just as feedback… as people are 
using more and more it here more people are complaining about this changes, 
as they also don’t note this small error message on the status bar and loose 
the work done when they notice it … :disappointed:
```

Per Hartanto Ario Widjaya (using zOSMF profile):

```
I should say that I only noticed the issue with USS files. I had several USS files 
back when I was doing the challenges for the IBM Z Student Contest a few months 
ago and I noticed that changes won't save after leaving the VSCode for some time.
One indicator for which I see to know whether the changes made to the USS files 
are saved remotely or not is the presence of a "Saving file" icon on the bottom left 
of VSCode (as attached on this post). If I don't see such icon, I would re-start VSCode 
and remade the changes, and it would saved correctly.
I am not aware whether such thing happened with normal z/OS SEQ data sets or PDS 
members, and I am not aware on how we can effectively replicate this issue.
```

**Expected behavior**

Any error returned should be displayed in a popup message to get the user's attention for awareness.

Issue #1948, opened 2022-10-05
Data corruption happens during update or modification operation


asatklichov:
Data corruption happens during update or modification operation once you work on multiple members using Zowe Explorer. 

How to reproduce: 
Open multiple members, and just try to copy paste some content from one to others, and meanwhile switch between members, close some of them or open new ones etc.  Or similar operation try to do on renaming members etc. 

As a result, some members are updated with content which you have not put on it, or un-expected overridden happens. And everything is written to MF, and you either lose the data or corrupt your member. 

I believe issue is during synchronization, especially working on multiple resources. 

Thanks

Issue #1800, opened 2022-05-12
Zowe Explorer does not react gracefully to degraded MF response times - There is a significant risk of code loss


jguillaumes:
** Description **

The extension does not react properly to degraded response times from the mainframe - There is a significant risk of code loss

When the response time from the ZOSMF server degrades the updates VSCODE sends to the mainframe begin to fail. The pending updates begin to queue:

![VSCodeCapture](https://user-images.githubusercontent.com/1552242/168050162-d4919efa-d703-4e2c-a0db-feeace7b8685.png)

Note all the "saving file" notifications correspond to the same file, in this case an USS C source code.

At this point a lot of "the remote file has been modified..." warnings appear. The warning window gets immediate focus, so if the user is typing it closes immediately, maybe with an unwanted option. Depending on the sequence of events, parts of the WHOLE of the code being edited can be lost. 

**To Reproduce**

I cannot reproduce it at will. As an educated guess, introducing a forced delay in the zOSMF APIs _could_ trigger this behaviour.

**Expected behavior**

The extension should gracefully manage a degraded response time situation. In any case, the source code **should** be preserved. Without knowing the internals, maybe throttling down the updates to ZOSMF could aleviate the problem.

**Screenshots**

See above.

**Desktop (please complete the following information):**

- OS: Ubuntu Linux 20.04 running under Windows 10 WSL2 (has also been observed under "native" windows)
- Zowe Explorer Version: 2.0.1
- (Optional) Zowe CLI Version:
- (Optional) Are you using Secure Credential Store? Not in this environment

**Additional context**

jguillaumes:
Additional information:
- This behaviour can be triggered by rapidly spamming CTRL-S and forcing multiple updates in fast succession.
- It can be avoided **disabling the Visual Studio Code autosave feature in the editor settings**.
- The autosave feature is **enabled by default**, with an autosave interval of 1 second. That means every second VSC sends the whole file or member to the mainframe invoking the PUT /zosmf/restfiles/ds service. That is unsustainable (imagine multiple developers typing on big source files sending their full content every second).
In my opinion the ZOWE extension should **NOT** use the autosave feature at all, or at least, use its own setting separated from the global VSC one.
- I suggest to add a clearly hightlighted recommendation to the welcome page directing the user to deactivate autosave if he is going to use the extension to edit mainframe files.

JillieBeanSim:
[...]

PR #2164, opened 2023-02-28
Fix data corruption issues and document file save flow

t1m0thyj:
## Proposed changes

* Created flowcharts that document ZE file save flow
* Fixed several issues identified that could lead to possible data corruption
  * The `onWillSaveTextDocument` event has some undesired behavior. Switched to `onDidSaveTextDocument` to fix the following:
    * #2099 - If the mainframe connection is slow and the upload API call consistently takes longer than 1.5 seconds, then VS Code will stop calling our `onWillSaveTextDocument` handler (see screenshot below) and we can't upload files anymore until VS Code is restarted.
    * #1948 - There is a potential race condition because 2 actions are running in parallel: (1) VS Code saves file to disk and (2) We upload file from disk to the mainframe. There is no guarantee that (1) has finished before (2), so if the filesystem is slow or there are a lot of concurrent save operations, we may be uploading an old version of the file.
    * #1800 - Previously we adopted `onWillSaveTextDocument` to fix this issue, so we need to ensure there are no regressions by switching to `onDidSaveTextDocument`. This PR avoids flooding the mainframe with requests by discarding intermediate save requests - all upload operations are processed sequentially in a queue, and uploading a file will be skipped if there is already a newer version of that same file to upload.
  * Wait for pending save/upload operations to complete before VS Code exits. There is a 5 second timeout on extension deactivation (according to https://github.com/Microsoft/vscode/issues/47881#issuecomment-381910587), so if the mainframe connection is slow there's no guarantee these operations will complete in time. I don't think there's anything we can do about this since VS Code doesn't provide a way to delay exiting. Worst case scenario is that the upload times out so we skip deleting the temp folder and the latest changes are cached on disk, which is significantly better than the previous case where pending uploads were totally lost when exiting VS Code.
  * Check if data set exists before saving should be performed for all DS types. Previously it was skipped for PDS datasets.

## Release Notes

Thanks to @traeok for co-authoring 🙂 

Milestone: 2.7.0

Changelog:
Fixed issues where document changes may fail to upload if the environment has a slow filesystem or mainframe connection, or when VS Code exits during an upload operation.

## Types of changes

## Checklist

## Further comments

image


packages/zowe-explorer/src/abstract/ZoweSaveQueue.ts
```diff
@@ -0,0 +1,89 @@
...
+    private static savingQueue: SaveRequest[] = [];
+
+    /**
+     * Iterate over the queue and process next item until it is empty.
+     */
+    private static async processNext() {
+        const nextRequest = this.savingQueue.shift();
+        if (nextRequest == null || this.savingQueue.some(({ savedFile }) => savedFile.fileName === nextRequest.savedFile.fileName)) {
+            return;
+        }
```
t1m0thyj, line 64: Should we add a test to verify that redundant save requests are discarded?

packages/zowe-explorer/src/extension.ts
```diff
@@ -71,6 +72,7 @@ export async function activate(context: vscode.ExtensionContext): Promise<ZoweEx
  * @export
  */
 export async function deactivate() {
+    await ZoweSaveQueue.all();
```
t1m0thyj, line 75: Should we add a test to verify that this method gets called on deactivate?

zFernand0:
This PR may also fix https://github.com/zowe/vscode-extension-for-zowe/issues/1116#issuecomment-1453892002

packages/zowe-explorer/src/shared/utils.ts
```diff
@@ -274,36 +274,32 @@ export async function willForceUpload(
...
-                const uploadResponse = Gui.withProgress(
-                    {
-                        location: vscode.ProgressLocation.Notification,
-                        title,
-                    },
-                    () => {
-                        return uploadContent(node, doc, remotePath, profile, binary, null, returnEtag);
+    Gui.infoMessage(localize("saveFile.info.confirmUpload", "Would you like to overwrite the remote file?"), {
+        items: [localize("saveFile.overwriteConfirmation.yes", "Yes"), localize("saveFile.overwriteConfirmation.no", "No")],
+    }).then(async (selection) => {
```
zFernand0, line 279: the `willForceUpload` method is already async.
I'm curious why we don't use `const selection = await ...`

packages/zowe-explorer/src/shared/utils.ts
```diff
@@ -274,36 +274,32 @@ export async function willForceUpload(
...
-            if (selection === localize("saveFile.overwriteConfirmation.yes", "Yes")) {
-                const uploadResponse = Gui.withProgress(
-                    {
-                        location: vscode.ProgressLocation.Notification,
-                        title,
-                    },
-                    () => {
-                        return uploadContent(node, doc, remotePath, profile, binary, null, returnEtag);
+    Gui.infoMessage(localize("saveFile.info.confirmUpload", "Would you like to overwrite the remote file?"), {
+        items: [localize("saveFile.overwriteConfirmation.yes", "Yes"), localize("saveFile.overwriteConfirmation.no", "No")],
```
zFernand0, line 278: I'm hoping that some day we reduce the duplication of string in our i18n files 😋 
ideally we will have some sort of shared strings file where things like `yes` and `no` will be included.

I'm not asking for this to be changed in this PR though 😋

packages/zowe-explorer/src/shared/utils.ts
```diff
@@ -274,36 +274,32 @@ export async function willForceUpload(
...
+    }).then(async (selection) => {
+        if (selection === localize("saveFile.overwriteConfirmation.yes", "Yes")) {
+            const uploadResponse = Gui.withProgress(
+                {
+                    location: vscode.ProgressLocation.Notification,
+                    title,
+                },
+                () => {
+                    return uploadContent(node, doc, remotePath, profile, binary, null, returnEtag);
```
zFernand0, line 287: Shouldn't we await the `uploadContent` call?

packages/zowe-explorer/src/shared/utils.ts
```diff
@@ -274,36 +274,32 @@ export async function willForceUpload(
...
+                }
+            );
+            uploadResponse.then((response) => {
```
zFernand0, line 290: Same comment as before regarding then vs await.
Not sure if there is a preference, but I rather not mix them if possible 😋

packages/zowe-explorer/src/shared/init.ts
```diff
@@ -116,34 +116,28 @@ export async function registerCommonCommands(context: vscode.ExtensionContext, p
...
                         "File was saved -- determining whether the file is a USS file or Data set.\n Comparing (case insensitive) "
                     ) +
-                        savedFile.document.fileName +
+                        savedFile.fileName +
                         localize("onDidSaveTextDocument2", " against directory ") +
                         globals.DS_DIR +
                         localize("onDidSaveTextDocument3", "and") +
                         globals.USS_DIR
                 );
-                if (!savedFile.document.isDirty) {
```
zFernand0, line 131: Just for my personal understanding, we are adding non-dirty files to the queue, right?

packages/zowe-explorer/src/extension.ts
zFernand0: thanks for adding the test 🥳

packages/zowe-explorer/src/dataset/actions.ts
```diff
@@ -1356,7 +1356,8 @@ export async function saveFile(doc: vscode.TextDocument, datasetProvider: api.IZ
     const profile = Profiles.getInstance().loadNamedProfile(sesName);
     if (!profile) {
         globals.LOG.error(localize("saveFile.log.error.session", "Couldn't locate session when saving data set!"));
-        return api.Gui.errorMessage(localize("saveFile.log.error.session", "Couldn't locate session when saving data set!"));
+        await api.Gui.errorMessage(localize("saveFile.log.error.session", "Couldn't locate session when saving data set!"));
+        return;
```
zFernand0, line 1360: curious why not `return await api.Gui.errorMessage`?
Also... should we type the return value of this (`saveFile`) function to `: Promise<void>`?

packages/zowe-explorer/src/dataset/actions.ts
```diff
@@ -1373,18 +1374,18 @@ export async function saveFile(doc: vscode.TextDocument, datasetProvider: api.IZ
...
-            // Checks if file still exists on server
-            const response = await ZoweExplorerApiRegister.getMvsApi(profile).dataSet(label, { responseTimeout: profile.profile?.responseTimeout });
-            if (!response.apiResponse.items.length) {
-                return api.Gui.errorMessage(
-                    localize("saveFile.error.saveFailed", "Data set failed to save. Data set may have been deleted on mainframe.")
-                );
-            }
-        } catch (err) {
-            await errorHandling(err, sesName, err.message);
+    const dsname = label.includes("(") ? label.slice(0, label.indexOf("(")) : label;
```
zFernand0, line 1377: Smart way to get the DS name.
Curious if this should be made a common utility (either in the CLI or the Zowe Explorer API)?
I know I had to create a similar one for the FTP plugin not long ago 😋 

---

Unrelated question, _what if_ we decide to change the label in order to add more complex descriptions (instead of using the `description` field)… will this be trimming undesired portions of the label?

packages/zowe-explorer/src/dataset/actions.ts
```diff
@@ -1373,18 +1374,18 @@ export async function saveFile(doc: vscode.TextDocument, datasetProvider: api.IZ
...
+    try {
+        // Checks if file still exists on server
+        const response = await ZoweExplorerApiRegister.getMvsApi(profile).dataSet(dsname, { responseTimeout: profile.profile?.responseTimeout });
+        if (!response.apiResponse.items.length) {
+            await api.Gui.errorMessage(
+                localize("saveFile.error.saveFailed", "Data set failed to save. Data set may have been deleted or renamed on mainframe.")
+            );
+            return;
```
zFernand0, line 1385: again, just out of curiosity, why not just have this in one line 😋

packages/zowe-explorer/__tests__/__unit__/shared/init.unit.test.ts
```diff
@@ -102,23 +103,45 @@ describe("Test src/shared/extension", () => {
...
-                name: "onWillSaveTextDocument:3",
-                parm: [{ document: { isDirty: true, fileName: "USS_DIR" } }],
-                mock: [{ spy: jest.spyOn(workspaceUtils, "handleSaving"), arg: [saveUSSFile, { isDirty: true, fileName: "USS_DIR" }, "uss"] }],
+                name: "onDidSaveTextDocument:3",
+                parm: [{ fileName: "USS_DIR", isDirty: true, uri: vscode.Uri.parse("") }],
+                mock: [
+                    {
+                        spy: jest.spyOn(ZoweSaveQueue, "push"),
+                        arg: [
+                            {
```
zFernand0, line 133: wow, thank for updating the `it`s.
I hope it wasn't too painful. 😋

packages/zowe-explorer/src/abstract/ZoweSaveQueue.ts
```diff
@@ -0,0 +1,89 @@
...
+// Set up localization
+import * as nls from "vscode-nls";
+nls.config({
+    messageFormat: nls.MessageFormat.bundle,
+    bundleFormat: nls.BundleFormat.standalone,
+})();
+const localize: nls.LocalizeFunc = nls.loadMessageBundle();
+
+interface SaveRequest {
+    uploadRequest: (document, provider) => Promise<void>;
```
zFernand0, line 27: should this be `(_document, _provider) => Promise<void>` to avoid lint warnings?
or can we update the rule to exclude `interface` definitions?

packages/zowe-explorer/src/abstract/ZoweSaveQueue.ts
```diff
@@ -0,0 +1,89 @@
...
+    savedFile: vscode.TextDocument;
+    fileProvider: IZoweTree<IZoweUSSTreeNode | IZoweDatasetTreeNode>;
```
zFernand0, line 29: This makes it easy to extend.
Curious what will happen if say an extender needed to use the ZoweSaveQueue for their own tree?

I'm wondering if we should use generic types.
If that makes some sense, we can worry about that if/when someone asks for it 😋

packages/zowe-explorer/src/dataset/actions.ts
traeok: Currently we are marking `return await` as a warning in ESLint, not sure if we would want to change this as part of #2068 or leave it as-is.

packages/zowe-explorer/src/utils/workspace.ts
```diff
@@ -131,12 +131,13 @@ export async function closeOpenedTextFile(path: string) {
...
 
-/**
- * Handle auto/regular save by prioritizing the last ongoing save of a series queued saves
- */
 const savingQueue = [];
 let latestSavedFile: vscode.TextDocument;
 let ongoingSave = false;
+/**
+ * Handle auto/regular save by prioritizing the last ongoing save of a series queued saves
+ * @deprecated Use ZoweSaveQueue class instead
```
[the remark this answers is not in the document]
JillieBeanSim: since this is Zowe Explorer code and not the ZE APIs I don't think we need to deprecate this and just change all calls to the new?

packages/zowe-explorer/src/dataset/actions.ts
zFernand0: Oh, that's great that we have that as a rule.
I'm a fan of one-liners. However, in this particular case (with `return await`) I have gotten confused sometimes and question myself on whether or not I meant to await it or just return the promise and let the caller deal with it.

I'm ok with leaving it as-is 😋 
Thanks for entertaining the conversation 🥳

packages/zowe-explorer/src/shared/utils.ts
t1m0thyj: I believe we don't want the `willForceUpload` to await the user's response to the upload confirmation. If we did, it would block all subsequent save requests in the queue from being processed.

packages/zowe-explorer/src/shared/init.ts
t1m0thyj: That's right, if a save event is triggered by VS Code for a non-dirty file, we will process it.

packages/zowe-explorer/src/shared/utils.ts
t1m0thyj: If we add `await` here we'd have to split up this line into 2 to satisfy the ESLint `no-return-await` rule. Is it ok to wait to handle this until we refine our ESLint rules (#2068)?

JillieBeanSim (approved):
LGTM! thanks @t1m0thyj for the fix

zFernand0 (approved):
LGTM! 😋

traeok (approved):

PR #2185, opened 2023-03-16
Fix for DS upload silent fail

JillieBeanSim:
## Proposed changes

If something goes wrong in data set upload it fails silently, no message for user and possible loss of data that didn't upload.

This PR will use a pop up message with task bar and file count of upload progress and if error is returned the message is shown to user and upload is halted.

## Release Notes

Milestone: 2.8.0

Changelog: Fixed issue with silent failures when uploading members into a data set.

## Types of changes

## Checklist

## Further comments

my connection was really slow when recording

https://user-images.githubusercontent.com/49491949/225982939-f2665ec5-6864-4876-bd7f-ec900e7aaa6c.mov


packages/zowe-explorer/src/shared/actions.ts
```diff
@@ -236,3 +236,20 @@ export async function resetValidationSettings(node: IZoweNodeType, setting: bool
...
+/**
+ * Reports the progress of a Gui.withProgress action results in visual update for users.
+ *
+ * @export
+ * @param {any} progress - Gui.withProgress progress
+ * @param {number} valueLength - values array length used to calculate progress
+ * @param {number} index - withProgress loop array's index
+ * @param {string} action - the action the progress is reported for, ie. Uploading
+ */
+export function reportProgress(progress: any, valueLength: number, index: number, action: string): void {
```
t1m0thyj, line 249: If this method can be used with any `Gui.withProgress` action, would it make sense to define it in the `Gui` namespace of the ZE API package?

t1m0thyj (commented):
Thanks for making the upload process more stable 🙂 I think its good that you re-implemented multiple uploads as a sequential operation rather than a parallel one - makes it easier to handle errors, and is consistent with the way we queue upload/download operations elsewhere in ZE.

Left a few comments, will approve after I get a chance to test it.

packages/zowe-explorer/src/dataset/actions.ts
```diff
@@ -174,14 +191,16 @@ export async function uploadDialog(node: ZoweDatasetNode, datasetProvider: api.I
...
         const prof = node.getProfile();
-        await ZoweExplorerApiRegister.getMvsApi(prof).putContents(doc.fileName, datasetName, {
+
+        const response = await ZoweExplorerApiRegister.getMvsApi(prof).putContents(doc.fileName, datasetName, {
             encoding: prof.profile?.encoding,
             responseTimeout: prof.profile?.responseTimeout,
         });
+        return response;
     } catch (e) {
         await errorHandling(e, node.getProfileName(), e.message);
```
traeok, line 205: It might have been the way I was testing, but I disconnected my internet in the middle of the upload and noticed that while there is an error dialog that comes back, there is also an error from `mDetails` being undefined in `errorHandling`:

image

I think if we would add a `?` to line 50 of `ProfilesUtils.ts`, then that should resolve the issue:
```ts
if (errorDetails?.mDetails !== undefined)
```

traeok (commented):
Thanks @JillieBeanSim , noticed a slight issue with `errorHandling` but I did some testing and everything else LGTM 🙂

packages/zowe-explorer/src/dataset/actions.ts
```diff
@@ -144,20 +145,36 @@ export async function uploadDialog(node: ZoweDatasetNode, datasetProvider: api.I
...
+            },
+            async (progress, token) => {
+                let index = 0;
+                for (const item of value) {
+                    if (token.isCancellationRequested) {
+                        api.Gui.showMessage(localize("uploadFile.uploadCancelled", "Upload action was cancelled."));
+                        break;
+                    }
+                    reportProgress(progress, value.length, index, "Uploading");
+                    const doc = await vscode.workspace.openTextDocument(item);
```
phaumer, line 164: For putContents all we need is the file path. Calling the VS Code openTextDocument() has a big performance impact. I can see in my log files that loading it as a document will actually call the language server, if you have one installed for the language of the file being uploaded, sending the file around, fully parsing it, etc. I would just hand down the file path to upload and nothing else as the files will not be opened in an editor at all.

phaumer (changes requested):
This works great, but I would like to remove the call that loads the file into the VS Code document cache, which is not needed, but can cost performance and memory.

With z/OSMF I see a weird quirk. When I create a PDS with the Zowe Explorer defaults it can only contain 29 members. But when I upload 30 then I get an error for the last one, but that member still appears in the tree and I can open it and it looks complete. When I try uploading more members it stops after the first one, with the same error, but also that first one appears in the tree afterwards. I did not see the same behavior with other data providers. For IBM's RSE API it reports the error for member 30 and only 29 members appear in the tree afterwards.

packages/zowe-explorer/src/dataset/actions.ts
JillieBeanSim: addressed not opening doc to get path when item itself was holding the information in commit https://github.com/zowe/vscode-extension-for-zowe/pull/2185/commits/11195f8623ebbe500ae2ac1e89b911cb5c4d5f7a

with the zOSMF specific issue, should we create a bug with zOSMF rest api?

packages/zowe-explorer/CHANGELOG.md
```diff
@@ -8,10 +8,13 @@ All notable changes to the "vscode-extension-for-zowe" extension will be documen
 
 ### Bug fixes
 
+- Fixed issue with silent failures when uploading members into a data set. [#2167](https://github.com/zowe/vscode-extension-for-zowe/issues/2167)
+
 ## `2.7.0`
 
 ### New features and enhancements
 
+- Added Job search label to the Jobs tree. Label will show current search query. [#2062](https://github.com/zowe/vscode-extension-for-zowe/pull/2064)
```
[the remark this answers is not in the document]
JillieBeanSim: addressed in commit https://github.com/zowe/vscode-extension-for-zowe/pull/2185/commits/11195f8623ebbe500ae2ac1e89b911cb5c4d5f7a

packages/zowe-explorer/src/dataset/actions.ts
JillieBeanSim: I've reached out to the zOSMF team in regards to the zosmf type specific bug.

phaumer (approved):
Thanks for the fixes. 

The one thing that should be polished is the File selection dialog. It seems to be opened showing the most recently used folder such as the temp directory where you uploaded last or a directory from a previous workspace you worked with. Then the user has to do a lot of clicking getting back to the current workspace, which is the most likely location for the files to upload imho. We could specify a default URI to the folder where the dialog should point to when opened, which could be the first workspace. See https://code.visualstudio.com/api/references/vscode-api#OpenDialogOptions

JillieBeanSim:
> https://code.visualstudio.com/api/references/vscode-api#OpenDialogOptions

That's a great idea, i did notice it opening in the temp folder myself.  created an [issue](https://github.com/zowe/vscode-extension-for-zowe/issues/2203) to tackle all uploadDialogs/openFiles at once

PR #2188, opened 2023-03-17
Fix errors related to Imperative overrides file

### LINKED ISSUE #2187: ZE overrides custom credential manager with invalid JSON
**Describe the bug**

When there is a custom credential manager defined in Imperative settings, Zowe Explorer overwrites it with `@zowe/cli` and in some cases writes invalid JSON to the file. For example:
```json
{
  "overrides": {
    "CredentialManager": "@zowe/cli"
  }
}r-for-zowe-cli"
  }
}
```

**To Reproduce**

1. Edit your `~/.zowe/settings/imperative.json` file to simulate that a credential manager plugin has been installed. The override name must be longer than the original name `@zowe/cli` - for example it can be `@zowe/secure-credential-store-for-zowe-cli` or `@zowe/secrets-for-kubernetes-for-zowe-cli`
2. Reload VS Code and notice that your profiles fail to load.
3. There is an error shown but it is misleading because it suggests there is a problem with your Zowe config, when the problem is really in the Imperative settings file.

**Expected behavior**

* When ZE launches it should create `imperative.json` if it doesn't exist, but should not overwrite the file if it already exists unless requested to.
* When ZE updates the contents of `imperative.json` it should remove all old contents from the file even if the new contents are shorter.
* When ZE fails to parse the contents of `imperative.json` it should show an error message that accurately conveys what happened.

t1m0thyj:
## Proposed changes

Fixes #2187 by implementing the following behavior:
* When ZE launches it should create `imperative.json` if it doesn't exist, but should not overwrite the file if it already exists unless requested to.
* When ZE updates the contents of `imperative.json` it should remove all old contents from the file even if the new contents are shorter.
* When ZE fails to parse the contents of `imperative.json` it should show an error message that accurately conveys what happened.

Also fixes a few other credential-related issues discovered when testing the K8s plugin:
* The utility method `readConfigFromDisk` was catching errors and rethrowing them with less information, which can make it difficult to debug problems with loading custom credential managers.
* When v1 profiles are in use, right-clicking on a session and selecting "Update Credentials" may update credentials for the wrong profile if another profile exists with a different type but the same name.

## Release Notes

Milestone: 2.8.0

Changelog: Fixed custom credential manager in `~/.zowe/settings/imperative.json` file being overwritten with invalid JSON.

## Types of changes

## Checklist

### DIFF
--- a/packages/zowe-explorer/src/Profiles.ts
+++ b/packages/zowe-explorer/src/Profiles.ts
@@ -555,7 +555,7 @@ export class Profiles extends ProfilesCache {
 
             return updSchemaValues;
         } catch (error) {
-            await errorHandling(error.message);
+            await errorHandling(error, profileName, error.message);
         }
     }
 
@@ -839,7 +839,7 @@ export class Profiles extends ProfilesCache {
             await readConfigFromDisk();
             return newProfileName;
         } catch (error) {
-            await errorHandling(error.message);
+            await errorHandling(error, profileName, error.message);
             ZoweExplorerExtender.showZoweConfigError(error.message);
         }
     }
--- a/packages/zowe-explorer/src/Profiles.ts
+++ b/packages/zowe-explorer/src/Profiles.ts
@@ -92,7 +92,7 @@ export class Profiles extends ProfilesCache {
             });
             let values: string[];
             try {
-                values = await Profiles.getInstance().promptCredentials(theProfile.name);
+                values = await Profiles.getInstance().promptCredentials(theProfile);
             } catch (error) {
                 errorHandling(
                     error,
@@ -844,7 +844,7 @@ export class Profiles extends ProfilesCache {
         }
     }
 
-    public async promptCredentials(sessionName: string, rePrompt?: boolean): Promise<string[]> {
+    public async promptCredentials(profile: string | zowe.imperative.IProfileLoaded, rePrompt?: boolean): Promise<string[]> {
         const userInputBoxOptions: vscode.InputBoxOptions = {
             placeHolder: localize("createNewConnection.option.prompt.username.placeholder", "User Name"),
             prompt: localize("createNewConnection.option.prompt.username", "Enter the user name for the connection. Leave blank to not store."),
@@ -856,7 +856,8 @@ export class Profiles extends ProfilesCache {
 
         const promptInfo = await ZoweVsCodeExtension.updateCredentials(
             {
-                sessionName,
+                sessionName: typeof profile !== "string" ? profile.name : profile,
+                sessionType: typeof profile !== "string" ? profile.type : undefined,
                 rePrompt,
                 secure: (await this.getProfileInfo()).isSecured(),
                 userInputBoxOptions,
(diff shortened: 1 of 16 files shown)


traeok (dismissed):
LGTM, thanks @t1m0thyj

traeok (dismissed):

packages/zowe-explorer/src/uss/actions.ts
```diff
@@ -64,7 +64,6 @@ export async function createUSSNode(node: IZoweUSSTreeNode, ussFileProvider: IZo
             filePath = `${filePath}/${name}`;
             await ZoweExplorerApiRegister.getUssApi(node.getProfile()).create(filePath, nodeType);
             if (isTopLevel) {
-                await Profiles.getInstance().refresh(ZoweExplorerApiRegister.getInstance());
```
traeok, line 67: Not really a suggestion, just curious to know why we've removed this 😅

traeok (approved):

packages/zowe-explorer/src/uss/actions.ts
t1m0thyj: The next line calls the `refreshAll` method, and the first thing that `refreshAll` does is call `await Profiles.getInstance().refresh(ZoweExplorerApiRegister.getInstance());` so we were doing it twice 😁

packages/zowe-explorer/src/Profiles.ts
```diff
@@ -839,12 +839,12 @@ export class Profiles extends ProfilesCache {
...
             return newProfileName;
         } catch (error) {
-            await errorHandling(error.message);
+            await errorHandling(error, profileName, error.message);
             ZoweExplorerExtender.showZoweConfigError(error.message);
         }
     }
 
-    public async promptCredentials(sessionName: string, rePrompt?: boolean): Promise<string[]> {
+    public async promptCredentials(profile: string | zowe.imperative.IProfileLoaded, rePrompt?: boolean): Promise<string[]> {
```
JillieBeanSim, line 847: Hey @t1m0thyj why would we keep the option of string | IProfileLoaded?  Should we just make it IProfileLoaded and update any calls to it?

zFernand0: I know this is only code in the zowe-explorer package.
But is it possible for extenders to call methods exported from the ZE's VSIX file (as a module)? (probably not recommended - if it's possible)

If that's the case, we may want to be careful with changing function signatures without maintaining the original implementation. 😋 

If we are not concerned about that, then we can  just replace the string and update our calls 😋

JillieBeanSim: we shouldn't have to worry about extenders trying to use ZE methods and should be able to adjust that code at any time unlike the ZE API package that is recommended for use to extend Zowe Explorer.

t1m0thyj: I agree that since this method isn't in the ZE API package, we shouldn't need to be concerned about making breaking changes to its signature.

I kept the option of passing a string because we call `promptCredentials` that way in our `errorHandling` method:
https://github.com/zowe/vscode-extension-for-zowe/blob/6fa16aad1897aa1fa339cc16fe10662991b724ec/packages/zowe-explorer/src/utils/ProfilesUtils.ts#L104

Within the `errorHandling` method we don't have access to an `IProfileLoaded` object, just the node label.
I'm open to refactoring this if we want but it would add scope to the PR since the `errorHandling` method is called around 70 times across our codebase.

Issue #2212, opened 2023-03-28
Add support for custom credential manager extensions


t1m0thyj:
As a developer of a VSCE that supplies a custom credential manager for Zowe,
I want ZE to detect that my extension is installed and load it at launch time,
So that secure credentials managed by ZE will use my custom credential manager.

**Resources**
* https://github.com/zowe/zowe-cli-secrets-for-kubernetes
* https://github.com/zowe/imperative/pull/937
* https://github.com/zowe/imperative/pull/957

t1m0thyj:
After discussing with the team, here are some implementation ideas:

Enhance Zowe Explorer to detect and load a custom credential manager at startup before loading CLI profiles.

1. It could read the contents of the Imperative settings file, and check the mapping mentioned above to see if the associated VS Code extension is installed.
    * Setting is defined in `~/.zowe/settings/imperative.json` or in team config files:
      `"CredentialManager": "Secrets for Kubernetes"`

2. If the credential manager is valid and an associated extension is installed, then ZE will load and use the custom credential manager extension instead of Keytar.

3. If the credential manager is valid and an associated extension is not installed, then ZE will prompt the user asking if they wish to install the extension from the VS Code Marketplace.
    * To verify that extension exists on marketplace, we could use: https://open-vsx.org/swagger-ui/index.html#/registry-api/getExtension

4. If the credential manager is invalid, then ZE will prompt the user asking if they wish to reset it to the default value (`@zowe/cli`) which uses Keytar.

PR #2230, opened 2023-04-11
Add support for custom credential manager extensions

rudyflores:
## Proposed changes

Solves #2212 

Add support for custom credential manager extensions. When a new custom credential manager extension is installed, Zowe Explorer will activate that extension and handle credentials for Zowe Profiles through the custom credential manager override set in the `imperative.json` file.

e.g: 
- User installs `Kubernetes secrets credential manager extension`
- `imperative.json` is set to `Secrets for Kubernetes`
- Zowe Explorer activates the extension and begins securing/storing credentials with Kubernetes secrets
- User uninstalls `Kubernetes secrets credential manager extension`
- Refresh is required for changes to happen
- `imperative.json` is set to the default `@zowe/cli`
- Zowe Explorer begins to secure credentials with Keytar

## Release Notes

Milestone: TBD

Changelog: Added support for custom credential manager extensions in Zowe Explorer

## Types of changes

## Checklist

## Further comments

This change will also remove the old code that will default `imperative.json` back to `@zowe/cli` if the unsecure profile is in use. From now on `imperative.json` should reflect what plugin is installed, the default is in use, or no credential manager is needed.


JillieBeanSim (approved):
Thanks @rudyflores for this enhancement

Would you mind adding a section in the wiki under Extenders section to explain the new addition please?

JillieBeanSim:
> @JillieBeanSim documentation written for custom credential manager support: https://github.com/zowe/vscode-extension-for-zowe/wiki/Secure-Credentials-for-Extenders#using-a-custom-credential-manager

Thanks @rudyflores, docs look great!

JillieBeanSim (dismissed):

packages/zowe-explorer/src/globals.ts
```diff
@@ -342,13 +342,18 @@ export function setSavedProfileContents(value: Uint8Array): void {
     SAVED_PROFILE_CONTENTS = value;
 }
 
-export async function setGlobalSecurityValue(): Promise<void> {
-    if (ISTHEIA && !SettingsConfig.isConfigSettingSetByUser(SETTINGS_SECURE_CREDENTIALS_ENABLED)) {
+export async function setGlobalSecurityValue(credentialManager?: string): Promise<void> {
+    const settingEnabled: boolean = SettingsConfig.getDirectValue(this.SETTINGS_SECURE_CREDENTIALS_ENABLED);
+    if (settingEnabled && credentialManager) {
+        PROFILE_SECURITY = credentialManager;
+        await SettingsConfig.setDirectValue(this.SETTINGS_SECURE_CREDENTIALS_ENABLED, true, vscode.ConfigurationTarget.Global);
```
t1m0thyj, line 349: If the setting is already enabled, do we need to set it here?

packages/zowe-explorer/src/utils/ProfilesUtils.ts
```diff
@@ -209,18 +210,77 @@ export class FilterDescriptor implements vscode.QuickPickItem {
...
+            imperative.CredentialManagerOverride.recordCredMgrInConfig(setting);
+        }
+    }
+
+    public static async getProfileInfo(envTheia: boolean): Promise<imperative.ProfileInfo> {
         ZoweLogger.trace("ProfilesUtils.getProfileInfo called.");
-        const mProfileInfo = new imperative.ProfileInfo("zowe", {
-            requireKeytar: () => getSecurityModules("keytar", envTheia),
+        const credentialManagerMap = ProfilesUtils.getCredentialManagerOverride();
+        const customCredentialManagerExtension = vscode.extensions.getExtension(credentialManagerMap?.credMgrZEName ?? "");
```
t1m0thyj, line 256: Instead of calling `getExtension` with an empty string, should we do something like this?
`const customCredentialManagerExtension = credentialManagerMap?.credMgrZEName && vscode.extensions.getExtension(credentialManagerMap.credMgrZEName)`

packages/zowe-explorer/src/utils/ProfilesUtils.ts
```diff
@@ -319,7 +380,7 @@ export class ProfilesUtils {
         let fileContent: string;
         const settingsFile = path.join(getZoweDir(), "settings", "imperative.json");
         try {
-            fd = fs.openSync(settingsFile, "r+");
+            fd = fs.openSync(settingsFile, "w+");
```
t1m0thyj, line 383: I think we want `r+` here to avoid overwriting the contents of the file if it already exists.

rudyflores: w+ seems to solve our current existing issue with this function which creates this broken imperative.json when loading a plugin of a custom credential mananger and reloading Zowe Explorer, for example:
```JSON
{
    overrides: {
             credentialManager: "my-credential-manager"
    }
}
}
}
```

t1m0thyj: I think the issue with a broken imperative.json was fixed in #2188:
> * When ZE updates the contents of `imperative.json` it should remove all old contents from the file even if the new contents are shorter.

If the `w+` works, perhaps we can simplify the code to remove the `catch` block? 😋

t1m0thyj:
Awesome work! Sorry for taking so long to review this PR 😢 Tested this PR by installing the K8s extension and it works 🎉
I do have a few questions:

In https://github.com/zowe/vscode-extension-for-zowe/issues/2212#issuecomment-1487286153 the following implementation was suggested:
> 1. It could read the contents of the Imperative settings file, and check the mapping mentioned above to see if the associated VS Code extension is installed.
>    * Setting is defined in `~/.zowe/settings/imperative.json` or in team config files:
      `"CredentialManager": "Secrets for Kubernetes"`
>
> 2. If the credential manager is valid and an associated extension is installed, then ZE will load and use the custom credential manager extension instead of Keytar.
>
> 3. If the credential manager is valid and an associated extension is not installed, then ZE will prompt the user asking if they wish to install the extension from the VS Code Marketplace.
>    * To verify that extension exists on marketplace, we could use: https://open-vsx.org/swagger-ui/index.html#/registry-api/getExtension
>
> 4. If the credential manager is invalid, then ZE will prompt the user asking if they wish to reset it to the default value (`@zowe/cli`) which uses Keytar.

This PR implements (1) and (2), but not (3) and (4). Do we still want to implement (3) and (4), and if so would we plan to do it in this PR or a separate one?

**Additional Comments:**
* After I installed the K8s plugin in Zowe CLI, I expected that ZE would prompt me to install the K8s extension at when VS Code launches. But it didn't prompt me, and it also reset my CredentialManager setting in `~/.zowe/settings/imperative.json` back to `@zowe/cli`. This behavior is undesirable IMO - as a Zowe user who wants to adopt the K8s credential manager, I would expect the "Secrets for Kubernetes" setting in my imperative.json file to be preserved. Hence my question above 🙂 
* Initially I had some trouble getting the K8s extension to work on Windows with a K8s cluster hosted in WSL, but was able to fix this by configuring VS Code to allow connections to `\\wsl$`. Created separate issues for this - see https://github.com/zowe/zowe-cli-secrets-for-kubernetes/issues/15 and https://github.com/zowe/zowe-cli-secrets-for-kubernetes/issues/16

rudyflores:
@t1m0thyj Regarding your comment, (1) and (2) should be implemented. As for (3) and (4), that was not really done yet since this can perhaps be addressed in a follow up PR? I think this PR is already a good starting point which we can perhaps refine to have those last two requirements implemented in order for us to provide the best experience when using custom credential managers :)

traeok (commented):
This looks awesome so far! I was able to store credentials in a secret on a Google Cloud k8s cluster. However, I ran into a couple issues when testing:

> After I installed the K8s plugin in Zowe CLI, I expected that ZE would prompt me to install the K8s extension when VS Code is launched. But it didn't prompt me, and it also reset my CredentialManager setting in ~/.zowe/settings/imperative.json back to @zowe/cli.

Like @t1m0thyj, I had a similar issue to the one described above, except I already had the plugin and VS extension installed. After reloading my VS Code window, the credential manager reset to `@zowe/cli`. After reinstalling both the CLI plugin and VS Code extension, the issue was resolved. However, I will try to update this review ASAP with steps to reproduce.

In addition, I also noticed that the credentials are stored under a single secret on Google Cloud, and I presume this will also be the case in a local environment. Since the credentials are stored in a way where the Zowe config path is the "key" for the object, @zFernand0 and I quickly discussed a scenario where he could potentially replace my credentials in the secret object:

1. Create a path that matches the file path my credentials are stored under
2. Set the `ZOWE_CLI_HOME` environment variable to fall under this path (e.g. if my config is `/a/b/c/zowe.config.json`, `ZOWE_CLI_HOME` would be `/a/b/c`).
3. Right click on a session in Zowe Explorer -> "Update Credentials"
4. After entering in the username and password, my stored credentials will be overwritten in the secret object w/ the new credentials provided.

Could we potentially avoid this by generating a unique hash on the user's end, and then chaining that with the file path to make a unique key? Or, could we generate a unique hash and simply use that as the key?

With either proposal, we would have to store the hash/identifier somewhere (such as in the Zowe config itself), but this would prevent someone from intentionally overwriting other user's credentials in the same cluster.

t1m0thyj (commented):
Will approve once merge conflicts are resolved 🙂

I have a question about this sequence: 
> * User installs `Kubernetes secrets credential manager extension`
> * `imperative.json` is set to `Secrets for Kubernetes`

Is it correct that in between those steps, refresh is required for changes to happen? If so that's fine - I think in the future we could investigate handling the `extensions.onDidChange` event in VS Code to prompt the user to refresh ZE or reload the window.

Regarding the issues mentioned earlier:
* The credential manager value in `~/.zowe/settings/imperative.json` changes back to "Secrets for Kubernetes" - this still happens for me but fixes itself after reloading the window
* The possibility for the same K8s secret to be shared by multiple users - this is related to the K8s extension not ZE so created a separate issue: https://github.com/zowe/zowe-cli-secrets-for-kubernetes/issues/18

traeok:
I opened VS Code this morning and noticed that since my k8s setup wasn't active, the CredentialManager failed to load the namespace and as a result, it failed to initialize the secure credential manager:

image

I expected the error to occur, but with 3 errors in the message it might not be obvious to end-users as to why the credential manager failed to load.

rudyflores:
> Will approve once merge conflicts are resolved 🙂
> 
> I have a question about this sequence:
> 
> > * User installs `Kubernetes secrets credential manager extension`
> > * `imperative.json` is set to `Secrets for Kubernetes`
> 
> Is it correct that in between those steps, refresh is required for changes to happen? If so that's fine - I think in the future we could investigate handling the `extensions.onDidChange` event in VS Code to prompt the user to refresh ZE or reload the window.
> 
> Regarding the issues mentioned earlier:
> 
> * The credential manager value in `~/.zowe/settings/imperative.json` changes back to "Secrets for Kubernetes" - this still happens for me but fixes itself after reloading the window
> * The possibility for the same K8s secret to be shared by multiple users - this is related to the K8s extension not ZE so created a separate issue: [Document that each user should have separate namespace zowe-cli-secrets-for-kubernetes#18](https://github.com/zowe/zowe-cli-secrets-for-kubernetes/issues/18)

Yes that is correct, currently the user needs to restart in order to see the changes reflected. I think in the future it would be nice to find a way to prompt for a reload if a credential manager is installed/uninstalled (currently only disabling the custom credential manager prompts for reload because VS Code prompts to do so).


Regarding the bullet points:

1. regarding that issue I think that's where we'll need to prompt for reload upon installing/uninstalling and also integrate the work discussed before regarding checking if the extension is in the VS Code marketplace and follow the CLI's configuration.
2. In the Kubernetes world a user will store a credential per namespace and this can also be documented, just not sure where this should be documented since this is related to the credential manager and not ZE?

t1m0thyj (approved):
LGTM, thanks @rudyflores!

zFernand0 (approved):
Thanks for addressing all my novice questions 😋 
LGTM! 😋

PR #2115, opened 2023-02-07
Make release workflow simpler and more automated

t1m0thyj:
## Proposed changes

This PR updates the Zowe Explorer deployment workflow to use [zowe-actions/octorelease](https://github.com/zowe-actions/octorelease).

Benefits of switching to Octorelease:
* It can automate commits that bump version and update changelog headers for us. For example:
  * Manual release PR: https://github.com/zowe/vscode-extension-for-zowe/pull/2089  vs  Automated release commit: https://github.com/t1m0thyj/vscode-extension-for-zowe/commit/363fc056a4268a7de0d5834f304b8e459f7dad57
  * Manual snapshot PR: https://github.com/zowe/vscode-extension-for-zowe/pull/2090  vs  Automated snapshot commit: https://github.com/t1m0thyj/vscode-extension-for-zowe/commit/64d6ef476b8eb2852b4fae74290e2821fb755acf
* The release process is implemented so that if any step fails, after fixing the issue you can just rerun the deployment workflow and any steps that succeeded previously will be skipped.
* The CLI squad already uses it to manage releases for all their repos, so less maintenance should be required if we share the same codebase 🙂 

TODO:
* [x] Merge https://github.com/zowe-actions/octorelease/pull/81
* [x] Replace the branch name `support-zowe-explorer` with `master` in workflow
* [ ] Update release process documentation in [the wiki](https://github.com/zowe/vscode-extension-for-zowe/wiki/Preparing-for-a-release)

## Release Notes

Milestone: 2.6.1

Changelog: N/A

## Types of changes

What types of changes does your code introduce to Zowe Explorer?
_Put an `x` in the boxes that apply_

- [ ] Bugfix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [x] Updates to Documentation or Tests (if none of the other choices apply)

## Checklist

_Put an `x` in the boxes that apply. You can also fill these out after creating the PR. If you're unsure about any of them, don't hesitate to ask. We're here to help! This checklist will be used as reference for both the contributor and the reviewer_

- [ ] I have read the [CONTRIBUTOR GUIDANCE](https://github.com/zowe/vscode-extension-for-zowe/wiki/Best-Practices:-Contributor-Guidance) wiki
- [ ] PR title follows [Conventional Commits Guidelines](https://www.conventionalcommits.org/en/v1.0.0-beta.2/)
- [ ] PR Description is included
[...]

PR #2231, opened 2023-04-11
[v1] Update release workflow with changes from main branch

t1m0thyj:
Backports changes to our release workflow from the main branch in #2115, #2136, #2137, #2138

This PR should be merged before #2227 so that v1.22.3 can be released using the updated process 🙂 

When reviewing, please check for any logic that was copied over by mistake and should not apply to the v1-lts branch.

### DIFF
--- a/.github/workflows/deployment.yml
+++ b/.github/workflows/deployment.yml
@@ -32,7 +32,7 @@ jobs:
       - name: Use Node.js LTS
         uses: actions/setup-node@v3
         with:
-          node-version: 'lts/*'
+          node-version: '14.x'
           cache: 'npm'
 
       - name: Install Yarn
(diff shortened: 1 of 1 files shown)


zFernand0 (approved):
LGTM! 😋

.github/release.config.js
```diff
@@ -0,0 +1,77 @@
...
+                // Use Lerna only for versioning and publish packages independently
+                npmPublish: false,
+            },
+        ],
+        [
+            "@octorelease/npm",
+            {
+                $cwd: "packages/zowe-explorer-api",
+                aliasTags: {
+                    "latest": ["zowe-v2-lts"],
```
traeok, line 47: Should this stay as `zowe-v2-lts` or should `latest` point to the latest v1 branch? Not familiar with how `aliasTags` is used here

traeok (approved):
LGTM, thanks @t1m0thyj

.github/release.config.js
t1m0thyj: Good catch, I think it's fine here because it defines the "zowe-v2-lts" tag as an alias that should be published anytime the "latest" tag is published. On lines 14-15 of the release config, we define "zowe-v1-lts" as the tag for the "v1-lts" Git branch. So defining the "latest" <-> "zowe-v2-lts" alias should be fine because it doesn't affect publishes of the "zowe-v1-lts" tag.

Issue #2166, opened 2023-03-01
Incorrect type in user settings for zowe.settings.version


phaumer:
**Describe the bug**

Zowe Explorer writes an invalid version number in 2.6.1 into the user settings file. The value should remain an integer 2 as the setting had not changed/do not require additional migration since 1.x.

<img width="485" alt="2023-03-01_13-30-41" src="https://user-images.githubusercontent.com/4391934/222270092-1e1de82c-816c-483a-9771-e124f485ca57.png">

**To Reproduce**

**Expected behavior**

Keep the 2 integer indicating that a migration from v1 to v2 had happened.

**Screenshots**

**Desktop (please complete the following information):**

- OS:
- Zowe Explorer Version:
- (Optional) Zowe CLI Version:
- (Optional) Are you using Secure Credential Store?

**Additional context**

phaumer:
@SG-A47 filed your issue here.

phaumer:
@zFernand0 @rudyflores I would propose increasing the priority as people will get this message every time there is a new release now and there no reason to use any other value than 2, unless we introduce more breaking changes.

<img width="1178" alt="2023-03-22_18-01-53" src="https://user-images.githubusercontent.com/4391934/227072247-f74767aa-c421-45e1-af1a-c1a50e1a3240.png">

PR #2281, opened 2023-05-16
[vNext] Migrate persistent settings to local storage

rudyflores:
## Proposed changes

Migration of Persistent VS Code user settings for Zowe Explorer, these include:

- `zowe.ds.history`
- `zowe.uss.history`
- `zowe.jobs.history`
- `zowe.commands.history`
- `zowe.cliLoggerSetting.presented`
- `zowe.settings.oldSettingsMigrated`
- `zowe.settings.localStorageMigrated`

These become stored in the following path which is secured to always exist according to VS Code's API and its contents are placed in sqlite database file named `state.vscdb` in the `Code/User/globalStorage` folder:

- [ExtensionContext.globalStorageUri](https://code.visualstudio.com/api/extension-capabilities/common-capabilities#data-storage)

**Note**
In order to deliver this enhancements in our next release and not v3, a migration script is in place as well. Refactoring was made to `SettingsConfig.ts` in order to migrate old persistent settings in VS Code settings to local storage and maintain the contents.

### Potential additions with this new code

with the introduction of `ZoweLocalStorage` API, in a future PR we could integrate this into the Zowe Explorer API and allow storage of other things in this singleton since it can be accesed anywhere in the code after initialization on activation. Examples of things we could store could be things such as team config profiles, contents of `.zowe` folder like `imperative.json`

### Testing this PR

In order to test this PR, run the debug extension host and notice how Zowe Explorer will activate, but now it should be reading and writing new history items in DS/USS/Jobs into local storage. Functionality can be tested by removing all persistent settings from VS Code settings in global scope **after** successfully activating Zowe Explorer the first time.

## Release Notes

Milestone: TBD

Changelog: Migrated VS Code persistent settings to local storage

## Types of changes

What types of changes does your code introduce to Zowe Explorer?
_Put an `x` in the boxes that apply_

- [ ] Bugfix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [x] Updates to Documentation or Tests (if none of the other choices apply)

## Checklist

_Put an `x` in the boxes that apply. You can also fill these out after creating the PR. If you're unsure about any of them, don't hesitate to ask. We're here to help! This checklist will be used as reference for both the contributor and the reviewer_

[...]

PR #2286, opened 2023-05-17
fix(settings): Coerce `zowe.settings.version` to an integer

traeok:
## Proposed changes

This PR fixes the `zowe.settings.version` value such that it will:

- use the correct value (2)
- fix any incorrect values for `zowe.settings.version`

Note that settings will only be migrated if there are old configuration options detected - this PR does not change this behavior.

## Release Notes

Milestone: 2.8.1

Changelog:

- Fixed issue where Zowe Explorer set incorrect values for `zowe.settings.version`. #2166 

## Types of changes

## Checklist


packages/zowe-explorer/src/utils/SettingsConfig.ts
```diff
@@ -67,15 +67,51 @@ export class SettingsConfig {
...
         const workspaceIsOpen = vscode.workspace.workspaceFolders !== undefined;
         const zoweSettingsExist = SettingsConfig.zoweOldConfigurations.length > 0;
 
+        // Update version value if there is a mismatch (coerces previous values into numeric ["2.6.1" -> 2])
+        if (globalVersion == null || globalVersion !== currentVersion) {
+            SettingsConfig.setDirectValue(globals.SETTINGS_VERSION, SettingsConfig.currentVersionNumber);
+        }
+        if (workspaceVersion != null && workspaceVersion !== currentVersion) {
+            SettingsConfig.setDirectValue(globals.SETTINGS_VERSION, SettingsConfig.currentVersionNumber, vscode.ConfigurationTarget.Workspace);
+        }
```
zFernand0, line 112: I'm not sure I agree with forcing the `zowe.settings.version` to match the ZE major version.
I can understand why, but it's possible that we may need to refactor a few things within the same major release.
Not saying we would do that this (or the next) release.

On the contrary, what if ZE settings are stable enough where the settings stay the same even after a new major release or ZE?

I guess I'm comparing this to the CLI zowe.schema.json where that is still at version `"1.0"`.

Feel free to ignore this if it sounds crazy. 😋

zFernand0 (approved):
LGTM! 😋 
left a comment, but nothing to worry about for now

JillieBeanSim (dismissed):
LGTM! thanks @traeok for the fix

zFernand0 (dismissed):

zFernand0:
@phaumer @JillieBeanSim, 
Do you see any reason why we should not archive the VSIX built with node 18 instead of the node16 one?

zFernand0 (dismissed):
LGTM! 😋 
Thanks for archiving the node18 built VSIX

JillieBeanSim:
@traeok @zFernand0 why remove testing with node 14?

Archiving 18 is fine, I believe it was set to 16 when we broke against 18

traeok:
> why remove testing with node 14?

@JillieBeanSim Node 14 is EOL, so I wasn't sure if we wanted to keep it in the matrix. If we still intend on supporting it I will definitely add it back to the workflow.

JillieBeanSim:
> @JillieBeanSim Node 14 is EOL, so I wasn't sure if we wanted to keep it in the matrix. If we still intend on supporting it I will definitely add it back to the workflow.

It is in maintenance mode and with our v2 we document having at least 14 for development so would be good to test against it.
Can remove from next branch though and change our supported versions for v3 release.

JillieBeanSim:
@traeok looking at that chart v3 should have 18 as lowest supported

packages/zowe-explorer/src/utils/SettingsConfig.ts
phaumer: I agree with @zFernand0 that if the next release 3.x does not require any migration that the value should stay 2 until there is a change, because changing this value is already a migration that can have a major impact (e.g. if the user has dozens of Git repos with workspace settings that will change requiring a commit.)

Supporting migration if there is not major version number update probably needs more discussion with use cases. In general breaking changes that require migration (even if semi-automatic) should be a major version update with the same argument as above as people keep their workspace settings in Git and require updates.

phaumer:
The latest VS Code still only ships with Node 16.17.1. We should build and package with that version.

traeok:
> The latest VS Code still only ships with Node 16.17.1. We should build and package with that version.

Thanks for mentioning that! It would be best to build/package with the same Node version as VS Code so we don't run into any Node-specific issues. Addressed in 7745c13

packages/zowe-explorer/src/utils/SettingsConfig.ts
traeok: @phaumer This logic won't exist in `next` (v3), as once we merge the PR for local storage #2281 we won't need this setting anymore. For now, this fix should help coerce any version strings that don't match to be updated to the integer `2`, and once `main` is updated with changes from `next` I will remove it 👍

JillieBeanSim (approved):

rudyflores (approved):

zFernand0 (commented):
I forgot to re-approve 😋 
LGTM though

Issue #2148, opened 2023-02-22
Add `madge` to CI to track circular dependencies


traeok:
**Is your feature request related to a problem? Please describe.**

Currently, we are not tracking dependencies between modules in our codebase. Adding a tool to graph and track these dependencies will help point out circular dependencies, as well as areas where module imports can be adjusted.

**Describe the solution you'd like**

Add the tool [`madge`](https://github.com/pahen/madge) to the CI pipelines. This will generate a visual graph of the module dependencies as well as other useful info.

PR #2320, opened 2023-06-01
feat(ci): Add `madge` to CI pipelines to track circular dependencies

traeok:
## Proposed changes

This PR adds `madge` to the scripts for all packages. This will help track any circular dependencies in the codebase. 
Implements #2148 

## Release Notes

Milestone: vNext

Changelog: 

- Add `madge` as a script in `package.json` to track circular dependencies in the codebase. #2148 

## Types of changes

## Checklist

### Notable changes to codebase/potential breaking changes

**API:**
- Tree node types `IZoweDatasetTreeNode, IZoweJobTreeNode, IZoweUSSTreeNode` moved into separate files in `src/tree` in the API. These types are still exported from `src/tree/index.ts`, so imports should be unaffected as long as users were properly importing these node types from `@zowe-explorer-api` and not the specific module file
- Moved `ICommon, IJes, IMvs, IUss` interfaces into `src/extend/interfaces.ts`. Users should now import these interfaces directly from `@zowe-explorer-api` instead of accessing them from `ZoweExplorerApi`:
  - `import { ZoweExplorerApi } from "@zowe-explorer-api" -> import { ICommon } from "@zowe-explorer-api"`
- `apiRegister` parameter for `ProfilesCache.refresh` has been changed to be optional. `apiRegister` was also changed to parent type `IRegisterClient` instead of `IApiRegisterClient`.
- `ZoweExplorerApi` module was dissolved and spread out in the new `src/extend` modules. This makes the module's use case more obvious and separates the extender functionality from other API definitions.
- `IApiRegisterClient` was renamed to generic type `IRegisterClient`. `IApiRegisterClient` is redefined as a type union of `IRegisterClient` in `src/extend/index.ts` to contain the `getExtenderInstance` functionality
  - This avoids a circular dependency between `ZoweExplorerApiRegister` and `ZoweExplorerExtender`.
  - This is worth noting because although this isn't a direct breaking change, users can now call `Profiles.refresh` without having to provide the API register - it will fallback to use `ZoweExplorerApiRegister.getInstance()` if `apiRegister` is `null/undefined`
  - Users might have to update imports of `IApiRegisterClient` to point to the correct module (`src/extend/index.ts`)

**Zowe Explorer**
- Not much of a breaking change, but `globals.LOG` has been merged into `ZoweLogger`. Uses of `globals.LOG` should be changed to use the `ZoweLogger` class as it also will print any output to our output channel in VS Code.
- To be continued...


traeok:
I think the Codecov patch coverage increased from the 87.4% above, but for some reason it is not triggering the Codecov step. You can confirm this by comparing local coverage results (covertura) with Codecov's - Codecov still shows 50% for two of the FTP API files, but the "uncovered lines" show as covered in covertura's results.

JillieBeanSim (commented):
Hey @traeok thanks for the addition of madge.  I was wondering if we should include the call in the build script too so devs don't have to run manually.  

I do still see some circular dependencies.  Do we have a follow up issue to correct them?  Also, once we have them all corrected can we have the build script fail when a new circular dependency is introduced?
Screen Shot 2023-06-05 at 3 48 46 PM

traeok:
> I do still see some circular dependencies. Do we have a follow up issue to correct them? Also, once we have them all corrected can we have the build script fail when a new circular dependency is introduced?

@JillieBeanSim I've reduced our circular dependencies from 10 to those 4 you've mentioned:

1. `globals.ts > utils/LoggerUtils.ts > utils/SettingsConfig.ts`
  - I'm not sure if there's much we can do about this one since `ZoweLogger` references values from `SettingsConfig`, and `SettingsConfig` will call `ZoweLogger.trace` for all of its functions. Even by moving these `SettingsConfig` functions into `ZoweLogger`, there will be a circular dep for the two files as long as we use the logger within `SettingsConfig`.
2. `utils/LoggerUtils.ts > utils/SettingsConfig.ts > utils/ZoweLocalStorage.ts`
  - The same circular dep from 1 applies here, because `ZoweLocalStorage` also imports the logger for tracing from `LoggerUtils`
3. `Profiles.ts > ZoweExplorerApiRegister.ts > ZoweExplorerExtender.ts`
  - This one is tough because I didn't want to reorganize both classes to the point where it would break all extenders in v3. `Profiles.ts` needs the API register to get access to the relevant APIs. Since the `ZoweExplorerApiRegister` class has a helper method to get the extender API, it imports `ZoweExplorerExtender`. In `ZoweExplorerExtender` we reference the `Profiles` class several times, causing this circular dep.
4. `Profiles.ts > ZoweExplorerApiRegister.ts > ZoweExplorerExtender.ts > utils/ProfilesUtils.ts`
  - The same applies here, except that `ProfilesUtils.ts` needs functions from `Profiles.ts`, so it spots the circular dependency again in this instance.

As for 1 and 2, I can propose moving the `ZoweLogger` instance into `globals.ts` to try to break the loop. For 3 and 4, I was hesitant to restructure these two classes, but would definitely be willing to do so if we would like to eliminate all circular deps.

We can discuss this in standup - for the time being I will allow the API and FTP extension `madge` scripts to fail if there are any circular deps. introduced.

traeok:
#2329 has been added to the repo in hopes to address the circular dependency between the API register and extender singleton.

awharn (dismissed):

zFernand0 (approved):

PR #2328, opened 2023-06-06
fix(profiles): avoid assigning `tokenType/tokenValue` to "direct-to-service" profiles

traeok:
## Proposed changes

`ProfileInfo.mergeArgsForProfile` is being called in `ZoweVsCodeExtension.updateCredentials` - as expected it returns a list of known arguments, and among those are the `tokenType, tokenValue` fields from the base profile. But, we assign `tokenType` and `tokenValue` even if the profile that was passed to `updateCredentials` did not contain a token. In addition, a user wouldn't be updating credentials if the profile used a token, so we should assume that the token should not be assigned in this context.

This PR changes `updateCredentials` so that we can provide an existing profile object to update. This avoids calling `mergeArgsForProfile` as we already have the profile built in this context. It also prevents `tokenType` and `tokenValue` from being assigned to direct-to-service profiles.

After discussing these proposed changes w/ @awharn , we determined that this fix is sufficient and can be introduced without any breaking changes.

## Release Notes

Milestone: 2.9.0

Changelog: 

- Fixed issue where "direct-to-service" configurations would get merged with token data from a base profile.

## Types of changes

## Checklist


awharn (dismissed):

zFernand0 (approved):
LGTM! 😋 
I don't know if we really need this to be published as a hotfix right away.
It might be good to still rebase it against `maintenance` though 😋

Issue #2262, opened 2023-04-28
Prevent users being locked out when session has invalid credentials


t1m0thyj:
**Steps to Reproduce**
1. Add a profile to Zowe Explorer that has an invalid or expired password.
2. Try to perform an action like searching for data sets with this profile.

**Actual Behavior**
An error message is shown, but I can easily get locked out of my MF account if I try to perform more actions with the same profile. If I click the "Cancel" button, the error message may reappear several times indicating that ZE kept using the invalid credentials for subsequent API calls.
![image](https://user-images.githubusercontent.com/22344007/235136725-3d28f47f-1d6f-458c-b74e-146370e05bb6.png)

**Expected Behavior**
1. The "Check Credentials" button is renamed to "Update Credentials" to make it clearer and more consistent with the rest of the UI.
2. For actions that perform multiple API calls (e.g., list archived data sets), hitting an Invalid Credentials error should fail fast and abort the remaining API calls.

Split this into enhancement request #2374 
3. For sessions with invalid credentials, we could show a 🔒 overlay on the tree icon. This "locked" state could be stored persistently in VS Code, so it is remembered after restarting. When attempting to perform an action with a locked session, a prompt could be shown with 2 options: "Update Credentials" to update them in ZE or "Unlock" if the credentials have been updated elsewhere.

phaumer:
I see this problem as well. It does not allow me to use any other profile either. Every time I click another profile it keeps prompting me about that expired one again. Workaround is to click Refresh.

JillieBeanSim:
@t1m0thyj I am going to pull #3 out of this into its own enhancement and have this follow bug fix.  Will mention this in enhancement request for locking

phaumer:
This is still not working for me. I am using 2.9.2 in Eclipse Che in which I do not want to store my password in the team config file, but use tokens. 
- I have two profiles and log on to both with tokens and run queries on both
- Then I log out of one 
- It immediately prompts me again for credentials
- When I click cancel it prompts me again when I use the other profile in any way. Then it become an endless loop that I cannot get out of. 
- When I give in an enter my credentials they are store in the the config file. If I used only tokens before that should somehow not happen and I should get warned about it. Perhaps, we need a setting in the profile that only allows storing tokens, but not credentials?

PR #2382, opened 2023-07-17
Fix for check creds bug

JillieBeanSim:
## Proposed changes

fix for #2262 check credentials loop when attempting to cancel the Check Creds (now `Update Credentials`) popup

*Note* Duplication is on test files

## Release Notes

Milestone: 2.9.2

Changelog: Fixed loop when user selects Cancel on the Check Credentials message.

## Types of changes

## Checklist


JillieBeanSim:
How did patch coverage drop from 100% in commit https://github.com/zowe/vscode-extension-for-zowe/pull/2382/commits/f1e216e451c2549d7b11abef17a383dbac5148e8 to 74.22% in following commit that only added CHANGELOG and updated localization?
Screen Shot 2023-07-17 at 1 29 35 PM

traeok (dismissed):
I'm running into a 401 error when trying to access my test LPAR w/ invalid credentials:

image

The "Update Credentials" prompt does not appear. I also noticed that there was an error with the `zowe.ds.pattern` command, so I've included that message in the screenshot as well.

traeok (commented):
I noticed that while the "Check credentials" loop is gone, it still tries to process a request and I get a 401 before the search prompt is dismissed/entered:

image

This happens in all trees. However, the "Update credentials" dialog does appear, and selecting "Cancel" works as intended and the dialog does not re-appear like it did before:

image

traeok (commented):
I'm running into an issue where the Data Sets and Job trees will prompt for credentials twice if the node is expanded with invalid credentials. It only seems to happen if I do the following:

1. Right-click -> "Update Credentials" on a session and provide fake/invalid credentials. 
2. Expand the session with a filter already set (either on Data Sets or Jobs)
3. Notice the prompt will appear twice

With the USS tree, this does not occur. It might have to do with the timing of when the session node is expanded and when the request is executed.

FixCheckCredsBug

JillieBeanSim:
> I'm running into an issue where the Data Sets and Job trees will prompt for credentials twice if the node is expanded with invalid credentials. It only seems to happen if I do the following:
> 
> 1. Right-click -> "Update Credentials" on a session and provide fake/invalid credentials.
> 2. Expand the session with a filter already set (either on Data Sets or Jobs)
> 3. Notice the prompt will appear twice
> 
> With the USS tree, this does not occur. It might have to do with the timing of when the session node is expanded and when the request is executed.
> 
> FixCheckCredsBug

@traeok from what I see in the video The first is from clicking the session node causing an expand of tree then the second is from the search filter.  The PR is to address the loop caused by search filter.  This is why I moved the expanding of the tree to end of methods since last comment.

traeok:
> from what I see in the video The double is from clicking the session node causing an expand of tree not from the search filter. The PR is to address the loop caused by search filter. This is why I moved the expanding of the tree to end of methods since last comment.

Ah okay, maybe this case should be addressed as a separate issue then? In terms of the search filter causing the loop, that appears to be resolved. If so I can approve so we can get this into 2.9.2 👍

JillieBeanSim:
> > from what I see in the video The double is from clicking the session node causing an expand of tree not from the search filter. The PR is to address the loop caused by search filter. This is why I moved the expanding of the tree to end of methods since last comment.
> 
> @JillieBeanSim Ah okay, maybe this case should be addressed as a separate issue then? In terms of the search filter causing the loop, that appears to be resolved. If so I can approve so we can get this into 2.9.2 👍

let me check, could also be from a previous search still attached to node.  If search fails we shouldn't attach the search criteria to the node, and maybe we are 😢

JillieBeanSim:
@traeok it's the search being added when failing let me fix other trees too

https://github.com/zowe/vscode-extension-for-zowe/assets/49491949/0dc84c5a-e0ce-4123-aa9e-aef12fd30d62

JillieBeanSim:
The other trees don't act like the data sets tree acted in video. pushed fix for that tree.  Thanks @traeok, the videos definitely helped.

traeok (commented):
The data set tree is fixed, thanks @JillieBeanSim !

I did find that the jobs tree does it sometimes, but it seems like it is related to the same scenario as the data sets (applying filter on expanded session node even if credentials are invalid), so it might be a similar fix.

FixCheckCredsBugJobs

JillieBeanSim:
@traeok I couldn't do the same as ds tree as they aren't handled the same but see the second call is coming from the syncsession call.  I don't believe that is required with watcher and the prompt method stores updates in allProfiles array.

JillieBeanSim:
This is what I am seeing in Jobs tree, on initial search with already incorrect credentials I only get prompted once with cancel.  If I update credentials do a successful search then change creds to be incorrect then I will see 2 pop ups.  This is still better than the loop we are seeing, but cannot locate the second call being initiated.

https://github.com/zowe/vscode-extension-for-zowe/assets/49491949/10455acf-464c-4247-a478-4916401ab913

JillieBeanSim:
@zFernand0 & @traeok I tested back to 2.8.x and see double prompt in jobs tree with tested scenarios, so was existing before the bad loop was introduced.  We should get this fix out and look at fixing old issues when we address the work we wanted to do with the trees and getting them handled the same way across all 3.  What do you think?

zFernand0 (approved):
Happy to get this PR merged in today (if we want to) 😋

traeok (approved):
Sorry for the delay; was out of office.

Sounds good, I think that its best to merge this fix and handle the jobs tree bug as a separate issue - LGTM, thanks @JillieBeanSim 😃

PR #2319, opened 2023-05-31
[vNext] Change activation event to onStartupFinished

JillieBeanSim:
## Proposed changes

Change activation event to onStartupFinished

## Release Notes

Milestone: vNext

Changelog: ZE & zFTP ext: Updated activation event to `onStartupFinished`

## Types of changes

What types of changes does your code introduce to Zowe Explorer?
_Put an `x` in the boxes that apply_

- [ ] Bugfix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [x] Updates to Documentation or Tests (if none of the other choices apply)

## Checklist

_Put an `x` in the boxes that apply. You can also fill these out after creating the PR. If you're unsure about any of them, don't hesitate to ask. We're here to help! This checklist will be used as reference for both the contributor and the reviewer_

- [x] I have read the [CONTRIBUTOR GUIDANCE](https://github.com/zowe/vscode-extension-for-zowe/wiki/Best-Practices:-Contributor-Guidance) wiki
- [x] PR title follows [Conventional Commits Guidelines](https://www.conventionalcommits.org/en/v1.0.0-beta.2/)
- [x] PR Description is included
- [ ] gif or screenshot is included if visual changes are made
- [x] `yarn workspace vscode-extension-for-zowe vscode:prepublish` has been executed
- [ ] All checks have passed (DCO, Jenkins and Code Coverage)
- [ ] I have added unit test and it is passing
- [ ] I have added integration test and it is passing
- [ ] There is coverage for the code that I have added
- [x] I have tested it manually and there are no regressions found
- [ ] I have added necessary documentation (if appropriate)
- [ ] Any PR dependencies have been merged and published (if appropriate)

Issue #1721, opened 2022-03-31
Remove `yarn build` from the top level `prepare` script


zFernand0:
In order to minimize the time spent on some build building and rebuilding over and over, we could remove the `yarn build` from the prepare script.
This may require some pipeline updates.

PR #2424, opened 2023-08-18
Migrate from yarn to pnpm

rudyflores:
## Proposed changes

Migrate to new package manager PNPM for improved install and build times for development vs Yarn

## Release Notes

Milestone: v3.0

Changelog: Changed Yarn package manager to PNPM package manager.

## Testing this PR

For testing this PR perform the following:

- run `pnpm i`
- run `pnpm build`
- run `pnpm package`
- run `pnpm test`
- run `pnpm fresh-clone` (this one uses the new concurrency feature!)
- try running any npm or yarn command (this should throw an error since only pnpm commands are allowed)

Make sure to test normal ZE functionality in the dev host as well as packaged extension to make sure everything is in order as well :) 

## Metrics

`fresh-clone` command:

**Yarn**
Screen Shot 2023-08-10 at 3 08 36 PM

**PNPM**
Screen Shot 2023-08-10 at 3 08 56 PM

`install` command:

**Yarn**
Screen Shot 2023-08-10 at 3 11 10 PM

**PNPM**
Screen Shot 2023-08-10 at 3 10 17 PM

`package` command:

**Yarn**
Screen Shot 2023-08-10 at 3 13 23 PM

**PNPM**
Screen Shot 2023-08-10 at 3 14 44 PM

## Types of changes

## Checklist


package.json
```diff
@@ -8,14 +8,11 @@
...
   "private": true,
-  "workspaces": [
-    "packages/*"
-  ],
   "engines": {
     "vscode": "^1.53.2"
   },
   "dependencies": {
-    "@zowe/cli": "7.16.6",
+    "@zowe/cli": "7.16.2",
```
awharn, line 15: Why is the dependency version of Zowe CLI decreasing here?

.npmrc
```diff
@@ -0,0 +1,4 @@
+registry=https://registry.npmjs.org/
+@zowe:registry=https://registry.npmjs.org/
```
awharn, line 2: On some of the other projects, the .npmrc is configured to pull Zowe CLI from Zowe's Artifactory server to prevent metric inflation on NPM. Is that something that we should consider doing here?

awharn (commented):
I have a couple of questions

package.json
rudyflores: Might've been from the cherry-picked old commits, I can go ahead and bump up the version, thanks for catching that.

.npmrc
rudyflores: I believe @zFernand0 suggested placing the `.npmrc` this way, Fernando did you have a reason to pull from this registry instead of the way @awharn  is suggesting?

zFernand0: Either one would work, as long as the Integrity checksums match what we get back from either registry.
My problem was mostly due to mismatched checksums.
I kind of agree with Andrew's point about metric inflation on NPM though. 😋

.github/workflows/changelog.yml
```diff
@@ -22,4 +22,3 @@ jobs:
         uses: awharn/check_changelog_action@v1
```
t1m0thyj, line 22: In order for [this action](https://github.com/awharn/check_changelog_action/blob/master/action.yml) to work correctly, I think it needs configuration to know this is a monorepo.

Instead of `yarnWorkspaces: true` we could set `lerna: true` since this repository has a lerna.json file 😋

.github/workflows/lint.yml
```diff
@@ -18,8 +18,8 @@ jobs:
           node-version: lts/*
 
       - run: |
-          npm install -g yarn
-          yarn install --frozen-lockfile
+          npm install -g pnpm
+          pnpm install --frozen-lockfile --shamefully-hoist=true
```
t1m0thyj, line 22: We may be able to remove both options - see comment above

This also applies to all the other workflows that call `pnpm install` 😋

package.json
```diff
@@ -8,9 +8,6 @@
     "url": "https://github.com/zowe/vscode-extension-for-zowe"
   },
   "private": true,
-  "workspaces": [
-    "packages/*"
-  ],
```
t1m0thyj, line 13: Our deployment workflow uses Lerna, which depends on `workspaces` being defined in package.json because our lerna.json file has "useWorkspaces" set to true.

I think either we'll need to leave the `workspaces` property here, or we could consider migrating from lerna v6 to v7 which removed the "useWorkspaces" setting (see [lerna@7 release notes](https://github.com/lerna/lerna/blob/main/CHANGELOG.md#700-2023-06-08))

packages/zowe-explorer/package.json
```diff
@@ -1913,26 +1913,26 @@
...
+    "build:integration": "pnpm createTestProfileData && pnpm license && tsc --pretty --project tsconfig-tests.json",
     "test:unit": "jest \".*__tests__.*\\.unit\\.test\\.ts\" --coverage",
     "test:theia": "mocha ./out/__tests__/__theia__/*.theia.test.js --timeout 50000 --reporter mocha-multi-reporters --reporter-options configFile=.mocharc.json",
-    "test": "yarn test:unit",
-    "vscode:prepublish": "gulp build && webpack --mode production && yarn updateStrings && yarn run compile-web",
+    "test": "pnpm test:unit",
+    "vscode:prepublish": "gulp build && webpack --mode production && pnpm updateStrings && pnpm run compile-web",
     "updateStrings": "node ../../scripts/stringUpdateScript.js && prettier --write --loglevel warn ./i18n package.nls*",
-    "package": "vsce package --yarn && node ../../scripts/mv-pack.js vscode-extension-for-zowe vsix",
+    "package": "vsce package --no-dependencies --allow-star-activation && node ../../scripts/mv-pack.js vscode-extension-for-zowe vsix",
```
t1m0thyj, line 1923: The `--allow-star-activation` shouldn't be necessary on the "next" branch because we stopped activating ZE that way in #2319

packages/samples/vue-webview-sample/package.json
```diff
@@ -21,8 +21,8 @@
     ]
   },
   "scripts": {
-    "vscode:prepublish": "yarn build",
-    "build": "npx tsc -p ./ && cd webviews/vue-sample && yarn && yarn build",
```
traeok, line 25: We'll still need `cd webviews/vue-sample && ...` as part of the `build` script or otherwise it will not build the `dist` folder for the webview.

packages/samples/vue-webview-sample/package.json
traeok, line 1: I noticed that the `vue-webview-sample` is not being picked up by the pnpm workspace - perhaps because it is in `packages/samples`? Regardless, we probably don't want to build samples as part of the workspace, so it's fine to leave this behavior as-is.

Note that the only way I was able to build the sample was with `pnpm install .` in the webview sample directory - but this might be how `pnpm` behaves around workspaces, so I'm not too concerned about it.

packages/zowe-explorer/package.json
```diff
@@ -1927,26 +1927,26 @@
     ]
   },
   "scripts": {
-    "build": "gulp build && yarn license && webpack --mode development && cd webviews/edit-attributes && yarn && yarn build",
```
traeok, line 1930: We will still need the `cd webviews/edit-attributes && pnpm install . && pnpm build` as part of the `build` script, or the feature won't be usable once packaged.

traeok (changes requested):
Looks great so far @rudyflores - just noticed a couple things while testing:

- During the `pnpm fresh-clone` command, I get an exit code 1 at the end and a warning about missing `node_modules`:
image
- The webview build steps were removed - we'll still need them in order to use the webviews once packaged or during development.

packages/samples/vue-webview-sample/package.json
t1m0thyj: After 2.11.0 is released and we merge the changes from master branch into next, then the `packages/samples` directory will be moved to a top-level `samples` directory because we did that in #2425. Maybe that will fix the problem with samples being picked up as part of the workspace?

packages/samples/vue-webview-sample/package.json
rudyflores: if you check `pnpm-workspace.yaml` we add the relevant workspaces for this exact command, which should bundle these packages without the need of manually adding these, you can test this by running the debugger and seeing that the webview works as intended.

packages/zowe-explorer/package.json
rudyflores: Check comment above on first comment made :)

packages/samples/vue-webview-sample/package.json
traeok: Ah, I stand corrected. I didn't notice that new `dist` folders were generated for the webviews when I last tested - but after testing again, the webviews are compiled as expected. Sorry for the confusion 😓

packages/zowe-explorer/package.json
traeok: Sorry, that was a mistake on my part - resolved

traeok (dismissed):
LGTM, thanks @rudyflores 😀 

Would it be possible to adjust the `fresh-clone` script so it doesn't error out at the end? I have the tendency to run `pnpm fresh-clone && pnpm install && ...` while switching between branches, and the `fresh-clone` script in its current state will short-circuit that expression 😅

rudyflores:
> LGTM, thanks @rudyflores 😀
> 
> Would it be possible to adjust the `fresh-clone` script so it doesn't error out at the end? I have the tendency to run `pnpm fresh-clone && pnpm install && ...` while switching between branches, and the `fresh-clone` script in its current state will short-circuit that expression 😅

Of course! I also plan to address this issue as well here in this PR, will push both these changes today :) https://github.com/zowe/vscode-extension-for-zowe/issues/1721

@zFernand0 FYI

traeok (dismissed):

t1m0thyj (changes requested):
The lint check seems to be failing with `ERR_PNPM_RECURSIVE_RUN_FIRST_FAIL`

rudyflores:
> The lint check seems to be failing with `ERR_PNPM_RECURSIVE_RUN_FIRST_FAIL`

Strangely enough I don't see this when running locally the command `pnpm lint` on both root and specific package for FTP, not sure if the check that's being run is using a specific version of something that shows an error we don't catch locally 🤔

rudyflores:
@traeok I think the `samples` packages in our repo should be maybe just moved to its own repository? otherwise we should define it in the `pnpm-workspaces.yaml` which will cause it to bundle and install when running `pnpm i && pnpm build`.

This is also why you see the check for the samples fail too

traeok:
@rudyflores We can definitely discuss moving the samples in tomorrow's stand-up - if we add more samples in the future, we are probably better off having a separate repo, so I'm in favor of that 👍

.github/workflows/zowe-explorer-samples.yml
```diff
@@ -38,8 +38,8 @@ jobs:
         with:
           node-version: ${{ matrix.node-version }}
 
-      # install yarn
-      - run: npm install -g yarn
+      # install pnpm
+      - run: npm install -g pnpm
 
-      - run: for dir in samples/*; do pushd $dir && yarn && yarn run compile && popd; done
+      - run: for dir in samples/*; do pushd $dir && pnpm i && pnpm run compile && popd; done
```
t1m0thyj, line 44: I think this workflow should be fixed if we add the `--ignore-workspace` flag:
* `pnpm i` -> `pnpm --ignore-workspace i`

t1m0thyj (dismissed):
Thanks @rudyflores! pnpm commands are running nice and fast for me locally.

JillieBeanSim (dismissed):
this all LGTM! thanks @rudyflores for all of the hard work that went into this change.  If all are on board for pnpm, then so am I 😄 .  With the amount of changes this introduces into the next branch it will be easier for PRs raised against main or maintenance to have another PR with same work possibly cherry-picked against next to avoid the conflicts that will come from the differences between the branches.

traeok (approved):
LGTM, thanks @rudyflores! `pnpm` is definitely quicker, looking forward to using it over `yarn` 🎉

zFernand0 (approved):
What Trae said 👆🏽 
LGTM! 😋

awharn:
I only have one comment about this PR, as the rest of this looks fantastic. Would anyone be against removing `--parallel` from the `test` script, and adding a `test:parallel` script? I only ask because the test data is getting mixed between packages, and personally, it becomes a little bit harder to track what is going wrong, and where.

image

awharn (approved):

zFernand0 (approved):

PR #2437, opened 2023-08-29
Refresh Zowe Explorer after token auth logout

rudyflores:
## Proposed changes

Refresh Zowe Explorer after a logout is performed for token based auth

## Release Notes

Milestone: 2.10.1

Changelog: Fixed issue with endless credential prompt loop when logging out 

## Types of changes

## Checklist

### DIFF
--- a/packages/zowe-explorer/src/utils/ProfilesUtils.ts
+++ b/packages/zowe-explorer/src/utils/ProfilesUtils.ts
@@ -137,20 +137,15 @@ export function isTheia(): boolean {
  * @returns {Promise<boolean>} a boolean representing whether token based auth is being used or not
  */
 export async function isUsingTokenAuth(profileName: string): Promise<boolean> {
-    const config = (await Profiles.getInstance().getProfileInfo()).getTeamConfig();
-    const baseProfile = config.properties.profiles[Profiles.getInstance().getDefaultProfile("base")?.name];
-    const profile = config.properties.profiles[profileName];
+    const baseProfile = Profiles.getInstance().getDefaultProfile("base");
     const isUsingZosmf = (await Profiles.getInstance().getLoadedProfConfig(profileName)).type === "zosmf";
-    if (isUsingZosmf && baseProfile && profile?.secure && baseProfile?.secure) {
-        return profile.secure.includes("tokenValue") || baseProfile.secure.includes("tokenValue");
+    if (isUsingZosmf && baseProfile) {
+        return (
+            (await Profiles.getInstance().getSecurePropsForProfile(profileName)).includes("tokenValue") ||
+            (await Profiles.getInstance().getSecurePropsForProfile(baseProfile?.name)).includes("tokenValue")
+        );
     }
-    if (isUsingZosmf && baseProfile && baseProfile?.secure) {
-        return baseProfile.secure.includes("tokenValue");
-    }
-    if (profile?.secure) {
-        return profile.secure.includes("tokenValue");
-    }
-    return false;
+    return (await Profiles.getInstance().getSecurePropsForProfile(profileName)).includes("tokenValue");
 }
 
 /**
--- a/packages/zowe-explorer/src/utils/ProfilesUtils.ts
+++ b/packages/zowe-explorer/src/utils/ProfilesUtils.ts
@@ -131,18 +131,26 @@ export function isTheia(): boolean {
     return false;
 }
 
+/**
+ * Function that checks whether a profile is using token based authentication
+ * @param profileName the name of the profile to check
+ * @returns {Promise<boolean>} a boolean representing whether token based auth is being used or not
+ */
 export async function isUsingTokenAuth(profileName: string): Promise<boolean> {
     const config = (await Profiles.getInstance().getProfileInfo()).getTeamConfig();
-    const defaultBaseProfileName = Profiles.getInstance().getDefaultProfile("base").name;
-    const baseProfile = config.properties.profiles[defaultBaseProfileName];
+    const baseProfile = config.properties.profiles[Profiles.getInstance().getDefaultProfile("base")?.name];
     const profile = config.properties.profiles[profileName];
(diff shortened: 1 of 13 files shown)


traeok (commented):
Thanks for looking into this, Rudy!

Is it possible to refresh the affected profile (or the Profile class as a whole) instead of re-activating the extension? Otherwise, users will have to wait for the extension to finish loading before interacting with other profiles.

traeok (approved):
LGTM, thanks for this fix and for discussing the profile refresh process offline @rudyflores!

Going to leave this here from our discussion. I feel that we would benefit from having access to the tree provider from the node, and it would allow us to fire events and do misc. actions without being within the scope of the tree provider variables. But, this might involve a good bit of refactoring, and it may not be worth implementing if we plan to switch to a FileSystemProvider.

---
_"I see that ssoLogout has one abstract function for all the providers, so we can't get access to it from there... But Timothy and I were discussing this in-person, it can be tedious to access the tree provider since you can't get it from the node directly (even though they are related...) We would definitely benefit from having a 'getTreeProvider' on the node or something like that, but this is slowly expanding the scope of this PR._

_I'm not opposed to just refreshing the extension for v2 - users might interpret the refresh as unexpected behavior, but it seems necessary unless we add a `treeProvider` variable to the nodes, and then set it when building them..."_

JillieBeanSim:
Still seeing a loop in the DS tree when trying to escape the prompt, I don't see this behavior in marketplace version

marketplace:
https://github.com/zowe/vscode-extension-for-zowe/assets/49491949/2e5aabe4-4300-4944-8f18-c43cecfce849

branch:
https://github.com/zowe/vscode-extension-for-zowe/assets/49491949/04ef8426-11d9-43be-a3d1-24268dbac5ff

JillieBeanSim (changes requested):
See comment, idk why when I dnd'ed the recordings it left links though

packages/zowe-explorer/src/utils/ProfilesUtils.ts
```diff
@@ -129,6 +131,17 @@ export function isTheia(): boolean {
     return false;
 }
 
+export async function isUsingTokenAuth(profileName: string): Promise<boolean> {
+    const config = (await Profiles.getInstance().getProfileInfo()).getTeamConfig();
+    const profile = await Profiles.getInstance().getLoadedProfConfig(profileName);
+
+    if (profile.type === "zosmf") {
+        return config.properties.profiles["base"].secure.includes("tokenValue");
+    }
```
t1m0thyj, line 140: There may be some additional cases we should handle for z/OSMF profiles:
* The token value may not always be stored in a base profile. This is the default behavior of ZE, but it could also be stored in the z/OSMF profile itself.
* The base profile may have a different name other than "base".

rudyflores:
@t1m0thyj made the base profile selected generic and will now select the default base profile in config (if existing) otherwise it will check the profile itself if using token based auth.

packages/zowe-explorer/src/utils/ProfilesUtils.ts
```diff
@@ -133,13 +133,16 @@ export function isTheia(): boolean {
 
 export async function isUsingTokenAuth(profileName: string): Promise<boolean> {
     const config = (await Profiles.getInstance().getProfileInfo()).getTeamConfig();
-    const profile = await Profiles.getInstance().getLoadedProfConfig(profileName);
+    const defaultBaseProfileName = Profiles.getInstance().getDefaultProfile("base").name;
+    const baseProfile = config.properties.profiles[defaultBaseProfileName];
+    const profile = config.properties.profiles[profileName];
```
t1m0thyj: This won't work to access the profile if it's a nested profile like `lpar1.zosmf`.

In order to check if a secure field exists for a profile, we could do it like this:
`config.api.secure.securePropsForProfile(profileName).includes("tokenValue")`

t1m0thyj (changes requested):

packages/zowe-explorer/src/utils/ProfilesUtils.ts
t1m0thyj: Sorry for the confusing comment, as @zFernand0 mentioned at standup we should probably be using the ProfileInfo API instead to ensure that both v1 and v2 profiles are supported 🙂  I don't think the ProfileInfo API has a convenient method to check if a secure property exists for a profile, but we could add one to the Profiles class:
```typescript
    public async getSecurePropsForProfile(profileName: string): Promise<string[]> {
        const profAttrs = await this.getProfileFromConfig(profileName);
        const mergedArgs = (await this.getProfileInfo()).mergeArgsForProfile(profAttrs);
        return [...mergedArgs.knownArgs, ...mergedArgs.missingArgs].filter((arg) => arg.secure).map((arg) => arg.argName);
    }
```

rudyflores:
****FYI****

Addressed the code smell by moving some of the `ssoLogin` code into two new functions, the functionality is the same just simplified the code since it had nesting issues from the code smell reported before due to many if blocks being nested.

JillieBeanSim (changes requested):
Thanks @rudyflores this is looking awesome!  I did see one thing when I tried to test but wasn't through the firewall.  I get an error and a success message and should only get the error.
Screen Shot 2023-09-08 at 8 15 50 AM

t1m0thyj (commented):
After I log out of APIML, the session I logged out of is collapsed, but the arrow next to it still shows an expanded state:
image

Is this expected behavior? It's kind of nice that when I log in again, both sessions automatically show contents. But not sure if it makes sense from a UX perspective to show the arrow in an expanded state next to items that have no children.

t1m0thyj (approved):
LGTM! Should this be PR be rebased on main branch if we plan to release it for 2.11.0?

JillieBeanSim:
@t1m0thyj will be merging these changes into https://github.com/zowe/vscode-extension-for-zowe/pull/2450 to bring all bugfixes that were for 2.10.1 to main for 2.11.0

PR #2459, opened 2023-09-12
Skip creating snapshot version after publish for next

t1m0thyj:
The prerelease published successfully with this workflow:
https://github.com/zowe/vscode-extension-for-zowe/actions/runs/6161558200

But the post-publish step to create a snapshot version failed. I think we can disable it for the "next" branch because it doesn't make sense to be creating "SNAPSHOT" versions on a prerelease branch.

### DIFF
--- a/.github/workflows/deployment.yml
+++ b/.github/workflows/deployment.yml
@@ -67,7 +67,7 @@ jobs:
           git add packages/*/CHANGELOG.md
 
       - name: Create Snapshot Version
-        if: steps.octorelease.outcome == 'success'
+        if: ${{ github.ref_name != 'next' && steps.octorelease.outcome == 'success' }}
         uses: zowe-actions/octorelease/script@v1
         env:
           VERSION_STRING: "%s-SNAPSHOT"


zFernand0 (approved):

zFernand0:
Doing this means that every PR that gets merged will trigger a new next release, right?

JillieBeanSim (approved):

JillieBeanSim:
> Doing this means that every PR that gets merged will trigger a new next release, right?

I hope not lol, we don't want that do we?

zFernand0:
> > Doing this means that every PR that gets merged will trigger a new next release, right?
> 
> I hope not lol, we don't want that do we?

I don't think it will be a problems since it's only pre-releases (and we can have as many as we want before the actual thing)

JillieBeanSim:
wouldn't it require the lerna version to change for publishing since npmjs and osvx will have the previous version already published?

awharn (approved):

t1m0thyj:
> wouldn't it require the lerna version to change for publishing since npmjs and osvx will have the previous version already published?

For our stable releases, we have to remove "-SNAPSHOT" from the version in lerna.json to trigger a publish.
This doesn't apply for prereleases though - the version in lerna.json will get auto-updated by the deploy workflow in the format `3.0.0-next.<timestamp>`.

> > Doing this means that every PR that gets merged will trigger a new next release, right?
>
> I hope not lol, we don't want that do we?

I forgot that the deployment workflow only runs when the following files change: https://github.com/zowe/vscode-extension-for-zowe/blob/6c5d66df164558387de262ae99c5aeb479d6ac74/.github/workflows/deployment.yml#L9-L12

Are we ok with updating the timestamp in lerna.json to trigger publishing a new prerelease (even though that's not the timestamp which will be used since it gets auto-updated)?

traeok (approved):

PR #2445, opened 2023-09-06
Enhance UX for managing credentials

t1m0thyj:
## Proposed changes

This PR refactors multiple context menu items:
* Update Credentials
* Log In to Authentication Service
* Log Out of Authentication Service

into a single menu item "Manage Credentials" which shows a quickpick for users to choose an authentication method.

image

## Release Notes

Milestone: 2.12.0

Changelog:
> Replaced multiple options for updating credentials with a single "Manage Credentials" option that prompts for preferred authentication type. [#2263](https://github.com/zowe/vscode-extension-for-zowe/issues/2263)

## Types of changes

What types of changes does your code introduce to Zowe Explorer?
_Put an `x` in the boxes that apply_

- [ ] Bugfix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Updates to Documentation or Tests (if none of the other choices apply)

## Checklist

_Put an `x` in the boxes that apply. You can also fill these out after creating the PR. If you're unsure about any of them, don't hesitate to ask. We're here to help! This checklist will be used as reference for both the contributor and the reviewer_

- [x] I have read the [CONTRIBUTOR GUIDANCE](https://github.com/zowe/vscode-extension-for-zowe/wiki/Best-Practices:-Contributor-Guidance) wiki
- [ ] PR title follows [Conventional Commits Guidelines](https://www.conventionalcommits.org/en/v1.0.0-beta.2/)
- [x] PR Description is included
- [x] gif or screenshot is included if visual changes are made
- [x] `yarn workspace vscode-extension-for-zowe vscode:prepublish` has been executed
- [ ] All checks have passed (DCO, Jenkins and Code Coverage)
- [x] I have added unit test and it is passing
- [ ] I have added integration test and it is passing
- [x] There is coverage for the code that I have added
- [ ] I have tested it manually and there are no regressions found
- [ ] I have added necessary documentation (if appropriate)
- [ ] Any PR dependencies have been merged and published (if appropriate)

## Further comments

In addition to improving UX, this should make it easier to support certificates (#2373).

PR #2472, opened 2023-09-21
Profile management

JillieBeanSim:
## Proposed changes

This is an update for profile management that we can build from to eliminate issues we have seen users face with different authentication methods

right-click menu
Screen Shot 2023-10-02 at 6 51 05 PM

basic auth:
Screen Shot 2023-10-02 at 7 05 23 PM

token auth:
Screen Shot 2023-10-02 at 6 52 53 PM

Screen Shot 2023-10-02 at 6 53 43 PM

no auth specified:
Screen Shot 2023-10-02 at 6 51 21 PM

Screen Shot 2023-10-02 at 6 51 39 PM

## Release Notes

Milestone: 2.12.0

Changelog: Introduce a new user interface for managing profiles via right-click action "Manage Profile".

## Types of changes

## Checklist


traeok (commented):
Tested and LGTM, thanks @JillieBeanSim! I like the UX for this, it cleans up the context menu quite a bit and is easy to navigate.

Just left a few comments regarding the `ProfileUtils` helpers and quick pick options.

packages/zowe-explorer/i18n/sample/src/Profiles.i18n.json
```diff
@@ -1,5 +1,6 @@
 {
   "profiles.operation.cancelled": "Operation Cancelled",
+  "profiles.manualEditMsg": "Editing or removal of profiles will need to be done manually.",
```
t1m0thyj, line 3: Thanks for adding this message 👍  Should we prefix it with something like `Config file has been opened in the editor.`? I know that should be obvious, but we want to make sure that users don't miss this and think they have to manually track down the location of the file and open it.

t1m0thyj (dismissed):
Thanks @JillieBeanSim for creating the new UI! I think it is an improvement over the one proposed in #2445 and it's nice to declutter the tree context menu even more 🙂 

---

After creating a config with `zowe config init`, I have a z/OSMF profile that connects direct-to-service, and a base profile that stores user and password. I can "Update Credentials" with the new UI, but there is no option to "Log in to Authentication Service" if I would like to switch from user/password to auth token. With the old UI, I could have done so.
<details><summary>Click to see config</summary>
<pre>
{
    "profiles": {
        "zosmf": {
            "type": "zosmf",
            "properties": {
                "port": 443
            },
            "secure": []
        },
        "tso": {
            "type": "tso",
            "properties": {
                "account": "",
                "codePage": "1047",
                "logonProcedure": "IZUFPROC"
            },
            "secure": []
        },
        "ssh": {
            "type": "ssh",
            "properties": {
                "port": 22
            },
            "secure": []
        },
        "base": {
            "type": "base",
            "properties": {
                "host": "example.com"
            },
            "secure": [
                "user",
                "password"
            ]
        }
    },
    "defaults": {
        "zosmf": "zosmf",
        "tso": "tso",
        "ssh": "ssh",
        "base": "base"
    },
    "autoStore": true
}
</pre>
</details> 

---

Tested with a different config that has a z/OSMF profile that connects through the APIML, and a base profile that stores the APIML token. It also contains a z/FTP profile that connects direct-to-service. When I manage the FTP profile, the new UI presents me with the option to use token auth. However FTP does not support token auth.

We can use `ZoweExplorerApiRegister` to detect whether a profile type supports tokens, and if not then hide the token auth options from the menu. This is how I did it in my PR: https://github.com/zowe/vscode-extension-for-zowe/blob/5a6fc3abdedf6d3503469b40820fcd00db82fefc/packages/zowe-explorer/src/utils/ProfilesUtils.ts#L367-L372

JillieBeanSim:
> After creating a config with `zowe config init`, I have a z/OSMF profile that connects direct-to-service, and a base profile that stores user and password. I can "Update Credentials" with the new UI, but there is no option to "Log in to Authentication Service" if I would like to switch from user/password to auth token. With the old UI, I could have done so.


Hey Timothy, thanks for testing.
I believe with this case and the current release if the profile has user and password they will get the pop up message about the profile not supporting login action because it has the basic auth info saved. The user is required to remove the user and password from the secure array to change authentication method.

Screen Shot 2023-09-27 at 12 34 27 PM

I will continue looking at other concerns.

zFernand0:
Dumb question:
Should "all things profile" be inside the "Manage Profile" option?
i.e. should we move "Hide Profile" and Disable/Enable Validation" inside the quick-pick menu?

zFernand0 (dismissed):
- Should we remove the login option for service profiles that have the "user" and "password" in the secure array? 
  May sound odd, but I was sort of expecting that option not to be there.

- When I click on the login option, and I cancel out of the user/password prompts, the "login successful message shows up... Should that be removed?
image

- Should the logout operation be removed when there is no tokenValue stored in the vault? otherwise the user might get this error message.
image

- This one might be on the `main` branch already and does not cause any error messages (other than the "Element with ID XYZ is already registered).

  - When we refresh the extension via the command palette, should we collapse all profiles?
image

JillieBeanSim:
@zFernand0 addressed feedback `element already registered` error that exists on main and is way out of scope for this PR.  The concern about the logout option being presented, that option is only presented when the tokenType exists since that exists only when logged in and removed upon logout. The tokenValue in the vault check is reused from Rudy's work to determine if the profile should have login option and no basic auth options.

JillieBeanSim:
A heads up with moving the validation and hiding options into menu.  These options only work in the treeview the action is initiated in, same as the main branch.  We could do follow up work to make these work across all 3 trees in the future.

zFernand0 (approved):
Changes LGTM! 😋 
Thanks for addressing the comments.

---

> The concern about the logout option being presented, that option is only presented when the tokenType exists since that exists only when logged in and removed upon logout. 

- Do we plan to possibly update this if we decide not to remove tokenType on logout?

> The tokenValue in the vault check is reused from Rudy's work to determine if the profile should have login option and no basic auth options.

- Would it be possible to hook into that work to determine whether or not to should the logout option?

---

I found something kind of odd (which also happens on main and the marketplace)

When clicking the search icon on an APIML profile, it invalidates the profile with the following errors
image

And when clicking a ZOSMF profile, it asks for username and password. And if you cancel out of the prompt, you still get prompted for the search filter (dataset pattern, ...)

JillieBeanSim:
Hey @zFernand0 thanks for re-testing.  Would you mind creating a bug issue for the errors mentioned that are pre-existing?

t1m0thyj (changes requested):
Description on some rows ends with a period, but not for other rows. Can we make this consistent? IMO without periods looks better 😋 
image

Also curious if we should move "Disable Profile Validation" down further in the list? Although it's good that we have it, I don't think this option is used very often compared to other ones like "Edit Profile".

zFernand0 (approved):

rudyflores (approved):
LGTM, thanks for this enhancement @JillieBeanSim love the new icons :))

PR #2476, opened 2023-09-23
updated with local filtering of jobs

SanthoshiBoyina:
## Proposed changes
 -> For filtering the jobs at the zOSMF session level a filter icon is added for the easier accessibility to filter jobs based on job name, job id and return code. 

 -> and also provided with an option 'Filter Jobs' at the right click menu of zOSMF session level to filter jobs based on job name, job id and return code. 

And the input box added here will be filtering the content of Jobs dynamically. That means for each and every character user enters it will be displaying results based on that, which is an enhancement over the old method of filtering once the user enters text and hits the enter button.

## Release Notes

Milestone:

Changelog:

## Types of changes

## Checklist


packages/zowe-explorer/src/job/actions.ts
```diff
@@ -541,3 +542,74 @@ export async function sortJobsBy(jobs: IZoweJobTreeNode, jobsProvider: IZoweTree
...
 }
+
+export async function filterJobs(jobsProvider: IZoweTree<IZoweJobTreeNode>): Promise<vscode.InputBox> {
+    let acutal_jobs;
+    let flag = false;
+    for (const level of jobsProvider.mSessionNodes) {
+        if (level.label === "zosmf") {
+            acutal_jobs = level.children;
+            if (level.collapsibleState === 1) {
+                vscode.window.showInformationMessage("Inorder to filter jobs,first populate them using search icon");
```
traeok, line 553: - `Gui.infoMessage` can be used here
- Consider using localization for this message so it can be translated 🙂

packages/zowe-explorer/src/job/actions.ts
```diff
@@ -541,3 +542,74 @@ export async function sortJobsBy(jobs: IZoweJobTreeNode, jobsProvider: IZoweTree
...
+                flag = true;
+            }
+        }
+    }
+    if (flag) return;
+
+    const inputBox = await vscode.window.createInputBox();
+    inputBox.placeholder = "Type here...";
```
traeok, line 561: We should probably localize this string as well

packages/zowe-explorer/src/job/actions.ts
```diff
@@ -541,3 +542,74 @@ export async function sortJobsBy(jobs: IZoweJobTreeNode, jobsProvider: IZoweTree
...
+        job.children = Spools;
+
+        await TreeViewUtils.expandNode(job, zoweFileProvider);
+        job.collapsibleState = vscode.TreeItemCollapsibleState.Expanded;
+        jobsProvider.refresh();
+    }
+
+    const actual_spools = job.children;
+    const inputBox = vscode.window.createInputBox();
+    inputBox.placeholder = "Type here...";
```
traeok, line 605: we can localize this string (as mentioned above) and reference it here

packages/zowe-explorer/src/job/actions.ts
```diff
@@ -541,3 +542,74 @@ export async function sortJobsBy(jobs: IZoweJobTreeNode, jobsProvider: IZoweTree
...
+}
+
+export async function filterSpools(
+    jobsProvider: IZoweTree<IZoweJobTreeNode>,
+    job: IZoweJobTreeNode,
+    zoweFileProvider: IZoweTree<IZoweNodeType>
+): Promise<vscode.InputBox> {
+    if (job["collapsibleState"] == 1) {
+        const spools = await getSpoolFiles(job);
+        const Spools = spools.map((spool) => {
```
traeok, line 584:
```suggestion
        const spools = (await getSpoolFiles(job)).map((spool) => {
```

Also, we should catch any errors that may be thrown from `getSpoolFiles` - if an error occurs, pass the error info to the `errorHandling` function

packages/zowe-explorer/src/job/init.ts
```diff
@@ -24,7 +24,9 @@ import { ZoweLogger } from "../utils/LoggerUtils";
 
 export async function initJobsProvider(context: vscode.ExtensionContext): Promise<IZoweTree<IZoweJobTreeNode>> {
     ZoweLogger.trace("job.init.initJobsProvider called.");
-    const jobsProvider: IZoweTree<IZoweJobTreeNode> = await createJobsTree(globals.LOG);
+    const data = await createJobsTree(globals.LOG);
+    const jobsProvider: IZoweTree<IZoweJobTreeNode> = data;
+    const zoweFileProvider: IZoweTree<IZoweNodeType> = data;
```
traeok, line 29: should be able to remove this and use `jobsProvider`

packages/zowe-explorer/src/job/actions.ts
```diff
@@ -541,3 +542,74 @@ export async function sortJobsBy(jobs: IZoweJobTreeNode, jobsProvider: IZoweTree
...
+        jobsProvider.refresh();
+    });
+    inputBox.show();
+    return inputBox;
+}
+
+export async function filterSpools(
+    jobsProvider: IZoweTree<IZoweJobTreeNode>,
+    job: IZoweJobTreeNode,
+    zoweFileProvider: IZoweTree<IZoweNodeType>
```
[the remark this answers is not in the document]
SanthoshiBoyina: Yes it will be working when we replace zoweFileProvider with jobsProvider. The problem comes in unit testing where in this line `await TreeViewUtils.expandNode(job, zoweFileProvider);` expandNode function  is of this type `TreeViewUtils.expandNode(node: IZoweTreeNode, provider: IZoweTree<IZoweNodeType>)` which accepts provider of type` IZoweTree<IZoweNodeType>`, so I have used zoweFileProvider of type` IZoweTree<IZoweNodeType>` instead of jobsProvider of type` IZoweTree<IZoweJobTreeNode>`.

packages/zowe-explorer/i18n/sample/src/job/actions.i18n.json
```diff
@@ -21,5 +21,7 @@
   "cancelJobs.notImplemented": "The cancel function is not implemented in this API.",
   "cancelJobs.notCancelled": "The job was not cancelled.",
   "cancelJobs.failed": "One or more jobs failed to cancel: {0}",
-  "cancelJobs.succeeded": "Cancelled selected jobs successfully."
+  "cancelJobs.succeeded": "Cancelled selected jobs successfully.",
+  "filterJobs.message": "Inorder to filter jobs,first populate them using search icon",
```
JillieBeanSim, line 25: Could we change this message to say the same as if profile was expanded without search in place.
`Use the search button to display jobs` 
Or something along those lines, possibly `A jobs search must be in place before local filtering is possible`

packages/zowe-explorer/i18n/sample/src/job/actions.i18n.json
```diff
@@ -21,5 +21,7 @@
...
+  "filterJobs.prompt.message": "Type here..."
```
JillieBeanSim, line 26: Could we explain what is to be entered here? Maybe something like `Enter local filter...`

JillieBeanSim (changes requested):
While testing I get the info message about needing a search in place to filter but do have a search in place.
Screen Shot 2023-09-28 at 9 18 26 AM

packages/zowe-explorer/src/job/actions.ts
```diff
@@ -574,42 +574,41 @@ export async function filterJobs(jobsProvider: IZoweTree<IZoweJobTreeNode>): Pro
...
+        const actual_spools = job.children;
+        const inputBox = vscode.window.createInputBox();
+        inputBox.placeholder = localize("filterJobs.prompt.message", "Enter local filter...");
+        inputBox.onDidChangeValue((query) => {
+            query = query.toUpperCase();
+            job["children"] = actual_spools.filter((item) =>
+                `${item["spool"].stepname as string}:${item["spool"].ddname as string} - ${item["spool"]["record-count"] as string}`.includes(query)
             );
-            return spoolNode;
+            jobsProvider.refresh();
```
traeok, line 607: We should refresh the `job` element here instead of the whole tree; otherwise, it will perform logic on other job nodes that are not a part of this filter operation.

This would also apply to line 571 in the `filterJobs` function - we can refresh the session (`level`) element for that case

packages/zowe-explorer/src/job/actions.ts
```diff
@@ -549,16 +549,16 @@ export async function filterJobs(jobsProvider: IZoweTree<IZoweJobTreeNode>): Pro
     for (const level of jobsProvider.mSessionNodes) {
         if (level.label === "zosmf") {
```
JillieBeanSim, line 550: We don't want to have any particular type calls like checking for zosmf profile type or if this is a check for profile name.  We want to be able to handle this for all profile types and names.  This may need to be a check on context value for if session node.

packages/zowe-explorer/src/job/actions.ts
JillieBeanSim: I tried both jobProvider.refresh and refreshElement(job) and don't see either of these working like I see the filter search at the session level.

JillieBeanSim (commented):
the inputBox.show then return of inputBox doesn't allow the input box to disappear after hitting enter, I have to escape the input box for it to leave the screen in both filter searches. One way to solve that is to to an 
`await vscode.window.showInputBox(inputBox)` and if wanting to return what is entered can store and return value or just have a void return.
Is there a way to clear the local filtering and return user to initial filter search?

rudyflores:
@JillieBeanSim do we have a milestone for this PR?

JillieBeanSim (approved):
This is great! thanks @SanthoshiBoyina

rudyflores (approved):
Thanks @SanthoshiBoyina !

traeok (changes requested):
Seems like the recent changes to this branch pulled in some other commits that were already merged into `main`. To fix this, it might require undoing the last merge attempt, and then running `git merge` again with the main branch. I'd be happy to help with resolving any merge conflicts that come up during the merge 🙂

PR #2490, opened 2023-10-02
Fix Coverity issue and use local ZE API in samples

t1m0thyj:
* Fixes a code smell flagged by Coverity in `ZoweDatasetNode.ts`
* Updates samples to reference ZE API with a symlink, so that Samples CI workflow will test for breaking changes to API


traeok (approved):
LGTM, thanks @t1m0thyj

awharn (approved):

Issue #2381, opened 2023-07-14
 Allow Zowe Explorer to check VS Code marketplace to install current credential manager


rudyflores:
**Is your feature request related to a problem? Please describe.**

In Zowe Explorer a user is able to use custom credential managers by installing a credential manager extension on VS Code that will change their imperative override setting. The issue  currently is that if the user does not have the credential manager installed, but the override does not exist it changes it back to the default credential manager 

**Describe the solution you'd like**

I would like for Zowe Explorer to check the VS Code marketplace for the availability of the extension and suggest to install it if available, otherwise change it back to the default credential manager.

**Describe alternatives you've considered**

**Additional context**

rudyflores:
Closing this issue since the PR above has addressed this

PR #2492, opened 2023-10-03
Update credential management flow

rudyflores:
## Proposed changes

Solves #2381 

This is a restructure of our existing `ProfilesUtils.ts` static class, it will now cover the following cases we meant to address when integrating custom credential managers, this were cases such as:

**Having `.vsix` of credential manager installed and `imperative.json` override not updated will now prompt the user to either change or stick to default `@zowe/cli` in their override setting**

Screenshot 2023-10-03 at 2 42 05 PM

**Having `imperative.json` override updated (Installed plugin from the CLI) and`.vsix` not installed will now prompt to install `.vsix` if missing**

Screenshot 2023-10-03 at 2 40 25 PM

### NOTE

This applies for any custom credential manager that is registered with imperative, not only the secrets for kubernetes plugin.

## Release Notes

Milestone: v2.12.0

Changelog: Added better way of handling credential management by installing/changing missing VSIX or override setting on `imperative.json`.

## Types of changes

## Checklist


JillieBeanSim (commented):
I have a quick question about this work and it's location. Should this work and these checks be in the ZE API package making it reusable for us as well as extenders?  I ask this because when thinking of extenders like the CICS extension, they may need to make considerations like this if handling profiles.  Right now they possibly call the ProfilesCache.getProfileInfo() that makes call to credential manager which is different than what we have here.

rudyflores:
> I have a quick question about this work and it's location. Should this work and these checks be in the ZE API package making it reusable for us as well as extenders? I ask this because when thinking of extenders like the CICS extension, they may need to make considerations like this if handling profiles. Right now they possibly call the ProfilesCache.getProfileInfo() that makes call to credential manager which is different than what we have here.

That's a fair point, I wouldn't be opposed to doing this change, the only thing we should confirm is the reason for having these function in `ProfileUtils.ts` to begin with, @zFernand0 @t1m0thyj any idea why we made this code in that location vs having it as code in the API originally?

packages/zowe-explorer/src/utils/ProfilesUtils.ts
```diff
@@ -225,23 +225,65 @@ export class FilterDescriptor implements vscode.QuickPickItem {
...
-        return imperative.CredentialManagerOverride.getCredMgrInfoByDisplayName(credentialManager.credMgrDisplayName);
     }
 
-    public static async activateCredentialManagerOverride(
+    /**
+     * Activate a vscode extension of a custom credential manager
+     * @param credentialManagerExtension The credential manager VS Code extension name to activate
+     * @returns Promise<imperative.ICredentialManagerConstructor> the constructor of the activated credential manager
+     */
+    private static async activateCredentialManagerOverride(
```
zFernand0, line 286: I'm curious if we should keep this methods public to avoid any breakage (if extenders have access to call these).
if not, please disregard this comment 😋

packages/zowe-explorer/src/utils/ProfilesUtils.ts
```diff
@@ -243,48 +285,171 @@ export class ProfilesUtils {
...
+                        if ((await Gui.showMessage(refreshMessage, { items: [refreshButton] })) === refreshButton) {
+                            await vscode.commands.executeCommand("workbench.action.reloadWindow");
+                        }
+                    }
+                }
+            }
+        );
+    }
+
+    public static async getProfileInfo(): Promise<imperative.ProfileInfo> {
```
JillieBeanSim, line 425: This is what I was referencing with ZE API.  We have a ProfilesCache.getProfileInfo() that makes the call to imperative to set the default credential manager.  If extenders like CICS are using this & their extension in an environment that would need the k8s ext for credentials shouldn't that be considered in the ProfilesCache.getProfileInfo() passing the correct information to imperative when an extenders calls the method.  If we don't they will always only get the default cred manager.

JillieBeanSim (changes requested):
see comment

packages/zowe-explorer/src/utils/ProfilesUtils.ts
rudyflores: It should be fine if this was switched to private, since it's in our Zowe Explorer package and no extenders can access this function :)

packages/zowe-explorer/src/utils/ProfilesUtils.ts
rudyflores: Created issue https://github.com/zowe/vscode-extension-for-zowe/issues/2525 for follow up work on this, since we need to add an optional parameter that extenders can pass in for loading either the default or custom credential manager, aside from that this PR is good to go!

JillieBeanSim:
@rudyflores can you check the failing Theia tests please, it seems to fail consistently

traeok (commented):
I did some testing and was able to reproduce both dialogs. Dialog options work as expected. Thanks @rudyflores!

I noticed one minor issue with the localized strings - looks like `vscode-nls` interpreted new lines from a couple of the strings and added it to the text. Other than that, LGTM

image

packages/zowe-explorer/i18n/sample/src/utils/ProfilesUtils.i18n.json
```diff
@@ -6,8 +6,18 @@
...
+  "ProfilesUtils.fetchRegisteredPlugins.customCredentialManagerFound": "Custom credential manager {0} found",
+  "ProfilesUtils.fetchRegisteredPlugins.message": "Do you wish to use this credential manager instead?",
+  "ProfilesUtils.fetchRegisteredPlugins.yes": "Yes",
+  "ProfilesUtils.fetchRegisteredPlugins.dontAskAgain": "Don't ask again",
+  "ProfilesUtils.promptAndHandleMissingCredentialManager.suggestInstallHeader": "Plugin of name \"{0}\" was defined for custom credential management on imperative.json file.",
+  "ProfilesUtils.promptAndHandleMissingCredentialManager.suggestInstallMessage": "Please install associated VS Code extension for\n        custom credential manager or revert to default.",
+  "ProfilesUtils.promptAndHandleMissingCredentialManager.revertToDefault": "Use Default",
+  "ProfilesUtils.promptAndHandleMissingCredentialManager.install": "Install",
+  "ProfilesUtils.promptAndHandleMissingCredentialManager.refreshMessage": "After installing the extension, please make sure to reload\n                             your VS Code window in order to start using the installed credential manager",
```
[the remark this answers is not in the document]
rudyflores: All good now!

zFernand0 (approved):
Good catch by Trae on the strings, I honestly didn't notice that.

- One slightly annoying thing is that you have to close/re-open (or `> Developer: Reload Window`) after selecting the `Use Default` button 😋 
  - It's possible that the above behavior is because I'm in a remote ssh env 😋

traeok (dismissed):

packages/zowe-explorer/src/utils/ProfilesUtils.ts
```diff
@@ -409,8 +408,8 @@ export class ProfilesUtils {
                     if (await vscode.env.openExternal(credentialManagerInstallURL)) {
                         const refreshMessage = localize(
                             "ProfilesUtils.promptAndHandleMissingCredentialManager.refreshMessage",
-                            `After installing the extension, please make sure to reload
-                             your VS Code window in order to start using the installed credential manager`
+                            `After installing the extension, please make sure to reload your VS Code window in order
+                             to start using the installed credential manager`
```
zFernand0, line 412: Not sure why the extra spacing doesn't show, but the notification looks good
image


Also, should the button say `Reload` instead of `Refresh`

packages/zowe-explorer/src/utils/ProfilesUtils.ts
```diff
@@ -390,13 +390,12 @@ export class ProfilesUtils {
         ZoweLogger.trace("ProfilesUtils.promptAndHandleMissingCredentialManager called.");
         const header = localize(
             "ProfilesUtils.promptAndHandleMissingCredentialManager.suggestInstallHeader",
-            `Plugin of name "{0}" was defined for custom credential management on imperative.json file.`,
+            "Plugin of name '{0}' was defined for custom credential management on imperative.json file.",
```
zFernand0, line 393: image

zFernand0 (dismissed):

packages/zowe-explorer/src/utils/ProfilesUtils.ts
rudyflores: hmm that makes sense! I think Refresh suits it better :)

packages/zowe-explorer/src/utils/ProfilesUtils.ts
rudyflores: Just wanted to confirm what the issue was here?

zFernand0: oh, sorry, just wanted to post an updated picture 😋

zFernand0 (approved):

traeok (approved):

JillieBeanSim (approved):
LGTM! thanks @rudyflores

Issue #2207, opened 2023-03-27
Use `vscode.FileSystemProvider` for content downloaded from mainframe


t1m0thyj:
Replace our usage of a temp download folder on disk with an in-memory file system provider. I think there are a lot of benefits that would result from this switch, but it is a significant change to ZE architecture so I suggest we start with a spike story to investigate/research it first and share results with the team before undertaking it.

**Benefits**
* Could hook into save event for better error handling/dirty detection
* Could use `vscode.open` for automatic double-click detection in trees
* More TBD

**Resources**
* https://code.visualstudio.com/api/extension-guides/virtual-documents
* https://github.com/microsoft/vscode-extension-samples/tree/main/fsprovider-sample
* https://github.com/liximomo/vscode-remote-fs/blob/master/src/core/RemoteFileSystemProvider.ts

t1m0thyj:
Spike for 23PI2 to investigate pros and cons of adopting the FIleSystemProvider API and document expected breaking changes.

VitGottwald:
An interesting aspect - from the extenders point of view - would be the ability to request dataset content or list of PDS members by reading ZOWE Explorer provided file system URL rather than having to issue download requests. Similarly for jobs and uss files.

t1m0thyj:
Proposed URI scheme to access resources through a FileSystemProvider:
`zowe://<treeId>/<profileName>/<resourcePath>?<options>`

* `<treeId>` is an ID associated with a tree provider that provides the resource ("ds", "uss", "jobs", "cics")
* `<profileName>` is the name of a profile to be used to access the resource (e.g., "lpar1.zosmf", "lpar1.zftp")
* `<resourcePath>` is a path to uniquely identify the resource (data set, USS file, job spool file, etc)
  * For "ds", it would be a data set (and member) name (e.g., `IBMUSER.PS`, `IBMUSER.PDS(MEMBER)`)
  * For "uss", it would be a USS path (e.g., `/u/ibmuser/test.txt`)
  * For "jobs", it could be similar to the name we currently use: `<jobname>.<jobid>.<ddname>`
* `<options>` allows request parameters to be passed (e.g., `binary=true`)

Advantages:
* No local filesystem required since we don't have to store files on disk in a temp folder anymore
* Breadcrumb at top of editor is a convenient alternative to trees for navigating mainframe resources
* Resource can be opened in ZE by issuing a CLI command - for example to open a data set:
[...]

PR #2522, opened 2023-10-18
VSC Compare function, not working with Favorites from Zowe Explorer

SanthoshiBoyina:
## Proposed changes

The feature of Visual Studio Code to compare 2 files, if the file on the mainframe has changed in the meantime, doesn't work if you open a Member under Favorites. It overrides the file automatically.  In this bug fix, we are able to trigger Compare function when ETAGs are mismatched.

## Release Notes

Milestone:

Changelog:

## Types of changes

## Checklist


rudyflores:
Is this for 2.12.0? @JillieBeanSim @SanthoshiBoyina

JillieBeanSim:
> Is this for 2.12.0? @JillieBeanSim @SanthoshiBoyina

I am not opposed to it being included if the tests are fixed and the coverage is good in time, not seeing the coverage report yet with failing tests

JillieBeanSim (changes requested):
I am seeing the diff editor open then when trying to save that I get the pop up to compare, I would assume one or the other happen but not both and the second right behind the first encounter.

traeok (commented):
Thanks for working on this bug fix @SanthoshiBoyina! 

I'm running into the same situation as @JillieBeanSim - the side-by-side diff appears, but the dialog with Compare/Overwrite does not. However, this occurs for non-favorited nodes as well. I think this is because calls to `willForceUpload` (in `src/shared/utils.ts`) were removed with a previous PR, and this function was responsible for showing the dialog:

```ts
Gui.infoMessage(localize("saveFile.info.confirmUpload", "Would you like to overwrite the remote file?"), {
        items: [localize("saveFile.overwriteConfirmation.yes", "Yes"), localize("saveFile.overwriteConfirmation.no", "No")],
    })
```
_(lines 284-286 in the mentioned file)_

I'm wondering if we still want to use this `willForceUpload` function somewhere? As a quick solution, if we add this dialog to the `compareFileContent` function in `src/shared/utils.ts`, or if we show the dialog before calling `compareFileContent`, it should fix the issue.

JillieBeanSim:
I don't think we should be forceUploading anything unless the decision came from user.

traeok:
> I don't think we should be forceUploading anything unless the decision came from user.

@JillieBeanSim I agree. My comment was mostly pointing out that the dialog we used to show before the diff is no longer there. We used to call a similar `willForceUpload` function (mentioned above) that contained a similar dialog, but this call was removed at some point in the past and now we do not show a dialog at all when a conflict occurs. This is a regression introduced in a previous release, since we used to show a dialog, but it is not a result of the work on this PR. This could always be added to the repo as a separate issue for a future PR to solve.

JillieBeanSim:
ok, if we had a regression maybe the popup can be added/moved to the shared.utils.compareFileContent() [method](https://github.com/zowe/vscode-extension-for-zowe/blob/464d7128ab890df5aa16ad20a45ea782f68201af/packages/zowe-explorer/src/shared/utils.ts#L362).  this will cut down on introducing duplicate code to bring it back in other cases and fix the duplication here in this PR

SanthoshiBoyina:
> I am seeing the diff editor open then when trying to save that I get the pop up to compare, I would assume one or the other happen but not both and the second right behind the first encounter.

VS Code has a built-in mechanism to show a notification when you try to save a dirty file that has been changed outside of VS Code or by another program
save-conflict
In order not to make this built-in vs code notification pop up, In the setting we need to configure `files.saveConflictResolution` to `overwriteFileOnDisk` option from the default `askUser` option.

traeok:
> > I am seeing the diff editor open then when trying to save that I get the pop up to compare, I would assume one or the other happen but not both and the second right behind the first encounter.
> 
> VS Code has a built-in mechanism to show a notification when you try to save a dirty file that has been changed outside of VS Code or by another program save-conflict In order not to make this built-in vs code notification pop up, In the setting we need to configure `files.saveConflictResolution` to `overwriteFileOnDisk` option from the default `askUser` option.

@SanthoshiBoyina Currently, PDS members and USS files are being written to temp files, so VS Code has no way of checking whether the document has a different e-tag than the mainframe. It calculates an e-tag for the temp. file on disk, but we do not use that e-tag when trying to upload file contents. Until we switch to something like a FileSystemProvider (#2207), we have to temporarily save the on-disk file, and then check with the mainframe to see if we have to do conflict resolution on our own. If there is a conflict, we un-save the file and then present a similar dialog.

rudyflores:
Seems like the issue here with the compare editor not showing is because the `node` being passed into `compareFileContent` is a directory, can see that the error that shows in the debug console when trying to click on `Compare` mentions that a directory was entered too, this happens right when `node.getUSSDocumentFilePath()` is called in line `377`

Issue #2499, opened 2023-10-06
zftp: Members aren't saved, message "Saving data set..." but nothing happens and never ends.


ssamayoa:
**Describe the bug**

zftp
Members aren't saved, message "Saving data set..." but nothing happens and never ends.
Some time works but most of the time doesn't. 
Used to work all the time but suddenly this erratic behavior began.

**To Reproduce**

1. Create new PDS
2. Create new member
3. Write something then save.
4. message "Saving data set..." but nothing happens and never ends.

**Expected behavior**
Members saved.

**Screenshots**

**Desktop (please complete the following information):**

- OS: Windows 10 22H2 10.0.19045.3448
- Zowe Explorer Version:  2.11.0
- Are you using Secure Credential Store? not sure.

**Additional context**

-  At start-up I get, since the beginning this message:

Error: Unable to initialize secure credentials plugin for Z Open Editor. Check that you configured the "Zowe Security: Credential Key" user preferences with the value "Zowe-Plugin". Error returned: "Failed to initialize secure credential manager"

- When "Zowe Explorer" shows  the data set list, I can see 2 tasks in SDSF:

OLS0009  STEP1             STC06059 OLS0009  
OLS0009  DBPROCBG TCP00027 TSU06101 OLS0009  
OLS0009  STEP1             STC06093 OLS0009  

DBPROCBG is my TN3270 session.

- I'm the administrator of the mainframe and haven't changed any configuration.
- Saved members (when works) does not have Size, Created, Changed or ID - should them?
- Where and/or how I can see what's happening?
- From CLI (zowe zos-ftp upload...) works fine.

ssamayoa:
**UPDATE:**

I configured TRACE level to "Zowe Explorer"'s log, connected to another z/OS and found that after the ds is saved this are the output:

[2023/10/07 10:10:54] [TRACE] dataset.utils.getProfileAndDataSetName called.
[2023/10/07 10:10:54] [TRACE] ZoweDatasetNode.getEtag called.
[2023/10/07 10:10:56] [TRACE] ZoweDatasetNode.setEtag called.

But connected to our main z/OS last line never show and the message "Saving data set..." stays for ever.

FWIW both z/OS are 2.4

What else I should check?

JillieBeanSim:
the startup error is from the secure credentials being used, which is default for Zowe.  This could cause other issues while trying to use the Zowe Profiles.

[...]

PR #2526, opened 2023-10-24
fix(FTP): One active FTP connection per operation; remove temp. files once finished

traeok:
## Proposed changes

This PR fixes a couple of issues with the FTP extension (found while working on FileSystemProvider):
- `FTPClient.connectWithArguments` was called twice before the first connection was closed in `putContents`. The fix closes this connection before opening a new one.
  - This seems like it might affect the behavior in #2499, but I haven't been able to reproduce the issue without the fix
- Temporary files created by `getContentsTag` were not removed after use, causing orphaned temp. files with remote content to persist in the user's temp folder.

## Release Notes

Milestone: 2.12.0 (or 2.13.0, I know there's a lot going out for this release)

Changelog:

- Fixed an issue with FTP where temp. files for e-tag comparison were not deleted after use.
- Fixed an issue with FTP where another connection attempt was made, even though one was already active.

## Types of changes

## Checklist


tiantn:
Hi @traeok, thank you for the fix. Could you also update the following line in red frame to ` if (!connection) {`? Thank you!
截屏2023-10-31 15 38 42

zFernand0:
> Hi @traeok, thank you for the fix. Could you also update the following line in red frame to ` if (!connection) {`? Thank you! 截屏2023-10-31 15 38 42

Hey @tiantn,
Do we expect falsy values from the plugin?

traeok:
> Hi @traeok, thank you for the fix. Could you also update the following line in red frame to ` if (!connection) {`? Thank you!

Hi @tiantn, no problem! I'm hesitant to invert this behavior, as `connection != null` will determine whether the connection object is valid before trying to call its functions.

Unless there is a need to check for other "falsy values," I think we can keep this check as-is 🙂

tiantn:
> > Hi @traeok, thank you for the fix. Could you also update the following line in red frame to ` if (!connection) {`? Thank you!
> 
> Hi @tiantn, no problem! I'm hesitant to change the current check, as `connection != null` will determine whether the connection object is valid before trying to call its functions.
> 
> Unless there is a need to check for other "falsy values," I think we can keep this check as-is 🙂

Hi @traeok , @zFernand0 , we consider the `undefined` case so want to change it.

zFernand0:
> > > Hi @traeok, thank you for the fix. Could you also update the following line in red frame to ` if (!connection) {`? Thank you!
> > 
> > 
> > Hi @tiantn, no problem! I'm hesitant to change the current check, as `connection != null` will determine whether the connection object is valid before trying to call its functions.
> > Unless there is a need to check for other "falsy values," I think we can keep this check as-is 🙂
> 
> Hi @traeok , @zFernand0 , we consider the `undefined` case so want to change it.

Hey @tiantn,

The `!=` operator should be equivalent to `connection !== undefined && connection !== null`.

Changing it to `!connection` will result in an error where:
- `null.close` throws `Uncaught TypeError: Cannot read properties of null (reading 'close')`
- `undefined.close` throws ``Uncaught TypeError: Cannot read properties of undefined (reading 'close')`

I believe you meant `if(connection) { connection.close() } `.
However that can lead to problems for truthy values where you might see this error: 
 - `Uncaught TypeError: connection.close is not a function`

tiantn:
> > > > Hi @traeok, thank you for the fix. Could you also update the following line in red frame to ` if (!connection) {`? Thank you!
> > > 
> > > 
> > > Hi @tiantn, no problem! I'm hesitant to change the current check, as `connection != null` will determine whether the connection object is valid before trying to call its functions.
> > > Unless there is a need to check for other "falsy values," I think we can keep this check as-is 🙂
> > 
> > 
> > Hi @traeok , @zFernand0 , we consider the `undefined` case so want to change it.
> 
> Hey @tiantn,
> 
> The `!=` operator should be equivalent to `connection !== undefined && connection !== null`.
> 
> Changing it to `!connection` will result in an error where:
> 
>     * `null.close` throws `Uncaught TypeError: Cannot read properties of null (reading 'close')`
> 
>     * `undefined.close` throws ``Uncaught TypeError: Cannot read properties of undefined (reading 'close')`
> 
> 
> I believe you meant `if(connection) { connection.close() } `. However that can lead to problems for truthy values where you might see this error:
> 
>     * `Uncaught TypeError: connection.close is not a function`
> 
> 
> In this case, a more thorough check could be `connection != null && connection instanceof T` 😋

Hi @zFernand0 , @traeok , sorry, I am wrong. I mean`if(connection)` or`if (connection !== null && connection !== undefined)`.

t1m0thyj (approved):
LGTM, thanks @traeok!

zFernand0:
> > > > > Hi @traeok, thank you for the fix. Could you also update the following line in red frame to ` if (!connection) {`? Thank you!
> > > > 
> > > > 
> > > > Hi @tiantn, no problem! I'm hesitant to change the current check, as `connection != null` will determine whether the connection object is valid before trying to call its functions.
> > > > Unless there is a need to check for other "falsy values," I think we can keep this check as-is 🙂
> > > 
> > > 
> > > Hi @traeok , @zFernand0 , we consider the `undefined` case so want to change it.
> > 
> > 
> > Hey @tiantn,
> > The `!=` operator should be equivalent to `connection !== undefined && connection !== null`.
> > Changing it to `!connection` will result in an error where:
> > ```
> > * `null.close` throws `Uncaught TypeError: Cannot read properties of null (reading 'close')`
> > 
> > * `undefined.close` throws ``Uncaught TypeError: Cannot read properties of undefined (reading 'close')`
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > I believe you meant `if(connection) { connection.close() } `. However that can lead to problems for truthy values where you might see this error:
> > ```
> > * `Uncaught TypeError: connection.close is not a function`
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > In this case, a more thorough check could be `connection != null && connection instanceof T` 😋
> 
> Hi @zFernand0 , @traeok , sorry, I am wrong. I mean`if(connection)` or`if (connection !== null && connection !== undefined)`.

I do believe that `connection != null` and `connection !== null && connection !== undefined` functionally behave the same way. 
The only reasons I see for changing it is because:
- Code Quality Tools:
  - If there is a syntactical analysis tool that complains about this
    - SonarCLoud doesn't seem to fail the quality check.
- Strict Mode Considerations:
  - If you're working in strict mode (`'use strict';`), using `!=` may throw a TypeError.
    - We haven't had any Javascript `strict;` users requesting that we avoid these.

If we want to have a more robust check, I believe that checking for `instanceof T` might be more impactful than switching from `1=` to `!==`.

That said, I'm ok if we decide to leave it as-is, or change it in the future.

If someone feels strongly one way or the other, I'm happy to approve a follow-up PR with these changes :yum:

zFernand0 (approved):
I believe I've expressed my point of view on the `!= null` vs. `!== null || !== undefined`  debate. Approving for now.

LGTM! :yum:

tiantn (approved):
Thank you @traeok and @zFernand0 !

JillieBeanSim:
Hey @tiantn @traeok @zFernand0 & @t1m0thyj, I see this has the approvals but before I merged I wanted to ask if we would like this out with the patch release 2.12.1?  If so we will want to rebase to maintenance or the [PR](https://github.com/zowe/vscode-extension-for-zowe/pull/2540) for merging all 2.12.0 changes into maintenance

PR #2538, opened 2023-10-31
fix(ds/sort): Fallback to sorting by name when values are equal

traeok:
_If we would prefer to keep this for a potential 2.12.1 release, then I will add a changelog. Otherwise, I think we can keep this as `no-changelog` as the feature will be released with 2.12._

---

## Proposed changes

I've cleaned up the sorting function a bit so that it falls back to sorting by name when 2 node values are equal.

In addition, I've adjusted the date checks slightly to also use Day.js (consistent with the rest of the date processing for this feature) for easier date comparison.

## Release Notes

Milestone: 2.12.0 (or 2.12.1)

Changelog:

- Fixed an issue where sorting functionality did not consider the sort direction when values were equal.

## Types of changes

## Checklist

### DIFF
--- a/packages/zowe-explorer/__tests__/__unit__/dataset/DatasetTree.unit.test.ts
+++ b/packages/zowe-explorer/__tests__/__unit__/dataset/DatasetTree.unit.test.ts
@@ -15,7 +15,15 @@ import * as fs from "fs";
 import * as zowe from "@zowe/cli";
 import { DatasetTree } from "../../../src/dataset/DatasetTree";
 import { ZoweDatasetNode } from "../../../src/dataset/ZoweDatasetNode";
-import { DatasetFilterOpts, Gui, IZoweDatasetTreeNode, ProfilesCache, ValidProfileEnum } from "@zowe/zowe-explorer-api";
+import {
+    DatasetFilterOpts,
+    DatasetSortOpts,
+    Gui,
+    IZoweDatasetTreeNode,
+    ProfilesCache,
+    SortDirection,
+    ValidProfileEnum,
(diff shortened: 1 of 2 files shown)


zFernand0 (commented):
Seems like this would work.
testing briefly before approving.

adam-wolfe (dismissed):
LGTM. Thanks Trae!

adam-wolfe:
One minor issue I am seeing that may be related to this is that if a member has no user ID, and you sort by User ID in the descending order, the member without a user ID will switch to the top and bottom of the list when you collapse and expand the PDS.

traeok:
> One minor issue I am seeing that may be related to this is that if a member has no user ID, and you sort by User ID in the descending order, the member without a user ID will switch to the top and bottom of the list when you collapse and expand the PDS.

Thanks for catching this, I've addressed this in [9235f28](https://github.com/zowe/vscode-extension-for-zowe/pull/2538/commits/9235f28b8bff80ad1d6816c881c1f7d2385888cc)

rudyflores (dismissed):

adam-wolfe (approved):
Looks good.

traeok:
I found an issue with how we generally build children in `ZoweDatasetNode` - we were only checking for `dsname` as part of the "existing node" check, where we should fallback to look for the `member` variable when `dsname` isn't present. Otherwise, the check for an existing PDS member node will always yield undefined. See 2da3995 for the change that was made.

zFernand0 (approved):
LGTM! 😋 
Thanks for addressing the comments

adam-wolfe (approved):

PR #2549, opened 2023-11-07
VSC Compare function, not working with Favorites from Zowe Explorer

SanthoshiBoyina:
## Proposed changes
The feature of Visual Studio Code to compare 2 files, if the file on the mainframe has changed in the meantime, doesn't work if you open a Member under Favorites. It overrides the file automatically. In this bug fix, we are able to trigger Compare function when ETAGs are mismatched.

## Release Notes

Milestone:

Changelog:

## Types of changes

## Checklist


traeok (commented):
Hi @SanthoshiBoyina, thanks for working on resolving conflict management - looks great so far.

After a discussion w/ @t1m0thyj, I do have a concern regarding the Data Set tree. On line 523 in the `openPS` function (`dataset/actions.ts`), I noticed that we set the e-tag regardless of its value in the response:

```ts
const response = await responsePromise;
node.setEtag(response?.apiResponse?.etag);
statusMsg.dispose();
```

If a response does not return an e-tag, it will set the e-tag to null, which will cause the first save to bypass the e-tag mismatch on the mainframe.

To avoid this, we'll need to add a check to see if `response?.apiResponse?.etag` is a valid value before calling `node.setEtag`.

t1m0thyj (changes requested):
Agree with @traeok's comment, here are steps for how I reproduced the issue:
1. Open a data set in preview mode by single-clicking on it in the tree
2. Make changes to the data set on the mainframe
3. Make changes to the data set locally in ZE
    This will switch the editor from preview mode (title italicized) to edit mode 👍 
    It will also reset the etag for the current node to null which we don't want 😢 
5. Save the changes and notice that no conflict is detected

IMO we should fix this before 2.12.1 to avoid potential data loss, potentially in this PR since its related to compare functionality? I also don't mind opening a separate PR to fix it if preferred 🙂

t1m0thyj:
Based on discussion at standup, I'll open another PR against this one with the fix for etag being set to null 🙂

JillieBeanSim (dismissed):
LGTM! thanks for the fix

rudyflores (dismissed):
Thanks @SanthoshiBoyina ! 

Only thing I see is some code coverage missing but not sure if this will be an issue since I know we want to get the release out?

t1m0thyj (changes requested):

t1m0thyj (approved):
LGTM, thanks @SanthoshiBoyina and @JillieBeanSim!

rudyflores (approved):
Nice catch @t1m0thyj !

PR #2559, opened 2023-11-09
Bring 2.12.1 changes from `maintenance` into `main`

rudyflores:
## Proposed changes

Bring in changes from `maintenance` branch into our `main` branch 

## Release Notes

Milestone: N/A

Changelog: N/A

## Types of changes

## Checklist


t1m0thyj (approved):

packages/zowe-explorer/src/dataset/actions.ts
```diff
@@ -1577,15 +1577,26 @@ export async function saveFile(doc: vscode.TextDocument, datasetProvider: api.IZ
     const ending = doc.fileName.substring(start);
     const sesName = ending.substring(0, ending.indexOf(path.sep));
     const profile = Profiles.getInstance().loadNamedProfile(sesName);
+    const fileLabel = doc.fileName.split("/").slice(-1)[0];
```
traeok, line 1580: This line should also be split by `path.sep` instead of a single forward slash - on Windows, this split operation may fail to return the right path segments.

traeok (dismissed):
We added a similar fix in v2.12.1 for the USS tree, but I noticed that the logic for opening data sets also has the "path separation" issue. This could affect the way data sets are displayed in the editor for Windows users - I haven't tested this action extensively, but we might need a v2.12.2 to resolve this issue.

packages/zowe-explorer/src/dataset/actions.ts
rudyflores: Was this something introduced with this PR or existing and planning to address for `v2.12.2`? @traeok

traeok: I just found this issue during my review, but it's not a part of this PR's changes. We'll want this in the v2.12.2 release - I just realized this line is only used for the file label, so it shouldn't be a huge regression, but the editor tabs might look odd on Windows for v2.12.1.

We can merge a fix into maintenance, and then update this PR with the changes from the base branch.

rudyflores: I think we can bring in the changes from a possible `2.12.2` if we do so as a separate PR since we may want to bring these changes into our `main` branch if that's ok with you?

JillieBeanSim (approved):
LGTM! thanks @rudyflores

packages/zowe-explorer/src/dataset/actions.ts
t1m0thyj: Addressed this in https://github.com/zowe/vscode-extension-for-zowe/commit/eed67ec8c9945038a75f8b54a015fa995bddaac6

PR #2602, opened 2023-11-29
add ability to add profile to all trees

rudyflores:
## Proposed changes

Solves https://github.com/zowe/vscode-extension-for-zowe/issues/2603

## Sample of change

https://github.com/zowe/vscode-extension-for-zowe/assets/68666202/a5de5e70-e59c-4773-b38d-0fea615833d5

## Release Notes

Milestone: v2.13.0

Changelog: Added ability to add profile to all trees or the current tree

## Types of changes

## Checklist


JillieBeanSim (approved):
LGTM! thanks @rudyflores  for adding this capability for adding profiles across all trees

packages/zowe-explorer-api/src/tree/IZoweTree.ts
```diff
@@ -63,14 +64,27 @@ export interface IZoweTree<T> extends vscode.TreeDataProvider<T> {
...
+     * @param provider tree provider to add to, undefined will add for all
      */
-    addSession(sessionName?: string, type?: string): Promise<void>;
+    addSession(sessionName?: string, type?: string, provider?: IZoweTree<IZoweTreeNode>): Promise<void>;
+
+    /**
+     * Adds a single session to the tree
+     * @param profile the profile to add to the tree
+     */
+    addSingleSession(profile: imperative.IProfileLoaded): Promise<void>;
```
zFernand0, line 75: should this be changed to an optional? (and maybe make it required in V3?)

packages/zowe-explorer-api/src/tree/IZoweTree.ts
```diff
@@ -63,14 +64,27 @@ export interface IZoweTree<T> extends vscode.TreeDataProvider<T> {
...
     /**
      * Edit a session to the container
      * @param node This parameter identifies the node that needs to be called
      */
     editSession(node: IZoweTreeNode, zoweFileProvider: IZoweTree<IZoweTreeNode>): Promise<void>;
 
+    /**
+     * Get sessions from persistent object of provider
+     */
+    getSessions(): string[];
```
zFernand0, line 86: same as before.
I know it could be easily implemented, but this may force people to make a change in a non-major version update.

packages/zowe-explorer/src/utils/ProfileManagement.ts
```diff
@@ -264,44 +297,8 @@ export class ProfileManagement {
...
-        if (!selection) {
-            return;
-        }
-        const [all] = this.getPromptChangeForAllTreesOptions();
-        return selection.label === all.label;
-    }
-
     private static async handleHideProfiles(node: IZoweTreeNode): Promise<void> {
-        const shouldHideFromAllTrees = await this.handleChangeForAllTrees(node);
+        const shouldHideFromAllTrees = await this.handleChangeForAllTrees(node.getLabel().toString(), false);
```
zFernand0, line 301: Curious if we can default the `checkPresence` boolean to false and make it optional for these methods.
- handleChangeForAllTrees
- promptChangeForAllTrees

packages/zowe-explorer/src/job/ZosJobsProvider.ts
```diff
@@ -218,53 +217,45 @@ export class ZosJobsProvider extends ZoweTreeProvider implements IZoweTree<IZowe
...
-            const theProfile: imperative.IProfileLoaded = Profiles.getInstance().loadNamedProfile(sessionName);
-            if (theProfile) {
-                await this.addSingleSession(theProfile);
-            }
-            for (const node of this.mSessionNodes) {
-                const name = node.getProfileName();
-                if (name === theProfile.name) {
-                    await resetValidationSettings(node, setting);
-                }
+        await super.addSession(sessionName, profileType, provider);
```
zFernand0, line 222: Now I understand why we did this.
Makes our codebase a lot more straightforward and easier to follow.
However, aren't we putting a bit of a burden on extenders by requiring them to implement such methods?

I could be misunderstanding the refactoring altogether though. so please correct me if this is not the case.

zFernand0 (changes requested):
Changes LGTM! 😋 
I did leave a comment about making a few methods optional instead of required.

zFernand0:
I forgot to mention that the functionality works great! 
LGTM! 😋

packages/zowe-explorer-api/src/tree/IZoweTree.ts
rudyflores: This function already existed in the definitions for trees such as `DataSetTree.ts`, `USSTree.ts` and `ZosJobsProvider.ts`. The problem was that it was never formally defined in the interface for access to this function from its parent class `ZoweTreeProvider.ts`.

A good example of this is installing the FTP extension and attempting to perform the `Add profile` action for all trees or singular tree and this should not show an issue

Another thing to note we reference this in our code but we don't really use an extender provided function, this is already part of our existing code but needs to be added to the interface for access

packages/zowe-explorer-api/src/tree/IZoweTree.ts
rudyflores: See comment above :)

packages/zowe-explorer/src/job/ZosJobsProvider.ts
rudyflores: No new code needs to be implemented, to break this down in a easier to understand way the behavior is as follows:

Scenario where `profile` is added from jobs tree view
1. User is prompted for either all or current tree
2. User selects `all trees`
3. `ZosJobsProvider.addSession` is called
4. Previous function now calls parent function of `ZoweTreeProvider.addSession` (or `super.addSession`)
5. Since all was selected, we enter the `else` case and add for each provider using the newly added static class for accessing all tree providers
6. Same old code runs for adding a profile to a tree view (this code already existed, its just now being used differently since it was moved to new functions and cleaned up a bit) 
7. `addSingleSession` is called per tree provider (each one has it defined in their class)

rudyflores:
@zFernand0 testing with zftp profile might help determine whether anything new is needed here or if breaking change is introduced, notice how no new code was introduced in that extension but it seems to work :))

rudyflores:
One more thing, since we don't really expose these functions to extenders anyways i'm curious if maybe these just need to be functions that are called directly in the parent and remove `addSession` from the three tree providers and see what we can do about handling `addSingleSession` access from the parent without having to add this new function in the interface 🤔 sounds like a v3 thing to do though

packages/zowe-explorer/src/utils/ProfileManagement.ts
zFernand0: nothing in particular, other than to save some extra characters when calling the method 😋

JillieBeanSim (approved):

likhithanimma1 (approved):
Tested and working as expected. Thankyou @rudyflores for this new feature

JillieBeanSim:
Merging as it seems concerns were explained via comment.

PR #2685, opened 2024-01-24
Migrate to Webpack v5

### LINKED ISSUE #2214: Replace Webpack 4 with newer bundler
Microsoft has upgraded away from Webpack 4 in their VSCE samples so we are falling a bit behind now 🙂 https://github.com/microsoft/vscode-extension-samples/blob/main/webpack-sample/package.json

For the v3 release, we could investigate alternatives like Webpack 6 or Turbopack (used in the K8s repo).

rudyflores:
## Proposed changes

Fixes #2214 

Webpack v5 migration docs: https://webpack.js.org/migrate/5/

Migrated our outdated webpack v4 to the newest release of webpack v5, this has introduced some really good things for our repo, as well as things that we should all know about when reviewing this PR:

## Code Splitting and Minimization

Code splitting using vendor strategy was used for optimizing load times on our project, see here for reference on how that was achieved: https://blog.jakoblind.no/code-split-vendors-with-webpack-for-faster-load-speed/

Minimization of our code was also integrated into our bundle for Zowe Explorer, which now greatly reduces the `extension.js` file size using Webpack's built in minimzer using Terser and ESBuild: https://webpack.js.org/plugins/terser-webpack-plugin/

## Bundling for VS Code for web and VS Code Desktop Versions

VS Code has a recommended way of setting up initially the `webpack.config.js`, this repo now follows (mostly) those recommendations. A thing to note is that our repo currently uses `fs` which are node.js API's that can't be used in the web and cause issues with bundling. We didn't notice this before with v4 since we were already bundling as `node` instead of VS Code's recommended `webworker` target to support both platforms in one bundle. See more on this here: https://code.visualstudio.com/api/working-with-extensions/bundling-extension

## Future things to address

Webpack no longer resolves polyfills automatically like in v4, which means if we switch the target to `webworker` we need to add those polyfill resolutions through fallbacks: https://webpack.js.org/configuration/resolve/#resolvefallback

Even if the work above is done, we need to discuss the approach to handle `fs` being used across Zowe Explorer and Zowe Explorer API in order to bundle for web and support that too (if we choose to support it).

## Release Notes

Milestone: v3 pre-releases

Changelog: Migrated to Webpack v5

## Types of changes

## Checklist


JillieBeanSim (changes requested):
the changes built and packaged without errors but I get this when trying to launch the debugger and that fails
Screenshot 2024-01-26 at 4 19 09 PM

t1m0thyj:
> the changes built and packaged without errors but I get this when trying to launch the debugger and that fails Screenshot 2024-01-26 at 4 19 09 PM

This happened to me recently too but I don't think its a problem with this PR - if you switched Node versions with `nvm` recently, might need to reinstall pnpm globally (`npm install -g pnpm`)?

JillieBeanSim (dismissed):
remembering to reinstall pnpm after each node update will have to be something I remember to do now lol don't even remember updating node since last time testing next. Thanks @t1m0thyj 
LGTM! thanks for the updates @rudyflores

traeok (changes requested):
Thanks for all your work on this @rudyflores! I do have a small request:

I ran into an issue that isn't directly related to this PR, but I realized I was unable to run the watch script after downgrading my Node version to v18.

On line 24 in `.vscode/tasks.json`, the type is set to "process" instead of "shell" - because of this, the task fails on Windows with OS error 193 because it tries to execute the Linux/macOS binary (which is not a valid Win32 application).

Can you please update this to "shell"? This should resolve the watch script failures on Windows.

rudyflores:
@traeok good catch! Should be all fixed now :)

traeok (dismissed):
LGTM, thanks for the tech. currency and for quickly resolving the `watch` task 😁

traeok:
I noticed after approving that the changelog headers have the `next` snapshot as the header instead of "Recent Changes" - I think that will resolve the failing changelog stage

traeok (approved):

JillieBeanSim (approved):
LGTM! thanks @rudyflores for this update

PR #2686, opened 2024-01-24
build(api): Enhance watch task to trigger for zowe-explorer-api changes

t1m0thyj:
**What It Does**
Updates our tsconfig and webpack config so that when the watch task is running, changes made to "zowe-explorer-api" package will trigger a new incremental build, and a complete rebuild of the project is no longer required.

Also fixes webpack config to load source maps correctly so that breakpoints can be set within "zowe-explorer-api" code.

See [here](https://medium.com/@nickexcell/using-typescript-project-references-with-ts-loader-and-webpack-part-1-5d0c3cd7c603) for more details about how TS project references can be used with Webpack.

**How to Test**
* Launch the debug task to run ZE or the FTP extension which starts a watch task in the terminal
* Make a change to source code inside the "zowe-explorer-api" package and save the file
* The watch task should automatically trigger to rebuild with code changes made to ZE API
* Set a breakpoint in a TS file in the "zowe-explorer-api" package and make sure it triggers


packages/zowe-explorer/webpack.config.js
```diff
@@ -50,6 +50,9 @@ const config = {
     resolve: {
         // support reading TypeScript and JavaScript files, 📖 -> https://github.com/TypeStrong/ts-loader
         extensions: [".ts", ".js"],
+        alias: {
+            "@zowe/zowe-explorer-api$": path.resolve(__dirname, "..", "zowe-explorer-api/src"),
```
rudyflores, line 54: Does this alias allow the incremental builds or just including in bundle when resolving? i'm curious about what this does :)

rudyflores (approved):
Thanks for this change @t1m0thyj ! This will improve our debugging/development speeds! Perhaps we could look into how this will migrate towards `next` after the migration to webpack v5 is merged :)

packages/zowe-explorer/webpack.config.js
t1m0thyj: This isn't needed for the incremental builds, but is one of 2 changes I had to make for breakpoints to work:
* The `resolve.alias` tells Webpack to reference TS files rather than JS files for any imports from the "@zowe/zowe-explorer-api" package, which allows us to set breakpoints directly in the TS files 🙂 
* The `devtoolModuleFilenameTemplate` needs to use absolute paths, because Webpack doesn't evaluate relative paths correctly for source files outside of the directory where webpack.config.js resides

traeok (approved):
This works great, thanks for looking into this @t1m0thyj!

JillieBeanSim (approved):
this is a great addition for development! thanks @t1m0thyj

Issue #2111, opened 2023-02-06
When the default profile has apiml token authentication, other profiles inaccessible


pablocarle:
**Describe the bug**

When the default profile has a base with apiml token authentication (`zowe auth login apiml`), other profiles in vscode start complaining that they are inactive and require authentication. If I try to authenticate, the extension displays an error: `This profile does not support token authentication` even if this profile was untouched.

**To Reproduce**

1. Have two profiles working, for different lpars, for example.
2. Update one of them to authenticate with an apiml instance (`zowe auth login apiml`) + update port and basePath on this profile. Set it to default profile (in base, zosmf, ssh, etc)
3. Try to use the profile that was not modified in vscode.

**Expected behavior**

If both profiles appear in vscode extension, both should work, one with token in apiml and the unchanged one with its authentication.

**Desktop (please complete the following information):**

- OS: MacOS
- Zowe Explorer Version: current (2.6.0)
- (Optional) Zowe CLI Version: 7.10.2
- (Optional) Are you using Secure Credential Store? Yes, default in v2

PR #2735, opened 2024-02-21
Restore accessibility to all profiles when default profile has APIML token authentication

likhithanimma1:
## Proposed changes

   Solves #2111 
   Fixes the issue of accessibility of all the profiles when the default profile is using the APIML token authentication

## Release Notes

Milestone:

Changelog:

## Types of changes

## Checklist

### DIFF
--- a/packages/zowe-explorer-api/src/profiles/ProfilesCache.ts
+++ b/packages/zowe-explorer-api/src/profiles/ProfilesCache.ts
@@ -542,7 +542,9 @@ export class ProfilesCache {
         return ((baseProfile?.profile?.host || baseProfile?.profile?.port) &&
             profile?.profile?.host &&
             profile?.profile?.port &&
-            (baseProfile?.profile.host !== profile?.profile.host || baseProfile?.profile.port !== profile?.profile.port ||(profile?.profile.user && profile?.profile.password)) &&
+            (baseProfile?.profile.host !== profile?.profile.host ||
+                baseProfile?.profile.port !== profile?.profile.port ||
+                (profile?.profile.user && profile?.profile.password)) &&
             profile?.profile.tokenType?.startsWith(zowe.imperative.SessConstants.TOKEN_TYPE_APIML)) as boolean;
     }
 
--- a/packages/zowe-explorer-api/src/profiles/ProfilesCache.ts
+++ b/packages/zowe-explorer-api/src/profiles/ProfilesCache.ts
@@ -542,7 +542,7 @@ export class ProfilesCache {
         return ((baseProfile?.profile?.host || baseProfile?.profile?.port) &&
             profile?.profile?.host &&
             profile?.profile?.port &&
-            (baseProfile?.profile.host !== profile?.profile.host || (profile?.profile.user && profile?.profile.password))&&
+            (baseProfile?.profile.host !== profile?.profile.host || (profile?.profile.user && profile?.profile.password)) &&
             profile?.profile.tokenType?.startsWith(zowe.imperative.SessConstants.TOKEN_TYPE_APIML)) as boolean;
     }
 
--- a/packages/zowe-explorer-api/src/profiles/ProfilesCache.ts
+++ b/packages/zowe-explorer-api/src/profiles/ProfilesCache.ts
@@ -539,11 +539,10 @@ export class ProfilesCache {
     }
 
     private shouldRemoveTokenFromProfile(profile: zowe.imperative.IProfileLoaded, baseProfile: zowe.imperative.IProfileLoaded): boolean {
-        return (baseProfile?.profile?.host &&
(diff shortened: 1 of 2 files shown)


packages/zowe-explorer-api/src/profiles/ProfilesCache.ts
```diff
@@ -539,11 +539,10 @@ export class ProfilesCache {
...
 
     private shouldRemoveTokenFromProfile(profile: zowe.imperative.IProfileLoaded, baseProfile: zowe.imperative.IProfileLoaded): boolean {
-        return (baseProfile?.profile?.host &&
-            baseProfile?.profile?.port &&
+        return ((baseProfile?.profile?.host || baseProfile?.profile?.port) &&
             profile?.profile?.host &&
             profile?.profile?.port &&
-            (baseProfile?.profile.host !== profile?.profile.host || baseProfile?.profile.port !== profile?.profile.port) &&
+            (baseProfile?.profile.host !== profile?.profile.host || (profile?.profile.user && profile?.profile.password)) &&
             profile?.profile.tokenType?.startsWith(zowe.imperative.SessConstants.TOKEN_TYPE_APIML)) as boolean;
```
traeok, line 546: There are a lot of factors in this boolean logic here 😅 

Would it make sense to move this line to the top so we return false before doing the other checks? If the profile doesn't have a token type, we can immediately return false.

JillieBeanSim:
should this be included in 2.15.0?

JillieBeanSim (approved):

zFernand0:
> should this be included in 2.15.0?

we could include it in 2.15.0

---

Also, I haven't approved the PR because I can still see the issue (when tested in the vNext version of this PR).
I believe there was a bit of info missing in the reproduction steps.
Where are the credentials coming from? (base or service)

- If the credentials were initially in the base profile, and you did a `zowe auth login apiml`, 
  - You will get a new base profile created (with the word `_apiml` as a suffix)
  - This new `base_apiml` profile gets assigned as the default
  - This will cause other service profiles to break (because there are no credentials in the default base profile anymore)
  - I was able to reproduce that error in this PR

If you want to avoid this from happening:
 - You must add the credentials on the service profile (or somewhere in the nested structure).
 - This works fine in the version out in the marketplace

packages/zowe-explorer-api/src/profiles/ProfilesCache.ts
```diff
@@ -539,11 +539,10 @@ export class ProfilesCache {
     }
 
     private shouldRemoveTokenFromProfile(profile: zowe.imperative.IProfileLoaded, baseProfile: zowe.imperative.IProfileLoaded): boolean {
-        return (baseProfile?.profile?.host &&
-            baseProfile?.profile?.port &&
+        return ((baseProfile?.profile?.host || baseProfile?.profile?.port) &&
             profile?.profile?.host &&
             profile?.profile?.port &&
-            (baseProfile?.profile.host !== profile?.profile.host || baseProfile?.profile.port !== profile?.profile.port) &&
```
traeok, line 546: Can we add this check back for comparing the base profile's port with the port of the profile that was passed in? I like the added check for user and password on the profile, but I think keeping the port check will help narrow the opportunity for error until something more comprehensive is incorporated into Zowe CLI/Imperative.

packages/zowe-explorer-api/src/profiles/ProfilesCache.ts
```diff
@@ -539,11 +539,10 @@ export class ProfilesCache {
...
+            (baseProfile?.profile.host !== profile?.profile.host || (profile?.profile.user && profile?.profile.password)) &&
```
zFernand0, line 545: Also, it might be good to keep the `port` check as well 😋

traeok (changes requested):

traeok:
> If you want to avoid this from happening:
> 
> * You must add the credentials on the service profile (or somewhere in the nested structure).
> * This works fine in the version out in the marketplace

I had tested this PR under the assumption that credentials were stored on the service profile. However, the details for #2111 do not mention whether credentials were stored on the service or base profile.

Considering that its unclear of what authentication method to use for a profile in that scenario, I think it would be best to accept this fix until a more comprehensive solution is introduced on the Zowe CLI/Imperative side (such as profile linking). However, I think we should also keep the port check until this solution is introduced.

packages/zowe-explorer-api/src/profiles/ProfilesCache.ts
t1m0thyj: Agreed, it might be good to keep the `port` check here as well as on line 542

traeok (approved):
LGTM, thanks @likhithanimma1 for the bugfix! 🎉

JillieBeanSim (approved):

PR #2777, opened 2024-03-12
feat(v3): FileSystemProvider implementation for Data Sets, USS and Jobs

traeok:
## Proposed changes

Check out the new page on ZE wiki for details on FSP infrastructure: https://github.com/zowe/zowe-explorer-vscode/wiki/FileSystemProvider

- [x] Implement the FileSystemProvider and add tests
  - [x] Base provider (for extenders to leverage to create their own providers 😋)
  - [x] Data Sets tree
  - [x] USS tree
    - [x] Drag and drop (same LPAR) 
    - [x] Drag and drop (same LPAR)
    - [x] Drag and drop (different LPARs) - _stretch goal_
      - [x] Files
      - [x] Folders
  - [x] Jobs tree
- [x] Implement the `onResourceChanged` function to allow extenders to **watch FileSystemProvider changes**
  - Extenders can add event listeners to each one of these events returned by the command.
---
**Misc. changes:**
- [x] Added a `zowe.placeholderCommand` command so that we can disable items in the tree view 😋 
  - this is helpful for the "no data sets found" and similar nodes that we create to give information
  - they will be greyed out now to differentiate between an "actionable node" and a placeholder/disabled one
- [x] Added a `MockedProperty` class to clean up mocked properties defined using `Object.defineProperty`
- [x] Added confirmation dialogs for Pulling from Mainframe and changing encoding if the contents are unsaved in the editor

## Release Notes

Milestone:

Changelog:

## Types of changes

## Checklist


anaxceron (approved):

packages/zowe-explorer-api/__mocks__/mockUtils.ts
zFernand0, line 1: Curious if it makes sense to try and reuse some of these mocks.

I know it may be required for the `__mocks__` to be at the same level as the project's `jest` configuration, but we could trick `jest` by just using symlinks? 😋

packages/zowe-explorer-api/src/extend/MainframeInteraction.ts
```diff
@@ -119,13 +119,23 @@ export namespace MainframeInteraction {
          * @param {string} ussFilePath
          * @param {zosfiles.IDownloadOptions} options
          */
-        getContents(ussFilePath: string, options: zosfiles.IDownloadOptions): Promise<zosfiles.IZosFilesResponse>;
+        getContents(ussFilePath: string, options: zosfiles.IDownloadSingleOptions): Promise<zosfiles.IZosFilesResponse>;
```
zFernand0, line 122: should we add a changelog item for this breaking change?

I believe this could impact some of the ongoing work happening in:
- #2783 

However, this PR should be merged first 😋

packages/zowe-explorer-api/src/extend/MainframeInteraction.ts
```diff
@@ -119,13 +119,23 @@ export namespace MainframeInteraction {
...
 
         /**
          * Uploads a given buffer as the contents of a file on USS.
          */
         uploadFromBuffer(buffer: Buffer, filePath: string, options?: zosfiles.IUploadOptions): Promise<zosfiles.IZosFilesResponse>;
 
+        /**
+         * Uploads a given buffer as the contents of a file on USS.
+         */
+        uploadFromBuffer(buffer: Buffer, filePath: string, options?: zosfiles.IUploadOptions): Promise<string | zosfiles.IZosFilesResponse>;
```
zFernand0, line 132: is it necessary to keep both instances?
Or could we just note the possible breaking change of the return type being a `string` too 😋

packages/zowe-explorer-api/src/extend/MainframeInteraction.ts
```diff
@@ -506,7 +516,7 @@ export namespace MainframeInteraction {
          * @param {string} command
          * @param {string} cwd
          * @param {boolean} flag
-         * @returns {string>}
+         * @returns {Promise<string>}
```
zFernand0, line 519: Should we update the other `issueXyzCommand` interfaces above?

I believe they all return a promise 😋

packages/zowe-explorer-api/src/fs/FsWatcher.ts
```diff
@@ -0,0 +1,89 @@
...
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Copyright Contributors to the Zowe Project.
+ *
+ */
+
+import { Disposable as VSDisposable, FileSystemWatcher, Uri, workspace } from "vscode";
+
+interface CoreWatchers {
```
zFernand0, line 14: Will this interface be helpful to export for exenters?
That way they know which `filewatchers` are available as part of the ZoweFsWatcher.

No strong feelings either way since the intention of the `watchers` is for internal use only (i.e. private with no getters)

packages/zowe-explorer-api/src/fs/FsWatcher.ts
```diff
@@ -0,0 +1,89 @@
...
+    uss: FileSystemWatcher;
+}
+
+export enum ZoweScheme {
+    DS = "zowe-ds",
+    Jobs = "zowe-jobs",
+    USS = "zowe-uss",
+}
+
+export class ZoweFsWatcher {
```
zFernand0, line 26: I may be reviewing this a bit early or too late... but don't we want to add all of these to the Changelog?

packages/zowe-explorer/__tests__/__unit__/ZoweNode.unit.test.ts
```diff
@@ -78,139 +79,139 @@ describe("Unit Tests (Jest)", () => {
...
-            label: "root",
-            collapsibleState: vscode.TreeItemCollapsibleState.Collapsed,
-            session,
-            profile: profileOne,
-        });
-        rootNode.dirty = true;
-        rootNode.contextValue = globals.DS_SESSION_CONTEXT;
-        rootNode.pattern = "SAMPLE, SAMPLE.PUBLIC, SAMPLE";
-        let rootChildren = await rootNode.getChildren();
+    // it("Testing that getChildren returns the correct Thenable<ZoweDatasetNode[]>", async () => {
```
zFernand0, line 82: Do we plan to implement these tests again?

I may run into these commented out tests in some other form later during the review process, but just wanted to raise the question in case we just forgot about them 😋

packages/zowe-explorer/__tests__/__unit__/dataset/DatasetTree.unit.test.ts
```diff
@@ -340,149 +344,152 @@ describe("Dataset Tree Unit Tests - Function getChildren", () => {
...
-
-        const children = await testTree.getChildren(testTree.mSessionNodes[1]);
-        expect(children.map((c) => c.label)).toEqual(sampleChildren.map((c) => c.label));
-        expect(children).toEqual(sampleChildren);
-        expect(spyOnDataSet).toHaveBeenCalled();
-        expect(spyOnDataSetsMatchingPattern).not.toHaveBeenCalled();
-        spyOnDataSet.mockRestore();
-        spyOnDataSetsMatchingPattern.mockRestore();
-    });
+    // TODO: FIXME
```
zFernand0, line 347: similar to a previous comment about skipped/commented tests 😋

packages/zowe-explorer/__tests__/__unit__/shared/init.unit.test.ts
```diff
@@ -145,58 +143,26 @@ describe("Test src/shared/extension", () => {
...
-                                fileProvider: "uss",
-                                savedFile: { fileName: "USS_DIR", isDirty: true, uri: vscode.Uri.parse("") },
-                                uploadRequest: saveUSSFile,
-                            },
-                        ],
-                    },
-                ],
-            },
             {
                 name: "onDidSaveTextDocument:4",
```
zFernand0, line 147: I know this testing scheme was far from ideal even though it helped simplify things at the time.
However, now that we have removed a few tests (`it`s) from the list could we update this to `onDidSaveTextDocument:2`

Sorry for being picky 😢

packages/zowe-explorer/__tests__/__unit__/uss/ZoweUSSNode.unit.test.ts
zFernand0, line 1: Seems like we were able to reduce the amount of test duplication. 
I think this is great as long as we are not dropping coverage too much 😋

zFernand0 (commented):
I'm trying to review this in multiple rounds.
so far I love what I see 😋

packages/zowe-explorer-api/src/extend/MainframeInteraction.ts
traeok: Nope - sorry, this was supposed to be replaced during a merge conflict and I didn't catch it.
Addressed in b5d9d3a1

packages/zowe-explorer-api/src/extend/MainframeInteraction.ts
traeok: Probably, I just noticed this while adding the `uploadFromBuffer` definition - addressed in b5d9d3a1 😋

packages/zowe-explorer-api/src/fs/FsWatcher.ts
traeok: Thanks for catching this, addressed in 755ba5f0

packages/zowe-explorer/__tests__/__unit__/uss/ZoweUSSNode.unit.test.ts
traeok: Project coverage is roughly the same as the `main` branch (93.27%) and slightly under the `next` branch (~93.30%).
I still need to add some coverage for the `handleDrop` function - this should get the coverage to match or even exceed the next branch 😋

packages/zowe-explorer/__tests__/__unit__/dataset/DatasetTree.unit.test.ts
traeok: I chose to comment out this test because the test setup had oversights - the test is not ideal because it creates a circular structure for the parent node, which will not occur in normal ZE scenarios. I can try to resolve this, but since the coverage is roughly the same and the test case is unclear anyway ("checking common run of function"), I'll likely remove it unless there are any objections. 🙂

packages/zowe-explorer/__tests__/__unit__/ZoweNode.unit.test.ts
traeok: These Data Set tests also had issues with circular structures. I don't mind writing new tests for this function though.

The USS node has coverage because we now use a function in the provider to build the children nodes. I will look into a similar implementation for Data Set nodes.

packages/zowe-explorer/__tests__/__unit__/shared/init.unit.test.ts
traeok: No worries, sorry I should have caught this while modifying these tests - addressed in cc4c0c40

packages/zowe-explorer-api/src/fs/FsWatcher.ts
traeok: This was added for stronger typing for the `watchers` variable in `ZoweFsWatcher`.

We can keep this internal since only `ZoweFsWatcher` uses it, but I've adjusted it to be a type.
Addressed in 0ca207b8

traeok:
> Hey @traeok thanks for working on this huge undertaking. Are we going to be wanting the WIP items included in this PR work or in another PR? If this one, should the PR be in draft state until ready?

The WIP items are stretch goals - maybe I should pull those out into a separate PR? That might make it easier from a review standpoint 🙂

packages/zowe-explorer/__tests__/__unit__/uss/ZoweUSSNode.unit.test.ts
traeok: Okay, project coverage is now sitting above the `next` branch at ~93.34% or so. 😋

zFernand0:
Lot's of commits 🤯 
Please be aware that after 200 commits, some of the GH endpoints stop sending all commit history and start paging them.
This may have changed since the last time we tested our automation workflows/scripts.

Not suggesting you to remove history or stop making small commits.
I do believe there is value in having small units of work being pushed to the branches 😋 

Just be aware that some weird behavior may start happening the closer we get to (and/or when we go past) the 200 commits

packages/zowe-explorer-api/__mocks__/mockUtils.ts
zFernand0: Created the issue below to capture this idea:
- https://github.com/zowe/zowe-explorer-vscode/issues/2790

packages/zowe-explorer/__tests__/__unit__/ZoweNode.unit.test.ts
zFernand0: I think you've done a pretty good job at keeping up with code-coverage.
Feel free to remove any commented out tests once you are done 😋

packages/zowe-explorer/__tests__/__unit__/dataset/DatasetTree.unit.test.ts
zFernand0: I think you've done a pretty good job at keeping up with code-coverage.
Feel free to remove any commented out tests once you are done 😋

packages/zowe-explorer-api/src/fs/types/abstract.ts
```diff
@@ -0,0 +1,149 @@
...
+
+export interface IFileSystemEntry extends vscode.FileStat {
+    name: string;
+    metadata: EntryMetadata;
+    type: vscode.FileType;
+    wasAccessed: boolean;
+    data?: Uint8Array;
+}
+
+export class FileEntry implements IFileSystemEntry {
```
zFernand0, line 82: Curious why we are implementing the interface vs converting `IFileSystemEntry` into a class and extending it.
If it's possible, I think it will reduce having to create/define the same properties in `FileEntry` (and `DirEntry`)

The reason I ask is because I don't see any abstract methods implemented in either of the classes.

packages/zowe-explorer-api/src/fs/types/datasets.ts
```diff
@@ -0,0 +1,57 @@
...
+
+import { DirEntry, EntryMetadata, FileEntry } from "../types";
+import { IProfileLoaded } from "@zowe/imperative";
+import { FileType } from "vscode";
+
+export class DsEntry extends FileEntry {
+    public name: string;
+    public metadata: DsEntryMetadata;
+    public type: FileType;
+    public wasAccessed: boolean;
```
zFernand0, line 20: I may be missing something very important, but do we need to re-difine every property defined in the extended class?
I get the metadata is being overloaded, but most/all other properties seem exactly the same 😋

packages/zowe-explorer-api/src/fs/types/jobs.ts
```diff
@@ -0,0 +1,48 @@
...
+    searchId: string;
+    owner: string;
+    prefix: string;
+    status: string;
+};
+
+export class SpoolEntry extends FileEntry {
+    public name: string;
+    public type: FileType;
+    public wasAccessed: boolean;
```
zFernand0, line 26: The comment about duplication/re-defining properties may apply to various classes across the board

packages/zowe-explorer-api/src/fs/types/uss.ts
```diff
@@ -0,0 +1,40 @@
...
+ */
+
+import * as vscode from "vscode";
+import { ConflictData, DirEntry, EntryMetadata, FileEntry } from ".";
+
+export class UssFile extends FileEntry {
+    public name: string;
+    public metadata: EntryMetadata;
+    public type: vscode.FileType;
+    public wasAccessed: boolean;
```
zFernand0, line 19: same as previous comment

packages/zowe-explorer-api/src/fs/utils/abstract.ts
```diff
@@ -0,0 +1,87 @@
...
+
+    // either the document doesn't exist or it isn't dirty
+    return true;
+}
+
+export function isDirectoryEntry(entry: any): entry is DirEntry {
+    return entry != null && entry["type"] === vscode.FileType.Directory;
+}
+
+export function isFileEntry(entry: any): entry is FileEntry {
```
zFernand0, line 81: Thanks for helping me understand type guards offline. I seriously need to keep up with TS development.

zFernand0 (commented):
Maybe 2 more rounds of reviews coming up 😋

packages/zowe-explorer-api/src/fs/types/abstract.ts
traeok: I'm hesitant to use a common class over an interface because a common class will introduce a layer of inheritance for the entries in the provider. With an interface, inheritance is avoided and it is just leveraged for type checking.

Another scenario that I thought of is that if we change the base class, we have to update the implementation for all subclasses. With an interface, we can add optional properties in the future that do not have to be defined by the classes that implement it.

Essentially, I think either option (base class or interface) would work here, so I'm willing to reconsider - but all in all, this interface is mostly for adding type checks without the overhead of a base class. If we see a benefit in changing this to a base class though, I'd be happy to do so 😁

packages/zowe-explorer-api/src/fs/types/datasets.ts
```diff
@@ -0,0 +1,57 @@
...
+export class DsEntryMetadata implements EntryMetadata {
+    public profile: IProfileLoaded;
+    public path: string;
+
+    public constructor(metadata: EntryMetadata) {
+        this.profile = metadata.profile;
+        this.path = metadata.path;
+    }
+
+    public get dsname(): string {
```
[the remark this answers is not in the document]
traeok: I think it was initially named as such to match the key in the API response, but I think that this makes more sense.
Renamed in commit 3224883e

packages/zowe-explorer-api/src/fs/types/abstract.ts
zFernand0: Thanks for providing the reasoning behind why we went with an interface instead of the possible extra overhead of the base class. I don't have a strong opinion one way or the other since they do seem comparable (in terms of memory footprint and other aspects). That being said, I still like this implementation very much. 😋 

No need to change it to a base class unless others have a strong opinion in favor of it 😋 
Resolving!

packages/zowe-explorer-api/src/fs/utils/jobs.ts
```diff
@@ -0,0 +1,27 @@
...
+
+import { IJobFile } from "@zowe/zos-jobs-for-zowe-sdk";
+import { IFileSystemEntry, JobEntry, SpoolEntry } from "../types";
+import { FileType } from "vscode";
+
+export function isJobEntry(entry: IFileSystemEntry): entry is JobEntry {
+    return entry != null && entry.type == FileType.Directory;
+}
+
+export function isSpoolEntry(entry: IFileSystemEntry): entry is SpoolEntry {
```
zFernand0, line 20: I'm really starting to see the benefits in using the typeguard. Not saying that we should restructure the code to prefer them over TS Generics like with the `IZoweTree` (since they may not be 1-1 replaceable due to some obscure edge case). However, maybe we should use them more often going forward to avoid unnecessary double/triple type-checks 😋

packages/zowe-explorer-api/src/profiles/ZoweExplorerZosmfApi.ts
```diff
@@ -126,14 +126,21 @@ export namespace ZoweExplorerZosmf {
...
-        public getContents(inputFilePath: string, options: zosfiles.IDownloadOptions): Promise<zosfiles.IZosFilesResponse> {
+        public getContents(inputFilePath: string, options: zosfiles.IDownloadSingleOptions): Promise<zosfiles.IZosFilesResponse> {
             return zosfiles.Download.ussFile(this.getSession(), inputFilePath, options);
         }
 
         public copy(outputPath: string, options?: Omit<object, "request">): Promise<Buffer> {
             return zosfiles.Utilities.putUSSPayload(this.getSession(), outputPath, { ...(options ?? {}), request: "copy" });
         }
 
+        public async move(oldPath: string, newPath: string): Promise<void> {
```
zFernand0, line 137: I know we added the breaking change interface-updates for the `MainframeInteraction.IUss`. But I couldn't find an entry in this PR's ZE-API's Changelog about adding the required `move` to `MainframeInteraction.IUss` or changing the signature of `getContents` in `MainframeInteraction.IMvs`.

Again, I could be missing these 😋

packages/zowe-explorer-api/src/tree/IZoweTree.ts
```diff
@@ -134,6 +134,11 @@ export interface IZoweTree<T> extends vscode.TreeDataProvider<T> {
      * @param favorite Node to refresh
      */
     refreshElement(node: IZoweTreeNode): void;
+    /**
+     * Signals that node data has changed in the tree view
+     * @param element to pass to event listener callback
+     */
+    nodeDataChanged(node: IZoweTreeNode): void;
```
zFernand0, line 141: Should this be added to the changelog?
As well as any required methods introduced by the extension of `vscode.TreeDragAndDropController` (if any)

packages/zowe-explorer-api/src/tree/IZoweTreeNode.ts
```diff
@@ -37,32 +37,11 @@ export type ZosEncoding = TextEncoding | BinaryEncoding | OtherEncoding;
  * @export
  * @interface IZoweTreeNode
  */
-export interface IZoweTreeNode {
-    /**
-     * The icon path or [ThemeIcon](#ThemeIcon) for the tree item.
-     */
-    iconPath?: string | vscode.Uri | { light: string | vscode.Uri; dark: string | vscode.Uri } | vscode.ThemeIcon;
+export interface IZoweTreeNode extends vscode.TreeItem {
```
zFernand0, line 40: Should we list all the properties removed from this interface in the changelog?
As well as those that just changed in their types due to the extension of the `vscodeTreeItem` 
For example:
- `tooltip: string | vscode.MarkdownString | undefined;` changed to `tooltip: string | vscode.MarkdownString;`

packages/zowe-explorer-api/src/utils/index.ts
```diff
@@ -9,5 +9,18 @@
...
 export * from "./Poller";
 export * from "./FileManagement";
+
+/**
+ * Getter to check dirty flag for nodes opened in the editor.
+ *
+ * NOTE: Only works for nodes that use resource URIs (see the `resourceUri` variable in IZoweTreeNode)
+ * @returns {boolean} whether the URI is open in the editor and unsaved
+ */
+export function isNodeInEditor(node: IZoweTreeNode): boolean {
```
zFernand0, line 24: This is also not in the changelog 😅 
If we don't plan to add it as a small enhancement in the changelog, should we make it `internal`

packages/zowe-explorer/tsconfig.json
```diff
@@ -4,13 +4,13 @@
     "target": "es6",
     "outDir": "out",
     "jsx": "react",
-    "lib": ["es6", "dom"],
```
zFernand0, line 7: I assume we will bring this back when we make an attempt to support `vscode.dev` / Web

packages/zowe-explorer/src/uss/utils.ts
```diff
@@ -78,19 +70,19 @@ export function disposeClipboardContents(): void {
...
-        } else {
-            node.setEncoding(taggedEncoding !== "untagged" ? { kind: "other", codepage: taggedEncoding } : undefined);
+            await node.setEncoding({ kind: "binary" });
+        } else if (taggedEncoding !== "untagged") {
+            await node.setEncoding({ kind: "other", codepage: taggedEncoding });
         }
     } else {
         const isBinary = await ussApi.isFileTagBinOrAscii(node.fullPath);
-        node.setEncoding(isBinary ? { kind: "binary" } : undefined);
+        await node.setEncoding(isBinary ? { kind: "binary" } : undefined);
```
zFernand0, line 86: It seems that we have special behavior defined for `encoding === null` based on this:
- `if (node.getEncoding() !== undefined) { return }`

IIRC, it means that we want to go with the default z/OS (MF/RSE/FTP) encoding.
> Thanks @t1m0thyj for confirming, I won't forget it again ([**_source_**](https://github.com/zowe/zowe-explorer-vscode/blob/206bf062390e8b58cd6364e485a4fe21ce4ba1f5/packages/zowe-explorer-api/src/tree/IZoweTreeNode.ts#L182-L187))

Shouldn't we set the encoding to `null` instead of `undefined` if it's not tagged and is not binary?
Otherwise, I can see how we could be calling the `autoDetectEncoding` method in an infinite loop trying to determine the encoding but never getting anywhere because we are the ones setting it to `undefined` 😋 

Please let me know if this makes no sense 😅

packages/zowe-explorer/src/utils/LocalFileManagement.ts
```diff
@@ -10,119 +10,50 @@
...
-        const downloadEtag = downloadResponse?.apiResponse?.etag;
-        if (node && downloadEtag !== node.getEtag()) {
-            node.setEtag(downloadEtag);
-        }
+    public static setCompareSelection(val: boolean): void {
+        LocalFileManagement.fileSelectedToCompare = val;
+        vscode.commands.executeCommand("setContext", "zowe.compareFileStarted", val);
+    }
+
+    public static reset(): void {
```
zFernand0, line 30: Should we rename this to be more specific as to what we are resetting? (maybe `resetCompareSelection`) 😋

packages/zowe-explorer/src/job/init.ts
```diff
@@ -9,29 +9,36 @@
...
+        vscode.commands.registerCommand("zowe.onJobChanged", (): vscode.Event<vscode.FileChangeEvent[]> => JobFSProvider.instance.onDidChangeFile)
+    );
+
     const jobsProvider = await createJobsTree(globals.LOG);
     if (jobsProvider == null) {
         return null;
     }
 
-    context.subscriptions.push(
-        vscode.commands.registerCommand("zowe.jobs.zosJobsOpenspool", (session, spoolNode) => jobActions.getSpoolContent(session, spoolNode))
```
zFernand0, line 33: Do we need to notify end-users that this command went away? (i.e. add this to the ZE changelog)

packages/zowe-explorer/src/globals.ts
```diff
@@ -23,7 +23,7 @@ export let USS_DIR: string;
 export let DS_DIR: string;
 export let CONFIG_PATH; // set during activate
 export let LOG: imperative.Logger;
-export const COMMAND_COUNT = 121;
+export const COMMAND_COUNT = 125;
```
zFernand0, line 26: Quite a few commands were added to the `package.json#contributes` property, but we only went up by 4.
It would seem that some commands are not being registered even though they might still be defined in the package.json file.

If the commands that are going away were exposed to end-users, don't we want to add something in the changelog about them going away?

zFernand0 (changes requested):
Alright, I'm down to the last 10 or so files.
only one more round 😋 

Marking this one as requested-changes because of some changelog related concerns/questions 😋

packages/zowe-explorer-api/__tests__/__unit__/fs/utils/abstract.unit.test.ts
```diff
@@ -0,0 +1,97 @@
...
+    });
+});
+
+describe("isDirectoryEntry", () => {
+    it("returns true if value is a DirEntry", () => {
+        const dirEntry = new DirEntry("testFolder");
+        expect(isDirectoryEntry(dirEntry)).toBe(true);
+    });
+
+    it("returns false if value is not a SpoolEntry", () => {
```
t1m0thyj, line 69: SpoolEntry -> DirEntry

packages/zowe-explorer-api/src/fs/utils/datasets.ts
```diff
@@ -0,0 +1,25 @@
...
+ *
+ * Copyright Contributors to the Zowe Project.
+ *
+ */
+
+import { IFileSystemEntry, DsEntry, MemberEntry, PdsEntry } from "../";
+import { FileType } from "vscode";
+
+export function isDsEntry(entry: IFileSystemEntry): entry is DsEntry {
+    return entry != null && entry.type == FileType.File;
```
t1m0thyj, line 16: I may be reading this wrong, but wouldn't this return true for a USS file entry too?

packages/zowe-explorer-api/src/fs/utils/jobs.ts
```diff
@@ -0,0 +1,27 @@
...
+ * Copyright Contributors to the Zowe Project.
+ *
+ */
+
+import { IJobFile } from "@zowe/zos-jobs-for-zowe-sdk";
+import { IFileSystemEntry, JobEntry, SpoolEntry } from "../types";
+import { FileType } from "vscode";
+
+export function isJobEntry(entry: IFileSystemEntry): entry is JobEntry {
+    return entry != null && entry.type == FileType.Directory;
```
t1m0thyj, line 17: Similar question to above - wouldn't this return true for a USS dir entry too?

t1m0thyj (commented):
Looks pretty good, thanks for undertaking this massive effort @traeok!

Haven't reviewed all the files yet, but left my first round of questions 🙂

packages/zowe-explorer/src/globals.ts
traeok: I added 3 developer commands - `zowe.onDatasetChanged, zowe.onUssChanged, zowe.onJobChanged`
- these are not added to the `contributes#commands` section of `package.json` because they don't have a title, and we don't want them in the command palette
- I noticed that I did set them to explicitly never show in the `commandPalette` section, but I don't believe that's necessary since they are not added to the first section I mentioned. Removed in upcoming commit

The other 3 commands are `zowe.diff.useLocalContent, zowe.diff.useRemoteContent, zowe.placeholderCommand`. I've added all of these to the `expectedCommands` array and those 3 are registered in that `contributes` section. 

I removed the `zowe.jobs.zosJobsOpenSpool` and `zowe.ds.ZoweNode.openPS` commands, so we went down by 2 - that leaves us at 125 commands 😋 

In an upcoming commit, I added changelog entries for the removal of both commands w/ info on how to build URIs to open data sets/jobs in v3 - sorry for not having added them previously 😅

packages/zowe-explorer/src/job/init.ts
traeok: Yep, sorry again about these missing changelog entries - I should have been more diligent with that.
Resolved in upcoming commit 🙂

packages/zowe-explorer/tsconfig.json
traeok: Honestly I'm unsure if we really need `dom` here - but I think we should be using es2020 (if possible) according to the VS Code extension [helloworld-sample](https://github.com/microsoft/vscode-extension-samples/blob/877375478be5382d834ad9ff8dd1e5f7ddf8eb65/helloworld-sample/tsconfig.json#L5) - `es6` is ECMAScript 2015, might have some benefits to upgrading 🤷‍♂️ 

But I'll look into this a bit more before resolving 😋

packages/zowe-explorer-api/src/tree/IZoweTreeNode.ts
traeok: I looked over the removed properties and added a changelog entry for the removal of the `encoding` property. Thank you for spotting that!

Regarding `tooltip` and other properties from `vscode.TreeItem`: Maybe I'm overlooking something, but the type signatures look to be the same for the ones I removed.

packages/zowe-explorer-api/src/tree/IZoweTree.ts
traeok: Not to sound like a broken record, but I'm sorry for these lack of changelog entries. Fixed in cae206c

packages/zowe-explorer-api/src/profiles/ZoweExplorerZosmfApi.ts
traeok: Added an entry for move, but as for `getContents` I wasn't sure whether that was breaking or not, I'll revisit this once I get to Timothy's feedback 😅

packages/zowe-explorer-api/__tests__/__unit__/fs/utils/abstract.unit.test.ts
traeok: Sorry about that. I was copying the template for these type guards since they all have the same general test cases, but I overlooked this. Resolved in an upcoming commit

packages/zowe-explorer-api/src/fs/utils/datasets.ts
traeok: Yep, you're right. I changed it to use `instanceof` since `DsEntry` is a class anyway. I think this was from a while back when the foundation was still an experiment of sorts.

packages/zowe-explorer-api/src/fs/utils/jobs.ts
traeok: Yep - thankfully `instanceof` saves the day here as well 😂  resolved in upcoming commit

packages/zowe-explorer-api/package.json
```diff
@@ -27,6 +27,7 @@
     "@zowe/zos-tso-for-zowe-sdk": "8.0.0-next.202403132009",
     "@zowe/zos-uss-for-zowe-sdk": "8.0.0-next.202403132009",
     "@zowe/zosmf-for-zowe-sdk": "8.0.0-next.202403132009",
+    "lodash.isequal": "^4.5.0",
```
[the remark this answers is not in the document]
traeok: Good call - I looked into this a bit more and it turns out we've had this in ZE for some time, not sure what I can do about that though for this PR because it stretches into both the edit attributes and edit history webviews - for the sake of scope creep we might want to save this for a future PR.

But, since this PR added it to ZE API, I've removed the reference in `BaseProvider`, replaced it with a similar implementation named `areContentsEqual` and removed `lodash.isequal` from the API `package.json`.

packages/zowe-explorer-api/CHANGELOG.md
```diff
@@ -9,6 +9,15 @@ All notable changes to the "zowe-explorer-api" extension will be documented in t
...
+- **Breaking:** Changed the type for the options parameter in the `getContents` function (`MainframeInteraction.IUss` interface) to `zosfiles.IDownloadSingleOptions`.
+- Added the `getEncoding` optional function to the `IZoweDatasetTreeNode` and `IZoweUSSTreeNode` interfaces.
+  - **Breaking:** Removed the `encoding` property from the `IZoweUSSTreeNode` interface in favor of the new `getEncoding` function.
+- Added an optional function `nodeDataChanged` to the `IZoweTree` interface to signal an event when a tree node needs updated.
+- Added the optional `vscode.DragAndDropController` interface to the `IZoweTree` interface.
+- Added a `ZoweScheme` enum to expose the core FileSystemProvider schemes for USS files, data sets and jobs.
+- Added optional function `move` to the `MainframeInteraction.IUss` interface to move USS folders/files from one path to another.
+- Added the `buildUniqueSpoolName` function to build spool names for Zowe resource URIs and VS Code editor tabs.
+- Added the `isNodeInEditor` function to determine whether a tree node is open in the editor.
```
anaxceron, line 20: Re: Lines 12-20: We should add pr/issue links to these.

anaxceron (dismissed):
Only one comment on the changelog -- thanks @traeok!

packages/zowe-explorer/tsconfig.json
traeok: So I did see that some popular extensions such as GitLens put `dom` in the `lib` array, so I've added it back in commit fb8cf5d.

traeok:
@t1m0thyj, regarding this:

> I'm wondering if this is truly a breaking change 🤔 Since IDownloadOptions extends IDownloadSingleOptions, doesn't that mean anyone who was previously passing an IDownloadOptions interface into the function will still be able to do so?

I looked into the types a bit more and it will not be breaking - that being said, I adjusted the changelog to remove `**Breaking**` but kept the entry there for awareness.

packages/zowe-explorer-api/src/profiles/ZoweExplorerZosmfApi.ts
traeok: I looked into this more and it won't be a breaking change. I added this to the changelog 👍

packages/zowe-explorer-api/CHANGELOG.md
```diff
@@ -9,6 +9,15 @@ All notable changes to the "zowe-explorer-api" extension will be documented in t
 - **Breaking:** Marked `getJobsByParameters` as a required function for the `MainframeInteraction.IJes` interface. [#2764](https://github.com/zowe/vscode-extension-for-zowe/pull/2764)
   - The new `getJobsByParameters` API is meant to replace `getJobsByOwnerAndPrefix`, and it adds new capabilities such as querying by status and limiting the amount of returned jobs.
 - **Breaking:** Removed string as a return type of the `uploadFromBuffer` method, since the z/OSMF API has been fixed to return a response object that includes an etag. [#2785](https://github.com/zowe/zowe-explorer-vscode/issues/2785)
+- Changed the type for the options parameter in the `getContents` function (`MainframeInteraction.IUss` interface) from `zosfiles.IDownloadOptions` to `zosfiles.IDownloadSingleOptions`. [#2207](https://github.com/zowe/zowe-explorer-vscode/issues/2207)
```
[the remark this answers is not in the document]
traeok: I've added some info on why this change was made to the `getContents` function. I've also updated some of the changelog items to clarify the purpose of the changes.

For any items that do not clarify the reasoning, I will add any necessary info to the ZE wiki. A lot of these changes are developer-oriented and were primarily done for refactoring purposes.

anaxceron (dismissed):
Thanks, Trae, looks good!

zFernand0 (dismissed):
Sorry for the delay.
Changed do LGTM!
Doing some sanity checks throughout the extension 😋

packages/zowe-explorer/src/dataset/init.ts
```diff
@@ -13,18 +13,29 @@ import * as vscode from "vscode";
...
+import { DatasetFSProvider } from "./DatasetFSProvider";
+import { posix as posixPath } from "path";
 
 export async function initDatasetProvider(context: vscode.ExtensionContext): Promise<DatasetTree> {
     ZoweLogger.trace("dataset.init.initDatasetProvider called.");
+
+    context.subscriptions.push(vscode.workspace.registerFileSystemProvider(ZoweScheme.DS, DatasetFSProvider.instance, { isCaseSensitive: true }));
+    context.subscriptions.push(
+        vscode.commands.registerCommand(
+            "zowe.onDatasetChanged",
```
t1m0thyj, line 34: Curious if rather than having a separate command for each tree, we could use a single command `zowe.onResourceChanged` that takes the resource type (dataset, ussFile, job) as an argument?

Or perhaps instead of using commands, we can expose the `onDidChangeFile` events on the `ZoweExplorerExtenderApi` class that extenders are able to import?

Sorry if I'm being picky - this seems like a non-standard usage of VS Code commands to register events so wondering if we can avoid it 😋

packages/zowe-explorer/src/extension.ts
```diff
@@ -39,14 +38,14 @@ export async function activate(context: vscode.ExtensionContext): Promise<ZoweEx
 
     await ProfilesUtils.initializeZoweProfiles((msg) => ZoweExplorerExtender.showZoweConfigError(msg));
     await Profiles.createInstance(ZoweLogger.imperativeLogger);
-    initializeSpoolProvider(context);
+    await Profiles.getInstance().getProfileInfo();
```
t1m0thyj, line 41: Can this line be moved inside `Profiles.createInstance`? It seems good to minimize the amount of code needed in `extension.ts` if possible 😋

packages/zowe-explorer/package.json
```diff
@@ -2097,12 +2123,10 @@
     "@zowe/zosmf-for-zowe-sdk": "8.0.0-next.202403132009",
     "@zowe/zowe-explorer-api": "3.0.0-next-SNAPSHOT",
     "dayjs": "^1.11.10",
+    "disposablestack": "^1.1.3",
```
t1m0thyj, line 2126: Should this be a dev dep instead of a prod dep, since it looks like we only use the package in unit tests?

packages/zowe-explorer/package.json
```diff
@@ -2079,12 +2107,10 @@
     "eslint-plugin-zowe-explorer": "3.0.0-next-SNAPSHOT",
     "event-stream": "^4.0.1",
     "expect": "^24.8.0",
-    "geckodriver": "^1.19.1",
     "glob": "^7.1.6",
-    "jsdom": "^16.0.0",
+    "jest-silent-reporter": "^0.5.0",
```
t1m0thyj, line 2111: Curious if this package is being used, since I don't see it referenced in the Jest config?

t1m0thyj (commented):
Looks pretty good, thanks @traeok for all your work on this! Left a few comments

packages/zowe-explorer-api/src/fs/types/uss.ts
```diff
@@ -0,0 +1,38 @@
...
+    public permissions?: vscode.FilePermission;
+
+    public constructor(name: string) {
+        super(name);
+    }
+}
+
+export class UssDirectory extends DirEntry {
+    public constructor(name?: string) {
+        super(name ?? "");
```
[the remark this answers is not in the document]
traeok: Yep, that's spot-on 👍

packages/zowe-explorer/package.json
traeok: It's not used in the current working tree, but I figured other contributors might find it useful to filter out a lot of the extraneous info from Jest results. I used this to clean up a bunch of failing tests without having to read in between the error outputs and the Jest metadata.

I can remove this if desired

packages/zowe-explorer/src/extension.ts
traeok: Sure, good call -- addressed in next commit

packages/zowe-explorer/CHANGELOG.md
```diff
@@ -6,6 +6,10 @@ All notable changes to the "vscode-extension-for-zowe" extension will be documen
 
 ### New features and enhancements
 
+- Implemented the FileSystemProvider for the Data Sets, Jobs and USS trees to handle all read/write actions as well as conflict resolution. [#2207](https://github.com/zowe/zowe-explorer-vscode/issues/2207)
+- **Breaking:** Removed the `zowe.jobs.zosJobsOpenSpool` command in favor of using `vscode.open` with a spool URI. See the [FileSystemProvider wiki page](https://github.com/zowe/zowe-explorer-vscode/wiki/FileSystemProvider#file-paths-vs-uris) for more information on spool URIs. [#2207](https://github.com/zowe/zowe-explorer-vscode/issues/2207)
+- **Breaking:** Removed the `zowe.ds.ZoweNode.openPS` command in favor of using `vscode.open` with a data set URI. See the [FileSystemProvider wiki page](https://github.com/zowe/zowe-explorer-vscode/wiki/FileSystemProvider#file-paths-vs-uris) for more information on data set URIs. [#2207](https://github.com/zowe/zowe-explorer-vscode/issues/2207)
```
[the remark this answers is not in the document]
traeok: Thanks for catching this, addressed in next commit

packages/zowe-explorer-api/src/fs/types/datasets.ts
```diff
@@ -0,0 +1,47 @@
...
+        super(name);
+        this.entries = new Map();
+    }
+}
+
+export class DsEntryMetadata implements EntryMetadata {
+    public profile: IProfileLoaded;
+    public path: string;
+
+    public constructor(metadata: EntryMetadata) {
```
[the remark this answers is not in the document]
traeok: `EntryMetadata` is a type, so `super` shouldn't be necessary here

packages/zowe-explorer-api/src/fs/BaseProvider.ts
```diff
@@ -0,0 +1,450 @@
...
+        this._profilesCache = profilesCache ?? ZoweVsCodeExtension.profilesCache;
+    }
+
+    /**
+     * Compares the data for 2 Uint8Arrays, byte by byte.
+     * @param a The first Uint8Array to compare
+     * @param b The second Uint8Array to compare
+     * @returns `true` if the arrays are equal, `false` otherwise
+     */
+    public static areContentsEqual(a: Uint8Array, b: Uint8Array): boolean {
```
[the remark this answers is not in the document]
traeok: While it's a more concise and declarative approach, the current implementation would be the imperative equivalent. Regardless, I like smaller code that accomplishes the same thing, so I've addressed it in the next commit

packages/zowe-explorer/src/job/ZoweJobNode.ts
```diff
@@ -123,13 +148,12 @@ export class ZoweJobNode extends ZoweTreeNode implements IZoweJobTreeNode {
                     label: vscode.l10n.t("There are no JES spool messages to display"),
                     collapsibleState: vscode.TreeItemCollapsibleState.None,
                     parentNode: this,
+                    profile: this.getProfile(),
                 });
-                noSpoolNode.iconPath = null;
+                noSpoolNode.iconPath = undefined;
```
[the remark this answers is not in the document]
traeok: Technically any falsy value will work, but I felt it was appropriate to change it since there were other references to `iconPath = undefined` from a few years ago, and because I've noticed the VS Code team [falling back to `undefined`](https://github.com/microsoft/vscode/blob/8558da133548b1a0b83f038e103ffa7f84f47da4/src/vs/workbench/api/common/extHostTreeViews.ts#L840) as well 😋 

and no problem!

packages/zowe-explorer-api/__mocks__/vscode.ts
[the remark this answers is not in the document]
traeok: I noticed that we've had duplicated module mocks between the two packages, and its impact was noticeable during this work (such as "which module mock needs this type?" when an error pops up from Jest). Since this was an issue before this PR, I think we should look into consolidating the module mocks as a whole in a separate issue, so I've filed #2817.

packages/zowe-explorer/src/dataset/init.ts
traeok: I've implemented a combination of your feedback by adding the `onResourceChanged` function in `ZoweExplorerApiRegister` that takes in a `ZoweScheme` as the parameter - it will return the corresponding event.

Thanks for the suggestions!

packages/zowe-explorer-api/src/fs/BaseProvider.ts
```diff
@@ -0,0 +1,450 @@
...
+        }, FS_PROVIDER_DELAY);
+    }
+
+    /**
+     * VScode utility functions for entries in the provider:
+     */
+
+    protected _lookup(uri: vscode.Uri, silent: false): IFileSystemEntry;
+    protected _lookup(uri: vscode.Uri, silent: boolean): IFileSystemEntry | undefined;
+    protected _lookup(uri: vscode.Uri, silent: boolean): IFileSystemEntry | undefined {
```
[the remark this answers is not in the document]
traeok: `stat` will throw an error if `_lookup` fails according to the spec, so no worries there 👍  error handling will be required for some of the functions due to the expected behavior of the `FileSystemProvider` interface.

As for the overloads, I'm not 100% sure if they are required, so I will look into removing them.

packages/zowe-explorer/package.json
t1m0thyj: I'm not opposed to having it since you found it useful 😋 Could we mention somewhere how to use it - maybe a note somewhere in dev docs, or an npm script that runs tests in silent mode?

traeok: Sure 😋 I'll add a script and a note in the wiki 👍

traeok: As discussed offline, there isn't a good place to mention this on the wiki - but I've added the script in the latest commit.

t1m0thyj (commented):
Noticed a few issues when doing manual testing:

* When I favorite a PDS or USS directory, I am unable to expand its children in the Favorites tree.
* In the USS tree, some drag and drop actions fail but indefinitely show the "Moving USS files..." message:
  * Dragging a parent folder into one of its children - probably shouldn't be allowed
  * Dragging a folder over a file - should move the folder into the file's parent folder

traeok:
> * When I favorite a PDS or USS directory, I am unable to expand its children in the Favorites tree.

Thanks @t1m0thyj for catching this, I updated the logic to use the profile name rather than trying to get it from the parent. It will still fall back to the session node's label if needed, since `profile` is an optional parameter on nodes.

adam-wolfe:
With a build of [c1e9558](https://github.com/zowe/zowe-explorer-vscode/pull/2777/commits/c1e9558b0439a99523d04f73e8f16579a1baab14), when I attempt to rename a USS directory, I run into two errors:
`Error running command zowe.uss.renameNode: zowe-uss:/home-dir/test12093. This is likely caused by the extension that contributes zowe.uss.renameNode.`

`Unable to rename node: EntryIsADirectory (FileSystemError): zowe-uss:/dev.zosmf/home-dir/test12093`

Log output:
```
[2024/04/08 16:20:39] [TRACE] USSTree.rename called.
[2024/04/08 16:20:39] [TRACE] ZoweUSSNode.getSessionNode called.
[2024/04/08 16:20:39] [TRACE] ZoweUSSNode.getSessionNode called.
[2024/04/08 16:20:39] [TRACE] USSTree.findFavoriteNode called.
[2024/04/08 16:20:39] [TRACE] USSTree.findMatchingProfileInArray called.
[2024/04/08 16:20:39] [TRACE] USSTree.getAllLoadedItems called.
[2024/04/08 16:20:39] [TRACE] USSTree.getChildren called.
[2024/04/08 16:20:39] [TRACE] ZoweUSSNode.getChildren called.
[2024/04/08 16:20:39] [TRACE] ZoweUSSNode.getSessionNode called.
[2024/04/08 16:20:39] [TRACE] ZoweUSSNode.getSessionNode called.
[2024/04/08 16:20:39] [TRACE] ZoweUSSNode.getChildren called.
[2024/04/08 16:20:39] [TRACE] USSTree.checkDuplicateLabel called.
[2024/04/08 16:20:40] [TRACE] USSTree.checkDuplicateLabel called.
[2024/04/08 16:20:40] [TRACE] USSTree.checkDuplicateLabel called.
[2024/04/08 16:20:40] [TRACE] ZoweUSSNode.rename called.
[2024/04/08 16:20:41] [TRACE] uss.utils.injectAdditionalDataToTooltip called.
[2024/04/08 16:20:41] [ERROR] EntryIsADirectory (FileSystemError): zowe-uss:/dev.zosmf/home-dir/test12092
{
  errorDetails: [EntryIsADirectory (FileSystemError): zowe-uss:/dev.zosmf/home-dir/test12092
  	at F._lookupAsFile (c:\Users\me\.vscode\extensions\zowe.vscode-extension-for-zowe-3.0.0-next-SNAPSHOT\out\src\main.extension.js:12:9202)
  	at F.getEncodingForFile (c:\Users\me\.vscode\extensions\zowe.vscode-extension-for-zowe-3.0.0-next-SNAPSHOT\out\src\main.extension.js:12:5302)
  	at s.getEncoding (c:\Users\me\.vscode\extensions\zowe.vscode-extension-for-zowe-3.0.0-next-SNAPSHOT\out\src\main.extension.js:92:18322)
  	at u (c:\Users\me\.vscode\extensions\zowe.vscode-extension-for-zowe-3.0.0-next-SNAPSHOT\out\src\main.extension.js:97:86)
  	at s.<anonymous> (c:\Users\me\.vscode\extensions\zowe.vscode-extension-for-zowe-3.0.0-next-SNAPSHOT\out\src\main.extension.js:92:20281)
  	at Generator.next (<anonymous>)
  	at w (c:\Users\me\.vscode\extensions\zowe.vscode-extension-for-zowe-3.0.0-next-SNAPSHOT\out\src\main.extension.js:92:13317)
  	at process.processTicksAndRejections (node:internal/process/task_queues:95:5)] {
    code: 'FileIsADirectory'
  },
  label: 'dev.zosmf',
  moreInfo: 'Unable to rename node:'
}
```

traeok:
> With a build of [c1e9558](https://github.com/zowe/zowe-explorer-vscode/pull/2777/commits/c1e9558b0439a99523d04f73e8f16579a1baab14), when I attempt to rename a USS directory, I run into two error messages: `Error running command zowe.uss.renameNode: zowe-uss:/home-dir/test12093. This is likely caused by the extension that contributes zowe.uss.renameNode.`
> 
> `Unable to rename node: EntryIsADirectory (FileSystemError): zowe-uss:/dev.zosmf/home-dir/test12093`

Thanks for the log output @adam-wolfe, that helped to find the source of the issue. When I pulled in the recent encoding changes on `next`, I missed a required check in `injectAdditionalDataToTooltip` to skip grabbing the encoding if its a USS directory.

Resolved in 73a0d3d 👍

packages/zowe-explorer-api/src/fs/BaseProvider.ts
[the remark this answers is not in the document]
traeok: Thanks for the suggestion 😋 
I've removed the two overloads in favor of:

```ts
protected _lookup(uri: vscode.Uri, silent: boolean = false): IFileSystemEntry | undefined {
```

anaxceron (approved):
The changelogs look good, thanks @traeok!

traeok:
@anaxceron Feel free to disregard the re-requested review, changelogs haven't changed since your last review 👍 
I will reach out if the changelogs are updated once more

t1m0thyj (commented):
Tested this PR with various z/OSMF operations and they all work 👍 

When I try to test this PR with v2.15.2 of the FTP extension, I encountered some issues:
* I can list data sets and USS files in the tree but attempting to open them results in an error:
    image
* When listing jobs I get the following error:
    image

The same zFTP profile works fine to view data sets in Zowe CLI. I may be doing something wrong here - do I need to use another version of the FTP extension like https://github.com/zowe/zowe-explorer-vscode/pull/2783?

traeok:
> Tested this PR with various z/OSMF operations and they all work 👍
> 
> When I try to test this PR with v2.15.2 of the FTP extension, I encountered some issues:
> 
> * I can list data sets and USS files in the tree but attempting to open them results in an error:

@t1m0thyj I've resolved this in an upcoming commit - some of the FTP APIs did not behave as expected and were missing some capabilities that are supported by `zos-node-accessor` 😋 

> * When listing jobs I get the following error:

As for this issue, I'm looking into this now 👍

JillieBeanSim:
I did see some strange behavior when first starting up this branch today with latest changes, and didn't experience in the past testing branch.  Got the pop up message about the logger setting, but the pop up was incorrect with TRACE in the settings.json file.  Was this setting moved to internal storage recently and why I may be seeing this?

traeok:
> I did see some strange behavior when first starting up this branch today with latest changes, and didn't experience in the past testing branch. Got the pop up message about the logger setting, but the pop up was incorrect with TRACE in the settings.json file. Was this setting moved to internal storage recently and why I may be seeing this? update: after selecting cancel in the pop up, logs are being written at TRACE level

@JillieBeanSim I'm not sure why this is happening, I'm wondering if it's coming from something recently merged into `next`? I'll double-check over the changed files, but from what I recall I did not adjust the behavior of the logger for this PR.

traeok:
> I am also seeing some strange behavior with drag & drop, it doesn't really follow copy/paste but more like cut/paste viewing my directories it moved from and to. I had to close and open the directory to see updated list of files, not auto updating atm. Also, quick question, is the drag and drop suppose to support multi-select? I can select many but only first selected moved when testing drag and drop.

Drag and drop performs a move operation, I think there is an open issue to also support copying when a modifier key is held (similar to how VS Code supports it in the Explorer view): #2808 

As for the multi-select support, it should support multiple files and folders, so I will do some further manual testing to make sure its working as expected.

traeok:
@JillieBeanSim in 79472d8 I've resolved the weird behavior you encountered with the UX for drag/drop. 
I also tested with a few files and a folder for the drag/drop operation but didn't run into any issues, but there could be a scenario that I've overlooked. I'll continue to perform more drag/drop tests as time permits 👍

zFernand0 (changes requested):
Trying to compare V2 with this PR and found something a bit strange when opening empty spool files 😋 

The following spool file open just fine (with no content) in V2
image

But the same spool file fails to load in this PR:
image

Not sure if this affects other areas though 😋

t1m0thyj (commented):
When saving a USS file opened via an FTP profile, the save operation hangs for me.

Everything else LGTM - tested misc actions with both z/OSMF and FTP 😁

packages/zowe-explorer/package.json
```diff
@@ -836,6 +839,21 @@
...
+        "icon": "$(check)"
+      },
+      {
+        "command": "zowe.diff.useRemoteContent",
+        "title": "%diff.useRemote%",
+        "icon": "$(discard)"
+      },
+      {
+        "command": "zowe.placeholderCommand",
+        "title": "%zowe.placeholderCommand%",
```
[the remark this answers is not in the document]
traeok: I like the idea and intended to remove the title, but once I did, an error occurred when debugging:
image

I'm unsure if this error is specific to the extension host, but I also don't want this to show in prod. Will test to confirm

traeok:
> When saving a USS file opened via an FTP profile, the save operation hangs for me.
> 
> Everything else LGTM - tested misc actions with both z/OSMF and FTP 😁

Thanks @t1m0thyj, resolved in bb4f232

t1m0thyj (commented):
Since I am traveling and had some VPN issues, I noticed that error messages have changed when a ds search fails. Not sure if the change in behavior is intentional? I think this could be addressed as a separate issue, and plan to approve the PR once my connection is fixed 😋 

| | **Before** | **After** |
| --- | --- | --- |
| zOSMF | image | image |
| zFTP | image | image |

t1m0thyj:
> Since I am traveling and had some VPN issues, I noticed that error messages have changed when a ds search fails. Not sure if the change in behavior is intentional? I think this could be addressed as a separate issue, and plan to approve the PR once my connection is fixed 😋

Discussed with @traeok, the zFTP `resourceUri` error was a red herring because the extension hadn't finished loading yet. For the change in zOSMF error message I filed a separate issue: https://github.com/zowe/zowe-explorer-vscode/issues/2847

traeok:
Discussed with @t1m0thyj offline: the first error shown in the "After" section for zFTP is an unrelated error caused by interacting with the extension before it's finished loading.

As for the difference in z/OSMF messaging, we might have to resolve this in a separate PR. I'll look into this to see where this change in messaging is coming from.

awharn (approved):

JillieBeanSim:
With approvals will go ahead and merge, we can open and address issues for anything found dealing with trees in v3 separately.

PR #2830, opened 2024-04-10
Enhance Unix command API to work with profile that doesn't require ssh

likhithanimma1:
## Proposed changes

Making the sshSession parameter optional for issueUnixCommand Api so that profiles that doesn't require ssh can also have it.

## Release Notes

Milestone:

Changelog:

## Types of changes

## Checklist

### DIFF
--- a/packages/zowe-explorer-api/src/extend/MainframeInteraction.ts
+++ b/packages/zowe-explorer-api/src/extend/MainframeInteraction.ts
@@ -506,10 +506,10 @@ export namespace MainframeInteraction {
          * @param {string} command
          * @param {string} cwd
          * @param {boolean} flag
-         * @returns {string>}
+         * @returns {string}
          * @memberof ICommand
          */
-        issueUnixCommand?(sshSession: zosuss.SshSession, command: string, cwd: string, flag: boolean): Promise<string>;
+        issueUnixCommand?(command: string, cwd: string, flag: boolean, sshSession?: zosuss.SshSession): Promise<string>;
         sshProfileRequired?(): boolean;
     }
 }
--- a/packages/zowe-explorer/src/command/UnixCommandHandler.ts
+++ b/packages/zowe-explorer/src/command/UnixCommandHandler.ts
@@ -345,12 +345,16 @@ export class UnixCommandHandler extends ZoweCommandProvider {
                     () => {
                         if (ZoweExplorerApiRegister.getCommandApi(profile).issueUnixCommand) {
                             // eslint-disable-next-line @typescript-eslint/no-unsafe-return
-                            return ZoweExplorerApiRegister.getCommandApi(profile).issueUnixCommand(
-                                this.sshSession,
-                                command,
-                                cwd,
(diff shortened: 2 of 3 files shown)


packages/zowe-explorer-api/src/extend/MainframeInteraction.ts
```diff
@@ -506,10 +506,10 @@ export namespace MainframeInteraction {
          * @param {string} command
          * @param {string} cwd
          * @param {boolean} flag
-         * @returns {string>}
+         * @returns {string}
          * @memberof ICommand
          */
-        issueUnixCommand?(sshSession: zosuss.SshSession, command: string, cwd: string, flag: boolean): Promise<string>;
+        issueUnixCommand?(command: string, cwd: string, flag: boolean, sshSession?: zosuss.SshSession): Promise<string>;
```
zFernand0, line 512: I think it'd be good to add this to the list of breaking changes in a changelog entry 😋

zFernand0 (changes requested):
I do think the idea of not having to provide an SSH session makes sense.
However, it needs to be replaced with something in order to use the `zosuss.Shell` API.

I believe that z/OSMF supports sending Unix commands, so is that what we want to use here?

Otherwise, I'm afraid I need a bit more details as to why we want to pass an `undefined` to the `executeSsh` method here:
```typescript
public async issueUnixCommand(command: string, cwd: string, flag: boolean, sshSession?: zosuss.SshSession): Promise<string> {
  let stdout = "";
  if (flag) {
    await zosuss.Shell.executeSshCwd(sshSession, command, '"' + cwd + '"', (data: string) => {
      stdout += data;
    });
  } else {
    await zosuss.Shell.executeSsh(sshSession, command, (data: string) => {
      stdout += data;
    });
  }
}
```

packages/zowe-explorer/src/command/UnixCommandHandler.ts
```diff
@@ -345,12 +345,16 @@ export class UnixCommandHandler extends ZoweCommandProvider {
...
+                            if (this.sshSession) {
+                                return ZoweExplorerApiRegister.getCommandApi(profile).issueUnixCommand(
+                                    command,
+                                    cwd,
+                                    this.pathInputConfirmationFlag,
+                                    this.sshSession
+                                );
+                            } else {
+                                return ZoweExplorerApiRegister.getCommandApi(profile).issueUnixCommand(command, cwd, this.pathInputConfirmationFlag);
+                            }
```
JillieBeanSim, line 357: I don't think we need an if/else here, if the sshSession is undefined shouldn't a single instance of the call suffice?

JillieBeanSim (changes requested):

packages/zowe-explorer/src/command/UnixCommandHandler.ts
likhithanimma1: Removed if/else

traeok (changes requested):

likhithanimma1:
Closing in favour of - #2866

PR #2838, opened 2024-04-12
Fix race condition in ProfilesCache refresh

### LINKED ISSUE #2831: ProfilesCache does not handle race conditions
**Describe the bug**

Although JavaScript and Node.js are single-threaded, multiple VS Code extensions can activate in parallel. Extensions that grab the static `ProfilesCache` instance from Zowe Explorer might compete with one another when performing operations on the cache (such as `refresh`), causing undefined behavior in the cache (such as missing or incorrect profile values, etc.)

**To Reproduce**

Have two or more extensions that access the `ProfilesCache` instance through Zowe Explorer's extension exports (`vscode.extensions.getExtension("zowe.vscode-extension-for-zowe").exports`). Try to call `refresh` on this cache instance for each extension, as soon as the extensions are activated.

**Expected behavior**

The ProfilesCache instance should fill the cache successfully during refresh, and return an instance with valid profiles.

- OS: Windows
- Zowe Explorer Version: 2.15.2

**Additional context**

A few internal developers have reported this issue. Due to the nature of VS Code extensions and asynchronous code in Node.js, we likely need to add multi-threading support to prevent any problems that may surface from ProfilesCache race conditions.

t1m0thyj:
## Proposed changes

Fixes #2831

## Release Notes

Milestone: 2.15.3

Changelog: Fixed an issue where `ProfilesCache` may return missing or incorrect profile values when multiple extensions call it during activation

## Types of changes

## Checklist

### DIFF
--- a/packages/zowe-explorer-api/src/profiles/ProfilesCache.ts
+++ b/packages/zowe-explorer-api/src/profiles/ProfilesCache.ts
@@ -70,7 +70,7 @@ export class ProfilesCache {
     public profilesValidationSetting: IValidationSetting[] = [];
     public allProfiles: zowe.imperative.IProfileLoaded[] = [];
     public profileTypeConfigurations: zowe.imperative.ICommandProfileTypeConfiguration[] = [];
-    protected allTypes: string[];
+    protected allTypes: string[] = [];
     protected allExternalTypes = new Set<string>();
     protected profilesByType = new Map<string, zowe.imperative.IProfileLoaded[]>();
     protected defaultProfileByType = new Map<string, zowe.imperative.IProfileLoaded>();
@@ -192,10 +192,7 @@ export class ProfilesCache {
     }
 
     public async refresh(apiRegister?: ZoweExplorerApi.IApiRegisterClient): Promise<void> {
-        this.allProfiles = [];
-        this.allTypes = [];
-        this.profilesByType.clear();
-        this.defaultProfileByType.clear();
+        const allProfiles: zowe.imperative.IProfileLoaded[] = [];
         let mProfileInfo: zowe.imperative.ProfileInfo;
         try {
             mProfileInfo = await this.getProfileInfo();
@@ -219,10 +216,15 @@ export class ProfilesCache {
                         // Step 3: Update allProfiles list
                         tmpAllProfiles.push(profileFix);
                     }
-                    this.allProfiles.push(...tmpAllProfiles);
+                    allProfiles.push(...tmpAllProfiles);
                     this.profilesByType.set(type, tmpAllProfiles);
                 }
-                this.allTypes.push(type);
+            }
+            this.allProfiles = allProfiles;
+            this.allTypes = allTypes;
+            for (const oldType of [...this.profilesByType.keys()].filter((type) => !allTypes.includes(type))) {
+                this.profilesByType.delete(oldType);
+                this.defaultProfileByType.delete(oldType);
             }
             // check for proper merging of apiml tokens
             this.checkMergingConfigAllProfiles();
--- a/packages/zowe-explorer/package.json
+++ b/packages/zowe-explorer/package.json
@@ -823,7 +823,7 @@
       },
       {
         "command": "zowe.placeholderCommand",
-        "title": "%zowe.placeholderCommand%",
+        "title": "",
         "enablement": "false"
       }
     ],
--- a/packages/zowe-explorer/package.json
+++ b/packages/zowe-explorer/package.json
@@ -2026,7 +2026,6 @@
     "eslint-plugin-zowe-explorer": "2.16.0-SNAPSHOT",
     "event-stream": "^4.0.1",
     "expect": "^24.8.0",
-    "geckodriver": "^1.19.1",
     "glob": "^7.1.6",
     "gulp": "^4.0.2",
     "gulp-cli": "^2.2.0",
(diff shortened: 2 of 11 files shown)


packages/zowe-explorer/package.json
```diff
@@ -823,7 +823,7 @@
       },
       {
         "command": "zowe.placeholderCommand",
-        "title": "%zowe.placeholderCommand%",
+        "title": "",
         "enablement": "false"
       }
```
zFernand0, line 828: Dumb questions: 😋 
- What's this?
- Can we remove it?

packages/zowe-explorer-api/src/profiles/ProfilesCache.ts
```diff
@@ -219,10 +216,15 @@ export class ProfilesCache {
...
+                    allProfiles.push(...tmpAllProfiles);
                     this.profilesByType.set(type, tmpAllProfiles);
                 }
-                this.allTypes.push(type);
+            }
+            this.allProfiles = allProfiles;
+            this.allTypes = allTypes;
+            for (const oldType of [...this.profilesByType.keys()].filter((type) => !allTypes.includes(type))) {
+                this.profilesByType.delete(oldType);
+                this.defaultProfileByType.delete(oldType);
```
zFernand0, line 227: I know this for-loop is based on a filtered version of `this.profileByType`, so doing `this.profileByType.delete(oldType)` should be fine.
I'm curious if we should check if `this.defaultProfileByType.has(oldType)` before trying to delete it. 😋

zFernand0 (approved):
Thanks for attempting to fix the theia tests. 😋 
Also, I can see how this is attempting to resolve the race condition and avoiding missing profile values (by not clearing `this.allTypes` on every refresh), but if you could provide some quick step to test how to get incorrect profile values, or maybe show this on a future standup?

Changes LGTM though! 😋

packages/zowe-explorer-api/src/profiles/ProfilesCache.ts
traeok: The current logic should be okay, as `delete` will [return false](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map/delete#return_value) if the element doesn't exist

packages/zowe-explorer/package.json
t1m0thyj: We use this to show tree nodes as disabled. VS Code will "gray out" a tree node that has a command assigned with "enablement = false". Thanks @traeok for suggesting this trick 😋

packages/zowe-explorer/package.json
```diff
@@ -823,7 +823,7 @@
       },
       {
         "command": "zowe.placeholderCommand",
-        "title": "%zowe.placeholderCommand%",
+        "title": "",
```
traeok, line 826: We'll probably want to switch back to a localized placeholder, as a side effect of the empty `title` was found while addressing feedback in #2777:

image

_(only affects extension host)_

traeok (commented):
LGTM, thanks @t1m0thyj for resolving this issue! I left a comment regarding the placeholder command based on our previous offline discussions

traeok (approved):

JillieBeanSim (approved):
LGTM! thanks @t1m0thyj

traeok:
Hi @roman-kupriyanov,

The second block of code you posted would be the right approach with the current state of the ProfilesCache. In `initForZowe` we do a `refresh` of the profiles cache, and during that refresh, we get all the types using `ProfilesCache.getAllProfileTypes`. If you do not call `registerCustomProfilesType` beforehand, it will not consider that type unless it has already been registered under one of the APIs (MVS, USS, or JES).

Ideally, I'd like to register those custom types as a part of the logic in `initForZowe` so that extenders don't have to call `registerCustomProfilesType` - rather, we can do it ourselves during initialization. I don't think this would be breaking since `allExternalTypes` is already a set, so continuing to call the `registerCustomProfilesType` function shouldn't affect anything.

I don't want to speak for @t1m0thyj, but those are my thoughts regarding extender initialization/preparing custom profile types 😋

roman-kupriyanov:
Hi @traeok,

Thanks for the quick response!

> I believe the second block of code you posted would be the right approach with the current state of the ProfilesCache. In `initForZowe` we do a `refresh` of the profiles cache, and during that refresh, we get all the types using `ProfilesCache.getAllProfileTypes`. If you do not call `registerCustomProfilesType` beforehand, it will not consider that type unless it has already been registered under one of the APIs (MVS, USS, or JES).

That was my assumption too since it makes total sense to have an external profile type registered before the attempt of refreshing the cache. I think, before that extra `reloadProfiles()` call was actually protecting us from the problems by refreshing one more time, when all the parameters are actually set.

> Ideally, I'd like to register those custom types as a part of the logic in `initForZowe` so that extenders don't have to call `registerCustomProfilesType` - rather, we can do it ourselves during initialization if the type doesn't already exist. I don't think this would be breaking since `allExternalTypes` is already a set, so continuing to call the `registerCustomProfilesType` function from an extender's perspective shouldn't affect anything.

Those were my thoughts exactly! Moreover, if the order of those calls is crucial to make it work right, I would expect it to be called internally by `initForZowe()` and at the right time to avoid making a mistake as the extender.

But I am not going to force any decision here as well and leave it up to @t1m0thyj  and the team. 

Thanks for the confirmation and investigation, cheers!

JillieBeanSim:
Thanks for testing the different scenarios of registering with ZE @roman-kupriyanov.  We will want to make sure our documentation matches the best solution for registering with ZE in zowe.org and our wiki if it's different than what we already have so other extenders can benefit from it. ❤️

awharn (approved):

PR #2839, opened 2024-04-12
Use Node 18 for v1 deployment since ovsx requires it

t1m0thyj:
Fixes the failing workflow: https://github.com/zowe/zowe-explorer-vscode/actions/runs/8661844096


JillieBeanSim (approved):
thanks @t1m0thyj

PR #2882, opened 2024-04-25
fix(ds): Support opening data sets in binary mode

### LINKED ISSUE #2696: Data set members may fail to open in binary mode
Data set members cannot always be opened in binary mode depending on the contents.

The following member can be opened in binary mode (as long as there is no trailing line) because it is exactly 80 bytes long:
```
Hello World                                                                    $
```

Attempting to open the following member in binary mode results in an error:
```
Hello World $
```

Here are the attributes of the PDS:

t1m0thyj:
## Proposed changes

Fixes #2696 by adding a prompt to download data sets in binary mode similar to the existing behavior for USS files.

## Release Notes

Milestone: 2.16.0

Changelog: Fixed issue where data sets or members containing binary content cannot be opened

## Types of changes

## Checklist


packages/zowe-explorer/src/dataset/DatasetTree.ts
```diff
@@ -1522,8 +1523,8 @@ export class DatasetTree extends ZoweTreeProvider implements IZoweTree<IZoweData
         }
     }
 
-    public async openWithEncoding(node: IZoweDatasetTreeNode): Promise<void> {
-        const encoding = await promptForEncoding(node);
+    public async openWithEncoding(node: IZoweDatasetTreeNode, encoding?: ZosEncoding): Promise<void> {
+        encoding = encoding ?? (await promptForEncoding(node));
```
zFernand0, line 1527: have you considered using `||=`? (e.g. `encoding ||= await prompt...`) 😋

packages/zowe-explorer/__tests__/__unit__/uss/ZoweUSSNode.unit.test.ts
```diff
@@ -1386,84 +1386,6 @@ describe("ZoweUSSNode Unit Tests - Function node.openedDocumentInstance()", () =
     });
 });
 
-describe("ZoweUSSNode Unit Tests - Function node.initializeFileOpening()", () => {
```
zFernand0, line 1389: thanks for cleaning up these tests after exporting/moving the function outside the node

packages/zowe-explorer/src/shared/utils.ts
```diff
@@ -544,3 +544,42 @@ export async function promptForEncoding(node: IZoweDatasetTreeNode | IZoweUSSTre
...
+                localize("zowe.shared.utils.initializeFileOpening.message", "Failed to open file as text. Re-download file as binary?"),
+                { items: [yesResponse, noResponse] }
+            );
+
+            if (response === yesResponse) {
+                const command = isZoweDatasetTreeNode(node) ? "zowe.ds.openWithEncoding" : "zowe.uss.openWithEncoding";
+                await vscode.commands.executeCommand(command, node, { kind: "binary" });
+            }
+        } else {
+            await Gui.showTextDocument(document, { preview: previewFile });
```
zFernand0, line 579: thanks for simplifying some of the logic too 😋

zFernand0 (commented):
Changes LGTM! 😋 
Left a small comment but nothing that needs to be changed.
Will approve after doing some more testing 😋

traeok (dismissed):
LGTM, thanks Timothy! I tested this with a "hello world" binary that I uploaded - in 2.15.4 I get an error message when trying to open it, but with this fix, I can open and save the PDS member as expected.

I noticed that you can right-click -> "Open with Encoding" on a member of a binary PDS and select a different encoding, but it fails to open and will prompt the user again to re-download as binary. However, for members stored with a text encoding, I can open them as either binary or text without issue, so the current behavior seems consistent to me 👍

packages/zowe-explorer/src/dataset/DatasetTree.ts
t1m0thyj: Typically I think we prefer using `??` over `||` - either one would probably work but the former has more strictly defined behavior. TIL that  [`??=`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing_assignment) is a thing, so I'll use that 😋

zFernand0 (approved):
LGTM, thanks @t1m0thyj! 😋

traeok (approved):

JillieBeanSim (approved):

JillieBeanSim:
was getting ready to merge then seen it was against maintenance branch, @t1m0thyj I thought you mentioned rebasing to main on Tues call?

PR #2896, opened 2024-05-02
Fix Coverity code smells in v3

t1m0thyj:
## Proposed changes

Fixes 4 code smells detected by Coverity in the v3 branch

## Release Notes

Milestone:

Changelog:

## Types of changes

## Checklist

### DIFF
--- a/packages/zowe-explorer/src/uss/UssFSProvider.ts
+++ b/packages/zowe-explorer/src/uss/UssFSProvider.ts
@@ -472,7 +472,7 @@ export class UssFSProvider extends BaseProvider implements vscode.FileSystemProv
             // If file names match, build the copy suffix
             let dupCount = 1;
             const extension = path.extname(fileName);
-            const baseNameForFile = path.parse(fileName)?.name;
+            const baseNameForFile = path.parse(fileName).name;
             let dupName = `${baseNameForFile} (${dupCount})${extension}`;
             while (fileList.find((file) => file.name === dupName) != null) {
                 dupCount++;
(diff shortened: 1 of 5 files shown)


packages/zowe-explorer/src/uss/UssFSProvider.ts
```diff
@@ -472,7 +472,7 @@ export class UssFSProvider extends BaseProvider implements vscode.FileSystemProv
             // If file names match, build the copy suffix
             let dupCount = 1;
             const extension = path.extname(fileName);
-            const baseNameForFile = path.parse(fileName)?.name;
+            const baseNameForFile = path.parse(fileName).name;
```
zFernand0, line 475: Interesting that we would add `?.` here.
I couldn't find which version of Node (since node12) would return a `null/undefined` value.

In any case, I would think we want to make sure that fileName is a string or surround the `baseNameForFile` assignment in a try/catch

packages/zowe-explorer/src/utils/ProfilesUtils.ts
```diff
@@ -229,20 +229,18 @@ export class ProfilesUtils {
...
-            }
-            if (this.PROFILE_SECURITY) {
-                imperative.CredentialManagerOverride.recordCredMgrInConfig(this.PROFILE_SECURITY);
-            }
+        if (settingEnabled && credentialManager) {
+            this.PROFILE_SECURITY = credentialManager;
+            return;
+        } else if (!settingEnabled) {
+            this.PROFILE_SECURITY = false;
+            ZoweLogger.info(vscode.l10n.t(`Zowe explorer profiles are being set as unsecured.`));
```
zFernand0, line 237: Technically this could be considered a bug fix.
I don't think that with the previous logic we would even get to this condition 😋 
Thanks for fixing 😋

zFernand0 (approved):
LGTM, thanks @t1m0thyj! 😋 

Left some comments but nothing that needs addressed right away 😋

zFernand0:
I don't think we have an open issue for addressing scan results in general, but I think this is a good practice to continue doing.

@t1m0thyj, Should we open an issue about the `settingEnabled` bug that was theoretically fixed here? Not sure if we really need it, but it might be good to capture it somewhere 😋

JillieBeanSim:
merging with the approvals.
also for follow up @t1m0thyj, here is the issue I mentioned in scrum that may be related https://github.com/zowe/zowe-explorer-vscode/issues/2063

Issue #2918, opened 2024-05-16
Vnext version 3.0 of Zowe Explorer uses a non-existent API


gejohnston:
**Describe the bug**

I noticed that in the next branch of ZE, the file zowe-explorer-vscode/packages/zowe-explorer-api/src/profiles
/ProfilesCache.ts uses the function `imperative.ConfigBuilder.convert`.

The function imperative.ConfigBuilder.convert no longer exists in the next branch of the Zowe CLI SDK. I suspect that the ZE team used an older snapshot of the CLI vNext SDK, and has not yet updated to use a newer snapshot.

I think that once ZE uses a newer CLI SDK version, ZE will not even compile. ZE code will have to be refactored to use a new CLI SDK function and interfaces before ZE will work again.

I open this issue so that the ZE team is aware of failures that will likely occur after they perform a routine update to a newer version of the CLI SDK.

**To Reproduce**

**Expected behavior**

**Screenshots**

**Desktop (please complete the following information):**

- OS:
- Zowe Explorer Version:
- (Optional) Zowe CLI Version:
- (Optional) Are you using Secure Credential Store?

**Additional context**

JTonda:
This will have to be addressed when we update our cli sdks.

PR #2932, opened 2024-06-04
[vNext] Update SDKs and address breaking changes

JillieBeanSim:
## Proposed changes

Update Zowe SDKs to 8.0.0-next.202407051717 and address breaking changes
#2918 

## Release Notes

Milestone:

Changelog:

## Types of changes

## Checklist


JillieBeanSim:
@t1m0thyj I will be opening an issue in @zowe/cli  for an issue I see where the converted configuration file isn't being written and also if the code to delete the @zowe/secure-credential-store plugin isn't going to be public imperative should handle the check and removal in it's APIs

@zFernand0 I tested the config file creation with this PR and still see that the base profile is missing

t1m0thyj:
> @t1m0thyj I will be opening an issue in @zowe/cli for an issue I see where the converted configuration file isn't being written (tested on mac and windows with same results) and also if the code to delete the @zowe/secure-credential-store plugin isn't going to be public imperative should handle the check and removal in it's APIs
> 
> issue: [zowe/zowe-cli#2170](https://github.com/zowe/zowe-cli/issues/2170)
> 
> @zFernand0 I tested the config file creation with this PR and still see that the base profile is missing

@JillieBeanSim Regarding the first issue, @gejohnston is investigating it.

Regarding the base profile missing, I pushed a commit that should fix this. It seems kind of redundant that the `ConfigBuilder` requires base profile definition to be passed in twice. Perhaps we could enhance the Zowe SDKs so that the old code would work 😋

zFernand0 (approved):
LGTM! 😋

packages/zowe-explorer/src/utils/ProfilesUtils.ts
```diff
@@ -571,4 +562,46 @@ export class ProfilesUtils {
         }
         throw new Error(vscode.l10n.t("Tree Item is not a Zowe Explorer item."));
     }
+
+    private static async convertV1Profs(): Promise<void> {
```
t1m0thyj, line 566: Noticed the logic for building a response message after converting v1 profiles has moved from ProfilesCache in ZE API to here. Do we think that extenders who wish to convert profiles don't need this code and plan to leave it up to them to process the response themselves?

t1m0thyj (commented):
Looks pretty good, thanks @JillieBeanSim! Left a few comments

awharn:
I tested some of the conversion functionality on Linux, and some of the output looks a bit odd once the conversion is complete. More specifically, the output appears to be very squished together. Not asking for a change, just raising a bit of awareness in case this wasn't seen in other operating systems.

image

t1m0thyj:
Pushed a commit that should make the message easier to read 😋 
image

@JillieBeanSim @zFernand0 I also ended up changing `ProfilesCache.convertV1ProfToConfig` to be a static method that takes an additional parameter. This seemed necessary because `Constants.PROFILES_CACHE` is not yet initialized at the time of converting profiles when ZE loads. Should I document this as a breaking change, or revert it?

zFernand0:
> JillieBeanSim zFernand0 I also ended up changing `ProfilesCache.convertV1ProfToConfig` to be a static method that takes an additional parameter. This seemed necessary because `Constants.PROFILES_CACHE` is not yet initialized at the time of converting profiles when ZE loads. Should I document this as a breaking change, or revert it?


Thanks for addressing the format of these messages. Curious if you had a chance to try and switch the order of these two calls in the `activate` function?
```typescript
    await ProfilesUtils.initializeZoweProfiles((msg) => ZoweExplorerExtender.showZoweConfigError(msg));
```

If the above is possible, then I think we don't need to perform this "breaking change".

---

On the other hand, I think the change is small enough that it may not be worth causing much disruption in out activation phase 😋

packages/zowe-explorer-ftp-extension/CHANGELOG.md
```diff
@@ -80,6 +81,7 @@ All notable changes to the "zowe-explorer-ftp-extension" extension will be docum
 ### Bug fixes
 
 - Fixed ECONNRESET error when trying to upload or create an empty data set member. [#2350](https://github.com/zowe/vscode-extension-for-zowe/issues/2350)
+- Update Zowe SDKs to `8.0.0-next.202407051717` for technical currency. [#2918](https://github.com/zowe/zowe-explorer-vscode/issues/2918)
```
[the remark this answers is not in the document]
JillieBeanSim: I think we can keep the pre-release version updates in CHANGELOGs, this may help extenders know the compatible CLI/SDK dependency version per Zowe Explorer pre-release.  We will create a nice clean CHANGELOG for the GA

packages/zowe-explorer/src/utils/ProfilesUtils.ts
JillieBeanSim: I believe we should keep the original response in the API call return so Extenders can manipulate the information how they wish.

packages/zowe-explorer/src/utils/ProfilesUtils.ts
```diff
@@ -504,15 +504,15 @@ export class ProfilesUtils {
...
             // eslint-disable-next-line max-len
-            "Zowe v1 profiles in use.\nZowe Explorer no longer supports v1 profiles, choose to convert existing profiles to a team configuration or create new."
+            "Zowe V1 profiles in use.\nZowe Explorer no longer supports V1 profiles. Choose to convert existing profiles to a team configuration or create new profiles."
         );
         ZoweLogger.warn(v1ProfileErrorMsg);
-        const createButton = vscode.l10n.t("Create New");
         const convertButton = vscode.l10n.t("Convert Existing Profiles");
-        Gui.infoMessage(v1ProfileErrorMsg, { items: [createButton, convertButton], vsCodeOpts: { modal: true } }).then(async (selection) => {
+        const createButton = vscode.l10n.t("Create New");
+        await Gui.infoMessage(v1ProfileErrorMsg, { items: [convertButton, createButton], vsCodeOpts: { modal: true } }).then(async (selection) => {
```
zFernand0, line 515: since we made this (`v1ProfileOptions`) function asynchronous, we should probably remove the nested `.then` here and operate on the returning value/selection of the `infoMessage` function.

zFernand0 (changes requested):
changes LGTM! 😋 
one small change requested just to clean some of the indentation/logic 😋

packages/zowe-explorer/__tests__/__unit__/utils/ProfilesUtils.unit.test.ts
```diff
@@ -347,18 +344,34 @@ describe("ProfilesUtils unit tests", () => {
...
+                    { msgFormat: imperative.ConvertMsgFmt.INDENT, msgText: "message text for testing." },
+                ],
+                profilesConverted: { zosmf: ["myzosmf"] },
+                profilesFailed: [
+                    { name: "zosmf2", type: "zosmf", error: "failed" as any },
+                    { name: null, type: "zosmf", error: "failed" as any },
+                ],
+            } as any);
+            Object.defineProperty(vscode.workspace, "openTextDocument", { value: jest.fn().mockReturnValue({}), configurable: true });
+            Object.defineProperty(Gui, "showTextDocument", { value: jest.fn(), configurable: true });
```
traeok, line 374: Using `Object.defineProperty` to define these functions might cause side-effects for other tests.
Can these be replaced with `jest.spyOn` or the use of `MockedProperty` class instead? With `MockedProperty`, you can call its dispose symbol to clean up the assignment:

```typescript
const showTextDocProperty = new MockedProperty(Gui, "showTextDocument", undefined, jest.fn());
// once the property is ready to be restored:
showTextDocProperty[Symbol.dispose]();
```

traeok (approved):
LGTM, thanks Billie! I left a comment about using `Object.defineProperty` but nothing that must be changed for this PR.

zFernand0 (approved):
LGTM! 😋
Thanks for addressing the feedback 🙏🏽

traeok (approved):

PR #2944, opened 2024-06-10
Adding the ability to switch between basic authentication and token-based authentication

SanthoshiBoyina1:
## Proposed changes
To add the ability to switch between basic authentication and token-based authentication.

## Release Notes

Milestone:

Changelog:

## Types of changes

## Checklist


adam-wolfe:
While this seems to be working, I'm wondering if there is an issue with adding tokenValue to the secure array for the z/OSMF profile when the base profile also has tokenValue in its secure array, i.e., which tokenValue would get updated if we reauthenticate in Zowe Explorer and what would be the implications for other Zowe extensions/CLI plugins that also go through the mediation layer to different services?

My config after switching from token auth to basic auth, then back to token auth for my dev.zosmf profile:

```
    "profiles": {
        "base": {
            "type": "base",
            "properties": {
                "host": "example.net",
                "port": 123,
                "rejectUnauthorized": false,
                "tokenType": "apimlAuthenticationToken"
            },
            "secure": [
                "tokenValue"
            ]
        },
        "dev": {
            "profiles": {
                "zosmf": {
                    "type": "zosmf",
                    "properties": {
                        "basePath": "/ibmzosmf/api/v1"
                    },
                    "secure": [
                        "tokenValue"
                    ]
                }
            }
        },
```

SanthoshiBoyina1:
> While this seems to be working, I'm wondering if there is an issue with adding tokenValue to the secure array for the z/OSMF profile when the base profile also has tokenValue in its secure array, i.e., which tokenValue would get updated if we reauthenticate in Zowe Explorer and what would be the implications for other Zowe extensions/CLI plugins that also go through the mediation layer to different services?
> 
> My config after switching from token auth to basic auth, then back to token auth for my dev.zosmf profile:
> 
> ```
>     "profiles": {
>         "base": {
>             "type": "base",
>             "properties": {
>                 "host": "example.net",
>                 "port": 123,
>                 "rejectUnauthorized": false,
>                 "tokenType": "apimlAuthenticationToken"
>             },
>             "secure": [
>                 "tokenValue"
>             ]
>         },
>         "dev": {
>             "profiles": {
>                 "zosmf": {
>                     "type": "zosmf",
>                     "properties": {
>                         "basePath": "/ibmzosmf/api/v1"
>                     },
>                     "secure": [
>                         "tokenValue"
>                     ]
>                 }
>             }
>         },
> ```

Especially, for the zosmf and apiml-zosmf profile I was not able to login using token-based authentication. The servers specific to IBM are not responding after the maintenance. I need to retest it and modify once the server is back.

packages/zowe-explorer/src/Profiles.ts
```diff
@@ -1237,6 +1237,120 @@ export class Profiles extends ProfilesCache {
...
+                if (creds != null) {
+                    const successMsg = localize(
+                        "handleSwitchAuthentication.switchFromTokenToBasicAuth.successful",
+                        "Login using basic authentication was successful for profile {0}.",
+                        typeof profile === "string" ? profile : profile.name
+                    );
+                    ZoweLogger.info(successMsg);
+                    Gui.showMessage(successMsg);
+                    await this.tokenAuthClearSecureArray(serviceProfile.name);
+                    await vscode.commands.executeCommand("zowe.extRefresh");
```
traeok, line 1333: Instead of doing an extension refresh, should we fire the `onProfilesUpdateEmitter` so that extenders are notified of the update?
Otherwise, running this command after switching auth. methods will reload ZE and its extenders, and they won't be notified of the change.

```suggestion
                    ZoweExplorerApiRegister.getInstance().onProfilesUpdateEmitter.fire(EventTypes.UPDATE);
```

traeok (commented):
Looks good, thanks Santhoshi! I have a quick question about the logic that's in place after switching authentication methods.

packages/zowe-explorer/src/Profiles.ts
```diff
@@ -1330,7 +1329,6 @@ export class Profiles extends ProfilesCache {
                     ZoweLogger.info(successMsg);
                     Gui.showMessage(successMsg);
                     await this.tokenAuthClearSecureArray(serviceProfile.name);
-                    await vscode.commands.executeCommand("zowe.extRefresh");
```
traeok: Can you please add this logic after line 1331 to fire the `onProfilesUpdate` event?

```ts
ZoweExplorerApiRegister.getInstance().onProfilesUpdateEmitter.fire(EventTypes.UPDATE);
```

Thanks!

traeok (changes requested):

traeok (dismissed):
LGTM, thanks @SanthoshiBoyina1!

zFernand0 (changes requested):
I also noticed the same behavior as Adam.
I think it is fine to add the token to the zosmf profile, however, I think the tokenType should also be added there instead of modifying the base profile.

I believe that some of these caveats could be addressed in a future PR, but perhaps we should avoid a possible regression (with logout not being present).

Here is a list of things that I believe could be improved:
- Change Authentication Method (CAM) should indicate the "from -> to" as it may not be clear in which direction we are going. (basic -> token or viceversa)
- If base profile already had basic-auth, performing a CAM stores the token type in the base profile and also the token value in two places (base profile and zosmf profile). This leave the config in a state where the apiml-zosmf profile is always considered to be using basic-auth.
-  the logout option is no longer present as part of this PR (related to https://github.com/zowe/zowe-explorer-vscode/issues/2633)

packages/zowe-explorer/src/Profiles.ts
```diff
@@ -1237,6 +1238,119 @@ export class Profiles extends ProfilesCache {
         }
     }
 
+    public async basicAuthClearSecureArray(profileName?: string): Promise<void> {
+        const profAttrs = await this.getProfileFromConfig(profileName);
+        const configApi = (await this.getProfileInfo()).getTeamConfig();
+        configApi.set(`${profAttrs.profLoc.jsonLoc}.secure`, ["tokenValue"]);
+        configApi.delete(`${profAttrs.profLoc.jsonLoc}.properties.user`);
+        configApi.delete(`${profAttrs.profLoc.jsonLoc}.properties.password`);
```
t1m0thyj, line 1246: This works when user & password are stored on the service profile directly, but it doesn't work if they are stored in a base profile.

Instead of hardcoding the JSON location where the secure properties are stored, we may want to use the `configApi.api.secure.securePropsForProfile` API: https://docs.zowe.org/stable/typedoc/classes/_zowe_imperative.configsecure.html#securepropsforprofile

t1m0thyj (changes requested):
Looks pretty good, thanks @SanthoshiBoyina1! Left a comment about removing user/password when switching from basic to token auth

zFernand0:
After discussion during standup, we should display a final confirmation dialog/quickpick to ensure that people do want to switch auth methods. After the confirmation, then we are good to modify the base profile.
And as a final sanity check, we should check the status of the base profile to make sure that it does not contains both auth methods (basic & token)

packages/zowe-explorer/src/Profiles.ts
t1m0thyj, line 1246: Related to Fernando's comment, when switching from basic auth to token auth this won't clear user/password if present in the base profile.

Instead of hardcoding the JSON location where the secure properties are stored, we may want to use the `configApi.api.secure.securePropsForProfile` API documented [here](https://docs.zowe.org/stable/typedoc/classes/_zowe_imperative.configsecure.html#securepropsforprofile) and warn/prompt the user before removing properties from a base profile 😋

adam-wolfe:
The functionality seems to be working in terms of switching the authentication methods in the config file. One thing I am seeing (that might be unrelated to this PR) is that the text displayed in the `Manage Profile` menu seems to not seem to reliably reflect the actual authentication method specified in the config.

Example 1 (Says I'm using basic auth, when the profile is using a token):
image

Example 2 (Says I have no authentication method, when profile is using username and password):
image

traeok:
Tested this and functionality LGTM, but also experiencing the same UX issue as Adam.
If the issue doesn't balloon too much in scope, I'm wondering if this bug could be resolved as a part of this PR? Users might be confused by the message when trying to manage their profile, as it doesn't accurately represent the authentication status.

That said, if the issue turns out to be more complex, I'd be okay with approving this and moving forward with merging 😋

JillieBeanSim:
may need to look into why Theia tests are failing on this branch, it's passing on other branches.

adam-wolfe (changes requested):
I believe the issue with detecting the wrong authentication method only occurs when using nested profiles. Also, after switching authentication methods several times, using a nested profile, `"user,password": []"` and `"tokenValue": []` are being added outside of the "profiles" object (down below defaults):

image

```
My config:
{
    "profiles": {
        "base": {
            "type": "base",
            "properties": {
                "host": "example.net",
                "port": 12345,
                "rejectUnauthorized": false
            },
            "secure": []
        },
        "dev": {
            "type": "base",
            "profiles": {
                "zosmf": {
                    "type": "zosmf",
                    "properties": {
                        "basePath": "/ibmzosmf/api/v1"
                    },
                    "secure": [
                        "user",
                        "password"
                    ]
                }
            }
        }
    },
    "defaults": {
        "zosmf": "dev.zosmf",
        "base": "base"
    },
    "autoStore": true,
    "CredentialManager": "@zowe/secure-credential-store-for-zowe-cli",
    "user,password": [],
    "tokenValue": []
}
```

adam-wolfe (approved):
Changes look good. It is correctly showing my current authentication method and I see nothing unusual happening in the config.

Thanks @SanthoshiBoyina1

traeok (approved):
Recent changes LGTM, was able to switch back and forth without any UX issues or unexpected config changes.

Thanks @SanthoshiBoyina1!

traeok (approved):

adam-wolfe (dismissed):

t1m0thyj (approved):
LGTM, thanks @SanthoshiBoyina1!

Tested this PR with a config file that had user/password in base profile. Switching first time stored token in base profile, and then switching back stored user/password directly on service profile. Although this is different from the original state of my config file, I think it is expected behavior 👍

zFernand0 (commented):
Could we wait for one more day before merging?
Just seeing a few things that may or may not be related.

t1m0thyj (approved):
LGTM, thanks @SanthoshiBoyina1!

traeok (approved):

zFernand0 (approved):
apologies for the delay.
I'll open the issue I mentioned in a few moments

t1m0thyj:
@SanthoshiBoyina1 @JillieBeanSim Should this feature be ported to the vNext branch?

SanthoshiBoyina1:
> @SanthoshiBoyina1 @JillieBeanSim Should this feature be ported to the vNext branch?

I will move this feature to vNext branch

t1m0thyj:
> I will move this feature to vNext branch

Done in #3062

PR #3002, opened 2024-07-15
updated getProfile method using latest instance of the token for PDS …

KaranP25:
…member

## Proposed changes

Issue was found in 'packages/zowe-explorer/src/dataset/ZoweDatasetNode.ts' in the openDS() method where it tries to fetch contents if local copy is not found... The code this.getProfile() is acquiring the bearer token where the token does not get refreshed after the expire time. To, fix it I saw in USS we are using "Profiles.getInstance().loadNamedProfile(this.getProfileName())" to call the profile and this does return the profile with the updated token.
https://github.com/zowe/zowe-explorer-vscode/issues/3001

## Release Notes

Milestone:

Changelog:

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] Enhancement (non-breaking change which adds or improves functionality)
- [ ] Breaking change (a change that would cause existing functionality to not work as expected)
- [ ] Documentation (Markdown, README updates)
- [ ] Other (please specify above in "Proposed changes" section)

## Checklist

**General**

- [ ] I have read the [CONTRIBUTOR GUIDANCE](https://github.com/zowe/zowe-explorer-vscode/wiki/Best-Practices:-Contributor-Guidance) wiki
- [ ] All PR dependencies have been merged and published (if applicable)
- [ ] A GIF or screenshot is included in the PR for visual changes
- [ ] The pre-publish command has been executed:
  - **v2 and below:** `yarn workspace vscode-extension-for-zowe vscode:prepublish`
  - **v3:** `pnpm --filter vscode-extension-for-zowe vscode:prepublish`

**Code coverage**

- [ ] There is coverage for the code that I have added
- [ ] I have added new test cases and they are passing
- [ ] I have manually tested the changes

**Deployment**

- [ ] I have added developer documentation (if applicable)
- [ ] Documentation should be added to Zowe Docs
  - If you're an outside contributor, please post in the [#zowe-doc Slack channel](https://openmainframeproject.slack.com/archives/CC961JYMQ) to coordinate documentation.
  - Otherwise, please check with the rest of the squad about any needed documentation before merging.
- [ ] These changes may need ported to the appropriate branches (list here):

PR #3019, opened 2024-07-23
feat(v3): Support logging in to multiple API ML instances

### LINKED ISSUE #2264: Support logging in to multiple APIML instances per config file
Currently we only support logging in to the APIML for base profiles, and there can only be one active base profile per team config file.
If nested profiles like the example below are defined, users may want to log in to multiple APIML instances per team config file.
Either Zowe Explorer should also support this scenario, or we should document it as a known limitation.

```json
{
    "profiles": {
        "lpar1": {
            "profiles": {
                "zosmf": {
                    "type": "zosmf",
                    "properties": {
                        "basePath": "/ibmzosmf/api/v1"
                    }
                }
            },
            "properties": {
                "host": "example1.com",
                "port": 7554,
                "rejectUnauthorized": true,
                "tokenType": "apimlAuthenticationToken"
            },
            "secure": [
                "tokenValue"
            ]
        },
        "lpar2": {
            "profiles": {
                "zosmf": {
                    "type": "zosmf",
                    "properties": {
                        "basePath": "/ibmzosmf/api/v1"
                    }
                }
            },
            "properties": {
                "host": "example2.com",
                "port": 7554,
                "rejectUnauthorized": true,
                "tokenType": "apimlAuthenticationToken"
            },
            "secure": [
                "tokenValue"
            ]
        }
    },
    "defaults": {
        "zosmf": "lpar1.zosmf"
    },
    "autoStore": true
}
```

t1m0thyj:
## Proposed changes

Resolves #2264 and ports #3002 to vNext branch

Also moves tests from the `packages/zowe-explorer/__tests__/__unit__/misc` folder into `Profiles.unit.test.ts` and `ZoweDatasetNode.unit.test.ts` where they seem to naturally fit

TODO:
* [x] Depends on https://github.com/zowe/zowe-cli/pull/2201
* [x] Depends on https://github.com/zowe/zowe-explorer-vscode/pull/2932
* [x] Depends on https://github.com/zowe/zowe-explorer-vscode/pull/2994
* [x] Determine desired behavior if someone has already stored token directly in zosmf profile with V2?
* [x] ~~Publish new vNext pre-release~~ (will do after #3021)
* [x] Fix behavior for multiple levels of nested profiles
* [x] Check that tokens are refreshed for PDS member nodes

## Release Notes

Milestone: V3

Changelog: Added support for logging in to multiple API ML instances per team config file

## Types of changes

- [x] Breaking change (a change that would cause existing functionality to not work as expected)

## Checklist

**General**

**Code coverage**

**Deployment**

## Further comments

The GIF below shows an example of multiple z/OSMF profiles using API ML on different LPARs in a single team config:
ze-multi-apiml


packages/zowe-explorer-api/src/profiles/ProfilesCache.ts
```diff
@@ -316,10 +316,15 @@ export class ProfilesCache {
     }
 
     // This will retrieve the base profile from imperative
-    public async fetchBaseProfile(): Promise<imperative.IProfileLoaded | undefined> {
+    public async fetchBaseProfile(profileName?: string): Promise<imperative.IProfileLoaded | undefined> {
         const mProfileInfo = await this.getProfileInfo();
         const baseProfileAttrs = mProfileInfo.getDefaultProfile("base");
-        if (baseProfileAttrs == null) {
+        const isUsingTokenAuth = (profName: string): boolean =>
+            mProfileInfo.getTeamConfig().api.secure.securePropsForProfile(profName).includes("tokenValue");
```
traeok, line 323: Is this arrow function necessary? It's only used once in the `if` statement below.

packages/zowe-explorer-api/src/vscode/ZoweVsCodeExtension.ts
```diff
@@ -212,13 +214,13 @@ export class ZoweVsCodeExtension {
         serviceProfile: string | imperative.IProfileLoaded,
         zeRegister?: Types.IApiRegisterClient, // ZoweExplorerApiRegister
         zeProfiles?: ProfilesCache // Profiles extends ProfilesCache
-    ): Promise<void> {
+    ): Promise<boolean> {
         const cache: ProfilesCache = zeProfiles ?? ZoweVsCodeExtension.profilesCache;
-        const baseProfile = await cache.fetchBaseProfile();
+        if (typeof serviceProfile === "string") {
+            serviceProfile = await ZoweVsCodeExtension.getServiceProfileForAuthPurposes(cache, serviceProfile);
+        }
```
traeok, line 221: Will this work as intended from a JavaScript/TypeScript perspective?
If `serviceProfile` is the `string` type, it won't update the parameter outside of the function. Maybe I'm misunderstanding the point of the assignment - if we don't use `serviceProfile` outside of the function, this is probably fine

traeok (changes requested):
Looks good so far, thanks Timothy! Had a couple questions regarding an arrow function and assignment of a parameter

packages/zowe-explorer-api/src/profiles/ProfilesCache.ts
t1m0thyj: Removed in 393169eeafd84442b52f4fcd03ca058b0124b398

packages/zowe-explorer-api/src/vscode/ZoweVsCodeExtension.ts
t1m0thyj: We were already setting `serviceProfile` this way, and AFAIK we don't use it outside of the function.

traeok: That's fair - feel free to disregard 🙂

packages/zowe-explorer-api/CHANGELOG.md
```diff
@@ -39,6 +39,8 @@ All notable changes to the "zowe-explorer-api" extension will be documented in t
 - Added the `onVaultUpdate` VSCode event to notify extenders when credentials are updated on the OS vault by other applications. [#2994](https://github.com/zowe/zowe-explorer-vscode/pull/2994)
 - Added the `onCredMgrsUpdate` VSCode event to notify extenders when the local PC's credential manager has been updated by other applications. [#2994](https://github.com/zowe/zowe-explorer-vscode/pull/2994)
 - **Breaking:** Updated most function signatures for exported programmatic interfaces. Changes make developing with the Zowe Explorer API more efficient for extenders by showing which properties they can expect when calling our APIs. [#2952](https://github.com/zowe/zowe-explorer-vscode/issues/2952)
+- Added support for logging in to multiple API ML instances per team config file. [#2264](https://github.com/zowe/zowe-explorer-vscode/issues/2264)
+- **Next Breaking:** Changed return type of `ZoweVsCodeExtension.logoutWithBaseProfile` method from `void` to `boolean` to indicate whether logout was successful.
```
zFernand0, line 43: Not sure if it's a breaking change when going from no return value to a return value (aside from the transpile/compile -time messages that may prevent a build from passing for extenders if they were returning the void value from this function on their own APIs) 😓

packages/zowe-explorer-api/src/vscode/ZoweVsCodeExtension.ts
```diff
@@ -180,11 +184,9 @@ export class ZoweVsCodeExtension {
             // If base profile already has a token type stored, then we check whether or not the connection details are the same
             (serviceProfile.profile.host === baseProfile.profile.host && serviceProfile.profile.port === baseProfile.profile.port);
         // If the connection details do not match, then we MUST forcefully store the token in the service profile
-        let profileToUpdate: imperative.IProfileLoaded;
+        let profileToUpdate = serviceProfile;
         if (connOk) {
-            profileToUpdate = baseProfile;
-        } else {
-            profileToUpdate = serviceProfile;
+            profileToUpdate = serviceProfile.name.startsWith(baseProfile.name + ".") ? { ...baseProfile, type: null } : baseProfile;
```
zFernand0, line 189: I guess this ternary check is in case the service profile happens to be nested inside of an real base profile and not just a nested profile (with type `null`)

packages/zowe-explorer-api/src/vscode/ZoweVsCodeExtension.ts
```diff
@@ -234,12 +236,15 @@ export class ZoweVsCodeExtension {
...
-            if (connOk) {
-                await cache.updateBaseProfileFileLogout(baseProfile);
-            } else {
-                await cache.updateBaseProfileFileLogout(serviceProfile);
-            }
+            const connOk =
+                serviceProfile.profile.host === baseProfile.profile.host &&
+                serviceProfile.profile.port === baseProfile.profile.port &&
+                !serviceProfile.name.startsWith(baseProfile.name + ".");
+            await cache.updateBaseProfileFileLogout(connOk ? baseProfile : serviceProfile);
```
zFernand0, line 243: I'm guessing that for logout purposes it doesn't matter if the service profile is nested inside an actual base profile? 😋

packages/zowe-explorer-api/src/vscode/ZoweVsCodeExtension.ts
```diff
@@ -234,12 +236,15 @@ export class ZoweVsCodeExtension {
...
+            return true;
+        } else {
+            Gui.errorMessage(`Logout failed: No base profile found to store SSO token for profile "${serviceProfile.name}"`);
```
zFernand0, line 246: Since this is the logout operation... should we change the wording a bit?
```suggestion
            Gui.errorMessage(`Logout failed: No base profile found to remove the SSO token for profile "${serviceProfile.name}"`);
```

packages/zowe-explorer-api/src/profiles/ProfilesCache.ts
```diff
@@ -316,10 +316,17 @@ export class ProfilesCache {
...
 
     // This will retrieve the base profile from imperative
-    public async fetchBaseProfile(): Promise<imperative.IProfileLoaded | undefined> {
+    public async fetchBaseProfile(profileName?: string): Promise<imperative.IProfileLoaded | undefined> {
         const mProfileInfo = await this.getProfileInfo();
         const baseProfileAttrs = mProfileInfo.getDefaultProfile("base");
-        if (baseProfileAttrs == null) {
+        const configApi = mProfileInfo.getTeamConfig().api;
+        if (
+            profileName?.includes(".") &&
```
zFernand0, line 324: I guess this is an opt-in feature. Meaning that extenders, calling the fetchBaseProfile directly, will have to provide the serviceProfile name or use our `log[in|out]WithBaseProfile` method 😋 

How should we let them know about it?
Will some TSDoc be warranted here? 😋

packages/zowe-explorer-api/src/profiles/ProfilesCache.ts
```diff
@@ -316,10 +316,17 @@ export class ProfilesCache {
...
+            (baseProfileAttrs == null || !configApi.secure.securePropsForProfile(baseProfileAttrs.profName).includes("tokenValue"))
```
zFernand0, line 325: It might just be me, but it took me a bit longer to understand what was happening here.

Would it be worth adding one or two lines describing this? 
Or maybe just a link to the diagram (in https://github.com/zowe/zowe-explorer-vscode/issues/2264#issuecomment-2236914511 - and highlight the path Yes-No-Yes 😋 )

zFernand0 (approved):
LGTM! 😋 
Left some comments/question to help me understand a bit better 😓

packages/zowe-explorer-api/src/vscode/ZoweVsCodeExtension.ts
t1m0thyj: Added comment here

packages/zowe-explorer-api/src/profiles/ProfilesCache.ts
t1m0thyj: Added TSDoc to the method and called out the change in ZE API release notes to inform extenders.

packages/zowe-explorer-api/src/profiles/ProfilesCache.ts
t1m0thyj: Added comment here

traeok (approved):
Changes LGTM - thanks Timothy for the enhancement!

awharn (commented):
How does this handle multiple levels of nesting? If a token is stored in level 1, and you do a login using level 3, does the token remain at level 1, or get moved to level 2?

t1m0thyj:
> How does this handle multiple levels of nesting? If a token is stored in level 1, and you do a login using level 3, does the token remain at level 1, or get moved to level 2?

Good catch - the token was getting moved to level 2. This should be fixed in https://github.com/zowe/zowe-explorer-vscode/pull/3019/commits/4708616435335f5ea11373a4192ccb4d8f4439f5 so the token now remains at its original level.

awharn (approved):

traeok (approved):

packages/zowe-explorer-api/src/profiles/ProfilesCache.ts
```diff
@@ -315,11 +315,28 @@ export class ProfilesCache {
         return baseProfile;
     }
 
-    // This will retrieve the base profile from imperative
-    public async fetchBaseProfile(): Promise<imperative.IProfileLoaded | undefined> {
+    /**
+     * Retrieves the base profile from Imperative to use for log in/out. If a
+     * nested profile name is specified (e.g. "lpar.zosmf"), then its parent
+     * profile is returned unless token is already stored in the base profile.
```
zFernand0, line 321: should we add this new behavior to the changelog as a new breaking change?

packages/zowe-explorer-api/src/profiles/ProfilesCache.ts
```diff
@@ -315,11 +315,28 @@ export class ProfilesCache {
...
+     * @returns IProfileLoaded object or undefined if no profile was found
+     */
+    public async fetchBaseProfile(profileName?: string): Promise<imperative.IProfileLoaded | undefined> {
         const mProfileInfo = await this.getProfileInfo();
         const baseProfileAttrs = mProfileInfo.getDefaultProfile("base");
-        if (baseProfileAttrs == null) {
+        const config = mProfileInfo.getTeamConfig();
+        if (
+            profileName?.includes(".") &&
+            (baseProfileAttrs == null || !config.api.secure.securePropsForProfile(baseProfileAttrs.profName).includes("tokenValue"))
```
zFernand0, line 331: Question about this line here.
In order for this function to return the default base profile (old behavior), the token only needs to be in the secure array of the base profile and not actually stored in the vault. Is that assumption correct?

Follow up question, Does that means that this function will return the default base profile even if I already have a token stored in the parent profile?

packages/zowe-explorer-api/src/profiles/ProfilesCache.ts
```diff
@@ -399,6 +416,20 @@ export class ProfilesCache {
         return allTypes;
     }
 
+    private getParentProfileForToken(profileName: string, config: imperative.Config): string {
+        const secureProps = config.api.secure.secureFields();
+        let parentProfile = profileName.slice(0, profileName.lastIndexOf("."));
+        let tempProfile = profileName;
+        while (tempProfile.includes(".")) {
+            tempProfile = tempProfile.slice(0, tempProfile.lastIndexOf("."));
+            if (secureProps.includes(`${config.api.profiles.getProfilePathFromName(tempProfile)}.properties.tokenValue`)) {
```
zFernand0, line 425: Nice way to handle complex nesting scenarios.

Seems like this works even for a token value stored in the QA level of `QA.Plex1.Lpar1.zosmf` 😋

zFernand0 (approved):
LGTM! 😋 
Left a few small questions, but nothing that should prevent this from being merged 😋 

Great job! 🥳

JillieBeanSim:
love the addition of this feature, the support for multiple API ML instances was a big hit when revealed for the CLI v3 at SHARE and is great to have this day 1 of ZE v3 too 😄

JillieBeanSim:
if other comments don't stop the merge, I think we can merge now but would like a CHANGELOG entry @t1m0thyj. can you share one here and it can be copied to the pre-release PR when staging release?

packages/zowe-explorer-api/src/profiles/ProfilesCache.ts
t1m0thyj: Good questions, sorry I didn't see this until after the PR was merged. Your assumption is correct 😁 

For the 2nd question, that's right. I think the reason was for backwards compatibility - however we probably want to give precedence to whichever profile has tokenValue already defined 😋

PR #3023, opened 2024-07-26
build(v3): Fix watch task to detect changes in ZE API package

t1m0thyj:
**What It Does**
Fixes regression of #2686 in the vNext branch after we migrated from `ts-loader` to `esbuild-loader`

Although `esbuild` doesn't officially support TS project references, we can trick it into supporting them by adding an `exports` section to ZE API package.json as suggested in https://github.com/evanw/esbuild/issues/1250#issuecomment-1463826174.

Also removes 2 Webpack plugins:
* `copy-webpack-plugin` - It tried to overwrite files on every incremental build which could cause file access errors on Windows. Replaced with a `copy-secrets` npm script that runs only on build.
* `tsconfig-paths-webpack-plugin` - It conflicts with the `conditionNames` configuration added in this PR.

**How to Test**
* Launch the debug task to run ZE or the FTP extension which starts a watch task in the terminal
* Make a change to source code inside the "zowe-explorer-api" package and save the file
* The watch task should automatically trigger to rebuild with code changes made to ZE API
* Set a breakpoint in a TS file in the "zowe-explorer-api" package and make sure it triggers

### DIFF
--- a/package.json
+++ b/package.json
@@ -51,7 +51,6 @@
     "rimraf": "^3.0.2",
     "terser-webpack-plugin": "^5.3.10",
     "ts-jest": "^29.0.3",
-    "tsconfig-paths-webpack-plugin": "^4.1.0",
     "tsx": "^4.9.3",
     "typescript": "^5.3.3",
     "webpack": "^5.89.0",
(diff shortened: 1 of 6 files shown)


traeok (approved):
Changes LGTM, was able to debug the API after making a change. Thanks Timothy for resolving this & the Secrets SDK binary issue 😋

PR #3027, opened 2024-07-30
feat: Table view framework and related facilities 

traeok:
## Proposed changes

- [x] Implement an extensible and customizable table view class (`Table.View`)
- [x] Add a builder facility for quickly building tables (`TableBuilder`)
- [x] Mediator facility for exposing tables to Zowe Explorer and other extenders (`TableMediator`)
- [x] Set up a panel view for showing tables in a fixed spot in Zowe Explorer 
  - [x] Expose a facility for updating the table in this view: `TableViewProvider`
- [x] Patch coverage for all mentioned facilities

### Unrelated changes

- [x] Fix small code warning with `zedc` for `zip_path` being unused

Note that some slight adjustments to this framework may occur to fully support the tabular jobs view in a separate branch. This branch only contains the framework and data structures for making, exposing and rendering generic tables.

[Dev. documentation on wiki](https://github.com/zowe/zowe-explorer-vscode/wiki/Table-Views)

## Release Notes

Milestone: vNext

Changelog:

- Implements the ability to render tables in Zowe Explorer
- Add table builder and mediator to facilitate customization and extensibility for table views

## Types of changes

- [x] Documentation (Markdown, README updates)

## Checklist

**General**

**Code coverage**

**Deployment**

  - Link: https://github.com/zowe/zowe-explorer-vscode/wiki/Table-Views


traeok:
The implementation of the `TableViewProvider` and some adjustments to the `WebView` class has brought down the code coverage. Going to work on this now and will mark as ready once resolved.

packages/zowe-explorer-api/__mocks__/vscode.ts
zFernand0, line 1: I know it's likely that the good TSDoc in this file is due to us copying the contents directly from VSCode's type definitions. 
However, I'm wondering if this is truly needed for our mock?

And a follow up question, why not have it (this great level of TSDoc) for the rest of our developer-facing interfaces and functions?  😋

packages/zowe-explorer-api/src/vscode/ui/utils/TableBuilder.ts
```diff
@@ -0,0 +1,224 @@
...
+ * ## Sharing tables
+ *
+ * Share a table during the build process by using the {@link TableBuilder.buildAndShare} function.
+ * This function will create the {@link Table.View} instance while also adding it to the mediator
+ * in Zowe Explorer's extender API. Extenders who would like to contribute to shared tables can do
+ * so by accessing the table by its unique ID.
+ */
+export class TableBuilder {
+    private context: ExtensionContext;
+    private data: Table.ViewOpts = {
```
zFernand0, line 37: Wondering if we should just leave this as undefined here and call `this.reset()` as part of the constructor.

it's just to reduce duplication 😋 

I know it's possible to want to access the data of a tableBuilder without initializing it first, so maybe this is not that important to change 😋

packages/zowe-explorer-api/src/vscode/ui/utils/TableBuilder.ts
```diff
@@ -0,0 +1,224 @@
...
+        this.data = { ...this.data, options: this.data.options ? { ...this.data.options, ...opts } : opts };
+        return this;
+    }
+
+    /**
+     * Set the title for the next table.
+     * @param name The name of the table
+     * @returns The same {@link TableBuilder} instance with the title added
+     */
+    public title(name: string): this {
```
zFernand0, line 74: Would it make sense to make this a setter? This also applies to a few other properties here.

packages/zowe-explorer-api/src/vscode/ui/utils/TableBuilder.ts
```diff
@@ -0,0 +1,224 @@
...
+    }
+
+    /**
+     * Adds rows to the table. Does not replace existing rows.
+     * @param rows The rows of content to add to the table
+     * @returns The same {@link TableBuilder} instance with the new rows added
+     */
+    public addRows(rows: Table.RowData[]): this {
+        this.data.rows = [...this.data.rows, ...rows];
+        return this;
```
zFernand0, line 96: I like that most (if not all) of these functions return `this`.
It helps to concatenate methods calls much faster.
I'm wondering if we should do more of `this` going forward 😋

packages/zowe-explorer-api/src/vscode/ui/utils/TableBuilder.ts
```diff
@@ -0,0 +1,224 @@
...
+        }
+        return this;
+    }
+
+    /**
+     * Add a row action to the next table.
+     * @param index The column index to add an action to
+     * @returns The same {@link TableBuilder} instance with the row action added
+     */
+    public addRowAction(index: number | "all", action: Table.ActionOpts): this {
```
zFernand0, line 175: I like the thought of passing in `all`

was this supposed to allow for all rows to have the same action?
If so, should we be checking for it?

I might be missing something here. 🤔

zFernand0 (changes requested):
First round of `weviews` comments, many more to come. 😋 
Code is LGTM so far! 🥳

packages/zowe-explorer-api/src/vscode/ui/utils/TableBuilder.ts
traeok: Good question - discussed offline and we do support the "all" keyword as an "index" into the record. It is strange but technically saves memory and keeps all the actions in one contiguous structure 😋

packages/zowe-explorer-api/src/vscode/ui/utils/TableBuilder.ts
traeok: Thanks! I figured the builder pattern would be useful here, but there might be a few other places in the codebase where this could be helpful 😁

packages/zowe-explorer-api/src/vscode/ui/utils/TableBuilder.ts
traeok: Currently all of the `TableBuilder` functions are meant to return the same instance, that way you can chain several methods to get the final table object:

```typescript
const table = new TableBuilder()
    .title("A table")
    .rows([{ a: 1, b: 2, c: 3 ])
    .columns([{ field: "a" }, { field: "b" }, { field: "c" }])
    .options({ pagination: false, paginationPageSize: 0 })
    .build();
```
By changing it to a setter, you would have to save the builder to an object, insert a new line to assign the title, and then continue re-assigning that object.

That being said, if we feel that its more appropriate to use setters for functions for specific properties (such as title), I wouldn't be opposed to incorporating those changes 😋

packages/zowe-explorer-api/src/vscode/ui/utils/TableBuilder.ts
traeok: I believe I assigned `data` to be an "empty-like" structure so that by default, the TableBuilder renders an empty table with no contents or title. If we set `data` to `undefined`, a user can create a TableBuilder instance and then call `build` on it, but the resulting table would not render any contents in the editor or panel.

packages/zowe-explorer-api/__mocks__/vscode.ts
zFernand0: yeah, let's not increase the scope of this PR since we don't know what else may happen during the mocks TSDoc cleanup 😋

packages/zowe-explorer-api/src/vscode/ui/utils/TableBuilder.ts
zFernand0: How would one create a TableBuilder instance?
`new TableBuilder(context)` ?


If that's the case, and assuming we add a `this.reset()` call in the constructor...
```
    public constructor(context: ExtensionContext) {
        this.reset();
        this.context = context;
    }
```
 I think that `this.data` will be populated if even if the default value is not defined at the class level.
 
 ---
 
 Otherwise, where do we plan to use the `reset()` function ourselves?  (playing devil's advocate) 😈

packages/zowe-explorer-api/src/vscode/ui/utils/TableBuilder.ts
zFernand0: Let's just keep it as-is. I think this is just a nuanced difference in the code style/structure.
I do like the concatenation as it reminds me of other personal, more structurally relaxed projects 😋 

---

> By changing it to a setter, you would have to save the builder to a let-scoped variable, insert a new line to assign the title, and then continue to re-assign that variable

The above is only true if we has a combination of these setters and concatenation-friendly methods. I understand the convenience of returning `this` everywhere, it's just a slightly different paradigm of how the rest of the codebase might be structured.

Technically having them all as setters may look something like this:

```typescript
const builder = new TableBuilder(); 
builder.title = "A table"; 
builder.rows = [ { a: 1, b: 2, c: 3} ] 
builder.columns = [{ field: "a" }, { field: "b" }, { field: "c" }]
builder.options = { pagination: false, paginationPageSize: 0 };
const table = builder.build();
```

And a slightly different approach might be to allow for the table properties to be provided at the constructor level instead of having to concatenate calls, or split statements.

```typescript
const table = new TableBuilder({ 
  title: "A table",
  rows: [ { a: 1, b: 2, c: 3} ],
  columns: [{ field: "a" }, { field: "b" }, { field: "c" }],
  options: { pagination: false, paginationPageSize: 0 },
}).build();
```

packages/zowe-explorer-api/src/vscode/ui/utils/TableBuilder.ts
traeok: That's a good point, sorry I overlooked the aspect of adding a `this.reset` call - I'll adjust that so that it uses reset instead 😋 

Developers (such as ourselves or extenders) can also use the `reset` function to clean up a builder between creation of two or more tables. It could also be useful to keep around in-memory and leveraged for making similar tables, calling `reset` between each build.

packages/zowe-explorer-api/src/vscode/ui/utils/TableBuilder.ts
traeok: My approach here was definitely a bit biased because it feels similar to coding in Rust, where libraries often use this pattern to allow chaining functions on one line 😆 That said, the other 2 options are definitely viable.

Something to keep in mind for the "setter" approach is that we'd also need to implement getters for each "property" to ensure some capabilities are still supported (e.g. adding new rows vs. replacing all rows). Definitely not a dealbreaker, just something to consider.

Also, the constructor approach definitely feels more TypeScript-like, but prevents making changes without constructing a new object.

packages/zowe-explorer-api/src/vscode/ui/utils/TableMediator.ts
```diff
@@ -0,0 +1,102 @@
...
+     */
+    public getTable(id: string): Table.View | undefined {
+        return this.tables.get(id);
+    }
+
+    /**
+     * Removes a table from the mediator.
+     *
+     * @param id The unique ID of the table to delete
+     * @returns `true` if the table was deleted; `false` otherwise
```
zFernand0, line 93: No changes requested here

---

I'm always conflicted about the return value of a delete-like method.
Depending on the use case:
- Someone might appreciate getting the deleted value back (like a `Stack.pop` operation)
- Or someone else might want `true` if the resource (table) didn't exist.

That said, I don't believe we use the return value of removeTable in the TableView.
So, it's fine to keep as is 😋

packages/zowe-explorer/src/extension.ts
```diff
@@ -43,6 +44,8 @@ export async function activate(context: vscode.ExtensionContext): Promise<ZoweEx
         uss: () => USSInit.initUSSProvider(context),
         job: () => JobInit.initJobsProvider(context),
     });
+
+    context.subscriptions.push(vscode.window.registerWebviewViewProvider("zowe-resources", TableViewProvider.getInstance()));
```
zFernand0, line 48: Not sure how many more webviewView providers we may have in the future, but should we move this to an `initViewProviders` method or maybe under the `initializeProviders` method?

The only reason this crossed my mind was because this will be the only subscription directly in the activate method. Not saying that it's a bad thing, it's just that I'm afraid this might set a precedence for increasing the complexity of the `activate` /main method. 😋

packages/zowe-explorer-api/src/vscode/ui/WebView.ts
```diff
@@ -80,18 +128,21 @@ export class WebView {
...
 
     /**
      * Disposes of the webview instance
      */
-    private dispose(): void {
-        this.panel.dispose();
+    protected dispose(): void {
+        if (this.panel) {
+            this.panel.dispose();
+        }
```
zFernand0, line 145: What do you think of...

```suggestion
        this.panel?.dispose();
```

packages/zowe-explorer-api/src/vscode/ui/utils/TableBuilder.ts
traeok: Thanks for the suggestion - addressed in 78e27ba

packages/zowe-explorer-api/src/vscode/ui/WebView.ts
```diff
@@ -43,35 +61,65 @@ export class WebView {
...
      * @param onDidReceiveMessage Event callback: called when messages are received from the webview
      */
-    public constructor(
-        title: string,
-        webviewName: string,
-        context: ExtensionContext,
-        onDidReceiveMessage?: (message: object) => void | Promise<void>,
-        retainContext?: boolean
-    ) {
+    public constructor(title: string, webviewName: string, context: ExtensionContext, opts?: WebViewOpts) {
```
zFernand0, line 64: Is this considered a breaking change? If so, should we add it to the changelog?

packages/zowe-explorer/src/webviews/src/utils.ts
```diff
@@ -0,0 +1,38 @@
...
+ *
+ */
+
+import { useEffect } from "preact/hooks";
+
+export function getVsCodeTheme(): string | null {
+    return document.body.getAttribute("data-vscode-theme-kind");
+}
+
+export const useMutableObserver = (target: Node, callback: MutationCallback, options: MutationObserverInit | undefined = undefined): void => {
```
zFernand0, line 18: is it ok to remove the default value for options here?
I believe that it will be `undefined` if not provided 😋

packages/zowe-explorer/src/trees/job/JobActions.ts
```diff
@@ -457,9 +454,15 @@ export class JobActions {
...
+        for (const session of sessionNodes) {
+            session.dirty = true;
+            await session.getChildren();
+            jobsProvider.refreshElement(session);
+        }
+
         if (failedJobs.length > 0) {
             // Display any errors from the API
-            await Gui.warningMessage(
+            Gui.warningMessage(
```
zFernand0, line 465: Should we leave a comment in the code as to way we are not awaiting these methods?
And maybe disable the eslint rule for that line 😋

packages/zowe-explorer-api/src/vscode/ui/WebView.ts
traeok: Addressed in ffc40b08

packages/zowe-explorer/src/webviews/src/utils.ts
traeok: Resolved in f8c068de

packages/zowe-explorer/src/trees/job/JobActions.ts
traeok: Solved in f8c068de

packages/zowe-explorer/src/extension.ts
traeok: That makes sense to me - I'll move it into the `registerCommonCommands` function for now, as other webviews are defined there. If we choose to implement another `WebviewViewProvider` at some point, we can create an `initViewProviders` method to consolidate the initialization & registration of those providers.

traeok: addressed in 861bb2c 😋

packages/zowe-explorer-api/src/vscode/ui/utils/TableMediator.ts
traeok: I could see value in both return types - this was modeled after the `Map.prototype.delete()` function, which returns `true/false` when deleting a key from the map. I figured that most users will not want the table view after removing it - in the case where they want to quickly create table views with a similar structure, they can hold onto a builder and then re-define its data before calling `TableBuilder.build`.

zFernand0 (approved):
I get some minor changes when running the prepublish script.
but aside from that, and all the other comments (which where mostly dumb questions for my own understanding)...
I do believe this LGTMerge! 😋

packages/zowe-explorer-api/src/vscode/ui/utils/TableBuilder.ts
zFernand0: I struggled a bit in my head on this one.
But I think this is what swayed me into keeping it as-is (with the method-chaining approach) was this comment right here...

> the constructor approach feels more akin to a typical TypeScript utility, but it prevents making changes without constructing a new builder object.

Thanks for the discussion Trae 🙏🏽 
And sorry for being difficult 😅

packages/zowe-explorer/src/webviews/src/utils.ts
```diff
@@ -0,0 +1,38 @@
...
+    const isWebUser =
+        (eventUrl.protocol === document.location.protocol && eventUrl.hostname === document.location.hostname) ||
+        eventUrl.hostname.endsWith(".github.dev");
+    const isLocalVSCodeUser = eventUrl.protocol === "vscode-webview:";
+
+    if (!isWebUser && !isLocalVSCodeUser) {
+        return false;
+    }
+
+    return true;
```
t1m0thyj, line 37: Could this be simplified? I believe `A || B` is identical to `!(!A && !B)` 😋 
```suggestion
    return isWebUser || isLocalVSCodeUser;
```

packages/zowe-explorer-api/package.json
```diff
@@ -37,7 +37,9 @@
     "@zowe/zos-tso-for-zowe-sdk": "8.0.0-next.202407232256",
     "@zowe/zos-uss-for-zowe-sdk": "8.0.0-next.202407232256",
     "@zowe/zosmf-for-zowe-sdk": "8.0.0-next.202407232256",
+    "deep-object-diff": "^1.1.9",
     "mustache": "^4.2.0",
+    "preact": "^10.16.0",
```
t1m0thyj, line 42: Do we need `preact` dependency in ZE API package? I don't see it imported anywhere inside the `packages/zowe-explorer-api` folder but am probably missing something 😋

packages/zowe-explorer-api/src/vscode/ui/TableView.ts
```diff
@@ -0,0 +1,571 @@
...
+
+        /**
+         * Access the unique ID for the table view instance.
+         *
+         * @returns The unique ID for this table view
+         */
+        public getId(): string {
+            if (this.uuid == null) {
+                this.uuid = randomUUID();
+            }
```
t1m0thyj, line 414: Not requesting to change anything, but the nullish coalescing assignment operator could be used here 😀 
```suggestion
            this.uuid ??= randomUUID();
```

packages/zowe-explorer-api/src/vscode/ui/TableView.ts
```diff
@@ -0,0 +1,571 @@
...
+            return this.updateWebview();
+        }
+
+        /**
+         * Update an existing row in the table view.
+         * @param index The AG GRID row index to update within the table
+         * @param row The new row content
+         * @returns Whether the webview successfully updated the new row
+         */
+        public async updateRow(index: number, row: RowData | null): Promise<boolean> {
```
t1m0thyj, line 478: Should we mention in the Typedoc for this method, that if `row` is null then the row will be deleted?

zedc/src/code/prepare.rs
```diff
@@ -81,6 +81,7 @@ fn code_cli_binary(dir: &Path) -> PathBuf {
 /// * `zip_path` - The path of the ZIP file to extract
 /// * `vsc_path` - The path where the archive should be extracted into
 ///
+#[allow(unused_variables)]
```
t1m0thyj, line 84: Thanks for adding this, I've noticed Rust Analyzer reporting it even without the file being open 😀

t1m0thyj (commented):
Looks pretty good, thanks @traeok! Left a few comments

packages/zowe-explorer-api/src/vscode/ui/utils/HTMLTemplate.ts
```diff
@@ -21,7 +21,7 @@ const HTMLTemplate: string = `
         <meta name="viewport" content="width=device-width, initial-scale=1.0" />
         <meta 
             http-equiv="Content-Security-Policy" 
-            content="default-src 'none'; font-src data:; img-src data: vscode-resource: https:; script-src 'nonce-{{ nonce }}';
+            content="default-src 'none'; font-src data:; img-src data: vscode-resource: https:; script-src 'unsafe-eval' 'nonce-{{ nonce }}';
```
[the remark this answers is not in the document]
traeok: That's correct. I think that's a good choice, I'll add some logic to handle that 👍

packages/zowe-explorer-api/src/vscode/ui/utils/TableBuilder.ts
```diff
@@ -0,0 +1,215 @@
...
+    private context: ExtensionContext;
+    private data: Table.ViewOpts;
+    private forWebviewView = false;
+
+    public constructor(context: ExtensionContext) {
+        this.reset();
+        this.context = context;
+    }
+
+    public isView(): this {
```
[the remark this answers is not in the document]
traeok: resolved in c9a3061

packages/zowe-explorer-api/src/vscode/ui/TableView.ts
```diff
@@ -0,0 +1,571 @@
...
+        filter?: boolean;
+        floatingFilter?: boolean;
+
+        // Headers
+
+        // "field" variable will be used as header name if not provided
+        headerName?: string;
+        headerTooltip?: string;
+        headerClass?: string | string[];
+        weapHeaderText?: boolean;
```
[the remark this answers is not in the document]
traeok: Thanks for catching this - resolved in c9a3061

packages/zowe-explorer-api/src/vscode/ui/TableView.ts
```diff
@@ -0,0 +1,571 @@
...
+    export type ActionOpts = Omit<Action, "condition"> & { condition?: Conditional };
+    export type ContextMenuOpts = Omit<ContextMenuOption, "condition"> & { condition?: Conditional };
+
+    // -- Misc types --
+    /** Value formatter callback. Expects the exact display value to be returned. */
+    export type ValueFormatter = (data: { value: ContentTypes }) => string;
+    export type Positions = "left" | "right";
+
+    /** The column type definition. All available properties are offered for AG Grid columns. */
+    export type Column = {
```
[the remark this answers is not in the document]
traeok: They do, but we would have to add `ag-grid-react` as a dependency of Zowe Explorer API, which would add ~1MB to all packages with it in the dependency tree. By replicating a similar set of properties, we can avoid adding the dependency just for type definitions.

packages/zowe-explorer-api/src/vscode/ui/utils/TableMediator.ts
```diff
@@ -0,0 +1,102 @@
...
+     */
+    public removeTable(instance: Table.Instance): boolean {
+        if (Array.from(this.tables.values()).find((table) => table.getId() === instance.getId()) == null) {
+            return false;
+        }
```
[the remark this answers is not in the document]
traeok: That's a good point - resolved in upcoming commit 👍

packages/zowe-explorer/src/webviews/src/utils.ts
traeok: This logic was already in the codebase from previous work, but I agree - will adjust 👍

packages/zowe-explorer-api/package.json
traeok: I don't believe so - I initially added this, with the plan that the API would have functional components as types for specific properties, but since I did not go that route, the dependency is unnecessary here.

Addressed in upcoming commit 👍

packages/zowe-explorer-api/src/vscode/ui/TableView.ts
traeok: Yes, good call, thanks - addressed in upcoming commit 👍

packages/zowe-explorer/src/webviews/src/utils.ts
traeok: Addressed in 4620206

packages/zowe-explorer-api/CHANGELOG.md
```diff
@@ -39,6 +39,8 @@ All notable changes to the "zowe-explorer-api" extension will be documented in t
...
 - Added the `onCredMgrsUpdate` VSCode event to notify extenders when the local PC's credential manager has been updated by other applications. [#2994](https://github.com/zowe/zowe-explorer-vscode/pull/2994)
 - **Breaking:** Updated most function signatures for exported programmatic interfaces. Changes make developing with the Zowe Explorer API more efficient for extenders by showing which properties they can expect when calling our APIs. [#2952](https://github.com/zowe/zowe-explorer-vscode/issues/2952)
+- Implemented support for building, exposing and displaying table views within Zowe Explorer. Tables can be customized and exposed using the helper facilities (`TableBuilder` and `TableMediator`) for an extender's specific use case. For more information on how to configure and show tables, please refer to the [wiki article on Table Views](https://github.com/zowe/zowe-explorer-vscode/wiki/Table-Views). [#2258](https://github.com/zowe/zowe-explorer-vscode/issues/2258)
+- **Breaking:** Consolidated WebView API options into a single object (`WebViewOpts` type), both for developer convenience and to support future options.
```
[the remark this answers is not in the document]
traeok: Done!

packages/zowe-explorer-api/src/vscode/ui/utils/HTMLTemplate.ts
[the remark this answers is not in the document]
traeok: Implemented in [67407c0](https://github.com/zowe/zowe-explorer-vscode/pull/3027/commits/67407c071ae76333114620540d2a2643426c4ec0)

traeok:
Hi @JillieBeanSim, sorry for the confusion - this PR implements the table and related facilities. We don't have a GitHub issue specific for the table framework, but I've attached it to the "jobs table" issue as it implements the generic components that will be used for it. To make it a bit easier for reviews, I have been working on the jobs table view in a separate PR that I'll be creating some time today/tomorrow 😁

JillieBeanSim (approved):
thanks @traeok for quick response, I found my issue with jobs search I mentioned so no worries for that, it was with my config.  Code looks good and can address any issues with follow up PR. Thanks!

packages/zowe-explorer/src/webviews/package.json
```diff
@@ -11,18 +11,22 @@
     "preview": "vite preview",
     "fresh-clone": "pnpm clean && rimraf node_modules",
     "clean": "rimraf dist || true",
-    "package": "echo \"webviews: nothing to package.\"",
+    "package": "node -e \"fs.accessSync(path.join(__dirname, 'dist'))\" && echo \"webviews: nothing to package.\" || pnpm build",
```
t1m0thyj, line 14: No change required, but we may be able to use a relative path here since I think npm scripts already run in the context of the current directory 😋 
```suggestion
    "package": "node -e \"fs.accessSync('dist')\" && echo \"webviews: nothing to package.\" || pnpm build",
```

t1m0thyj (approved):
LGTM, thanks @traeok!

PR #3093, opened 2024-09-05
fix(list): Do not fetch directory URIs in `stat`

traeok:
## Proposed changes

There was an oversight when implementing the remote lookup functionality for Data Sets and USS. The `stat` implementations should never make an API call unless we are explicitly fetching the URI with the `fetch` query, or if the entry is a file (`stat` is called for files open in the editor).

Considering this causes a balloon in API calls when listing data sets and USS folders, it would be best to get this fix into the upcoming pre-release, or at the very least get the fix merged before v3 GA so that our users don't encounter this problem 😅 

**How to test:** 
**NOTE:** _Please test on a dev LPAR as this may spam the mainframe with API calls_
List a large data set in Zowe Explorer. In this PR it will load very quickly - in the last pre-release it will take quite some time to complete (or it won't finish loading at all)

## Release Notes

Milestone: vNext

Changelog:

- Fixed issue where listing data sets or USS files would cause a drastic increase in API calls, causing delays or a complete halt in Zowe Explorer.

## Types of changes

## Checklist

**General**

**Code coverage**

**Deployment**


packages/zowe-explorer/src/trees/uss/UssFSProvider.ts
```diff
@@ -133,7 +135,7 @@ export class UssFSProvider extends BaseProvider implements vscode.FileSystemProv
         }
         const response = await ZoweExplorerApiRegister.getUssApi(profile).fileList(ussPath);
         // If request was successful, create directories for the path if it doesn't exist
-        if (!this.exists(uri)) {
+        if (response.success && !(keepRelative ?? false) && response.apiResponse.items?.[0]?.mode?.startsWith("d") && !this.exists(uri)) {
```
zFernand0, line 138: Curious why we don't just default the keepRelative parameter to false in the function signature here 😋 
Maybe that way we avoid a few operators 😋

packages/zowe-explorer/src/trees/uss/UssFSProvider.ts
```diff
@@ -161,6 +163,22 @@ export class UssFSProvider extends BaseProvider implements vscode.FileSystemProv
...
             return entry;
         }
+        if (entry == null && resp?.success) {
+            // if entry is null, listFiles did not create a new directory entry - this is a file
+            let parentDir = this._lookupParentDirectory(uri, true);
+            if (parentDir == null) {
+                const parentPath = path.posix.join(uri.path, "..");
+                const parentUri = uri.with({ path: parentPath });
+                await vscode.workspace.fs.createDirectory(parentUri);
+                parentDir = this._lookupParentDirectory(uri, false);
```
zFernand0, line 173: I assume this will throw if the immediate parent (`..`) is not a directory.
And that's why we don't need to `parentDir?.metadata =`

packages/zowe-explorer/src/trees/dataset/DatasetFSProvider.ts
```diff
@@ -218,11 +220,13 @@ export class DatasetFSProvider extends BaseProvider implements vscode.FileSystem
...
         let entryIsDir = entry != null ? entry.type === vscode.FileType.Directory : false;
+        const uriPath = uri.path.substring(uriInfo.slashAfterProfilePos + 1).split("/");
+        const pdsMember = uriPath.length === 2;
         if (!entryExists) {
-            const resp = await ZoweExplorerApiRegister.getMvsApi(uriInfo.profile).dataSet(path.posix.basename(uri.path), { attributes: true });
+            const resp = await ZoweExplorerApiRegister.getMvsApi(uriInfo.profile).dataSet(uriPath[0], { attributes: true });
             if (resp.success) {
                 const dsorg: string = resp.apiResponse?.items?.[0]?.dsorg;
-                entryIsDir = dsorg?.startsWith("PO") ?? false;
+                entryIsDir = pdsMember ? false : dsorg?.startsWith("PO") ?? false;
```
zFernand0, line 229: Interestingly enough, the `??` gets evaluated first, then the `? :`
That's why this works without having to add parenthesis.

Curious if we should enforce parenthesis with an eslint rule on mixed operators 😋

packages/zowe-explorer/src/trees/dataset/DatasetFSProvider.ts
```diff
@@ -380,7 +389,28 @@ export class DatasetFSProvider extends BaseProvider implements vscode.FileSystem
      * @returns The data set's contents as an array of bytes
      */
     public async readFile(uri: vscode.Uri): Promise<Uint8Array> {
-        const file = this._lookupAsFile(uri);
+        let ds: DsEntry | DirEntry;
```
zFernand0, line 392: curious if it would be possible to combine some of the try/catch logic between the DS and USS file providers on their readFile functions 😋 

it feels somewhat similar 😋

zFernand0 (approved):
Changes make sense to me! 🙏🏽 
Thanks for addressing the issue so quickly! 🥳 
LGTM! 😋

packages/zowe-explorer/src/trees/uss/UssFSProvider.ts
traeok: Yep that's correct - `createDirectory` will throw if the parent already exists as a file entry.

packages/zowe-explorer/src/trees/dataset/DatasetFSProvider.ts
traeok: Hmm.. I intended on the `? :` being evaluated first, but didn't consider that `??` might take priority.
That said, it seems to have the same effect - if `pdsMember` is true, the entry is not a dir - otherwise it will check the dsorg.

packages/zowe-explorer/src/trees/uss/UssFSProvider.ts
traeok: I agree, writing the fallback to false felt weird even though it has a similar effect to the default parameter value 😄
I think it makes sense to fix this here for developer clarity - resolved in commit fd904aba

packages/zowe-explorer/src/trees/dataset/DatasetFSProvider.ts
traeok: I agree, I think some of this logic can be consolidated into "abstract" filesystem functions and reused elsewhere. While I'd like to look into that, it'll probably expand the scope of the PR, so I can create an issue to track that 🙂

packages/zowe-explorer/src/trees/dataset/DatasetFSProvider.ts
zFernand0: a quick way to test is:
- prints false: `node  -e "console.log((true ? undefined : null) ?? false)"` 
- prints undefined: `node  -e "console.log(true ? undefined : null ?? false)"`

zFernand0 (approved):
LGTM! 😋

anaxceron (approved):

packages/zowe-explorer/src/trees/dataset/DatasetFSProvider.ts
traeok: Filed as #3099

JillieBeanSim (approved):
This works great and solves the hang mentioned on call last week retrieving data set members.  Thanks @traeok for the fix!

Issue #3121, opened 2024-09-18
Do not specify language IDs for data sets opened in the Zowe Explorer v3 tree view


benjamin-t-santos:
**Is your feature request related to a problem? Please describe.**

In v2, Zowe Explorer downloaded sequential data sets to temporary on-disk storage. These downloaded files had extensions (`.cbl`, `.pli`, etc.) that allowed extenders to associate the extension with a language.

In v3, data sets are associated with a language by Zowe Explorer using the VS Code API. In `onDidOpenTextDocument`of `DatasetFSProvider`, opened documents have their language set to the ID returned by `getLanguageId(parentPath)`. For example, PL/I programs have their language set to `pli`.

Not adding a file extension prevents extenders or users from specifying their own [file associations](https://code.visualstudio.com/docs/languages/overview#_add-a-file-extension-to-a-language). Setting the ID forces extenders to register languages in their extension manifest with the same language IDs as Zowe Explorer to have their extension recognize data sets opened in the tree view.

**Describe the solution you'd like**

Zowe Explorer should not decide the language ID for data sets. Extenders should decide this - either by providing an extension for data sets opened in Zowe Explorer (as was done in v2) or by granting users access to the `DatasetFSProvider` to specify their own language IDs.

**Describe alternatives you've considered**

You could request extenders to register new languages that use the exact IDs Zowe Explorer uses. I am not sure how else to satisfy the file associations problem.

**Additional context**

If a solution is agreed upon, I'd be happy to help contribute.

phaumer:
Zowe Explorer should be completely language agnostic without trying to guess extensions or language ids.

Language ids are case sensitive and we had discussion in the past between the vendors about which variant to use to avoid conflicts. E.g. in this issue it was mentioned that Microfocus was using `COBOL`, IBM was using `cobol` and Broadcom want to use a different name such as `Cobol` or `che4z.cobol` (decision was not posted): <https://github.com/IBM/zopeneditor-about/issues/70>

traeok:
Hi all,

After discussing with the squad, it makes the most sense to use the previous behavior from v2 where we assigned a file extension to data set resources. I've made some changes in #3127 to add those extensions back - this should allow LSPs to continue leveraging file associations as usual. 

As a side note, we would like to hide the file extension from the breadcrumbs/tab for a cleaner UX, but we'll have to wait for the custom tab labels to be available for developers.

PR #3127, opened 2024-09-19
fix: add extensions back to Data Set resources; resolve USS copy/paste bug

traeok:
## Proposed changes

- Fixed issue where file extensions were removed from data sets, causing language detection to sometimes fail for Zowe Explorer extenders. [#3121](https://github.com/zowe/zowe-explorer-vscode/issues/3121)
- Fixed an issue where copying and pasting a file/folder in the USS tree would fail abruptly, displaying an error. [#3128](https://github.com/zowe/zowe-explorer-vscode/issues/3128)

## Release Notes

Milestone: vNext

Changelog: See `Proposed changes`

## Types of changes

## Checklist

**General**

**Code coverage**

**Deployment**


phaumer:
Why are even the file extensions required? If a user stores JCL file in a data set that has COB or CBL in the name Zowe Explorer automatically assumes it is a COBOL file. If we leave Zowe Explorer agnostic then the user can define their own rules in "file.associations" using glob patterns there. Is there a limitation in the new virtual file system that requires using extensions?

traeok:
> Why are even the file extensions required? If a user stores a JCL file in a data set that has COB or CBL in the name Zowe Explorer automatically assumes it is a COBOL file. If we leave Zowe Explorer agnostic then the user can define their own rules in "file.associations" using glob patterns there. Is there a limitation in the new virtual file system that requires using extensions?
> 
> I know we had this in v1 and v2, but now is the opportunity to remove it (perhaps with a switch to turn it back on).

@phaumer They aren't necessarily required, but this means that all users will have to define file associations for their data sets in order to get language detection. Adding a file extension allows VS Code to determine which LSPs are viable. Alternatively, we can assign a language ID, but since we can't determine which language IDs are viable, the team felt it made sense to leverage the current behavior exhibited in v2 (adding file extensions).

Ideally, we can get rid of this altogether, but this would be a breaking change for both users and extenders.

packages/zowe-explorer-api/src/fs/types/datasets.ts
```diff
@@ -47,8 +47,18 @@ export class DsEntryMetadata implements EntryMetadata {
         this.path = metadata.path;
     }
 
+    private extensionRemovedFromPath(): string {
+        for (const ext of [".c", ".jcl", ".cbl", ".cpy", ".inc", ".pli", ".sh", ".rexx", ".xml", ".asm", ".log"]) {
+            if (this.path.endsWith(ext)) {
+                return this.path.replace(ext, "");
+            }
+        }
```
zFernand0, line 55: should we loop through the list, or just remove the last extension if found in the path?

example:
`test.log.rexx.jcl` will result in `test` 😢 

I know that's a bit unrealistic, but how about a dataset that contains log related information for a JCL job that executes REXX 😅 

`HLQ.LOG.REXX.JCL`

packages/zowe-explorer-api/src/fs/types/datasets.ts
```diff
@@ -47,8 +47,18 @@ export class DsEntryMetadata implements EntryMetadata {
         this.path = metadata.path;
     }
 
+    private extensionRemovedFromPath(): string {
```
zFernand0, line 50: should we make this public/export ?

phaumer:
> @phaumer They aren't necessarily required, but this means that all users must define file associations for their data sets in order to get language detection _unless_ the LSP itself supports automatic detection. Adding a file extension allows VS Code to determine which LSPs are viable. 

Yes, it should be down to the extenders providing the actual language support to make that determination imho. In Z Open Editor we generate files.associations based on our user settings and would not want Zowe Explorer make these decisions for us.

To make it less breaking we could provide a switch to re-enable this (for users who want to continue using it) and could even add an extension point in the SDK to inject name manipulation code on open for extenders, but the current behavior introduced language specific behavior that language extenders might not want (which I complained about already for v1 :-) ).

traeok:
> To make it less breaking we could provide a switch to re-enable this (for users who want to continue using it) and could even add an extension point in the SDK to inject name manipulation code on open for extenders, but the current behavior introduced language specific behavior that language extenders might not want (which I complained about already for v1 :-) ).

I think it makes sense to add a switch, but since the cut-off for breaking changes in Zowe v3 was the end of August, we'll have to keep this default behavior in place. In v3, we can look into an enhancement that would allow users to opt out of this behavior.

packages/zowe-explorer-api/src/fs/types/datasets.ts
zFernand0: I missed the return statement 😅

zFernand0 (dismissed):

anaxceron (dismissed):

traeok:
Marking as draft while I bump patch coverage from changes.

packages/zowe-explorer-api/src/fs/types/datasets.ts
traeok: Good call - fixed in c66f196

traeok:
Okay, should be good to go now - please disregard the SonarCloud warning for duplication as it only occurs in test code.

phaumer:
I understand. I created https://github.com/zowe/zowe-explorer-vscode/issues/3131 for the follow-up.

benjamin-t-santos:
I tested out opening data sets - extensions are being added like they were in v2 👍

anaxceron (dismissed):

traeok:
> I tested out opening data sets - extensions are being added like they were in v2 👍

Great - thanks for testing!

packages/zowe-explorer/src/trees/dataset/DatasetUtils.ts
```diff
@@ -67,45 +67,45 @@ export class DatasetUtils {
     }
 
     /**
-     * Get the language ID of a Data Set for use with `vscode.languages.setTextDocumentLanguage`
+     * Get the file extension for a Data Set (or data set member) based on its name or its PDS name.
      */
-    public static getLanguageId(label: string): string | null {
+    public static getExtension(label: string): string | null {
```
t1m0thyj, line 72: Curious if we'd want to define a mapping of extensions to ds qualifiers as a constant? For example:
```javascript
const EXT_MAPPING = {
  ".c": ["C"],
  ".jcl": ["JCL", "JCLLIB", "CNTL", "PROC", "PROCLIB"]
}
```

This could simplify the code below, as well as in `extensionRemovedFromPath` which could call `Object.keys` to iterate over all the extensions.

traeok: I had restored the previous implementation for this fix, but I think it makes sense to clean it up. I like this idea, but I used a map instead so I can get the extension & array of matches using `Map.entries`.

Addressed in d7a30e4

packages/zowe-explorer/src/trees/uss/UssFSProvider.ts
```diff
@@ -649,11 +660,7 @@ export class UssFSProvider extends BaseProvider implements vscode.FileSystemProv
                 }
             }
         } else {
-            const fileEntry = this.lookup(source, true);
-            if (fileEntry == null) {
-                return;
-            }
-
+            const fileEntry = this.lookup(source);
```
zFernand0, line 663: Dumb question: 
Isn't this a breaking change in behavior?

My theory:
Having this function (copyTree) throw an error may ultimately result in us stopping a `pasteUssTree` operatation. Compared to returning right away, we might've allowed for a partial upload. However, I believe it's probably best to fail right away instead of having to keep track of which URIs failed to then report that in an error notification (which I don't think we were doing anyways)

packages/zowe-explorer/src/trees/dataset/DatasetFSProvider.ts
```diff
@@ -561,13 +542,16 @@ export class DatasetFSProvider extends BaseProvider implements vscode.FileSystem
     }
 
     public async delete(uri: vscode.Uri, _options: { readonly recursive: boolean }): Promise<void> {
-        const entry = this.lookup(uri, false);
+        const entry = this.lookup(uri, false) as DsEntry | PdsEntry;
         const parent = this._lookupParentDirectory(uri);
```
zFernand0, line 546: It would seems that the `this.lookup` method is called with the same `uri` many times (twice in this scenario)

I see that the lookup operation is not very complex, but curious if there is a way to reduce the number of times we have to perform those activities 😅 

Not requesting changes here though 😋

packages/zowe-explorer/src/trees/dataset/DatasetUtils.ts
```diff
@@ -67,45 +67,23 @@ export class DatasetUtils {
...
-            if (split[i] === "XML") {
-                return "xml";
-            }
-            if (split[i] === "ASM" || split[i].indexOf("ASSEMBL") > -1) {
-                return "asm";
-            }
-            if (split[i] === "LOG" || split[i].indexOf("SPFLOG") > -1) {
-                return "log";
+            for (const [ext, matches] of DS_EXTENSION_MAP.entries()) {
+                for (const match of matches) {
```
zFernand0, line 78: Thanks for cleaning up the implementation 🙏🏽

zFernand0 (commented):
LGTM! 😋 
Left a few questions, mostly for my own understanding.
I don't believe that any of my comments should require changes to be made.

packages/zowe-explorer/src/trees/dataset/DatasetFSProvider.ts
```diff
@@ -55,29 +55,6 @@ export class DatasetFSProvider extends BaseProvider implements vscode.FileSystem
         return DatasetFSProvider._instance;
     }
 
-    /**
-     * onDidOpenTextDocument event listener for the dataset provider.
-     * Updates the opened document with the correct language ID.
-     * @param doc The document received from the onDidOpenTextDocument event.
-     */
-    public static async onDidOpenTextDocument(this: void, doc: vscode.TextDocument): Promise<void> {
```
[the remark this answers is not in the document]
traeok: Yep, that's correct.

packages/zowe-explorer/src/trees/dataset/DatasetFSProvider.ts
```diff
@@ -495,7 +475,8 @@ export class DatasetFSProvider extends BaseProvider implements vscode.FileSystem
 
         try {
             if (!entry) {
-                entry = new DsEntry(basename, FsDatasetsUtils.isPdsEntry(parent));
+                const isPdsMember = FsDatasetsUtils.isPdsEntry(parent);
+                entry = new DsEntry(basename, isPdsMember);
```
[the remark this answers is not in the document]
traeok: When we use `path.parse` within the Data Sets provider, we are accessing the `name` property which is not always the same result as calling `path.posix.basename`.

packages/zowe-explorer/src/trees/dataset/DatasetFSProvider.ts
```diff
@@ -604,11 +588,15 @@ export class DatasetFSProvider extends BaseProvider implements vscode.FileSystem
         const newName = path.posix.basename(newUri.path);
 
         try {
-            if (FsDatasetsUtils.isPdsEntry(entry)) {
+            if (FsDatasetsUtils.isPdsEntry(entry) || !entry.isMember) {
                 await ZoweExplorerApiRegister.getMvsApi(entry.metadata.profile).renameDataSet(oldName, newName);
             } else {
                 const pdsName = path.basename(path.posix.join(entry.metadata.path, ".."));
```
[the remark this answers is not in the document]
traeok: See my above comment on `path.posix.basename` vs `path.parse` in the DS provider 😋 
Happy to discuss in more detail offline

packages/zowe-explorer/src/trees/uss/UssFSProvider.ts
traeok: When copying from a `source` URI to `destination` URI, we have to make sure that the source URI exists in the file system. If the lookup fails within this branch, it's because the URI is pointing to a single file - and since that file no longer exists in the file system, and we cannot copy it as a result. That said, we will still do a partial upload for any part of the tree that's being uploaded, since this function is recursive, and it will simply fail at that file. Once the failure occurs, we'll stop copying the remainder of the tree, so the behavior should be the same.

packages/zowe-explorer/src/trees/dataset/DatasetFSProvider.ts
traeok: Lookup is called twice here but with different URIs:

- `uri` parameter is passed to `this.lookup` to get its entry
- `uri` parameter is passed to `this.lookupParentDirectory`, which builds a URI "one level upward" and then calls `this.lookup` on the new URI that was built

packages/zowe-explorer/src/trees/dataset/DatasetInit.ts
```diff
@@ -182,7 +182,6 @@ export class DatasetInit {
                 await datasetProvider.onDidChangeConfiguration(e);
             })
         );
-        context.subscriptions.push(vscode.workspace.onDidOpenTextDocument(DatasetFSProvider.onDidOpenTextDocument));
```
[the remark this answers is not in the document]
traeok: Yep! 😁

packages/zowe-explorer/src/trees/dataset/DatasetFSProvider.ts
```diff
@@ -604,11 +588,15 @@ export class DatasetFSProvider extends BaseProvider implements vscode.FileSystem
         const newName = path.posix.basename(newUri.path);
 
         try {
-            if (FsDatasetsUtils.isPdsEntry(entry)) {
+            if (FsDatasetsUtils.isPdsEntry(entry) || !entry.isMember) {
```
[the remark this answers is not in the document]
zFernand0: Interesting! 🥳 
Nice to see some smarts built-in into the language.

It makes sense then to access the `isMember` property directly to avoid type guard checking, thus avoid having to type-case `entry` in the else condition 🙏🏽 

Thanks for explaining! 🙏🏽

packages/zowe-explorer/src/trees/dataset/DatasetFSProvider.ts
zFernand0: That makes sense.
It's nice to see that it just works by doing `..` (at the end) to traverse upward 🙏🏽 

Thanks for explaining 🥳

packages/zowe-explorer/src/trees/dataset/DatasetFSProvider.ts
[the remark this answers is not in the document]
zFernand0: That makes sense.
In the case of `stat`, where we call the `MvsApis.dataSets`, we want to make sure to strip out the file extension. (hence the use of `parse(...).name` in those scenarios)
Thanks for clarifying! 🥳

packages/zowe-explorer/src/trees/uss/UssFSProvider.ts
zFernand0: Hey, 
Thanks for taking the time to explain offline.
Partial upload will still occur and based on how the recursion iterates through folders, it will be clear to end users where we stopped uploading. 🙏🏽 
Again, thanks for clarifying 🥳

zFernand0 (approved):
Small merge conflict, but LGTM! 😋

JillieBeanSim (approved):
LGTM! thanks @traeok

zFernand0 (approved):
LGTM! 😋

PR #3137, opened 2024-09-23
Refactor format of l10n strings to standard format

rudyflores:
## Proposed changes

Refactored format of l10n strings to use the standard `vscode.l10n.t(message, args)` format instead of the `vscode.l10n.t({ 
message, args, comments })` format, which introduced issues on certain cases.

## Release Notes

Milestone: v3.0.0

Changelog: Changed format of l10n strings to standard format

## Types of changes

## Checklist

**General**

**Code coverage**

**Deployment**


rudyflores:
> Thanks @rudyflores for this localization fix. Should we also update the items in JobTableView.ts as well or does it behave differently in table views? Screenshot 2024-09-23 at 10 37 28 AM

I believe this might be a potential issue too, thanks for finding that! Not sure why that didn't appear when searching :o

zFernand0 (commented):
Does this mean that we are removing the `context` in POEditor for those contributing translations? 
image

I might be confusing things here though. Here is what VSCode documents:
image

traeok:
Does the previous function call not work for us due to issues with translations or our own logic, or is this something that could be resolved upstream in [microsoft/vscode-l10n](https://github.com/microsoft/vscode-l10n)?

I tried debugging this issue (changing languages with the translations in place), but it seems like the error occurs from within the `vscode-l10n` package itself.

JillieBeanSim:
> Does the previous function call not work for us due to issues with translations or our own logic, or is this something that could be resolved upstream in [microsoft/vscode-l10n](https://github.com/microsoft/vscode-l10n)?
> 
> I tried debugging this issue (changing languages with the translations in place), but it seems like the error occurs from within the `vscode-l10n` package itself.

If it is found to be upstream, should we have the workaround in place until fixed upstream?

traeok:
> If it is found to be upstream, should we have the workaround in place until fixed upstream?

Definitely in favor of implementing a workaround if it can be resolved upstream - we might want to file this as a bug in the l10n repo

t1m0thyj:
> Does this mean that we are removing the `context` in POEditor for those contributing translations? image
> 
> I might be confusing things here though. Here is what VSCode documents: image

@zFernand0 This should not affect the `context` value shown in POEditor. The `context` is a translation ID defined only for strings coming from `package.nls.json`. The comments are unused by POEditor as far as I know 😋 

@rudyflores Thanks for the refactor! Are you able to provide more details about "issues on certain cases" to help with testing this PR?

rudyflores:
> > If it is found to be upstream, should we have the workaround in place until fixed upstream?
> 
> Definitely in favor of implementing a workaround until it can be resolved upstream - if it hasn't been filed as a bug in the l10n repo, we could make an issue with steps to reproduce.
> 
> That said, if we're removing context from translators within POEditor, then it might be easier to wait for the fix so that we don't have to add the comments back once its resolved. If the comments are unused, then feel free to disregard - we can always keep the changes in place.

I think we should definitely have this workaround in place, otherwise we can't provide localization for any user's outside of english when we did in the past, which means we are breaking in terms of the support for that :/

rudyflores:
> > Does this mean that we are removing the `context` in POEditor for those contributing translations? image
> > I might be confusing things here though. Here is what VSCode documents: image
> 
> @zFernand0 This should not affect the `context` value shown in POEditor. The `context` is a translation ID defined only for strings coming from `package.nls.json`. The comments are unsupported/unused by POEditor as far as I know 😋
> 
> @rudyflores Thanks for the refactor! Are you able to provide more details about "issues on certain cases" to help with testing this PR?

Yes! @t1m0thyj so it fails if you change an l10n string to use the object format, so for example if you do that for any l10n string in our activation code and switch the language, Zowe Explorer will fail to activate

t1m0thyj:
I don't know if we want to remove all the l10n comments from our code. The npm docs for `@vscode/l10n-dev` suggest that using comments is good practice, but they need to be stripped out in the translated bundle file:
image

For ZE, the `downloadPoeditorL10n.js` script already handles this - it creates l10n bundle files in a format that has the comments removed. For testing purposes we could add a small script to the repo that strips out comments for any l10n bundle file:
```
const fs = require("fs");
const bundleJsonFile = process.argv[2];
const l10nBundle = JSON.parse(fs.readFileSync(bundleJsonFile, "utf-8"));
for (const [k, v] of Object.entries(l10nBundle)) {
    if (typeof v === "object") {
        l10nBundle[k] = l10nBundle[k].message;
    }
}
fs.writeFileSync(bundleJsonFile, JSON.stringify(l10nBundle, null, 2) + "\n");
```

If this would be a satisfactory solution, I'd prefer to keep the comments in our code as they may add value for human translators 😋

rudyflores:
> I don't know if we want to remove all the l10n comments from our code. The npm docs for `@vscode/l10n-dev` suggest that using comments is good practice, but they need to be stripped out in the translated bundle file: image
> 
> For ZE, the `downloadPoeditorL10n.js` script already handles this - when downloading translations from POEditor it creates l10n bundle files in a format that has the comments removed. For testing purposes we could add a small script to the repo that strips out comments for any l10n bundle file:
> 
> ```
> const fs = require("fs");
> const bundleJsonFile = process.argv[2];
> const l10nBundle = JSON.parse(fs.readFileSync(bundleJsonFile, "utf-8"));
> for (const [k, v] of Object.entries(l10nBundle)) {
>     if (typeof v === "object") {
>         l10nBundle[k] = l10nBundle[k].message;
>     }
> }
> fs.writeFileSync(bundleJsonFile, JSON.stringify(l10nBundle, null, 2) + "\n");
> ```
> 
> If this would be a satisfactory solution, I'd prefer to keep the l10n comments in our code as they may add value for human translators 😋

Thoughts on this? @phaumer @JillieBeanSim

phaumer:
I cannot change the output format of our automated translation. If there is content that breaks it then we indeed need to remove from the translation input files.

t1m0thyj:
> I cannot change the output format of our automated translation. If there is content that breaks it then we indeed need to remove from the translation input files.

We could enhance the ZE npm script `generateLocalization` to strip out the comments so that the translation input file consumed by the pipeline would be in the right format.

t1m0thyj (commented):
Should we close this in favor of #3149?

rudyflores:
> Should we close this in favor of #3149?

@phaumer what do you think of this approach instead?

zFernand0:
> > Should we close this in favor of #3149?
> 
> @phaumer what do you think of this approach instead?

Hey all,
Now that #3149 is merged, is this PR ok to close for now?

zFernand0 (commented):
Waiting on a decision (on whether to close this PR) before reviewing 😋

JTonda:
Closing in favor of https://github.com/zowe/zowe-explorer-vscode/pull/3149

Issue #3125, opened 2024-09-19
Edit team config fails with global & project level present


JillieBeanSim:
**Describe the bug**

Project team config file is now named zowe.config.json instead of zowe.config.user.json. I looked at Profiles.createZoweSchema() from the Zowe Explorer source on github for the latest v3 tag, and could not understand how it got changed. Also, the ZE UI action to edit a team config file fails for project config if there are both project and global configs present. The rename might be why.

The name change was intentional to align with Zowe CLI, but the project vs global bug should be investigated.

v3.0.0-next.202409091409

split from https://github.com/zowe/zowe-explorer-vscode/issues/3089#issuecomment-2347244119

**To Reproduce**

**Expected behavior**

**Screenshots**

**Desktop (please complete the following information):**

- OS:
- Zowe Explorer Version:
- (Optional) Zowe CLI Version:
- (Optional) Are you using Secure Credential Store?

**Additional context**

PR #3138, opened 2024-09-23
fix(V3): Edit team config fails with global & project level present

SanthoshiBoyina1:
## Proposed changes

Solves #3125 
To fix Edit team config fails with global & project level present

## Release Notes

Milestone:

Changelog:

## Types of changes

## Checklist

**General**

**Code coverage**

**Deployment**


JillieBeanSim (approved):
LGTM! thanks for the fix @SanthoshiBoyina1

packages/zowe-explorer/src/configuration/Profiles.ts
```diff
@@ -464,6 +464,7 @@ export class Profiles extends ProfilesCache {
                 if (choice === "project") {
                     rootPath = workspaceDir.uri.fsPath;
                     global = false;
+                    user = true;
```
traeok, line 467: Would this cause a regression in behavior? In v2 we were accidentally creating a user config when someone selects the project option. This would bring back that behavior where it will create a user config instead of a normal config.

t1m0thyj (dismissed):
Will approve once Trae's comment is addressed - creating `zowe.config.user.json` files should not be the default behavior, as the purpose of User Config is solely to override values in `zowe.config.json`.

zFernand0 (commented):
left a small comment on the changelog 😅

packages/zowe-explorer/CHANGELOG.md
```diff
@@ -15,6 +15,7 @@ All notable changes to the "vscode-extension-for-zowe" extension will be documen
 - Fixed issue where file extensions were removed from data sets, causing language detection to sometimes fail for Zowe Explorer extenders. [#3121](https://github.com/zowe/zowe-explorer-vscode/issues/3121)
 - Fixed an issue where copying and pasting a file/folder in the USS tree would fail abruptly, displaying an error. [#3128](https://github.com/zowe/zowe-explorer-vscode/issues/3128)
 - To fix Strange behaviour of Job label under Job Favorites. [#2632](https://github.com/zowe/zowe-explorer-vscode/issues/2632)
+- To fix Edit team config fails with global & project level present. [#3125](https://github.com/zowe/zowe-explorer-vscode/issues/3125)
```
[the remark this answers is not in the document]
anaxceron: Can we explain what we fixed? As in: 
"Fixed the undefined Job label to **correct a typo** under Job Favorites?"
"Fixed the Edit profile operation to **DO WHAT** when global and project levels are present"

JillieBeanSim:
Hey all, I tried a different approach.  Let me know what you think @traeok @t1m0thyj @zFernand0

zFernand0 (dismissed):
LGTM! 😋 

Left a small comment about an edge-case. 😅 
But it's a bit rare for someone to want to update their config by opening their zoweDir and selecting project as their choice 😋

packages/zowe-explorer/src/configuration/Profiles.ts
```diff
@@ -530,15 +530,15 @@ export class Profiles extends ProfilesCache {
             switch (choice) {
                 case "project":
                     for (const file of existingLayers) {
-                        if (file.user) {
+                        if (!file.path.includes(FileManagement.getZoweDir())) {
```
zFernand0, line 533: I believe that the only scenario that this won't work is if someone opens their `ZOWE_CLI_HOME` folder and wants to edit the configuration  and select project config.
Since the file.path will in fact be the same as the getZoweDir(), it won't trigger the openConfigFile 😅 

That is a rare enough scenario though.

traeok (commented):
Thanks Santhoshi and Billie for looking into this!

I'm wondering if we can do the opposite of what we're doing for the `global` case in the switch statement? If we only open the layers that aren't marked as global, that will cover the project layers (both normal and user configs).

```typescript
case "project":
    for (const file of existingLayers) {
        if (!file.global) {
            await this.openConfigFile(file.path);
        }
    }
```

I did some manual testing and noticed a couple things regarding the Edit Config option:
- If a global user config is present, it does not open both configs because `showTextDocument` defaults to previewing the file, so the second call to open the user config causes the regular config file to disappear in the editor. This behavior also occurs when selecting the Project option while a project user config is present. To resolve this, we can add options to the `Gui.showTextDocument` call in `Profiles.openConfigFile`: `{ preview: false }`
- As Fernando mentioned, opening the `.zowe` folder or `ZOWE_CLI_HOME` and then selecting "Project" from the Edit Config quick pick will cause the "Team config opened" message to appear, but the config never opens. If we filter out the layers based on unique paths, we can handle all scenarios even with using the above `!file.global` check:

```typescript
public async editZoweConfigFile(): Promise<void> {
    ZoweLogger.trace("Profiles.editZoweConfigFile called.");
    const configLayers = await this.getConfigLayers();
    const uniquePaths = new Set();
    const existingLayers = configLayers.filter((layer) => {
        const normalized = path.normalize(layer.path);
        if (!uniquePaths.has(normalized)) {
            uniquePaths.add(normalized);
            return true;
        }

        return false;
     });
     // ...rest of function
```

packages/zowe-explorer/CHANGELOG.md
```diff
@@ -29,6 +29,7 @@ All notable changes to the "vscode-extension-for-zowe" extension will be documen
 - Fixed issue where Search operation did not prompt for credentials if profile contains expired token. [#2259](https://github.com/zowe/zowe-explorer-vscode/issues/2259)
 - Fixed issue where inactive status was not displayed for profiles loaded from Global Config. [#3134](https://github.com/zowe/zowe-explorer-vscode/issues/3134)
 - Fixed issue where switching from token-based authentication to user/password would cause an error for nested profiles. [#3142](https://github.com/zowe/zowe-explorer-vscode/issues/3142)
+- Fixed issue seen editing team configuration when global and project level team configuration exists. [#3125](https://github.com/zowe/zowe-explorer-vscode/issues/3125)
```
anaxceron, line 32: This seems a little vague. Can we explain what specific issue we're seeing?

anaxceron (approved):
Looks great, thanks @SanthoshiBoyina1!

packages/zowe-explorer/src/configuration/Profiles.ts
```diff
@@ -530,15 +539,15 @@ export class Profiles extends ProfilesCache {
...
+                        if (!file.global) {
                             await this.openConfigFile(file.path);
                         }
                     }
                     Gui.showMessage(this.manualEditMsg);
                     break;
                 case "global":
                     for (const file of existingLayers) {
-                        if (file.global) {
+                        if (file.path.includes(FileManagement.getZoweDir())) {
```
t1m0thyj, line 550: I think the change on this line could be reverted to address Fernando's comment 😋 
```suggestion
                        if (file.global) {
```

traeok (commented):
As Timothy mentioned, it looks like we'll need to change the logic back for `case "global"` so that it handles opening files in the Zowe dir as well.

Will approve once resolved 😋

traeok (approved):

JillieBeanSim (approved):
LGTM! thanks @SanthoshiBoyina1 for the fix

PR #3174, opened 2024-10-03
feat: Detect JWT expiration in token values and prompt user if expired

traeok:
## Proposed changes

Implements JWT expiration checks for profiles using a token for authentication.

- Prompts the user if the JWT could be successfully decoded from the token **and** the expire time/date has elapsed
- Skips LTPA token values (cannot be decoded w/o private key, only given to servers generating the tokens)

## How to test

1. Ensure that dependencies are up-to-date using `pnpm install`
2. Log into a profile using API ML or z/OSMF using JWT token authentication.
3. Wait for the token to expire ⏳ 
4. Once expired, attempt to search again - notice the login prompt will appear instead of the search prompt, as the token has expired

## Release Notes

Milestone: 3.1.0

Changelog:

- Added expired JWT token detection for profiles in each tree view (Data Sets, USS, Jobs). When a user performs a search on a profile, they are prompted to log in if their token expired. [#3175](https://github.com/zowe/zowe-explorer-vscode/issues/3175)

## Types of changes

## Checklist

**General**

**Code coverage**

**Deployment**


traeok:
Failing stage `ze-build (20.x, ubuntu-latest)` is due to an unrelated, transient failure in the integration tests.

adam-wolfe:
Tried it out. Looks great.

packages/zowe-explorer/src/trees/ZoweTreeProvider.ts
```diff
@@ -302,6 +304,46 @@ export class ZoweTreeProvider<T extends IZoweTreeNode> {
...
+        try {
+            const payloadJson = JSON.parse(Buffer.from(tokenParts[1], "base64url").toString("utf8"));
+            if ("exp" in payloadJson) {
+                const expireDate = dayjs.unix(payloadJson.exp);
+                if (expireDate.isBefore(dayjs())) {
+                    await AuthUtils.promptUserForSsoLogin(profileName);
+                }
+            }
+        } catch (err) {
+            return;
```
zFernand0, line 343: Curious if this should be moved to the SDKs 😅 
Or at least be made available to extenders 😅

zFernand0 (changes requested):
the functionality LGTM! 😋 

requesting changes to prevent this PR from being merged before we publish 3.0.1 😅

packages/zowe-explorer/src/trees/ZoweTreeProvider.ts
traeok: I like this idea, thanks for the suggestion! I moved this function implementation into Imperative's `ProfileInfo` class.
Here is the PR: https://github.com/zowe/zowe-cli/pull/2298

Once that's approved & merged, we can adopt it in this PR - this would mean that extenders also get access to it directly from the facility that they already use for profiles ^^

packages/zowe-explorer/CHANGELOG.md
```diff
@@ -6,6 +6,8 @@ All notable changes to the "vscode-extension-for-zowe" extension will be documen
 
 ### New features and enhancements
 
+- Added expired JWT token detection for profiles in each tree view (Data Sets, USS, Jobs). When a user performs a search on a profile, they are prompted to log in if their token expired. [#3175](https://github.com/zowe/zowe-explorer-vscode/issues/3175)
```
anaxceron, line 9: Should we spell out JWT?

anaxceron (approved):
Looks good, @traeok. Had one question on the changelog.

zFernand0 (dismissed):
LGTM! 😋

packages/zowe-explorer/CHANGELOG.md
traeok: Addressed in 6d9c2e4

zFernand0 (approved):
LGTM! 😋

anaxceron (approved):

pnpm-lock.yaml
t1m0thyj, line 1: Zowe SDK updates look good, curious why `geckodriver` and all its deps were added to the lockfile 🤔

packages/zowe-explorer/src/trees/ZoweTreeProvider.ts
```diff
@@ -266,6 +266,7 @@ export class ZoweTreeProvider<T extends IZoweTreeNode> {
                 Profiles.getInstance().validProfile = Validation.ValidationType.UNVERIFIED;
             }
         }
+        await ZoweTreeProvider.checkJwtTokenForProfile(node.getProfileName());
```
t1m0thyj, line 269: Curious if we'd want to move this inside `Profiles.getInstance().checkCurrentProfile` so that we can set the validation status to `inactive` when token is expired?

After further thought, maybe not and this is a dumb question 😋 I believe validation status is only used to determine if the host/port are accessible, so we don't need to update it for invalid or expired credentials.

pnpm-lock.yaml
traeok: This is a good question - when I last merged, I grabbed the lockfile from the `main` branch and then ran `pnpm install` to get the latest deps. I can try this again to see if there was a mishap during the last merge attempt.

packages/zowe-explorer/src/trees/ZoweTreeProvider.ts
traeok: While I like this idea, I'm on the fence for changing this if what you mentioned about validation is true:

>  I believe validation status is only used to determine if the host/port are accessible

Maybe we can track this as a separate enhancement to get the team's consensus on updating the validating profiles logic? I don't believe we do this for basic credentials, as validation currently passes even with invalid credentials - the invalid credentials error does not appear until after the search is submitted. But if we want to do this for tokens, we should probably do something similar for basic creds 😋

pnpm-lock.yaml
traeok: I'm not sure what I did the first time, but doing this again results in a different lockfile. Sorry for the confusion. This should be addressed in 70fc546

JillieBeanSim (dismissed):
LGTM! thanks @traeok for the feature

t1m0thyj (dismissed):
LGTM, thanks @traeok!

Note to others testing this PR: You may need to run `pnpm install` to install the new version of Imperative that adds the `ProfileInfo.hasTokenExpiredForProfile` method 😋

t1m0thyj (approved):

awharn (approved):

PR #3232, opened 2024-10-11
Test PR

zFernand0:
To be closed soon


zFernand0:
This is just to confirm that DCO should be fine going forward.

Here is what my sign-offs will look like
image


And here is what the squashed commit will look like from the GH UI
image

PR #3233, opened 2024-10-14
fix(ci): Use `ubuntu-22.04` images to get access to libsecret

traeok:
## Proposed changes

`libsecret` was removed from the Ubuntu 24.04 runner images. For the time being, we can use the `ubuntu-22.04` image in CI until we get clarification on whether this change was intentional.


t1m0thyj (approved):
LGTM, thanks @traeok! We'll probably need to make a similar change in the CLI repo 😋

zFernand0 (approved):

awharn (approved):

PR #3245, opened 2024-10-15
ci: only notify users for merge-by during scheduled run

traeok:
Fixes a bug in the `build-table-and-notify.js` script so that:
- Merge-by notifications are only posted once per PR (sorry to all that received the extra notification)
- The notifications are only posted when the `schedule` event runs for the workflow


t1m0thyj (approved):
LGTM, thanks @traeok!

zFernand0 (approved):
LGTM! 😋

PR #3274, opened 2024-10-23
fix: "Show Config" button in error dialog does not work during initialization

### LINKED ISSUE #3273: 3.0.2 regression for `Show Config` when json errors exist
**Describe the bug**

This issue was resolved in 3.0.1, but has reappeared in 3.0.2.

**To Reproduce**

1. Introduce a json error into your config. 
2. Reload VS Code. 
3. Click the "Show Config" button in the warning that appears.
4. Nothing happens.

**Expected behavior**

When the `Show Config` button is clicked, the user should see their config.

**Screenshots**

No-config-shown

- OS:
- Zowe Explorer Version:

**Additional context**

traeok:
## Proposed changes

Resolves #3273 

This issue can be reproduced in v3.0.2 if you have a Zowe team configuration file with errors when initializing Zowe Explorer. The prompt with "Error encountered when loading your Zowe config" would appear, but the "Show Config" button did nothing when clicked.

This PR not only fixes the issue, but also adds an integration test to ensure this dialog still works during initialization moving forward.

Finally, this fixes a transient failure encountered with one of the integration tests for updating credentials, so the (Ubuntu, 20.x) stage should not fail now unless there is truly an issue during the test.

## Release Notes

Milestone: 3.0.3

Changelog:

- Fixed an issue during initialization where a broken team configuration file caused the "Show Config" action in the error dialog to stop working. [#3273](https://github.com/zowe/zowe-explorer-vscode/issues/3273)

## Types of changes

## Checklist

**General**

**Code coverage**

**Deployment**

- [ ] These changes may need ported to the appropriate branches (list here): This may need ported to v2 - if I can reproduce an issue, I will create a port.


adam-wolfe:
While this does solve the issue with the Show Config button, after fixing the config and saving changes, trying to add a profile to the trees or refresh the trees results in an error:
image

traeok:
> While this does solve the issue with the Show Config button, after fixing the config and saving changes (and without reloading the VS Code window), trying to add a profile to the trees or refresh the trees results in an error (with nothing added to the logs):

@adam-wolfe This appears to be the behavior since v3.0.0 and is not related to the regression from 3.0.1. That said, I can look into resolving it as part of this PR as well.

traeok:
@adam-wolfe I found another place where `getProfileInfo` was being called without handling exceptions, should be working as expected now 👍

anaxceron (dismissed):

adam-wolfe (dismissed):
Looks good. 

After I fix my config and save changes, only one profile is added to the trees, but if I reload the window, all of my profiles are added to the trees. I'm not sure that this is new behavior though.

adam-wolfe (approved):

zFernand0:
> Looks good.
> 
> After I fix my config and save changes, only one profile is added to the trees, but if I reload the window, all of my profiles are added to the trees. I'm not sure that this is new behavior though.

I ran into the same issue. After fixing the error and saving... the profiles are not loaded. However, it's kind of expected since the watcher was not initialized due to the activation error 😅

zFernand0 (changes requested):
LGTM! 😋 

Thanks for adding e2e tests 🙏 


Requesting changes just because there is a missing license header in this file:
`packages/zowe-explorer/__tests__/__integration__/bdd/step_definitions/dialogs/ShowConfigErrorDialog.steps.ts`

traeok:
Sorry, I tested this again in sandbox and ran into the activation error mentioned above - turns out that I was missing another `null` check in a separate place where `getProfileInfo` is called. 

With this change, fixing the profile and saving causes the default profile to be added to all trees. In addition, it resolves the additional "Activation failed" error and the extension successfully activates even if profiles cannot be loaded.

Thanks for testing and the feedback @adam-wolfe @zFernand0, it's been helpful. I did not see the additional activation error in Dev Host, but was able to reproduce it in the sandbox environment.

---

After looking over the current logic for `ProfilesUtils.getProfileInfo`, I think we should consider refactoring it:
- it does not handle errors unrelated to credential manager initialization and doesn't return anything in those cases (implicitly returns `undefined`)
- it initializes a new `ProfileInfo` instance and reads profiles from disk every time it is called, instead of caching the instance that was first created

For now, I've added a couple safe guards:

- `debug` logger call into the `catch` block of `setupDefaultCredentialManager` so we can get additional context at the time an error occurs in that function
- added typedoc to clarify that the function throws, so that other devs are aware before trying to use it without a try/catch block. 

I chose not to make additional changes to this function for now (such as always returning a `profileInfo` instance), as I see its being used in a few different places - I don't want to alter the behavior in those areas.

packages/zowe-explorer/__tests__/__integration__/bdd/features/dialogs/ShowConfigErrorDialog.feature
```diff
@@ -0,0 +1,7 @@
+Feature: Show Config Error Dialog
+
+Scenario: Initializing Zowe Explorer with a broken profile
+    When a user opens Zowe Explorer
+    Then the Show Config dialog should appear
```
zFernand0, line 5: I'm afraid that this step didn't work (as-is).
I had to introduce an error in the config file in order for the test to pass.

Curious if we should add a few steps:
```suggestion
    When a user selects "Edit Team Configuration File" # existing step
    Then it will open the config in the editor #existing step
    When the user introduces and error and saves the configuration file # new step
    Then the Show Config dialog should appear
```

zFernand0 (approved):
LGTM! 😋 

Now I do see the default profile loading after saving (same as Adam), which indicates that the watcher still works 🙏 

I don't think it's a big deal to have to reload extensions (or vscode itself), hence why I'm approving 🙏 

I also left a small comment on the bdd tests, but nothing that should be addressed right away.  I'm curious how the CI works though 😅 

re: the refactoring.
Thanks for adding the null checks 🙏 
However, I'm still trying to wrap my head around what the refactoring will entail. 😅 
I could benefit from a quick discussion at some point today 🙏 

Sorry for the long review, but LGTM! 😋

traeok:
> I don't think it's a big deal to have to reload extensions (or vscode itself), hence why I'm approving 🙏

Are you encountering a reload during your testing? I didn't add any reloads to the logic, so I just want to make sure its behaving properly for you. The extension will activate even if profiles failed to load, and the watcher is registered during the first activation; so once your config is fixed you should be able to continue using ZE as normal.

> re: the refactoring. Thanks for adding the null checks 🙏 However, I'm still trying to wrap my head around what the refactoring will entail. 😅 I could benefit from a quick discussion at some point today 🙏

Definitely happy to discuss offline 😋

packages/zowe-explorer/__tests__/__integration__/bdd/features/dialogs/ShowConfigErrorDialog.feature
zFernand0: Thanks for the quick discussion offline 🙏

adam-wolfe (approved):

JillieBeanSim (approved):
LGTM! thanks @traeok

zFernand0 (approved):

PR #3281, opened 2024-10-28
fix: Only show "No configs detected" prompt if ZE opened

### LINKED ISSUE #3280: Zowe Explorer Prompts Credentials for non z/OS Projects
If I open VS Code for a project that has nothing to do with mainframe, I continuously encounter this prompt:

image

It seems like this prompt should only occur when I actually try to open the Zowe Explorer Extension 

**Describe the bug**

**To Reproduce**

**Expected behavior**

**Screenshots**

- OS:
- Zowe Explorer Version:

**Additional context**

traeok:
## Proposed changes

Fixes #3280

## Release Notes

Milestone: 3.0.3

Changelog:

- Fixed issue where Zowe Explorer would present the "No configs detected" notification before it was used in a workspace without a Zowe config. [#3280](https://github.com/zowe/zowe-explorer-vscode/issues/3280)

## Types of changes

## Checklist

**General**

**Code coverage**

**Deployment**

### DIFF
--- a/packages/zowe-explorer/src/trees/dataset/DatasetInit.ts
+++ b/packages/zowe-explorer/src/trees/dataset/DatasetInit.ts
@@ -20,6 +20,7 @@ import { SharedActions } from "../shared/SharedActions";
 import { SharedContext } from "../shared/SharedContext";
 import { SharedInit } from "../shared/SharedInit";
 import { SharedUtils } from "../shared/SharedUtils";
+import { ProfilesUtils } from "../../utils/ProfilesUtils";
 
 export class DatasetInit {
     public static async createDatasetTree(log: imperative.Logger): Promise<DatasetTree> {
@@ -30,6 +31,12 @@ export class DatasetInit {
         return tree;
     }
 
+    public static async datasetTreeVisibilityChanged(this: void, e: vscode.TreeViewVisibilityChangeEvent): Promise<void> {
+        if (e.visible) {
+            await ProfilesUtils.promptUserWithNoConfigs();
+        }
+    }
+
     public static async initDatasetProvider(context: vscode.ExtensionContext): Promise<DatasetTree> {
         ZoweLogger.trace("DatasetInit.initDatasetProvider called.");
         context.subscriptions.push(vscode.workspace.registerFileSystemProvider(ZoweScheme.DS, DatasetFSProvider.instance, { isCaseSensitive: true }));
@@ -37,6 +44,8 @@ export class DatasetInit {
         if (datasetProvider == null) {
             return null;
         }
+
+        datasetProvider.getTreeView().onDidChangeVisibility(DatasetInit.datasetTreeVisibilityChanged);
 
         context.subscriptions.push(
             vscode.commands.registerCommand("zowe.all.config.init", async () => {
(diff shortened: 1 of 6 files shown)


packages/zowe-explorer/src/trees/dataset/DatasetInit.ts
```diff
@@ -30,6 +31,12 @@ export class DatasetInit {
         return tree;
     }
 
+    public static async datasetTreeVisibilityChanged(this: void, e: vscode.TreeViewVisibilityChangeEvent): Promise<void> {
+        if (e.visible) {
+            await ProfilesUtils.promptUserWithNoConfigs();
+        }
+    }
```
zFernand0, line 38: I like the idea of having a set of actions that can be performed as long as a specific tree is visible.

I know this is to fix one specific issue, but...

I'm curious if we should either do the same kind of visibility detection across the board, or in the shared init.

The only reason why I'm asking is because the `ProfilesUtils.promptUserWithNoConfigs` doesn't feel to be DS specific  😅

zFernand0 (changes requested):
LGTM! 😋 

Left a quick request, and please feel free to dismiss my requested-changes if you believe it doesn't make much sense for this fix 😋

packages/zowe-explorer/src/trees/dataset/DatasetInit.ts
traeok: This dialog is mostly to help new users who don't know that they don't have any Zowe client configurations. This means that when the Zowe Explorer panel is opened for the first time, it will prompt the user if no client configurations are detected. Since the Data Sets tree is the first to initialize, the event listener will do the prompting if no client configs were detected once the data sets tree is opened.

While this isn't specific to the data sets tree, I figured that it made sense to use that tree for the event subscription as it is the first tree that's initialized. I don't think there's a way to detect when Zowe Explorer is opened from the Side Bar - if there was, I think that would be the best approach. That's why I put the event on one of the tree views as it will initialize when the ZE icon is clicked in the Side Bar.

zFernand0: Makes sense to keep using the first tree in our activations sequence.
I just hope we find this whenever we need to make other "if-visible" operations 😋

zFernand0 (approved):
LGTM! 😋

anaxceron (approved):

packages/zowe-explorer/src/utils/ProfilesUtils.ts
```diff
@@ -28,6 +28,7 @@ export enum ProfilesConvertStatus {
 
 export class ProfilesUtils {
     public static PROFILE_SECURITY: string | boolean = Constants.ZOWE_CLI_SCM;
+    private static noConfigDialogShown: boolean = false;
```
t1m0thyj, line 31: Should we rename this variable so it doesn't start with the word `no`? Please feel free to ignore, I'm definitely being picky but this was a bit hard to read because it seemed like a double negative 😋 

```suggestion
    private static createConfigDialogShown: boolean = false;
```

t1m0thyj (approved):
LGTM, thanks @traeok!

JillieBeanSim (approved):
LGTM! thanks for this fix @traeok ❤️

zFernand0 (approved):
LGTM as-is 😋 
Up to y'all if we should wait for this final comment to be addressed:
- https://github.com/zowe/zowe-explorer-vscode/pull/3281#discussion_r1823352803

PR #3321, opened 2024-11-14
Fix infinite loop when fetching USS resources with stat()

benjamin-t-santos:
## Proposed changes

Extenders can fetch a USS resource on demand using `vscode.workspace.fs.stat()` with the `fetch=true` query. When done with a USS file that has not expanded in the USS tree view, Zowe Explorer gets stuck in an infinite loop of calling `stat()` over and over again. Here's how it happens:

1. `vscode.workspace.fs.stat()` is called for a USS resource with the fetch query set to true. Example: `vscode.workspace.fs.stat(zowe-uss:/zosmf/user/test.txt)`
2. `UssFSProvider.remoteLookupForResource()` is called to fetch the file from the host.
3. This eventually calls `UssFSProvider.fetchEntries()`. `fetchEntries()` checks if the parent directory of the requested file is in the file system. If it's not, `await vscode.workspace.fs.createDirectory(parentUri)` is called on line 191 to create it.
4. `stat()` is called again for the `parentUri` with a `fetch=true` query as a result of calling `vscode.workspace.fs.createDirectory()`. (I believe this is due to the VS Code file system base implementation. It surprised me but looking at the call stack that is what I see)
5.  `UssFSProvider.remoteLookupForResource()` is called to fetch the directory from the host. This eventually calls `UssFSProvider.fetchEntries()`
6. When `listFiles()` is called in `fetchEntries()`, if the resource requested is a directory that does not already exist in the file system, we call `vscode.workspace.fs.createDirectory()` with the directory's URI. This is the same exact URI that is used when calling `createDirectory()` in step 3. 

Basically, Zowe Explorer keeps calling `vscode.workspace.fs.createDirectory()` for the parent of the requested file over and over again without ever resolving. This does not happen when using tree views because of how directories are expanded from the top-down. In this case, I as an extender am requesting a file from a parent directory that has not been created.

There are improvements we can make to improve the `UssFSProvider` for extenders, but with the goal of trying to get a fix for this issue into the v3.0.3 release, I tried to keep my changes minimal. I added code that sets the query of the `parentUri` to an empty string when calling `vscode.workspace.fs.createDirectory()` in `fetchEntries()`. This prevents it from entering the infinite loop by not trying to fetch the requested resource's parent.

This change will not impact the tree view, as those views do not make use of the fetch query. This change will only fix functionality for extenders.

## Release Notes

Milestone:

Changelog: 

## Types of changes

## Checklist

**General**

**Code coverage**

**Deployment**

### DIFF
--- a/packages/zowe-explorer/CHANGELOG.md
+++ b/packages/zowe-explorer/CHANGELOG.md
@@ -28,6 +28,7 @@ All notable changes to the "vscode-extension-for-zowe" extension will be documen
 - Fixed issue where users were not prompted to enter credentials if a 401 error was encountered when opening files, data sets or spools in the editor. [#3197](https://github.com/zowe/zowe-explorer-vscode/issues/3197)
 - Fixed issue where profile credential updates or token changes were not reflected within the filesystem. [#3289](https://github.com/zowe/zowe-explorer-vscode/issues/3289)
 - Fixed issue to update the success message when changing authentication from token to basic through the 'Change Authentication' option.
+- Fixed an issue where fetching a USS file using `UssFSProvider.stat()` with a `fetch=true` query would cause Zowe Explorer to get stuck in an infinite loop.
 
 ## `3.0.2`
 
--- a/packages/zowe-explorer/src/trees/uss/UssFSProvider.ts
+++ b/packages/zowe-explorer/src/trees/uss/UssFSProvider.ts
@@ -148,7 +148,6 @@ export class UssFSProvider extends BaseProvider implements vscode.FileSystemProv
         try {
             response = await ZoweExplorerApiRegister.getUssApi(profile).fileList(ussPath);
             // If request was successful, create directories for the path if it doesn't exist
-            //why are creating directories for the path
             if (response.success && !keepRelative && response.apiResponse.items?.[0]?.mode?.startsWith("d") && !this.exists(uri)) {
                 await vscode.workspace.fs.createDirectory(uri.with({ query: "" }));
             }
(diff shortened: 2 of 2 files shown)


traeok (approved):
Thanks for catching this Benjamin, the fix is much appreciated. I tested this with a basic sample extension and the fix looks good to me. Please disregard the failing `ze-build (22.x, ubuntu-22.04)` check as its unrelated.

---

It seems that the `vscode.workspace.fs` wrappers perform some extra checks before creating the directory - my guess is that they probably `stat` the folder to make sure it doesn't exist before trying to create it. When calling `createDirectory` directly within the providers, this step doesn't occur, so the infinite loop isn't encountered. 

Admittedly, the `stat` function has caused more grief than expected with the file system provider implementations, especially when it comes to network requests. I also did not expect VS Code to call it as much as it does, and this particular case seems to be a side effect of preserving the `fetch` query on the original URI  😵‍💫

benjamin-t-santos:
@traeok Thanks for the review. Is it possible to include this fix in the v3.0.3 release?

JillieBeanSim:
> @traeok Thanks for the review. Is it possible to include this fix in the v3.0.3 release?

@zFernand0 @t1m0thyj @adam-wolfe what do y'all think too? I believe we are including a similar fix for another tree.

t1m0thyj (approved):
LGTM, thanks @benjamin-t-santos!

t1m0thyj:
> @zFernand0 @t1m0thyj @adam-wolfe what do y'all think too? I believe we are including a similar fix for another tree.

@JillieBeanSim This fix looks like a simple one that would be good to include 👍

JillieBeanSim:
thanks I will get it ported

Issue #3238, opened 2024-10-14
Edit attributes refresh not updating values


JillieBeanSim:
**Describe the bug**

Edit attributes refresh not updating/reverting values

**To Reproduce**

1. Go to USS tree view and enter a filter search, locating a file to view Attributes
2. Right-click on file and select Edit Attributes
3. Change some values but do not apply, instead click the refresh
4. See attributes not updated/reverted after refresh

**Expected behavior**

In this scenario, I would assume the refresh would work similar to cancel updating the webview back to the original values.

**Desktop (please complete the following information):**

- OS: macos
- Zowe Explorer Version: 3.0.1
- (Optional) Zowe CLI Version:
- (Optional) Are you using Secure Credential Store?

**Additional context**

PR #3354, opened 2024-12-10
Fix Refresh button in Unix Attributes View

JillieBeanSim:
## Proposed changes

bug fix for Edit Attributes webview refresh button when changes not applied, reverting the view to original values.
https://github.com/zowe/zowe-explorer-vscode/issues/3238

## Release Notes

Milestone: 3.1.0

Changelog: Fixed an issue with edit UNIX file attributes refresh button. [#3238](https://github.com/zowe/zowe-explorer-vscode/issues/3238)

## Types of changes

## Checklist

**General**

**Code coverage**

**Deployment**


packages/zowe-explorer/src/trees/uss/USSAttributeView.ts
```diff
@@ -45,14 +43,11 @@ export class USSAttributeView extends WebView {
...
             case "refresh":
                 if (this.canUpdate) {
-                    this.onUpdateDisposable = this.ussNode.onUpdate(async (node) => {
-                        await this.attachTag(node);
-                        await this.panel.webview.postMessage({
-                            attributes: await this.ussNode.getAttributes(),
-                            name: node.fullPath,
-                            readonly: this.ussApi.updateAttributes == null,
-                        });
-                        this.onUpdateDisposable.dispose();
```
traeok, line 55: I think we'll still need the `onUpdate` event listener. Once the node has been updated in the tree, the attributes in the webview are updated because this callback will be fired in `getChildren` ([see line 233 in ZoweUSSNode.ts](https://github.com/zowe/zowe-explorer-vscode/blob/b6dd3b50f69aee9660a2ae6c39b9a6646adb3e66/packages/zowe-explorer/src/trees/uss/ZoweUSSNode.ts#L233)). Then, on lines 53-57 in this file, the refresh is called on the tree node to get the new attributes.

It seems like removing this causes every other refresh call to work - for example, if I change the permissions in my terminal and then refresh in ZE, the first refresh does not show any changes, but the second one pulls the right permissions:
ZE-EditAttributesRefresh

I think we'll want to keep the `onUpdate` callback as it was working as expected in v2. It seems like maybe the `onUpdateEmitter` is not firing as expected, which is why the refresh doesn't update the webview with the current logic on `main`.

traeok (changes requested):
Thanks for looking into this Billie! I noticed that in some scenarios the Refresh button still does not work, I posted a GIF and some context below.

packages/zowe-explorer/src/trees/uss/USSAttributeView.ts
zFernand0: I'm seeing the same behavior as Trae.
Also, the GIF shows that the `Apply changes` button is enabled the second time the refresh button is clicked, even though nothing the user (I) didn't make changes 😅

traeok:
Considering the timing of the holidays, I've "snoozed" (disabled) the merge-by comments until everyone is back in the office.

JillieBeanSim:
closing this as this solution doesn't work for both cases, still investigating

PR #3409, opened 2025-01-23
fix: call Mutex.waitForUnlock instead of locking profile before request

### LINKED ISSUE #3408: Automatically locking the user profile in the Zowe filesystem providers prevents extenders from accessing multiple resources in parallel
**Describe the bug**

A profile locking/unlocking mechanism was added to all Zowe filesystem providers in https://github.com/zowe/zowe-explorer-vscode/pull/3370. This new functionality completely prevents users from accessing z/OS resources in parallel, as the new mutex only allows one API call at a time for each profile.

**To Reproduce**

It is not unusual for a COBOL member to have thousands of copybooks. Previously, we could download the contents of multiple members quickly by initiating multiple calls to the `DatasetFSProvider` using `await vscode.workspace.fs.readFile(pdsMemberUri)`. When dealing with thousands of copybooks, fetching member content in parallel in this way is necessary.

Now, all these calls despite being made asynchronously are handled one at a time because of the profile lock. It is _very_ slow compared to previous performance. I tested fetching the contents of 10 members and it took 4x as long on v3.1.0 versus marketplace v3.0.3

**Expected behavior**

The profile should **not** be automatically locked for filesystem operations. On the [wiki](https://github.com/zowe/zowe-explorer-vscode/wiki/Handling-profiles-and-related-errors) page for the AuthHandler, I see that this functionality is supposed to be opt-in. The filesystem is intended to be used by extenders, so I believe the same should apply (or at the very least there should at least be some way to opt-out). 

In my case, I perform a connection test before initiating multiple filesystem calls to avoid 401 errors, so the lock does not add any benefit.

**Screenshots**

- OS: Mac
- Zowe Explorer Version: 3.1.0

**Additional context**

In the case above, I am specifically referring to the lock used in `DatasetFSProvider.fetchDatasetAtUri()` that results in one `getContents()` call running at a time.

traeok:
## Proposed changes

Fixes #3408

Sorry for the oversight @benjamin-t-santos @phaumer - I did not intend to limit extenders to sequential processing, but I had inadvertently done so by forcing the filesystem to acquire the mutex for each request. With these changes, the filesystem only waits for the mutex to be unlocked *after* a 401 error is encountered *and* if the extender explicitly enables their profile type to use locks (using `AuthHandler.enableLocksForType`). This allows extenders to:

- perform their own credential management (detecting your API's credential error, then prompting for credentials), 
- leverage Zowe Explorer's credential prompting/log in functionality as usual (default behavior) without using/waiting for the locks
- or opt-in to the locks, which prevents repeated requests for REST API services w/ invalid credentials

## Release Notes

Milestone: 3.1.0

Changelog: N/A since we have not yet released a new version with the bug

## Types of changes

## Checklist

**General**

**Code coverage**

**Deployment**

### DIFF
--- a/packages/zowe-explorer-api/__tests__/__unit__/profiles/AuthHandler.unit.test.ts
+++ b/packages/zowe-explorer-api/__tests__/__unit__/profiles/AuthHandler.unit.test.ts
@@ -17,6 +17,23 @@ import { AuthPromptParams } from "@zowe/zowe-explorer-api";
 
 const TEST_PROFILE_NAME = "lpar.zosmf";
 
+describe("AuthHandler.disableLocksForType", () => {
+    it("removes the profile type from the list of profile types w/ locks enabled", () => {
+        AuthHandler.disableLocksForType("zosmf");
+        expect((AuthHandler as any).enabledProfileTypes.has("zosmf")).toBe(false);
+    });
+});
+
+describe("AuthHandler.enableLocksForType", () => {
+    it("adds the profile type to the list of profile types w/ locks enabled", () => {
+        AuthHandler.enableLocksForType("sample-type");
+        expect((AuthHandler as any).enabledProfileTypes.has("sample-type")).toBe(true);
+
+        // cleanup for other tests
+        (AuthHandler as any).enabledProfileTypes.delete("sample-type");
+    });
+});
+
 describe("AuthHandler.waitForUnlock", () => {
     it("calls Mutex.waitForUnlock if the profile lock is present", async () => {
         const mutex = new Mutex();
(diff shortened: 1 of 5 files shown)


benjamin-t-santos:
Just chiming in to say I tested this PR and it works great. Fixed the issue I described in https://github.com/zowe/zowe-explorer-vscode/issues/3408 and the performance is the same as v3.0.3.

t1m0thyj (approved):
LGTM, thanks @traeok for the quick fix!

JillieBeanSim (approved):
Thanks @traeok for the fix! LGTM

PR #3453, opened 2025-02-10
Unable to open a Sequential DS after renaming a favourited Sequential DS

SanthoshiBoyina1:
## Proposed changes

Fixes https://github.com/zowe/zowe-explorer-vscode/issues/3345

## Release Notes

Milestone:

Changelog:

## Types of changes

## Checklist

**General**

**Code coverage**

**Deployment**


anaxceron (dismissed):

traeok (changes requested):
Thanks for looking into this Santhoshi!

Is the `newUri` parameter a required change to fix this behavior? Extenders have to update their implementation to support the proposed function definitions for `renameNode` and `renameFavorite`, which is a breaking change as these functions are required. We should have enough information to generate a new URI based on the new label and the existing profile on the node.

JillieBeanSim (dismissed):
LGTM! thanks for the fix @SanthoshiBoyina1

packages/zowe-explorer/src/configuration/Definitions.ts
```diff
@@ -67,6 +67,7 @@ export namespace Definitions {
         session?: imperative.Session;
         profile?: imperative.IProfileLoaded;
         contextOverride?: string;
+        contextValue?: string;
```
t1m0thyj, line 70: Do we need to add `contextValue` here? It feels redundant since it's already defined on tree nodes, and the `IZoweTreeOpts` type (which is only used in constructors) has `contextOverride` to support overriding the default context value.

traeok, line 70: I believe `contextOverride` serves the same purpose, unless I'm missing additional context. Is this new property needed?

traeok (commented):
This fix LGTM, thanks Santhoshi! I had a quick question before approving, but the changes make sense to me.

t1m0thyj (commented):
Looks pretty good, thanks @SanthoshiBoyina1! Left a question

anaxceron (approved):

packages/zowe-explorer/src/configuration/Definitions.ts
SanthoshiBoyina1: Earlier I added it for fix some failing test cases. But now it's working fine even after reverting back the change.

anaxceron (approved):

traeok (approved):

JillieBeanSim (approved):
LGTM! thanks @SanthoshiBoyina1 for the fix

zFernand0 (approved):
LGTM! 😋 
---

great catch on this edge case 🤯 
Thanks for the fix! 🙏

Issue #3447, opened 2025-02-08
OpenRecentMember command problems


JWaters02:
**Describe the bug**

There are two parts to this bug:
1. The keybinding to open it does not work
2. The function itself does not work (special edge case!)

## Part 1

When I use the keybinding ctrl+alt+r to run the command `zowe.openRecentMember`, nothing happens.

However, the keybinding (ctrl+alt+p) does work for its cousin, the command `zowe.searchInAllLoadedItems`, despite their entries in package.json being identical.

For the benefit of the doubt, I temporarily switched the zowe.searchInAllLoadedItems command to run the openRecentMemberPrompt function, and when pressing that keybind (ctrl+alt+p), it opened fine.

Perhaps some kind of key conflict? These are my keys:

![Image](https://github.com/user-attachments/assets/a568c205-e138-4ed6-b22b-b24ea60d379c)

## Part 2

Once I got the function running using that temporary switch as above, the menu would appear with all the files in my history as expected, however, when selecting a file, I always get the warning message `Cannot read properties of undefined (reading 'pattern')`.

I debugged the issue and found that the `sessionNode` was not being found in my profile, which dropped through the code as null and eventually lead to this warning. `sessionNode` is not being found in my profile because my profile name is now simply different to how it is in the file history, which is where the recent files quick pick gets its items from.

Much easier to show a screenshot, so here, you can see that my current profile name is `zosmf_zxplore`, while the profile name that is being expected, is `zxplore`. 

![Image](https://github.com/user-attachments/assets/2da59109-7aec-46c6-812c-2e67504505cd)

However, you may notice, that in the above screenshot, the USS file in history and my USS profile name is the same. So why is it still not working then?

Because of a seperate oversight in the code:

https://github.com/zowe/zowe-explorer-vscode/blob/eeeaf84a62a0a83b076b49e6bc7a1eb5ed9f0abc/packages/zowe-explorer/src/trees/shared/SharedActions.ts#L190

The file history stores the profile name in all caps, while my profile name is in all lower, so it is checking all caps profile name with all lower one and does not find a match, so `sessionNode` is also not being found. Notice that this is not an issue for a matching dataset profile though, because it does actually do the lower case conversion for both first!

https://github.com/zowe/zowe-explorer-vscode/blame/eeeaf84a62a0a83b076b49e6bc7a1eb5ed9f0abc/packages/zowe-explorer/src/trees/shared/SharedActions.ts#L195

**To Reproduce**

[...]

Issue #3446, opened 2025-02-07
Edit History Webview (V3): Items are not being added to file history


adam-wolfe:
**Describe the bug**

In Zowe Explorer V3 (3.0.3 and 3.1.0), files, data sets, and jobs are not being added to the file history in the Edit History Webview.

This functionality seems to work as expected in Zowe Explorer 2.18.0.

**To Reproduce**

1. Open any USS file, data set (or member), or spool file.
2. Open the Edit History Webview and check File History.
3. Note that the file history does not contain the file/ds/spool that you recently opened.

**Expected behavior**

After opening a file, the file should be listed in the file history.

**Screenshots**

**Desktop (please complete the following information):**

- OS: Win 10
- Zowe Explorer Version: 3.0.3, 3.1.0
- (Optional) Zowe CLI Version:
- (Optional) Are you using Secure Credential Store?

**Additional context**

JWaters02:
Firstly, it seems like `addFileHistory` is not implemented yet for Jobs: https://github.com/zowe/zowe-explorer-vscode/blob/eeeaf84a62a0a83b076b49e6bc7a1eb5ed9f0abc/packages/zowe-explorer/src/trees/job/JobTree.ts#L895

I did a little debugging and noticed that when you open a dataset, `addFileHistory`, and above that, `openDs`, are never called unless you specifically *only* open the dataset with encoding - which calls the `openWithEncoding` command then function, which does called `openDs` with `addFileHistory`

What it seems like, is that when the file is opened in the editor (by calling its URI), there is simply nothing that tells zowe explorer that the file has been opened, so subsequently it does not do any logic for actually adding the file to the history.

JWaters02:
> [@adam-wolfe](https://github.com/adam-wolfe) & [@JWaters02](https://github.com/JWaters02) Should this issue be closed with the bug fix [PR](https://github.com/zowe/zowe-explorer-vscode/pull/3441) or is there more to be done for this?

No, this bug is seperate to the fixes I made in #3441 

That PR fixes some issues related to refreshing within the edit history view itself, while this bug is to do with it not actually adding any  new items into the history in the first place (a much more major issues in my opinion).

PR #3477, opened 2025-02-21
Bug fixes on open recent member command

JWaters02:
## Proposed changes

Fixes both parts of #3447 

1. Keybind `ctrl + alt + r` has a couple of conflicts with system keybinds, so when you press the keybinding, vscode does not even attempt to call the command. I noticed that ctrl + alt + t is free (and I have quite a lot of extensions installed, so hopefully it is fine? Unless macOS has something). So changed it to that.
2. Fixed the bug with profile not being found causing an error
3. Fixed the bug where profiles being in the wrong case on uss files would be not found
4. Added open recent member into the command palette (this is nice QoL for me, a command palette enjoyer)

## Release Notes

Milestone: 3.1.2

Changelog:
- The "Open Recent Member" command can now be triggered from the command palette.
- Fixed issue where selecting a recent member to open from a profile name that no longer exists would cause an error. [#3447](https://github.com/zowe/zowe-explorer-vscode/issues/3447)
- Fixed issue with a conflicting keybinding for `Open Recent Member`, changed keybinding to `Ctrl`+`Alt`+`t` for Windows and `⌘ Cmd`+`⌥ Opt`+`t` for macOS. [#3447](https://github.com/zowe/zowe-explorer-vscode/issues/3447)

## Types of changes

- [x] Other (please specify above in "Proposed changes" section): Keybinding change

## Checklist

**General**

**Code coverage**

**Deployment**

- [ ] These changes may need ported to the appropriate branches (list here): 3.1.2

## Further comments

No mention of open recent member command in zowe docs, so nothing to be changed there in relation to keybinds changes.


JWaters02:
Why did DCO check fail? I signed off my commit with my GPG key. You can see the verified tag on my commit... is it because the DCO CI checks only for the "signed off by email" and does not take into account GPG keys?

packages/zowe-explorer/CHANGELOG.md
```diff
@@ -11,6 +11,7 @@ All notable changes to the "vscode-extension-for-zowe" extension will be documen
 - Implemented drag and drop functionality of data sets within and across LPARs. [#3413](https://github.com/zowe/zowe-explorer-vscode/pull/3413)
 - Renamed "Zowe Explorer z/OS Console" to "z/OS Console". [#3445](https://github.com/zowe/zowe-explorer-vscode/pull/3445)
 - Adopted ZE APIs new `directConnectLogin` and `directConnectLogout` methods for login and logout actions NOT using the tokenType `apimlAuthenticationToken`. [#3346](https://github.com/zowe/zowe-explorer-vscode/issues/3346)
+- The "Open Recent Member" command can now be triggered from the command palette.
```
traeok, line 14: Can you link your PR on this changelog entry? It identifies where the change was made and helps us give proper credits in the GitHub release 😋

traeok:
Thanks for working on this @JWaters02 - I'm running into an issue where trying to use the "Open Recent Member" results in an error that the "Node does not exist. It may have been deleted":

ZE-OpenRecentMember-Issue

traeok (dismissed):

JWaters02:
> Thanks for working on this @JWaters02 - I'm running into an issue where trying to use the "Open Recent Member" results in an error that the "Node does not exist. It may have been deleted":

Thanks I'll look into it

JWaters02:
> I'm running into an issue where trying to use the "Open Recent Member" results in an error that the "Node does not exist. It may have been deleted":

This appears to be a seperate bug where the member being saved is only saving its name to the history, not its full PDS + member name like it used to be done in v2. 

image

I don't think this is technically within scope of this PR as this PR only fixes open recent member, while the bug is caused by the saving to history, which would probably be related to #3446 

But I'll look into it anyway as it does directly break this feature, and if it's an easy one I'll probably include it in this PR anyway.

JWaters02:
Okay, fixed the bug in openDs. Therefore, this bug isn't related to #3446 because in that bug, openDs is not being called at all. So I will include this fix in this PR.

packages/zowe-explorer/src/trees/shared/SharedActions.ts
```diff
@@ -187,13 +187,21 @@ export class SharedActions {
             if (pattern.indexOf("/") > -1) {
                 // USS file was selected
                 const filePath = pattern.substring(pattern.indexOf("/"));
-                const sessionNode = ussTree.mSessionNodes.find((sessNode) => sessNode.getProfileName() === sessionName);
+                const sessionNode = ussTree.mSessionNodes.find((sessNode) => sessNode.label.toString().toLowerCase() === sessionName.toLowerCase());
```
zFernand0, line 190: I guess I haven't had a need to do this, but do we not support case-sensitive profile names? 😋

zFernand0 (approved):
Changes LGTM! 😋

packages/zowe-explorer/src/trees/dataset/ZoweDatasetNode.ts
```diff
@@ -689,7 +689,13 @@ export class ZoweDatasetNode extends ZoweTreeNode implements IZoweDatasetTreeNod
...
                 }
                 if (datasetProvider) {
-                    datasetProvider.addFileHistory(`[${this.getProfileName()}]: ${this.label as string}`);
+                    let criteria = `[${this.getProfileName()}]: `;
+                    if (SharedContext.isDsMember(this)) {
+                        criteria = `${criteria}${this.getParent().label as string}(${this.label as string})`;
+                    } else {
+                        criteria = `${criteria}${this.label as string}`;
+                    }
+                    datasetProvider.addFileHistory(criteria);
```
traeok, line 698: Thanks for fixing how the recent member is saved 🙏

traeok (commented):
The issue I encountered is now resolved. Thanks again for including that bug fix!

However, I've run into an issue where, after opening a data set member, the member is no longer added to that list:

ZE-OpenRecentMember-Issue2

This doesn't happen all the time, but seems to be after I open a file without selecting an encoding.

---

Also, maybe we can discuss this as a parking lot in the next ZE standup, but I do have some concerns about how we match against the session name (coercing using `toLowerCase`). Maybe this could be resolved separately since I could see this being "breaking" / involving more refactoring to address recent members that were already added in all caps.

JWaters02:
> This doesn't happen all the time, but seems to be after I open a file without selecting an encoding.

Yes, this is a bug reported in #3446 where I found that openDs is not being called when opening a file by just clicking on it, but opening it by opening with encoding does work as it explicitly calls the openDs function which adds the file to history.

packages/zowe-explorer/src/trees/shared/SharedActions.ts
JWaters02: Good point, left a comment about it below https://github.com/zowe/zowe-explorer-vscode/pull/3477#issuecomment-2704986548

traeok:
> > This doesn't happen all the time, but seems to be after I open a file without selecting an encoding.
> 
> Yes, this is a bug reported in #3446 where I found that openDs is not being called when opening a file by just clicking on it, but opening it by opening with encoding does work as it explicitly calls the openDs function which adds the file to history.

Ah I see, thanks for clarifying on that - I think we'll need to refactor how we add the data set to file history. Ultimately, we want the node's command to stay as `vscode.open` so that a double-click operation opens the file and a single click opens the file in preview mode. When we used a different command on the node, the user experience was not great - errors would bubble up from VS Code when double clicking on files (which is still the case in ZE v2), and we had no way of accurately determining whether the user double-clicked the node, as our "open data set" action would run twice 😬 not ideal...

I think we can leverage the `onDidOpenTextDocument` event to know when to add the resource to the file history. But I'm definitely not requesting that as part of this PR, just wanted to give context on why we moved away from calling a function to open nodes in the editor 😋 We'll look into this as a separate bug.

packages/zowe-explorer/src/trees/shared/SharedActions.ts
```diff
@@ -187,13 +187,21 @@ export class SharedActions {
...
+                if (!sessionNode) {
+                    Gui.showMessage(vscode.l10n.t("Profile not found."));
```
t1m0thyj, line 192: In the case that the profile is not found, should we delete the invalid recent entry from the list? Perhaps we could also add a prompt to the message asking the user if they want to keep or delete the invalid entry.

Just an idea, not a request changes for this PR - we could address this as a future enhancement 😋

t1m0thyj (approved):
LGTM, thanks @JWaters02!

zFernand0 (approved):
LGTM! 😋

traeok (approved):

Issue #3088, opened 2024-09-04
Expose logic for creating new team configs/schemas


traeok:
**Is your feature request related to a problem? Please describe.**

From the perspective of a Zowe Explorer extender, I would have to re-implement most of the logic for creating a new team config and schema. This would likely involve duplicating code that already exists in Zowe Explorer's codebase in order to access it.

**Describe the solution you'd like**

It would be helpful to provide some of this logic in ZE API, if possible, to avoid code duplication and to allow this feature to be leveraged by extenders.

**Describe alternatives you've considered**

Copy/pasting the code from Zowe Explorer itself, but this feels redundant and is not easy to maintain.

**Additional context**

_Follow up on discussion in issue #3078_

zFernand0:
This will be helpful for the CICS extension

FrankSu1996:
It would be nice to have the ability to add a custom profile if needed in the created team config instead of always the default profile attributes.

JillieBeanSim:
https://github.com/zowe/zowe-explorer-vscode/issues/3229 seems related to this one

sosadchuk:
Hello, thx for implementing this!

May I ask you to point me to some usage examples please?

I'm trying to init a team config along with our custom profiles, and even though I register them and see returned via `profilesCache.getConfigArray()`, it seems like the ProfileCache instance that is used with `createTeamConfiguration()` doesn't get them during this call `knownCliConfig.push(...ZoweVsCodeExtension.profilesCache.getConfigArray());`


My code sample:
```
private profileMetadata: imperative.ICommandProfileTypeConfiguration[]; // contains 2 profile definitions

  private async initProfiles(): Promise<void> {
    const types = this.profileMetadata.map((item) => item.type);

    const api = ZoweVsCodeExtension.getZoweExplorerApi(ZOWE_EXPLORER_API_VERSION);
    this.profilesCache = api.getExplorerExtenderApi().getProfilesCache();

    for (const type of types) {
      await api.getExplorerExtenderApi().initForZowe(type, this.profileMetadata);
      this.profilesCache.registerCustomProfilesType(type);
    }

    await this.profilesCache.refresh();
  }
```

PR #3520, opened 2025-03-17
Feature: creation of team config moved to ZE API and adopted by ZE

JillieBeanSim:
## Proposed changes

Feature: creation of team config moved to ZE API and adopted by ZE
created `...-ext` unit test file as adding `jest.mock("@zowe/imperative")` which is needed for these unit tests broke many of the existing in OG unit test file.

## Release Notes

Milestone: 3.2.0

Changelog:

- ZE API - Added new API, `ZoweVsCodeExtension.createTeamConfiguration`, allowing extenders to use the same prompting process adopted by Zowe Explorer for Zowe team configuration creation including all registered profile types. [#3088](https://github.com/zowe/zowe-explorer-vscode/issues/3088)
- ZE - Adopted new Zowe Explorer API, `ZoweVsCodeExtension.createTeamConfiguration` for Zowe team configuration creation. [#3088](https://github.com/zowe/zowe-explorer-vscode/issues/3088)

## Types of changes

## Checklist

**General**

**Code coverage**

**Deployment**


packages/zowe-explorer/src/configuration/Profiles.ts
```diff
@@ -690,6 +627,7 @@ export class Profiles extends ProfilesCache {
         await this.openConfigFile(filePath);
     }
 
+    // eslint-disable-next-line @typescript-eslint/explicit-function-return-type
     public async profileValidationHelper(theProfile: imperative.IProfileLoaded, getStatus: (...args: unknown[]) => Promise<string>) {
```
traeok, line 631: I saw that this line was added because the function has been in place for some time without a return type. Thoughts on updating the return type for this function to be explicit? It also eliminates the need for the ESLint comment.

```suggestion
    public async profileValidationHelper(theProfile: imperative.IProfileLoaded, getStatus: (...args: unknown[]) => Promise<string>): Promise<string> {
```

packages/zowe-explorer-api/src/profiles/ProfilesCache.ts
```diff
@@ -38,19 +43,38 @@ export class ProfilesCache {
...
-            const index = this.profileTypeConfigurations.findIndex((ele) => ele.type == item.type);
+            const index = ProfilesCache.sessionProfileTypeConfigurations.findIndex((ele) => ele.type == item.type);
             if (index !== -1) {
-                this.profileTypeConfigurations[index] = item;
+                // eslint-disable-next-line deprecation/deprecation
+                ProfilesCache.sessionProfileTypeConfigurations[index] = this.profileTypeConfigurations[index] = item;
             } else {
+                // eslint-disable-next-line deprecation/deprecation
                 this.profileTypeConfigurations.push(item);
+                ProfilesCache.sessionProfileTypeConfigurations.push(item);
```
traeok, line 65: I noticed we introduced a new static property for the `profileTypeConfigurations`. Is this property required? This question is more for my own clarification, as we already have a non-static property that holds the same data.

packages/zowe-explorer-api/src/profiles/ProfilesCache.ts
```diff
@@ -38,19 +43,38 @@ export class ProfilesCache {
...
 
+    /**
+     * Returns an array of Zowe Explorer registered profile types and core Zowe supported profile types meta-data ,
+     * used with Zowe team configuration actions.
+     *
+     * @returns {imperative.ICommandProfileTypeConfiguration[]}
+     */
     public getConfigArray(): imperative.ICommandProfileTypeConfiguration[] {
-        return this.profileTypeConfigurations;
+        return ProfilesCache.sessionProfileTypeConfigurations;
```
traeok, line 77: This change won't directly impact us, but I do wonder if this affects extenders who extend `ProfilesCache` for their own use cases...

Since subclasses of `ProfilesCache` are likely uncommon, and the change could be argued as a bug fix, I'm not specifically requesting changes for this - just something that came to mind 😋

packages/zowe-explorer-api/src/profiles/ProfilesCache.ts
JillieBeanSim: Hey @traeok, I had to add a static constant to hold the registered values.  When I tried the adoption with the exiting constant I would only create a config with the default values that we add in the method, the registered array was always empty.  Having a static constant solved the issue.

packages/zowe-explorer-api/src/profiles/ProfilesCache.ts
JillieBeanSim: in the sense of extender, this change will fix a bug where they were not receiving the registered extender types just the core types we add in the method.

JillieBeanSim:
found an already existing issue/enhancement to bring localization into the ZE API package https://github.com/zowe/zowe-explorer-vscode/issues/3283

packages/zowe-explorer-api/src/profiles/ProfilesCache.ts
zFernand0: Will there be any concerns with a possible de-sync? 

Hypothetically, a VSCode element (quickpick, ...) may need to be disposed off in order to get updated contents from a static object/variable. 😋

zFernand0 (dismissed):
Changes LGTM! 😋 
---

My only concern might be the possible loss of logs and localization 😋 
But other than that, this LGTM! 🥳

JillieBeanSim:
@zFernand0 I started a discussion for the squad about the lack of logging with more functionality being moved into the ZE API package
https://github.com/zowe/zowe-explorer-vscode/discussions/3553

zFernand0 (approved):
LGTM! 😋

packages/zowe-explorer-api/src/profiles/ProfilesCache.ts
traeok: I'm inclined to roll with the static array for now, as the intention of `ZoweVsCodeExtension.profilesCache` is to expose a single instance of the `ProfilesCache` across all extenders. However I agree with Fernando's concerns that for other instances of the `ProfilesCache`, this may change the expected results of this array when that instance is refreshed. That said, since we currently expose the profiles cache as a property and don't advertise creating a separate instance, this should work fine in most cases. 

Something to consider for v4: if we find that properties in the cache need to be static to resolve desync scenarios, we should consider making the `ProfilesCache` itself a singleton so that only a single instance is used across all extenders & ZE itself. Currently, anyone can create an instance of the `ProfilesCache` class to access profiles, but its not always going to be up-to-date with the instance that Zowe Explorer uses - this could be mitigated to some extent w/ a singleton implementation.

traeok (approved):
LGTM, thanks Billie!

PR #3525, opened 2025-03-18
fix: Only check if token has expired when profile's API supports it

traeok:
## Proposed changes

Fixes an issue where the profile expiration check had continued executing for profiles that do not explicitly support tokens through their API implementations.

## Release Notes

Milestone: 3.1.2

Changelog:

- Fixed an issue where the token expiration check was performed on profiles whose APIs do not support a token type

## Types of changes

## Checklist

**General**

**Code coverage**

**Deployment**

### DIFF
--- a/packages/zowe-explorer/src/trees/ZoweTreeProvider.ts
+++ b/packages/zowe-explorer/src/trees/ZoweTreeProvider.ts
@@ -340,9 +340,9 @@ export class ZoweTreeProvider<T extends IZoweTreeNode> {
      *
      * @param profileName The name of the profile to check the JWT token for
      * @returns
-     * `true` if:
-     * - the user attempted to log in
-     * - the profile does not have a token
+     * `true` if one of these conditions are met:
+     * - the token expired and the user clicked log in
+     * - the profile does not have a token or does not support tokens
      * - the token has not expired on the profile
      *
      * `false` if:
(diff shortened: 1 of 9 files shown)


anaxceron (approved):

packages/zowe-explorer/src/trees/ZoweTreeProvider.ts
```diff
@@ -339,17 +340,30 @@ export class ZoweTreeProvider<T extends IZoweTreeNode> {
      *
      * @param profileName The name of the profile to check the JWT token for
      * @returns
-     * `true` if:
-     * - the user attempted to log in
-     * - the profile does not have a token
+     * `true` if one of these conditions are met:
+     * - the token expired and the user clicked log in
+     * - the profile does not have a token or does not support tokens
      * - the token has not expired on the profile
```
zFernand0, line 346: Thanks for updating.
Still a bit confusing to return `true` if:
- the token expired (and the suer clicked log in) 
- the token has not expired

zFernand0 (approved):
Changes LGTM! 😋

packages/zowe-explorer/src/trees/ZoweTreeProvider.ts
traeok: Yeah I'm not a huge fan of the scenarios where the function returns true/false, I can look into refactoring it further 😅

traeok: Following up on your concerns for the function:

- I've renamed the function to `checkJwtForProfile` and explicitly defined what a JWT (JSON Web Token) is in the typedoc. It also avoids the function from essentially being named "check(JsonWebToken)TokenForProfile" 😆 
- I've refactored the function logic to return a `JwtCheckResult` enum value as a `boolean` return type didn't feel right here, especially now that the "successful" scenarios have become more specific. From making a change to the enum, I noticed an oversight with the boolean check. Now, we only mark the profile as inactive when either of these conditions are true:
  - the user responds to "token expired" prompt, but login fails 
  - the user cancelled the "token expired" prompt
- I've cleaned up the typedoc to clarify the explicit results of invoking the function.

Thanks for the feedback - this should serve as a more robust approach.

zFernand0 (approved):
LGTM! 😋 
---

Thanks for refactoring that function and using an enum 🙏 
This makes the logic a lot easier to follow 🥳

JillieBeanSim (approved):
LGTM! thanks @traeok

PR #3526, opened 2025-03-18
Add data set search case sensitivity (#3455)

### LINKED ISSUE #3455: Add case-sensitive option to searching for a string in data sets

Currently, searching is case-insensitive. In some cases, it would be useful to perform case-sensitive searching to limit the number of results returned to only those results that match the case of the specified query.

Let users specify whether the search should be case-sensitive.

**Additional context**

In Zowe CLI there is a `--case-sensitive` option for the `zowe files search ds` command.

awharn:
## Proposed changes

Resolves #3455 by adding a quick pick option to prompt users for case sensitivity of searches

## Release Notes

Milestone: 3.2.0

Changelog: Added case sensitive search options to Data Set searches

## Types of changes

## Checklist

**General**

**Code coverage**

**Deployment**

## Further comments

Screen Recording 2025-03-18 at 11 08 11 AM


anaxceron (approved):

zFernand0 (approved):
LGTM! 😋

adam-wolfe:
Looks good. Wondering if we should consider running this by UX

JWaters02:
Minor nitpick, is it necessary (or good UX) for there to be two steps when searching? I.e. currently, you enter the search string, press enter, then another quick pick step appears with both a string input box and the case sensitive option. Even though it says in the input field to select an option to search on, since you can still type stuff in there it isn't obvious that the input field is acting as a search bar for the quick pick selections (an unfortunate quick pick reality that cannot be disabled).

Just as an easier comparison, here's the two ways it could work. The major downside of the second option is that since the quick pick multi-select is in the same place as the place you put the search string, your search string is also filtering on the quick pick options, which is also a bit confusing. I wish quick pick had an option to disable filtering on options entirely, but I don't think it does ☹️ So both options feel a bit naf. Which one is the least worst? 😅 

https://github.com/user-attachments/assets/b06ac756-98c8-49c4-960f-cda31fe2689e

packages/zowe-explorer/src/trees/dataset/DatasetActions.ts
```diff
@@ -1890,10 +1891,32 @@ export class DatasetActions {
...
         }
 
+        // Handle various search options
+        const caseSensitiveQuickPick = await Gui.showQuickPick(
+            [
+                {
+                    label: vscode.l10n.t("Case Sensitive"),
+                    description: vscode.l10n.t("Perform the search with case sensitivity"),
+                    iconPath: new vscode.ThemeIcon("case-sensitive"),
+                } as vscode.QuickPickItem,
```
traeok, line 1904: It seems that this cast is not required since it does not change the type inferred by `Gui.showQuickPick`, although it does make the type inference look cleaner when hovering over the function. Not requesting any changes for this, but just something I observed when reviewing

traeok (dismissed):
LGTM, thanks Andrew for the enhancement!

awharn:
@JWaters02 I like the second option, but my only concern was that there has been a little discussion about keeping a history of searches. If we end up doing that, combining them into a single prompt might not be feasible. That being said, if we want to establish that we are not going to do that, I quite like that solution, keeping things in one input.

JWaters02:
> @JWaters02 I like the second option, but my only concern was that there has been a little discussion about keeping a history of searches. If we end up doing that, combining them into a single prompt might not be feasible. That being said, if we want to establish that we are not going to do that, I quite like that solution, keeping things in one input.

Ah yeah you're right, that's a good point. The first option would indeed make sense with search history. I think therefore what is currently in this PR is fine. It's a couple of months until 3.2 right, so a search history feature may be implemented before then anyway 👍

packages/zowe-explorer-api/CHANGELOG.md
```diff
@@ -6,6 +6,7 @@ All notable changes to the "zowe-explorer-api" extension will be documented in t
 
 ### New features and enhancements
 
+- Updated `Gui.showQuickPick` to support selecting multiple `quickPickItem`s. [#3526](https://github.com/zowe/zowe-explorer-vscode/pull/3526)
```
anaxceron, line 9: - We should explain what `Gui.showQuickPick` is: "Updated the `Gui.showQuickPick` **package/rocket ship/whatever** to support ... "
- We should avoid pluralizing  `quickPickItem` like this ^^: " ... selecting more than one  `quickPickItem`."

anaxceron (commented):
Made a comment on a changelog item. Not sure if it's related to this PR or not, but there you have it 🤷

t1m0thyj (commented):
LGTM, thanks @awharn! Waiting to approve until the UX has been finalized 🙂

packages/zowe-explorer/src/trees/dataset/DatasetSearch.ts
```diff
@@ -0,0 +1,427 @@
...
+            DatasetSearch.searchQuickPickReset();
+        });
+    }
+
+    private static searchQuickPickReset(): void {
+        DatasetSearch.searchUpdateOptionsLabel();
+
+        // This statement is required to show the quick pick items again after the search options are edited.
+        // eslint-disable-next-line no-self-assign
+        DatasetSearch.searchQuickPick.items = DatasetSearch.searchQuickPick.items; // Do not remove this line.
```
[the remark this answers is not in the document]
awharn: Yes, this does assign items to itself. However, as can be confirmed by @zFernand0, @t1m0thyj, and @traeok, VSCode has some weird behavior when showing a quick pick for a second time. Without this line, all of the quick pick items disappear from the dropdown, hiding the user's history and preventing them from accessing the search options.

adam-wolfe:
Would it be possible to allow the user to modify the query after they select it from the recent searches (like we do for listing data sets)? It's a minor nitpick, but if I need to adjust something in a regex, I have to retype the whole regex.

awharn:
An updated screen recording on the functionality and Quick Picks.

https://github.com/user-attachments/assets/60fdb93b-d36e-4e1c-8c74-82ed9b1f026e

JWaters02:
> An updated screen recording on the functionality and Quick Picks.

Looks really good!

anaxceron (commented):
Made some comments re: changelogs

t1m0thyj (commented):
Changes LGTM, thanks @awharn! Only comments are on the changelog

Edit: Resolved my comments since I see they are duplicates of Ana's 😋

anaxceron (approved):
Great changelogs @awharn 😸

t1m0thyj (approved):

traeok (dismissed):
LGTM, thanks for the enhancements Andrew!

traeok (approved):

t1m0thyj (approved):

PR #3616, opened 2025-04-24
Job Spool Pagination Bug fixes

pujal0909:
## Proposed changes

https://github.com/zowe/zowe-explorer-vscode/issues/3600
1. Fixed a bug where the load more button did not work for spool files lines that are greater than the default.
2. Changed toggle and default records settings to be under Zowe -> jobs -> paginate to be consistent with the data sets pagination feature.
3. Key binding `ctrl + l` updated to work as expected now.

To test functionality:

1. Run the debugger
2. In the Zowe explorer settings, check off the setting to enable job spool pagination. You can set the number of records to fetch when polling spool files.
3. Create a shell script and submit a member with JCL that points to the script.
4. View the submitted job's output. You can click on the load more code lens to load additional contents of the job spool. `Ctrl + l` can be used as well to load contents.

## Release Notes

Milestone:

Changelog:

## Types of changes

## Checklist

**General**

**Code coverage**

**Deployment**


packages/zowe-explorer-api/CHANGELOG.md
```diff
@@ -21,6 +21,8 @@ All notable changes to the "zowe-explorer-api" extension will be documented in t
 
 ### Bug fixes
 
+-BugFix: Fixed an issue with job spool pagination where the load more button did not work for spools files greater than the default record length. [#3600](https://github.com/zowe/zowe-explorer-vscode/issues/3600)
```
JWaters02, line 24: Is this changelog entry necessary since the feature does not exist yet?

JWaters02:
That works great, keybind included! 🙂 

I still believe that a must have bug fix is to hide the load more button if there are no more records to load. 

Couple minor bugs:
- It is fetching the settings number of records + 1. So default of 100 is fetching 101 records, or 1 is fetching 2 records.
- If you change jobs > paginate > records to fetch to `0` (or less) this warning appears and then the pagination feature breaks even if I change the setting back to 100. It works again once I restart vscode. People shouldn't be setting it to 0, but the feature breaking after changing the setting after a warning/error is a case that could crop up again in a future edge case that I think is worth looking into fixing.
image

zFernand0:
> I still believe that a must have bug fix is to hide the load more button if there are no more records to load.

Hey @JWaters02,
I like the idea, but would you agree that we can only remove the codeLens if the job has finished (Queue: `OUTPUT`)?
Otherwise, I don't believe we can determine whether we will receive new records or not.

JWaters02:
> > I still believe that a must have bug fix is to hide the load more button if there are no more records to load.
> 
> Hey @JWaters02, I like the idea, but would you agree that we can only remove the codeLens if the job has finished (Queue: `OUTPUT`)? Otherwise, I don't believe we can determine whether we will receive new records or not.

Oh yeah good point, didn't think of that. Yeah, I think that would have to be the case.

JWaters02:
Great, all seems to work well now... except one more edge case (probably as a result of a previous bug fix); if you set records to fetch to 1 and try to open any spool file, this error appears instead:
image

packages/zowe-explorer/src/trees/job/JobFSProvider.ts
```diff
@@ -225,18 +227,23 @@ export class JobFSProvider extends BaseProvider implements vscode.FileSystemProv
...
         const query = new URLSearchParams(uri.query);
         let recordRange = "";
+        const recordsToFetch = SettingsConfig.getDirectValue<number>("zowe.jobs.paginate.recordsToFetch") ?? 0;
+
         if (query.has("startLine")) {
             const startLine = parseInt(query.get("startLine")!);
-            const endLine = query.has("endLine")
-                ? (parseInt(query.get("endLine")!))
-                : (startLine + (SettingsConfig.getDirectValue<number>("zowe.jobs.paginate.recordsToFetch") ?? 0)) - 1;
+            const endLine = query.has("endLine") ? parseInt(query.get("endLine")!) : startLine + (recordsToFetch - 1);
```
traeok, line 234: Would the same problem also occur here when records to fetch is 1?
I know its highly unlikely that someone would use this number for that setting, but if `recordsToFetch` is 1 here, then `endLine` ends up being the same value as `startLine` rather than `startLine + 1`

packages/zowe-explorer/src/trees/job/JobFSProvider.ts
```diff
@@ -225,18 +227,23 @@ export class JobFSProvider extends BaseProvider implements vscode.FileSystemProv
...
+            console.log(endLine);
```
traeok, line 235: Looks like there are `console.log` statements on line 235, 237 and 262 from testing

packages/zowe-explorer/src/trees/job/JobFSProvider.ts
```diff
@@ -262,11 +271,25 @@ export class JobFSProvider extends BaseProvider implements vscode.FileSystemProv
...
+                const content = bufBuilder.read()?.toString("utf-8") ?? "";
+                this.allLines = content.split(/\r?\n/).filter((line) => line.trim() !== "");
+                this.currentLine = 0;
+            }
+            if (this.currentLine < this.allLines.length) {
+                const line = this.allLines[this.currentLine];
+
+                spoolEntry.data = Buffer.concat([spoolEntry.data, Buffer.from(line + "\n", "utf-8")]);
+                this.currentLine += 1;
+            }
```
traeok, line 286: I think we'll need to avoid putting `allLines` and `currentLine` directly on the jobs FS provider as this would cause undefined behavior with 2 spool files open & pagination enabled.

Question for my own clarification: is there a reason why we need different logic here when the records to fetch is 1?

It seems like if one record was pulled using record range, the same method of `Buffer.concat(...)` on line 288 would work to add the new line to the existing contents. But maybe I'm missing something 😋

traeok (dismissed):
Thanks Pujal! I have a couple requests, but happy to discuss in more detail offline 👍

pujal0909:
@JWaters02 
Thanks for catching that edge case of when records to fetch is set to 1! The API doesn't support fetching a single record since the record range is inclusive. To avoid this error and since it is unlikely a user would set it to 1, I set the minimum allowed value for records to fetch to 2 in the settings.

adam-wolfe:
One thing I'm seeing: if I disable job spool pagination, open a spool file, and click load more, the contents of the spool file are added to the editor (so that the contents are duplicated)

packages/zowe-explorer/src/trees/job/JobInit.ts
```diff
@@ -161,9 +161,10 @@ export class JobInit {
...
-        const paginationEnabled = SettingsConfig.getDirectValue<boolean>("zowe.jobs.settings.pagination");
-        if (paginationEnabled) {
-            context.subscriptions.push(vscode.commands.registerCommand("zowe.jobs.loadMoreRecords", (doc) => JobActions.loadMoreRecords(doc)));
+        const paginationEnabled = SettingsConfig.getDirectValue<boolean>("zowe.jobs.paginate.enabled");
+        const recordsToFetch = SettingsConfig.getDirectValue<number>("zowe.jobs.paginate.recordsToFetch");
+        if (paginationEnabled && recordsToFetch > 1) {
+            context.subscriptions.push(vscode.commands.registerCommand("zowe.jobs.loadMoreRecords", (uri) => JobActions.loadMoreRecords(uri)));
             vscode.languages.registerCodeLensProvider(
                 { scheme: ZoweScheme.Jobs },
                 new PaginationCodeLens("zowe.jobs.loadMoreRecords", (doc) => JobFSProvider.instance.supportSpoolPagination(doc))
```
traeok, line 170: I'm seeing the same behavior as Adam. As part of resolving this, I think we should update `JobFSProvider.supportSpoolPagination` so that it returns false if pagination is disabled, so that the code lens is hidden in the editor when its off.

zFernand0 (commented):
LGTM! :yum:
---
There is one edge case that I can't figure out exactly how to reproduce, but here is what I'm seeing...
image

I'll provide a quick update once I find where that's coming from (._.❓)

UPDATE: 
This also happens on `main`, meaning it wasn't not introduced by the pagination efforts on spool files 
I'm still confused though (._.?)

zFernand0 (approved):

traeok:
> UPDATE: This also happens on `main`, meaning it wasn't not introduced by the pagination efforts on spool files I'm still confused though (._.?)

Feel free to correct me if I'm wrong @zFernand0, but we already have the job spool pagination merged into main so its possible that this bug was recently introduced. Can you reproduce this with the v3.1.2 tag? If so, feel free to disregard, but would like to rule out the possibility of a regression 😅

zFernand0:
> Can you reproduce this with the v3.1.2 tag? 

On the V3.1.2 tag I'm not able to list any spool files 🤔 
image

I can reproduce the issue when testing with the v3.1.2 version on the marketplace, the `main` branch, and this PR. (._.?)
It's still a bit confusing as to what's causing the spool contents to go away. The contents show for like half a second, but then the editor goes blank and shortly after the `Load more...` appears. There seems to be some sort of race condition happening in `main`, thus I don't believe it is related to this PR 🙏

traeok (approved):
Thanks for addressing my feedback Pujal! LGTM

traeok:
> With the latest changes, if I enable pagination, set records to fetch to something low like 5, open a spool file and click `load more` nothing happens (when I would expect Zowe Explorer to get the next 5 records).
> 
> Also, if I set records to fetch to 3, Zowe Explorer gets 4 records.

Struggling to reproduce this scenario but continuing to investigate - I tested with a spool that prints 1 to 1000, each on a new line (using a shell script, BPXBATCH). If I open the spool file and the last line is the number 5, and I set records to fetch to 5, it adds the following lines to the end of the doc:

```
6
7
8
9
10
```

adam-wolfe:
Disregard my comment. VS Code was failing to install the extension. Looks good to me now.

traeok (approved):

awharn (approved):

PR #3625, opened 2025-04-29
Fix duplicate nodes when adding profiles

jace-roell:
## Proposed changes

Extenders that may add `base` or `ssh` to the profile cache type array will result in duplicate nodes for ssh or base profile types when adding a profile to a tree.

Adding an ssh profile from an extender before the fix:
image

This fix changes the profile types to a set to avoid duplicate types and thus duplicate nodes

Adding an ssh profile from an extender after the fix:
image

## Release Notes

Milestone:

Changelog:

## Types of changes

## Checklist

**General**

**Code coverage**

**Deployment**

### DIFF
--- a/packages/zowe-explorer-api/CHANGELOG.md
+++ b/packages/zowe-explorer-api/CHANGELOG.md
@@ -4,6 +4,8 @@ All notable changes to the "zowe-explorer-api" extension will be documented in t
 
 ## TBD Release
 
+- BugFix: Resolved a bug where extenders adding `ssh` or `base` to `allTypes` on `ProfileCache` will result in duplicate nodes when adding a profile of that type in Zowe Explorer. [#3625](https://github.com/zowe/zowe-explorer-vscode/pull/3625)
+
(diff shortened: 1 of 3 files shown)


traeok (commented):
Tested w/ an extension that contributes SSH and its working as expected 😋 also saw the changes working in-person - thanks Jace!
I was wondering if we should have a unit test for this 🤔 I notice that the patch coverage is unaffected, but it would be nice to have a test to prevent this from happening in the future

jace-roell:
> Tested w/ an extension that contributes SSH and its working as expected 😋 also saw the changes working in-person - thanks Jace! I was wondering if we should have a unit test for this 🤔 I notice that the patch coverage is unaffected, but it would be nice to have a test to prevent this from happening in the future

Good point! I added a unit test to ensure no duplicate profile types exist after running profCache.refresh(). I also compared against main and the new test does fail as seen below. 
image

traeok (approved):
Thanks for adding the unit test! LGTM

Issue #3664, opened 2025-05-20
cached profile information with use of node.getProfile


JillieBeanSim:
**Describe the bug**

instead of node.getProfile we should use Profiles.loadNamedProfile to get the latest information about the profile from profilesCache to avoid having out of date information

**Expected behavior**

have up to date profile information for each action within Zowe Explorer

**Screenshots**

**Desktop (please complete the following information):**

- OS:
- Zowe Explorer Version:
- (Optional) Zowe CLI Version:
- (Optional) Do you have secure credentials enabled?
- (Optional) Are you using a custom credential manager?

**Additional context**

Issue #3662, opened 2025-05-20
ZE FileSystemProvider `stat` function throws the authentication error for correct credentials


hohy:
**Describe the bug**

When calling the `stat` function of the Zowe Explorer `FileSystemProvider` while logged out of the Authentication Service, a `FileSystemError` is thrown even though correct credentials are provided during the authentication prompt. The error indicates an authentication failure, but subsequent `stat` calls with the same parameters succeed without error.

**To Reproduce**

1. Open VSCode
2. Configure `zosmf` profile but DO NOT log in to Authentication Service
3. Call ZE FS provider with code similar to this
```ts
try {
  await vscode.workspace.fs.stat(
    vscode.Uri.parse(`zowe-ds:/${profileName}/${dsn}?fetch=true`)
  );
} catch (error) {
  vscode.window.showErrorMessage(error.stack);
}
```
4. Zowe Explorer prompts for login to the Authentication Service
5. Enter correct credentials
6. ZE displays Login to authentication service successful dialog.
7. Despite the successful login, the `stat` function throws the following error:
```
Error (FileSystemError): Rest API failure with HTTP(S) status 401 This operation requires authentication.
```
8. Repeating the same `stat` call with the same parameters returns the correct result without error.

**Expected behavior**
The initial `stat` call should succeed if correct credentials are provided during the authentication prompt, even if the user was not previously logged in. No error should be thrown in this scenario.

**Screenshots**

![Image](https://github.com/user-attachments/assets/784793e6-1077-4dae-b9da-d2bd9f00bbe3)

**Desktop (please complete the following information):**

- OS: macOC
- Zowe Explorer Version: 3.2.0
- (Optional) Zowe CLI Version:
- (Optional) Do you have secure credentials enabled?
- (Optional) Are you using a custom credential manager?

**Additional context**

traeok:
This behavior is seen because the current file system implementation does not retry requests after a user has acknowledge an authentication failure. The enhancement proposed in #3726 resolves this scenario and could serve as a general application for various functions in the filesystem.

Issue #3307, opened 2024-11-11
Consolidate profile facilities and improve caching efficiency


traeok:
**Is your feature request related to a problem? Please describe.**

Potentially related: #2577 #3303 

**Describe the solution you'd like**

We should reduce the amount of times we read from disk when using the `ProfilesCache` facility. We also should cache more information in the profiles cache, or at least cache both the `ProfilesCache` and `ProfileInfo` instances, re-using them wherever possible.

**Describe alternatives you've considered**

We can continue to use the facilities available in the codebase, but the issue of excessive profile I/O operations and the scattered function implementations will continue to fester until it becomes a primary bottleneck in some Zowe Explorer actions/operations.

PR #3673, opened 2025-05-29
Node cached profile fix

JillieBeanSim:
## Proposed changes

updated instances of node.getProfile() to use loadNamedProfile to get latest from ProfilesCache in ZE API
solution is zosmf and other extenders that return 401 specific atm, zftp bug raised due to not returning 401.
can test by opening a file within one of the trees, change creds to profile in another tree to incorrect and make a change in open file and save. Should be notified of invalid creds and accept to get prompted for correct creds then attempt to make change to file and save again.  Last attempt should be successful showing that the lowest tree nodes are grabbing updated profile information for z/OS actions

## Release Notes

Milestone: 3.2.2

Changelog: Fixed issue seen with out of date profile information tied to tree nodes. [#3664](https://github.com/zowe/zowe-explorer-vscode/issues/3664)

## Types of changes

## Checklist

**General**

**Code coverage**

**Deployment**

### DIFF
--- a/packages/zowe-explorer/src/utils/AuthUtils.ts
+++ b/packages/zowe-explorer/src/utils/AuthUtils.ts
@@ -62,7 +62,7 @@ export class AuthUtils {
 
             const authOpts: AuthPromptParams = {
                 authMethods: Constants.PROFILES_CACHE,
-                imperativeError: new imperative.ImperativeError({ msg: String(err), stack: err.stack }),
+                imperativeError: err as unknown as imperative.ImperativeError,
                 isUsingTokenAuth: await AuthUtils.isUsingTokenAuth(profile.name),
                 errorCorrelation,
             };


JillieBeanSim:
> Thanks @JillieBeanSim for your work on this!
> 
> I understand why we want to call `Profiles.getInstance().loadNamedProfile(node.getProfileName())` to obtain the latest version of the profile rather than a cached one. Would it be easier for us to update the implementation of `node.getProfile` so that we don't have to replace it everywhere?

Quickly trying a couple things with this approach and end up with numerous circular dependencies. I'll continue looking for a possible way other than what I have implemented here

packages/zowe-explorer-api/CHANGELOG.md
```diff
@@ -49,6 +49,7 @@ All notable changes to the "zowe-explorer-api" extension will be documented in t
 - Generalized and updated the README. [#3367](https://github.com/zowe/zowe-explorer-vscode/issues/3367)
 - Added ability to pass the `OverrideWithEnv` property to the `imperative.ProfileInfo` class when the setting "Override with Environment Variables" is enabled. When enabled, environment variables such as `ZOWE_OPT_USER` and `ZOWE_OPT_PASSWORD` take priority over properties specified in the team configuration. [#3591](https://github.com/zowe/zowe-explorer-vscode/pull/3591)
 - Implemented the `Paginator` class, providing a generic facility for extenders to paginate tree nodes within their own tree views. [#3585](https://github.com/zowe/zowe-explorer-vscode/pull/3585)
+- Added an optional parameter of ProfileCache instance to ZoweTreeNode.getProfile to fix an issue seen with out of date profile information within tree views. [#3664](https://github.com/zowe/zowe-explorer-vscode/issues/3664)
```
t1m0thyj: Should we use codeblocks here?
```suggestion
- Added an optional parameter of `ProfileCache` instance to `ZoweTreeNode.getProfile` to fix an issue seen with out of date profile information within tree views. [#3664](https://github.com/zowe/zowe-explorer-vscode/issues/3664)
```

anaxceron: @t1m0thyj -- yes, we should use backticks for both. We should also describe what `ZoweTreeNode.getProfile` is: " ... instance t the `ZoweTreeNode.getProfile` cat/car/bar to fix an issue ... "

traeok:
Thanks Billie for looking into this. I just saw your latest reply and wanted to ask while its fresh in my mind:

> found the issue with extenders at the file level, it lies in the error produced in FS Provider but authChecks are never called. making draft and will update to cover these scenarios

I was wondering if this error seen is related to #3662? I'm currently pursuing a fix that throws a more explicit error if the user cancels the authentication dialog shown from a call to the filesystem. It's meant to help extenders handle the case where the authentication prompt is cancelled. In combination with the retry functionality proposal in #3726, I'm thinking this may resolve some of the concerns encountered by extenders.

JillieBeanSim:
Hey @traeok I found the changes I had in FSProvider isn't needed, just an update in AuthUtils we used to check for 401 error string but now only check for errCode.  The FTP ext will still need update to add an error handler and return a 401 error instead of pass failed error but fixes other extenders that are recently broke. So @t1m0thyj 's part makes sure nodes all throughout each tree are grabbing the latest profile information and this addition makes sure to handle all instances of 401 error returned as we used to only rely on the string but had changed to only rely on error code.  Now we check for both

packages/zowe-explorer-api/src/tree/ZoweTreeNode.ts
traeok, line 1: I'm not necessarily opposed to this change if we want to support different/multiple profile caches, but if we have to pass an optional cache here to resolve extender issues, this suggests that the same "Profiles" instance is not used amongst Zowe Explorer and all its extenders. 

I think #3307 was aimed to resolve this, but I see it was closed as stale. Should we reopen it to address this problem more broadly? I think for a complete resolution, it would involve breaking changes, but it could be done for v4. Maybe then we could consider the idea of code-base dependency injection for the profiles cache, so that same instance can be more easily passed around/reused rather than depending on a single instance (which could be considered tight coupling).

packages/zowe-explorer-api/CHANGELOG.md
```diff
@@ -49,6 +49,7 @@ All notable changes to the "zowe-explorer-api" extension will be documented in t
 - Generalized and updated the README. [#3367](https://github.com/zowe/zowe-explorer-vscode/issues/3367)
 - Added ability to pass the `OverrideWithEnv` property to the `imperative.ProfileInfo` class when the setting "Override with Environment Variables" is enabled. When enabled, environment variables such as `ZOWE_OPT_USER` and `ZOWE_OPT_PASSWORD` take priority over properties specified in the team configuration. [#3591](https://github.com/zowe/zowe-explorer-vscode/pull/3591)
 - Implemented the `Paginator` class, providing a generic facility for extenders to paginate tree nodes within their own tree views. [#3585](https://github.com/zowe/zowe-explorer-vscode/pull/3585)
+- Added an optional parameter of `ProfileCache` instance to `ZoweTreeNode.getProfile` to fix an issue seen with out of date profile information within tree views. [#3664](https://github.com/zowe/zowe-explorer-vscode/issues/3664)
```
traeok:
```suggestion
- Added an optional `profilesCache` parameter to `ZoweTreeNode.getProfile` to fix an issue seen with out of date profile information within tree views. When provided, the tree node uses that cache to obtain profile information. [#3664](https://github.com/zowe/zowe-explorer-vscode/issues/3664)
```

packages/zowe-explorer/__tests__/__unit__/trees/dataset/ZoweDatasetNode.unit.test.ts
traeok, line 1: Thanks for refactoring this test case to avoid `Object.defineProperty`, I know there are several across our tests but every change helps 😋

packages/zowe-explorer/__tests__/__unit__/trees/job/JobActions.unit.test.ts
```diff
@@ -999,7 +999,7 @@ describe("Jobs Actions Unit Tests - Function submitMember", () => {
                     }
                 );
             }
-            expect(mocked(Profiles.getInstance)).toHaveBeenCalledTimes(2 * (o + 1));
+            expect(mocked(Profiles.getInstance)).toHaveBeenCalledTimes(4 * (o + 1) + 2);
```
traeok, line 1002: (question) I'm wondering if doing a `mockClear` on `mocked(Profiles.getInstance)` before the for loop would cause the number of calls to satisfy the previously-expected value in the test. Not explicitly requesting changes for this as there may be additional context for why the number of calls has increased.

packages/zowe-explorer/__tests__/__unit__/trees/uss/USSTree.unit.test.ts
```diff
@@ -861,6 +861,7 @@ describe("USSTree Unit Tests - Function filterBy", () => {
         const blockMocks = await createBlockMocks();
         globalMocks.isUsingTokenAuth.mockResolvedValueOnce(true);
         const sessionWithCred = createISession();
+        globalMocks.mockProfilesInstance.loadNamedProfile.mockReturnValue({ name: "ussTestSess2" } as any);
```
traeok: I think we should try to fully stub test profiles (when possible) to ensure proper processing for the tree node logic:

```suggestion
        globalMocks.mockProfilesInstance.loadNamedProfile.mockReturnValue(createIProfile());
```

This ensures that name, type, etc. are present in the profile object before a function tries to do logic on it. Ran into a similar issue with tests missing that data in #3727

packages/zowe-explorer-api/src/tree/ZoweTreeNode.ts
```diff
@@ -74,8 +75,11 @@ export class ZoweTreeNode extends vscode.TreeItem {
      *
      * @returns {imperative.IProfileLoaded}
      */
-    public getProfile(): imperative.IProfileLoaded {
-        return this.profile ?? this.getParent()?.getProfile();
+    public getProfile(profilesCache?: ProfilesCache): imperative.IProfileLoaded {
```
traeok, line 78: I noticed that the function definition was updated for this class but the interface is not. Do we want to allow extenders to pass this `profilesCache` to `getProfile` in their own tree node implementations?

packages/zowe-explorer/src/utils/AuthUtils.ts
```diff
@@ -61,7 +62,7 @@ export class AuthUtils {
                 await AuthHandler.waitForUnlock(profile);
             } else {
                 // Lock the profile and prompt the user for authentication by providing login/credential prompt options.
-                await AuthHandler.lockProfile(profile, authOpts);
+                await AuthHandler.lockProfile(profile, authOpts as any);
```
traeok: If `authOpts` should be of type `AuthPromptParams`, so the `as any` cast is not required here

traeok (commented):
Thanks for working on this Billie! I'm mostly requesting changes because of the `as any` cast in production code, would like to avoid that if necessary. Also left a couple comments to consider for future ZE versions/changes

packages/zowe-explorer-api/src/tree/ZoweTreeNode.ts
JillieBeanSim: @traeok In the Zowe Explorer trees the same cache is used for all registered with the trees.  for extenders using the APIs to initialize their own trees, they would need to adopt the changes in the ZE API and pass their profiles cache to the request.

traeok: Thanks for the clarification. You're right that for the trees registered directly within Zowe Explorer, the same cache is used. My concern is more about how this works for extenders and the potential for data inconsistency when they consume our APIs. I agree with your changes to workaround the issue, I'm just suggesting that the root cause is more involved.

I believe the core issue is that the `ProfilesCache` is not a singleton - so it does not enforce that the same instance is used by extenders. If we are noticing a desync that's resolved by providing our Profiles instance (rather than what's exposed to extenders), this suggests that the way the profiles cache is currently handled or exposed in Zowe Explorer is causing desync. If all extenders and Zowe Explorer used the same profiles cache, there wouldn't be a desync because we'd always be using the latest profiles.

Happy to discuss more in a GH discussion or offline - I'm not requesting for changes here, just wondering if we should reopen & revisit the issue I linked. I believe a more robust, long-term (v4) solution would be to refactor ProfilesCache into a singleton and expose the singleton instance through the Zowe Explorer API register, guaranteeing that Zowe Explorer and all its extenders share a single, consistent cache.

JillieBeanSim: that discussion can be for a v4 change but I don't think we can accomplish that kind of change in a non breaking way but open for discussion. This solution is the least invasive way to make sure that the latest profile info is being pulled since the nodes themselves were not getting updated, especially at the file level if the file was previously opened in the editor

traeok: I agree that it would need to be done as a breaking change in v4 (see initial comment).

You may want to see if #3735 resolves some of the behavior you're seeing. I found another getter for the profiles cache in ZE API where we keep making new ProfilesCache instances. I could see this contributing to the desync because we use it in multiple places in the API, and its not the same instance as the one we expose from Zowe Explorer.

packages/zowe-explorer/src/utils/AuthUtils.ts
JillieBeanSim: the previous code that you were the author of is trying to pass an Error as an ImperativeError causing the type to not match, even using `as unknown as AuthPromptParams` caused build error. not sure why the build error didn't show until I made a change to the file 🤔 
If I set the type in the assignment I get the error, but will look into fixing it.
Screenshot 2025-06-27 at 11 31 54 AM

traeok: Maybe there's something strange going on with how TypeScript is interpreting the types - as you can see on line 36 in the function we do an `instanceof` check to make sure that `err` is an ImperativeError, then we pass it into `authOpts`. I don't have the build error on my end, not sure what's going on here... 😕 

Since the `instanceof` check has to pass before we build these options, try casting `err as imperative.ImperativeError`, that should have the same effect and is considered a valid cast.

traeok: What's strange is, my IntelliSense shows `err` is the right type for me:

image

And it compiles even if I add `: AuthPromptParams` after `const authOpts` (no errors during webpack):

image

packages/zowe-explorer/__tests__/__unit__/trees/uss/USSTree.unit.test.ts
JillieBeanSim: addressed but a bit diff since test looking for particular profile name

packages/zowe-explorer/src/utils/AuthUtils.ts
JillieBeanSim: hope what i pushed is a good solution, tested and all still works as expected

packages/zowe-explorer/__tests__/__unit__/trees/job/JobActions.unit.test.ts
JillieBeanSim: clearing mocks doesn't change anything here

packages/zowe-explorer/src/utils/AuthUtils.ts
JillieBeanSim: there are 2 checks now in the check other than instance check, an existing and the one I added
err.message.includes("All configured authentication methods failed"))) ||
            err.message.includes("HTTP(S) status 401")

traeok: Ah I see the problem. We should do an instanceof check to make sure it's an Imperative error (from the RestClient) and then do the message comparison. But since the error is expected to be an ImperativeError for the auth prompts, I think the initial authOpts object is valid.

packages/zowe-explorer/src/utils/AuthUtils.ts
```diff
@@ -33,10 +42,11 @@ export class AuthUtils {
...
         if (
-            err instanceof imperative.ImperativeError &&
-            profile != null &&
-            (Number(err.errorCode) === imperative.RestConstants.HTTP_STATUS_401 ||
-                err.message.includes("All configured authentication methods failed"))
+            (err instanceof imperative.ImperativeError &&
+                profile != null &&
+                (Number(err.errorCode) === imperative.RestConstants.HTTP_STATUS_401 ||
+                    err.message.includes("All configured authentication methods failed"))) ||
+            err.message.includes("HTTP(S) status 401")
```
traeok, line 49: Let's include the new message check beside the other one so the `instanceof` check is still satisfied:

```suggestion
            (err instanceof imperative.ImperativeError &&
                profile != null &&
                (Number(err.errorCode) === imperative.RestConstants.HTTP_STATUS_401 ||
                    err.message.includes("All configured authentication methods failed") || err.message.includes("HTTP(S) status 401"))
            )
```

packages/zowe-explorer/src/utils/AuthUtils.ts
```diff
@@ -50,9 +60,9 @@ export class AuthUtils {
                 },
             });
 
-            const authOpts = {
+            const authOpts: AuthPromptParams = {
                 authMethods: Constants.PROFILES_CACHE,
-                imperativeError: err,
+                imperativeError: err as unknown as imperative.ImperativeError,
```
traeok, line 65: With the suggestion above, this is no longer necessary. Also, casting `err` as `unknown as imperative.ImperativeError` does get rid of the error, but its just a more explicit way of doing `as any` because it discards the type of the original parameter

traeok:
```suggestion
                imperativeError: err,
```

packages/zowe-explorer/src/utils/AuthUtils.ts
JillieBeanSim: @traeok the requirement of the error to be an imperative error as of 7 months ago when change was implemented would be considered a breaking change if extenders are returning errors straight from the rest api right?

traeok: Sorry if I've inspired any confusion, we have been casting all incoming errors as an `imperativeError` since v3.0.0 just based on the presence of `mDetails`, see here: https://github.com/zowe/zowe-explorer-vscode/blob/72bbc17b4fbc27b53f605613998f2bc7e07120e4/packages/zowe-explorer/src/utils/AuthUtils.ts#L30-L31

The way this was being cast is not a safe assumption because it assumes the error is an ImperativeError just based on the presence of that property. But this also confirms that our logic has checked for Imperative errors before handling authentication, at least since v3 was released and maybe even back in v2 (would need to double-check).

> if extenders are returning errors straight from the rest api right?

Errors from the RestClient are Imperative errors, in addition the logic in `AuthUtils.errorHandling` had a similar check to see if it was an Imperative error before handling authentication:

https://github.com/zowe/zowe-explorer-vscode/blob/72bbc17b4fbc27b53f605613998f2bc7e07120e4/packages/zowe-explorer/src/utils/AuthUtils.ts#L29

packages/zowe-explorer/src/utils/AuthUtils.ts
traeok: Please disregard, we've been incorrectly casting errors as imperative errors since v3 release, so might as well keep the cast in place to avoid breaking changes...

packages/zowe-explorer/__tests__/__unit__/trees/job/JobActions.unit.test.ts
traeok: You had mentioned in a previous comment that the `loadNamedProfile` is called more now. That explains the increase in calls. Thanks for clarifying.

packages/zowe-explorer/src/utils/AuthUtils.ts
traeok: I updated in the comment below, you could consider it a bug that we've always been casting errors as imperative errors (since v3 release), doing an instanceof check is more proper... but it depends on whether you want to consider that a breaking change or just doing a proper check (bug fix)

Either way, feel free to keep the solution as is, thanks for digging into the checks/workarounds

traeok (approved):

zFernand0 (approved):
LGTM! 😋 
---
Apologies for the delay on this PR 🙏

Issue #3653, opened 2025-05-15
Drag and drop may cause undesired data conversion


zFernand0:
**Describe the bug**

When moving a binary data (from a PDS member, a PDS that contains a binary member, or a Sequential data), the contents get converted to the EBCDIC.

**To Reproduce**

1. Create a PDS member and/or a sequential data set (with binary data)
2. Try to drag & drop it to another LPAR.
3. Notice the binary data was converted to EBCDIC

**Expected behavior**

To preserve the data without altering its encoding

**Notes**

- The same may happen with allocation details of the data set being transferred.
- VSAM data sets get transferred as empty sequential data sets.

PR #3687, opened 2025-06-05
Fixing binary dataset transfer via drag and drop across LPARs

ATorrise:
## Proposed changes

1.Destination PDS now posses the original DCB attributes (RECFM, LRECL, BLKSIZE, etc.) as the source PDS. 
2.Data is now transferred as binary to prevent unnecessary/unwanted conversion to text which would corrupt the data.
3.Encoding is something the user has to decide when opening their file to be read on the end-system. Zowe does not convert encodings during transfer.

## How to Test
1. Create a binary PDS, add however many members you want. Change the encoding via ZVSCE to binary.
2. Drag and drop that PDS to a different LPAR.
3. Ensure that you can view all members of that PDS and that no data was lost. Use a Hex Editor to ensure file contents are preserved. 

## Release Notes

Milestone:

Changelog:
Fixed: Cross-LPAR PDS transfers now preserve original DCB attributes and prevent binary data corruption.

## Types of changes

## Checklist

**General**

- [X] I have read the [CONTRIBUTOR GUIDANCE](https://github.com/zowe/zowe-explorer-vscode/wiki/Best-Practices:-Contributor-Guidance) wiki

**Code coverage**

- [X] I have manually tested the changes

**Deployment**


packages/zowe-explorer-api/src/fs/types/datasets.ts
```diff
@@ -33,7 +34,7 @@ export const DS_EXTENSION_MAP: Map<string, (string | RegExp)[]> = new Map([
 
 export class DsEntry extends FileEntry implements DsEntryProps {
     public metadata: DsEntryMetadata;
-
+    public encoding?: ZosEncoding;
```
traeok: This `encoding` property is already defined in the `FileEntry` class, I think we should avoid re-defining it here.

packages/zowe-explorer-api/src/tree/IZoweTreeNode.ts
t1m0thyj, line 1: Modifying property types of interfaces in the `zowe-explorer-api` package constitutes a breaking change. And this `encoding` property is deprecated so we should avoid using it if possible.

I think we'll need to keep encoding defined as a string here, and either:
1. Call the `node.getEncoding()` method (which returns type `ZosEncoding`) instead of `node.encoding`
2. Or if that doesn't work because `node.getEncoding()` references the `DatasetFSProvider` and the ds path/URI has changed after a Drag-n-Drop operation, we should find another workaround that doesn't require accessing a deprecated property.

t1m0thyj (dismissed):
Thanks for looking into this 🙂  Left a comment regarding use of deprecated stuff

packages/zowe-explorer/src/trees/dataset/DatasetFSProvider.ts
```diff
@@ -532,6 +538,11 @@ export class DatasetFSProvider extends BaseProvider implements vscode.FileSystem
             throw vscode.FileSystemError.FileNotFound(vscode.l10n.t("Profile does not exist for this file."));
         }
 
+        // If an encoding is provided, set it on the entry (only for DsEntry)
+        if (ds && options?.encoding && !("entries" in ds)) {
+            (ds as DsEntry).encoding = options.encoding;
+        }
```
traeok, line 543: - Typically, we only set the encoding on the entry through the "Open with Encoding" menu. Once the user gives us a new encoding through this menu, we set it directly on the filesystem entry using the `setEncoding` function on the node. I would avoid updating the entry's encoding based on any parameters passed to `readFile` as it could introduce a desync: for example, if we (or an extender) calls this with a different encoding, it would update the entry's encoding and wipe out the user's previously-selected encoding.
- I think we should avoid adding an `options` parameter to this function as extenders won't be able to pass it - they use our file systems through the `vscode.workspace.fs` namespace. The `readFile` function in this namespace does not have this `options` object, so it can't be passed to our provider. In this case, only we could call it (internally) with this option.
  - If we want to support passing an encoding as part of the request, we could add support for a query parameter "encoding" that would respect that encoding instead. But I'm not sure its necessary for this bugfix.

packages/zowe-explorer/src/trees/dataset/DatasetFSProvider.ts
```diff
@@ -647,6 +668,11 @@ export class DatasetFSProvider extends BaseProvider implements vscode.FileSystem
                     return;
                 }
 
+                // If an encoding is provided, set it on the entry
+                if (options.encoding) {
+                    entry.encoding = options.encoding;
+                }
```
traeok, line 672: Related: please check out my comment above about how setting encoding within the file system could surface some other issues

packages/zowe-explorer/src/trees/dataset/DatasetTree.ts
```diff
@@ -154,13 +163,20 @@ export class DatasetTree extends ZoweTreeProvider<IZoweDatasetTreeNode> implemen
                 if (entry == null) {
                     if (sourceNode.contextValue === Constants.DS_MEMBER_CONTEXT) {
                         dsname = destUri.path.match(/^\/[^/]+\/(.*?)\/[^/]+$/)[1] + "(" + (sourceNode.getLabel() as string) + ")";
-                        await ZoweExplorerApiRegister.getMvsApi(destinationInfo.profile).createDataSetMember(dsname, {});
+                        // Detect binary and set allocation attributes
+                        const allocAttrs = SharedContext.isBinary(sourceNode)
+                            ? { recordFormat: "U", lrecl: 0, blksize: 32760 }
+                            : {};
```
traeok: I could be missing something here, but can we detect if the data set is binary by checking if `entry.encoding` is equal to `{ kind: "binary" }`?

traeok (changes requested):
Thanks for working on this Amber! I have a few requests for changes and figured I should share them before I start manually testing.

packages/zowe-explorer/src/trees/dataset/DatasetTree.ts
```diff
@@ -190,8 +227,8 @@
                 return;
             }
 
-            if (!recursiveCall) {
-                // Delete any files from the selection on the source LPAR
+            if (!recursiveCall && writeSucceeded) {
```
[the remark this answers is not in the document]
zFernand0: This is a good comment. If we don't return inside the `catch` block, then the `writeSucceeded` will be necessary.

packages/zowe-explorer-api/src/fs/types/datasets.ts
zFernand0: That's a good catch. :yum:
Thank you! :pray:

packages/zowe-explorer-api/src/tree/IZoweTreeNode.ts
zFernand0: I agree that modifying the type of the encoding property may be a breaking change.
Also, we should avoid using a deprecated property.

Thanks for catching these! :pray:

zFernand0 (commented):
Partially addressing review comments

packages/zowe-explorer-api/src/tree/IZoweTreeNode.ts
zFernand0: Update:

Trae has a good point about our intentions of calling the `readFile` with parameters that only Zowe Explorer may be able to provide internally.

We will look for other ways to address the drag+drop issue(s).

Resolving this comment for now! 😋

packages/zowe-explorer-api/src/fs/types/datasets.ts
```diff
@@ -12,6 +12,7 @@
 import type { Types } from "../..";
 import { DirEntry, EntryMetadata, FileEntry } from "./abstract";
 import { IProfileLoaded } from "@zowe/imperative";
+import { ZosEncoding } from "../../tree/IZoweTreeNode";
```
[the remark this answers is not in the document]
zFernand0: Makes sense :yum:

packages/zowe-explorer/src/trees/dataset/DatasetTree.ts
zFernand0: That's a good point...
We could simplify this with:
```
const binary = (await sourceNode.getEncoding())?.kind === "binary";
```

ATorrise:
> [#3653 (comment)](https://github.com/zowe/zowe-explorer-vscode/issues/3653#issuecomment-2996280236) Comment explaining initial misunderstanding when developing a solution for this bug.
> 
> **TLDR:** You cannot safely mix binary and text (EBCDIC) files in a PDS because the system applies the same DCB to all members. Even with RECFM=U, binary content can get corrupted due to padding or misinterpretation.
> 
> @zFernand0 @t1m0thyj @traeok @adam-wolfe @awharn
> ...
> My question now is, how to proceed?
> ...

Through this greater understanding ive changed course and found that the destination PDS is not being allocated with the same DCB attributes as the source + binary flag isn't always being correctly set for member transfers.

Still unsure about whether or not I should make changes to UI, disabling "open with binary" unless within a PDS of recfm U and lrecl 0. Allowing this possibility might make users think our tool is broken because we are actually more constrained than we are letting on

ATorrise:
https://github.com/zowe/zowe-explorer-vscode/pull/3731 new PR

PR #3696, opened 2025-06-09
feat: Add support for AuthOrder :yum:

zFernand0:
## Proposed changes

Add support for `authOrder` in all authentication-related operations:
- Login
- Logout
- Switch (basic -> token)
- Switch (token -> basic)
- Tooltip synchronization
- Zowe CLI (or manual) edits to authOrder

## How to test

First, create a config file that can connect to the APIML

- Login: Notice a new property (authOrder) in your service profile
  - Check the tooltip (it should say `Token-Authentication`)
- Switch from Token to Basic: Notice the same authOrder property being modified
  - Check the tooltip (it should say `Basic-Authentication`)
- Switch from Basic to Token: Notice the same authOrder property being modified
  - Check the tooltip (it should say `Token-Authentication`)
- Logout: Notice the authOrder property stays the same
  - The tooltip should say `Unknown` for the authentication method.
  - You should be able to either add credentials or login again 😋 
  
Also, feel free to use the CLI in between any steps above. And feel free to change the order in which you perform the operations above. 😋 

If you see something, say something! 😉 

## Release Notes

Milestone: 3.3.0

Changelog: Allows users to specify the desired authentication order for each service profile.

## Types of changes

## Checklist

**General**

**Code coverage**

**Deployment**

### DIFF
--- a/packages/zowe-explorer/src/utils/AuthUtils.ts
+++ b/packages/zowe-explorer/src/utils/AuthUtils.ts
@@ -25,6 +25,8 @@ import {
 import { Constants } from "../configuration/Constants";
 import { ZoweLogger } from "../tools/ZoweLogger";
 import { SharedTreeProviders } from "../trees/shared/SharedTreeProviders";
+import { SettingsConfig } from "../configuration/SettingsConfig";
+import { SharedContext } from "../trees/shared/SharedContext";
 
 interface ErrorContext {
     apiType?: ZoweExplorerApiType;
@@ -392,7 +394,12 @@ export class AuthUtils {
         }
         if (nodeToRefresh) {
             nodeToRefresh.dirty = true;
-            void nodeToRefresh.getChildren().then(() => SharedTreeProviders.getProviderForNode(nodeToRefresh).refreshElement(nodeToRefresh));
+            const shouldPaginate =
+                SharedContext.isDatasetNode(nodeToRefresh) &&
+                SettingsConfig.getDirectValue<number>(Constants.SETTINGS_DATASETS_PER_PAGE, Constants.DEFAULT_ITEMS_PER_PAGE) > 0;
+            void nodeToRefresh
+                .getChildren(shouldPaginate)
+                .then(() => SharedTreeProviders.getProviderForNode(nodeToRefresh).refreshElement(nodeToRefresh));
         }
     }
 
--- a/packages/zowe-explorer-api/src/tree/IZoweTreeNode.ts
+++ b/packages/zowe-explorer-api/src/tree/IZoweTreeNode.ts
@@ -187,6 +187,11 @@ export interface IZoweDatasetTreeNode extends IZoweTreeNode {
     children?: IZoweDatasetTreeNode[];
 
     /**
+     * Whether the node is currently being updated through a filter prompt
+     */
+    inFilterPrompt?: boolean;
+
+    /**
      * Retrieves child nodes of this IZoweDatasetTreeNode
      *
      * @returns {Promise<IZoweDatasetTreeNode[]>}
--- a/packages/zowe-explorer/CHANGELOG.md
+++ b/packages/zowe-explorer/CHANGELOG.md
@@ -41,6 +41,7 @@ All notable changes to the "vscode-extension-for-zowe" extension will be documen
 - Prevented data loss by modifying the save process to handle data sets with records that exceeded the logical record length. [#3791](https://github.com/zowe/zowe-explorer-vscode/issues/3791)
 - Fixed cross-LPAR PDS transfers to preserve original DCB attributes and prevent binary data corruption. Data is now transferred as binary, and the encoding is respected if previously set by the user. [#3731](https://github.com/zowe/zowe-explorer-vscode/pull/3731)
 - Fixed copy/paste functionality of data sets for extender-type profiles by using `SharedContext` helper functions. [#3815](https://github.com/zowe/zowe-explorer-vscode/pull/3815)
+- Fixed issue where the `getChildren` function made requests repeatedly while the data set filter prompt was open, causing repeated authentication prompts and excessive requests for listing data sets. [#3807](https://github.com/zowe/zowe-explorer-vscode/issues/3807)
 
 ## `3.2.2`
 
--- a/packages/zowe-explorer/__tests__/__unit__/trees/dataset/DatasetTree.unit.test.ts
+++ b/packages/zowe-explorer/__tests__/__unit__/trees/dataset/DatasetTree.unit.test.ts
@@ -1900,66 +1900,6 @@ describe("Dataset Tree Unit Tests - Function datasetFilterPrompt", () => {
 
         expect(mocked(Gui.showMessage)).toHaveBeenCalledWith("No selection made. Operation cancelled.");
     });
-    it("Checking adding of new filter error is caught on getChildren", async () => {
-        const globalMocks = createGlobalMocks();
-        const blockMocks = createBlockMocks(globalMocks);
-
-        mocked(vscode.window.showQuickPick).mockResolvedValueOnce(new FilterDescriptor("\uFF0B " + "Create a new filter"));
-        mocked(vscode.window.showInputBox).mockResolvedValueOnce("HLQ.PROD1.STUFF");
-        mocked(vscode.window.createTreeView).mockReturnValueOnce(blockMocks.treeView);
-        const testTree = new DatasetTree();
-        testTree.mSessionNodes.push(blockMocks.datasetSessionNode);
-        Object.defineProperty(testTree.mSessionNodes[1], "getChildren", {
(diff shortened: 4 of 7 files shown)


packages/zowe-explorer/src/utils/AuthUtils.ts
```diff
@@ -19,11 +19,13 @@ import {
     ErrorCorrelator,
     ZoweExplorerApiType,
     AuthHandler,
+    ZoweExplorerZosmf,
     AuthPromptParams,
 } from "@zowe/zowe-explorer-api";
 import { Constants } from "../configuration/Constants";
 import { ZoweLogger } from "../tools/ZoweLogger";
 import { SharedTreeProviders } from "../trees/shared/SharedTreeProviders";
+// import { ZoweExplorerApiRegister } from "../extending/ZoweExplorerApiRegister";
```
t1m0thyj: Should this import be removed?

zFernand0: resolved as part of 95a5911

packages/zowe-explorer-api/src/profiles/AuthHandler.ts
```diff
@@ -103,6 +103,7 @@ export class AuthHandler {
 
     /**
      * Function that checks whether a profile is using token based authentication
+     * @deprecated Use AuthUtils.sessTypeFromProfile and/or AuthUtils.sessTypeFromSession, which will adhere to authOrder.
```
traeok: The extender-facing function was deprecated, but the suggested functions cannot be used as `AuthUtils` is in ZE and not ZE API 🤔

Is there a way that we can provide `sessTypeFromProfile` or `sessTypeFromSession` in ZE API?

packages/zowe-explorer/src/tools/ZowePersistentFilters.ts
```diff
@@ -177,11 +177,9 @@ export class ZowePersistentFilters {
      */
     public addSortSetting(node: IZoweDatasetTreeNode, setting: Sorting.NodeSort): void {
         ZoweLogger.trace("PersistentFilters.addSortSettings called.");
-        if (node && setting) {
```
traeok, line 180: I know the related code was added separately but thanks for addressing this 🙏

packages/zowe-explorer/src/management/ProfileManagement.ts
```diff
@@ -40,24 +40,17 @@ export class ProfileManagement {
...
-                break;
-            }
+        if (sessTypeFromProf === imperative.SessConstants.AUTH_TYPE_BASIC) {
+            ZoweLogger.debug(`Profile ${profile.name} is using basic authentication.`);
+            selected = await this.setupProfileManagementQp(imperative.SessConstants.AUTH_TYPE_BASIC, node);
+        } else if (sessTypeFromProf === imperative.SessConstants.AUTH_TYPE_TOKEN || sessTypeFromProf === imperative.SessConstants.AUTH_TYPE_BEARER) {
+            ZoweLogger.debug(`Profile ${profile.name} is using token authentication.`);
+            selected = await this.setupProfileManagementQp(imperative.SessConstants.AUTH_TYPE_TOKEN, node);
+        } else {
+            ZoweLogger.debug(`Profile ${profile.name} authentication method is unkown.`);
```
traeok: (nitpick) 😋 
```suggestion
            ZoweLogger.debug(`Profile ${profile.name} authentication method is unknown.`);
```

packages/zowe-explorer/src/configuration/Profiles.ts
```diff
@@ -97,37 +98,67 @@ export class Profiles extends ProfilesCache {
...
+     * @returns True if the profile has a secure token, false otherwise.
+     */
+    private async profileHasSecureToken(theProfile: imperative.IProfileLoaded): Promise<boolean> {
+        const teamConfig = (await this.getProfileInfo()).getTeamConfig();
+        const profName = teamConfig.api.profiles.getProfilePathFromName(theProfile.name);
+        const tokenValue = profName + ".properties.tokenValue";
+        return teamConfig.api.secure.secureFields().includes(tokenValue);
+    }
+
     public async checkCurrentProfile(theProfile: imperative.IProfileLoaded, node?: Types.IZoweNodeType): Promise<Validation.IValidationProfile> {
```
traeok, line 114: It seems like `node` is now passed to this function in various places. Was this change required to keep the same functionality as before? Or are we performing additional logic now that requires the node for full effect?

traeok (changes requested):
Thanks for all your work on this Fernando! I have a question for clarification and a request regarding a deprecated API, but the changes make sense to me.

packages/zowe-explorer/src/configuration/Profiles.ts
traeok: As discussed offline we need to pass `node` to this function to ensure that the node tooltip is kept up to date as the user leverages different features. Thanks for the discussion!

packages/zowe-explorer-api/src/profiles/AuthHandler.ts
zFernand0: That is an extremely good observation. 🙏 
I moved the new functions here since no circular dependencies were found. 🙏 

resolved in 98d0240086c622506a1ebe06b89ec36596437ba9

traeok: Thanks for the quick resolution! I like how we could move it without any overhead, it's nice to see how modular auth order can be 😋

packages/zowe-explorer-api/src/profiles/AuthHandler.ts
```diff
@@ -101,9 +102,45 @@ export class AuthHandler {
...
+     * Function that returns the session associated with the specified profile.
+     *
+     * @param {imperative.IProfileLoaded} profile The profile to be inspected.
+     *
+     * @returns {imperative.Session}
+     *      The session associated with the specified profile
+     */
+    public static getSessFromProfile(profile: imperative.IProfileLoaded): imperative.Session {
+        return new ZoweExplorerZosmf.CommonApi(profile).getSession();
+    }
```
traeok: This would make a new instance of `CommonApi` every time this function is called, right? Then shortly after, it is destroyed by the garbage collector after every call.. 🤔 

Curious to see the performance implications but I can understand the reason for the change. Could we access the API instances through the API register instead? that might help eliminate creating new class instances 😋

zFernand0: resolved in 606005c

packages/zowe-explorer-api/src/profiles/AuthHandler.ts
```diff
@@ -111,7 +111,7 @@ export class AuthHandler {
      *      The session associated with the specified profile
      */
     public static getSessFromProfile(profile: imperative.IProfileLoaded): imperative.Session {
-        return new ZoweExplorerZosmf.CommonApi(profile).getSession();
+        return ZoweVsCodeExtension.getZoweExplorerApi().getCommonApi(profile).getSession(profile);
```
traeok, line 116: If the profile has no available APIs, then the `getCommonApi` function throws an error:

https://github.com/zowe/zowe-explorer-vscode/blob/1335f8501ee23087d703322862f0357b578472fa/packages/zowe-explorer/src/extending/ZoweExplorerApiRegister.ts#L340-L346

Should we add a `@throws` to the TSDoc so extenders are aware of the implications of passing a profile that does not have any APIs available for it? Its unlikely that an extender would pass a profile whose type doesn't have any registered APIs, but could help identify issues with type registration if the error is documented.

This function is also called in `sessTypeFromProfile` so it may be worth adding a `@throws` there too 😅

traeok (approved):
Thanks Fernando! I had a question about adding `@throws` to the TSDoc for these functions, but going to approve since there are other places in the code where we aren't using that doc keyword.

packages/zowe-explorer-api/src/profiles/AuthHandler.ts
zFernand0: I didn't think of that 😅 

Thanks for mentioning it! 🙏 

Resolved in 67b50f665 🥳

JillieBeanSim (changes requested):
thanks for working on this @zFernand0 I do see issues with the change auth type.  it worked well the first time changing from basic to token but when I attempted to change back the request didn't work or prompt for creds. the second attempt to change back to basic prompted for creds and updated the config file but the following action to enter a search filter prompted me for credentials. after putting in credentials again when prompted the tree sessions showed action in the background like filter search initiated then I was prompted for credentials again

looking at debug console I see this error 
`rejected promise not handled within 1 second: Error: Zowe Explorer Profiles Cache error: Could not find profile named: Users.`

JillieBeanSim (approved):
I do experience the same on main, will test marketplace as well. approving and will create separate issue

Issue #3726, opened 2025-06-24
Implement "Retry request" functionality


traeok:
**Is your feature request related to a problem? Please describe.**

Currently, we stop making any requests to the mainframe in the event of an authentication error, waiting for the user to address authentication before the profile is unlocked. However, if an extender is making a request to the filesystem and the user has already updated credentials on their profile, the request still fails and the extender has to retry the request.

**Describe the solution you'd like**

We should offer a feature "Retry requests" that allows the user to configure the max amount of retry requests that they want to use for their environment. We should make this setting **opt-in** and ensure the default for this setting is a low amount (e.g. 3 or less) to prevent introducing lockouts.

**Describe alternatives you've considered**

The extender can retry the request again, but its not clear whether the user had acknowledged the credential prompt, making the retry more risky.

PR #3727, opened 2025-06-24
fix(FSP): Throw explicit `AuthCancelledError` if auth. prompt cancelled

traeok:
## Proposed changes

Part 1 to resolving #3662. To completely satisfy the request, #3726 needs to be implemented (opt-in feature) to allow automatic retries in filesystem by Zowe Explorer with a set threshold.

With this change, the FSPs now throw an `AuthCancelledError`, allowing extenders to handle these error scenarios with additional context:
- User cancels SSO login
- User cancels credential input
- User dismisses authentication prompt entirely

If the user cancels an authentication prompt, they have two ways to resolve it:
- Right-click on the profile, select "Manage Profile" -> "Update Credentials" / "Log in..."
- Performing an action in the filesystem for a profile that's still locked and not yet re-authenticated results in a prompt before the request occurs. This avoids locking out the user, only allowing the filesystem request to continue once they've confirmed/re-entered their credentials or completed their auth service login.

## Release Notes

Milestone: 3.2.2

Changelog:

## Types of changes

## Checklist

**General**

**Code coverage**

**Deployment**

### DIFF
--- a/packages/zowe-explorer/CHANGELOG.md
+++ b/packages/zowe-explorer/CHANGELOG.md
@@ -25,6 +25,7 @@ All notable changes to the "vscode-extension-for-zowe" extension will be documen
 - Fixed an issue where clicking the refresh icon beside a profile in the Unix System Services (USS) view had no effect. [#3693](https://github.com/zowe/zowe-explorer-vscode/issues/3693)
 - Fixed an issue where clicking the Cancel button in the Save Credentials dialog triggered a 401 error prompting to update credentials. [#3713](https://github.com/zowe/zowe-explorer-vscode/pull/3713)
 - Resolved an issue where secure credentials were inadvertently converted to non-string data types. [#3728](https://github.com/zowe/zowe-explorer-vscode/issues/3728)
+- Changed the jobs pagination setting to be disabled by default and the default records to fetch from 100 to 1000 due to customer feedback. [#3743](https://github.com/zowe/zowe-explorer-vscode/issues/3743)
 - Fixed an issue where a `401 Unauthorized` error is thrown by the FileSystemProvider implementations when the user cancels or closes an authentication prompt. Now, the filesystem throws a more explicit `AuthCancelledError` in this scenario. [#3662](https://github.com/zowe/zowe-explorer-vscode/issues/3662)
 
 ## `3.2.1`
--- a/packages/zowe-explorer/package.json
+++ b/packages/zowe-explorer/package.json
@@ -2018,13 +2018,13 @@
         },
         "zowe.jobs.paginate.enabled": {
           "type": "boolean",
-          "default": true,
+          "default": false,
           "description": "%zowe.jobs.paginate.enabled%",
           "scope": "window"
         },
         "zowe.jobs.paginate.recordsToFetch": {
           "type": "integer",
-          "default": 100,
+          "default": 1000,
           "description": "%zowe.jobs.paginate.recordsToFetch%",
           "scope": "window",
           "minimum": 2
--- a/packages/zowe-explorer/src/webviews/package.json
+++ b/packages/zowe-explorer/src/webviews/package.json
@@ -25,7 +25,7 @@
     "@vscode/webview-ui-toolkit": "^1.2.2",
     "ag-grid-community": "^32.0.2",
     "ag-grid-react": "^32.0.2",
-    "dompurify": "^3.2.5",
+    "dompurify": "^3.2.6",
     "es-toolkit": "^1.16.0",
     "marked": "^15.0.11",
     "preact": "^10.16.0",
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -243,7 +243,7 @@ importers:
         version: 0.2.3
       ts-node:
         specifier: ^10.9.2
-        version: 10.9.2(@types/node@24.0.4)(typescript@5.8.3)
+        version: 10.9.2(@types/node@20.17.6)(typescript@5.8.3)
       wdio-vscode-service:
         specifier: ^6.1.3
         version: 6.1.3(webdriverio@8.43.0)
@@ -352,8 +352,8 @@ importers:
         specifier: ^32.0.2
         version: 32.0.2(react-dom@18.3.1)(react@18.3.1)
       dompurify:
-        specifier: ^3.2.5
-        version: 3.2.5
+        specifier: ^3.2.6
+        version: 3.2.6
       es-toolkit:
         specifier: ^1.16.0
         version: 1.16.0
@@ -2874,12 +2874,6 @@ packages:
     dev: true
     optional: true
 
-  /@types/node@24.0.4:
-    resolution: {integrity: sha512-ulyqAkrhnuNq9pB76DRBTkcS6YsmDALy6Ua63V8OhrOBgbcYt6IOdzpw5P1+dyRIyMerzLkeYWBeOXPpA9GMAA==}
-    dependencies:
-      undici-types: 7.8.0
-    dev: true
-
   /@types/normalize-package-data@2.4.4:
     resolution: {integrity: sha512-37i+OaWTh9qeK4LSHPsyRC7NahnGotNuZvjLSgcPzblpHB3rrCJxAOgI5gCdKm7coonsaX1Of0ILiTcnZjbfxA==}
     dev: true
@@ -4372,13 +4366,6 @@ packages:
       dequal: 2.0.3
     dev: true
 
-  /aria-query@5.3.2:
-    resolution: {integrity: sha512-COROpnaoap1E2F000S62r6A60uHZnmlvomhfyT2DlTcrY1OrBKn2UhH7qn5wTC9zMvD0AY7csdPSNwKP+7WiQw==}
-    engines: {node: '>= 0.4'}
-    requiresBuild: true
-    dev: true
-    optional: true
-
   /array-buffer-byte-length@1.0.1:
     resolution: {integrity: sha512-ahC5W1xgou+KTXix4sAO8Ki12Q+jf4i0+tmk3sC+zgcynshkHxzpXdImBehiUYKKKDwvfFiJl1tZt6ewscS1Mg==}
     engines: {node: '>= 0.4'}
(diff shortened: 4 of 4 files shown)


packages/zowe-explorer-api/CHANGELOG.md
```diff
@@ -16,6 +16,8 @@ All notable changes to the "zowe-explorer-api" extension will be documented in t
 - Added support for asynchronous operations when using integrated terminals. [#3640](https://github.com/zowe/zowe-explorer-vscode/issues/3640)
 - Added support for TTY-dependent scripts when using integrated terminals. [#3640](https://github.com/zowe/zowe-explorer-vscode/issues/3640)
 - Implemented graceful handling for non-existent named profiles within the loadNamedProfile function. [#3678](https://github.com/zowe/zowe-explorer-vscode/pull/3678)
+- Added an option `throwErrorOnCancel` to `AuthPromptParams`. When this option is `true`, an error is thrown when the user cancels or dismisses an authentication prompt. [#3662](https://github.com/zowe/zowe-explorer-vscode/issues/3662)
```
anaxceron: Please add descriptions for both `throwErrorOnCancel` and `AuthPromptParams`.

packages/zowe-explorer/src/trees/dataset/DatasetFSProvider.ts
```diff
@@ -149,6 +150,8 @@ export class DatasetFSProvider extends BaseProvider implements vscode.FileSystem
         const profileEntry = this._lookupAsDirectory(uri, false) as FilterEntry;
 
         // Wait for any ongoing authentication process to complete
+        const profile = Profiles.getInstance().loadNamedProfile(uriInfo.profile.name);
+        await AuthUtils.reauthenticateIfCancelled(profile);
```
t1m0thyj: Why do we need to load the profile now several places across the FS Providers, instead of reusing `uriInfo.profile`?

traeok: There were a few instances where I could remove the call to `loadNamedProfile`, so I pushed up a commit with those changes. However there are multiple places where `uriInfo` isn't defined, so I've kept those calls as-is. `getInfoFromUri` does a call to `loadNamedProfile` to get the profile so it should have the same performance implications without doing the extra logic to get the other properties.

Another ProfilesCache pain point: Loading a named profile shouldn't be an expensive operation since each profile name is currently treated as unique - it should be constant-time access through something like a map. Currently the cache uses an array of profiles, which is expensive to iterate (worst case linear time for searching by name) especially when we could treat the name as a hash (at least until something like profile UUIDs exist). If loading a cached profile was constant-time access in the cache, then I think the implications of calling this function would be negligible and it would also largely benefit extenders that use it to access profiles.

packages/zowe-explorer-api/CHANGELOG.md
traeok: Got it - does this sound better?

```suggestion
- Added an option `throwErrorOnCancel` to the `AuthPromptParams` interface (defines parameters passed to authentication prompt functions). When this option is `true`, an error is thrown when the user cancels or dismisses an authentication prompt. [#3662](https://github.com/zowe/zowe-explorer-vscode/issues/3662)
```

anaxceron: Yes, thank you @traeok! 😸

anaxceron (approved):

t1m0thyj (approved):
LGTM, thanks @traeok!

anaxceron (approved):

JWaters02 (changes requested):
Hi Trae, I am just testing now, I notice that clicking cancel on the prompt immediately pops up the same prompt again. Clicking cancel again then does actually cancel. This is the same bug as in 3.2.1 and not clear if this PR is also attempting to fix that bug. 

On the topic of the new note added to the dialog, I am reading it and wondering if you could replace "profile" with the profile name or similar? Even though I understand these instructions, it is still a bit difficult to follow at first glance - if I were a typical mainframer end user, would I even know what a profile is (I did not until I started doing ZE development)? Also I think the "NOTE" part being in all caps seems a bit shouty/puts the note into a more confrontational tone which is not how it should read. So maybe the below could be something like: 

> Selecting "Cancel" or closing this dialog blocks further actions with the <profile name> profile. To unblock, select the Zowe Explorer icon, right-click the <profile name> profile, and select "Manage Profile" to re-authenticate.

image

traeok:
> Hi Trae, I am just testing now, I notice that clicking cancel on the prompt immediately pops up the same prompt again. Clicking cancel again then does actually cancel. This is the same bug as in 3.2.1 and not clear if this PR is also attempting to fix that bug.

Thanks for testing @JWaters02 - can you provide context on what operation you were performing that caused the prompt to appear twice? It might be specific to the action you're taking.

JWaters02:
> > Hi Trae, I am just testing now, I notice that clicking cancel on the prompt immediately pops up the same prompt again. Clicking cancel again then does actually cancel. This is the same bug as in 3.2.1 and not clear if this PR is also attempting to fix that bug.
> 
> Thanks for testing @JWaters02 - can you provide context on what operation you were performing that caused the prompt to appear twice? It might be specific to the action you're taking.

When I attempt to search on my profile, the prompt appears. I click cancel. The prompt appears again. I click cancel. The prompt no longer appears.

traeok:
Sorry for any troubles caused by the mainframe lockout @JWaters02 - with the latest commit, it should be good to go.

traeok:
> Btw what are your thoughts on this? I assume you either didn't see it or disagree based on not comitting any changes about it but if so wondered why?

Sorry Joshua, I missed that part of your message, was doing too much context switching. Going to address that shortly.

traeok:
> Hmm, I am still getting the cancel prompt popping up twice issue. I am definately on latest commit as I have the newest cancel note. 
> 
> 
> 
> Bit of extra info. So click search data sets, the quick pick comes up, immediately followed by the update creds prompt. When I click cancel it pops up once more. Once it is fully cancelled, I then press enter on the quick pick to search data sets, where the prompt then comes up again. This time when I cancel, the prompt comes up, cancel and it comes up a third time. I repro'd this reliably. So now it comes up twice when attempting to search and thrice when actually searching. 

Thanks for the additional context, looks like the data sets tree is still searching too early before the user submits the pattern. You shouldn't see the update creds prompt more than once and it should only show after the search is made. Will look into this further

zFernand0 (dismissed):
Changes make sense, but I'm having a hard time understanting how to test this PR.

image

Based on the error message, I expected the profile to be locked (meaning no new z/osmf request will go out unless I took action) 🤔 

I can get to the explicit AuthCancelledError
image

But I'm still getting locked out of the mainframe.
am I confusing something here? 🤔 

---

> Hmm, I am still getting the cancel prompt popping up twice issue. I am definately on latest commit as I have the newest cancel note.

Also, there is a second call to z/OS where the AuthCancelledError is not thrown when I cancel the prompt in the second prompt 🤔 

The second prompt happens when expanding the profile, which may trigger a `getChildren` twice with my previous search 😅 

---

Even though I'm a bit confused on this, I'm still approving the PR since it did achieve what was set out to do. Throw the AuthCancelledError if the user cancelled the (first) prompt.

traeok:
> Also, there is a second call to z/OS where the AuthCancelledError is not thrown when I cancell the prompt in the second request 🤔
> 
> Even though I'm a bit confused on this, I'm still approving the PR since it did achieve what was set out to do. Throw the AuthCancelledError if the user cancelled the (first) prompt.

The redundant prompt that you and Joshua are seeing is not related to the changes on this PR. It's a valid bug, yes, but the PR is meant to address how the cancel is handled within the filesystem.

I plan to look into the second prompt, if the fix is too complicated we should part it out into a separate PR so we can get this in for the patch release..

JWaters02:
So just to clarify, the cancel prompt re-appearing will be fixed in a seperate PR? 

Do we know if this PR causes the issue of cancelling but still getting locked out of the mainframe?

traeok:
> So just to clarify, the cancel prompt re-appearing will be fixed in a seperate PR?

That might be best as I haven't had the chance to look into it yet, I'm juggling a few different issues at the moment. This change was intended to be specific to the filesystem and is more for a developer, we now return an explicit `AuthCancelledError` message. If the user had just cancelled the last auth dialog and the developer takes another action in the filesystem, the user is prompted before a request is made. I tried to introduce some quick fixes to address the behavior you're seeing, but the root cause is turning out to be more complicated than just resetting a flag 😅 

> Do we know if this PR causes the issue of cancelling but still getting locked out of the mainframe?

I don't see how the changes would be introducing this problem as I'm not adding new requests to the mainframe with these changes. I do see the issue with the redundant prompt, but its specific to listing in the data sets tree, whereas this change was specific to prompts coming from the filesystem. This duplicate prompt is a valid bug, but it is because the data sets tree has an issue with duplicate refresh when applying a search pattern.

zFernand0:
I agree with a follow up PR to resolve the duplicate request when listing data sets.

Also, the mainframe lock-out still occurs, but not because of this PR 🙏 

That said, I believe this PR is good to re-review purely for the AuthCancelledError which is informative for extenders 😋

packages/zowe-explorer-api/CHANGELOG.md
```diff
@@ -16,6 +16,8 @@ All notable changes to the "zowe-explorer-api" extension will be documented in t
 - Added support for asynchronous operations when using integrated terminals. [#3640](https://github.com/zowe/zowe-explorer-vscode/issues/3640)
 - Added support for TTY-dependent scripts when using integrated terminals. [#3640](https://github.com/zowe/zowe-explorer-vscode/issues/3640)
 - Implemented graceful handling for non-existent named profiles within the loadNamedProfile function. [#3678](https://github.com/zowe/zowe-explorer-vscode/pull/3678)
+- Added an option `throwErrorOnCancel` to the `AuthPromptParams` interface (defines parameters passed to authentication prompt functions). When this option is `true`, an error is thrown when the user cancels or dismisses an authentication prompt. [#3662](https://github.com/zowe/zowe-explorer-vscode/issues/3662)
```
anaxceron, line 19: What does the parenthetical do? Is it explaining what the interface does?

anaxceron (commented):
Left a comment seeking some clarification on a changelog entry.

packages/zowe-explorer-api/CHANGELOG.md
traeok: It is some additional context to clarify what the interface is used for

zFernand0: The parenthetical provides a bit more information regarding about the `AuthPromptParams` interface 😋

PR #3731, opened 2025-06-26
Fixing binary dataset transfer via drag and drop across LPARs

ATorrise:
## Proposed changes

1.Destination PDS now posses the original DCB attributes (RECFM, LRECL, BLKSIZE, etc.) as the source PDS.
2.Data is now transferred as binary to prevent unnecessary/unwanted conversion to text which would corrupt the data.
3.Encoding is something the user has to decide when opening their file to be read on the end-system. Zowe does not convert encodings during transfer.

## How to Test
Create a binary PDS, add however many members you want. Change the encoding via ZVSCE to binary.
Drag and drop that PDS to a different LPAR.
Ensure that you can view all members of that PDS and that no data was lost. Use a Hex Editor to ensure file contents are preserved.

https://github.com/user-attachments/assets/27dda6ae-b488-4e3a-93e0-7897e215f98d

## Release Notes

Milestone: 3.3?

Changelog: Fixed cross-LPAR PDS transfers to preserve original DCB attributes and prevent binary data corruption. Data is now transferred as binary, and encoding is user-selected at file open time. The encoding query parameter is now respected during write operations. [#3731]

## Types of changes

## Checklist

**General**

**Code coverage**

- [X] I have manually tested the changes

**Deployment**


zFernand0 (commented):
Hey, 
Thanks for keeping the DCB the same when moving the PDS 🥳 

---

I might be doing something wrong, but I'm hitting some data-loss issues when moving a data set like this one:
image

Plus, I get quite a few error messages popping up 😢 
image

And finally, this is what it looks like in the target (and all members have no data in them) 😢 
image

packages/zowe-explorer/src/trees/dataset/DatasetFSProvider.ts
```diff
@@ -588,10 +588,18 @@ export class DatasetFSProvider extends BaseProvider implements vscode.FileSystem
...
             const profile = Profiles.getInstance().loadNamedProfile(entry.metadata.profile.name);
-            const profileEncoding = entry.encoding ? null : profile.profile?.encoding; // use profile encoding rather than metadata encoding
+            const profileEncoding = entry.encoding ? null : profile.profile?.encoding;
+
+            let binary = false;
+            if (encoding === "binary") {
+                binary = true;
+            } else if (entry.encoding?.kind === "binary") {
+                binary = true;
+            }
```
traeok: We should use `const` when possible, this is a good use case for it:

```suggestion
            const binary = encoding === "binary" || entry.encoding?.kind === "binary";
```

packages/zowe-explorer/src/trees/dataset/DatasetFSProvider.ts
```diff
@@ -664,7 +672,8 @@ export class DatasetFSProvider extends BaseProvider implements vscode.FileSystem
                 }
 
                 if (entry.wasAccessed || content.length > 0) {
-                    const resp = await this.uploadEntry(entry as DsEntry, content, forceUpload);
+                    // Pass encodingParam to uploadEntry
```
traeok: (nitpick) Can we remove this comment? 😋

packages/zowe-explorer/src/trees/dataset/DatasetFSProvider.ts
```diff
@@ -625,10 +633,10 @@ export class DatasetFSProvider extends BaseProvider implements vscode.FileSystem
             throw vscode.FileSystemError.FileExists(uri);
         }
 
-        // Attempt to write data to remote system, and handle any conflicts from e-tag mismatch
```
traeok, line 672: I don't think we should remove this comment because it provides context for the logic below

packages/zowe-explorer/src/trees/dataset/DatasetTree.ts
```diff
@@ -112,16 +112,26 @@ export class DatasetTree extends ZoweTreeProvider<IZoweDatasetTreeNode> implemen
...
-                // create a PDS on remote
+                // find source PDS attributes and make destination PDS have same attributes
+                const sourceAttributesResponse = await ZoweExplorerApiRegister.getMvsApi(sourceInfo.profile).dataSet(dsname, {
+                    attributes: true,
+                    responseTimeout: sourceInfo.profile?.profile?.responseTimeout,
+                });
+                const { dsname: dsnameSource, ...rest } = sourceAttributesResponse.apiResponse.items[0];
+
+                // create a PDS on remote with the same attributes as source PDS
+                const transformedAttrs = (zosfiles.Copy as any).dataSetCrossLPAR({}, rest);
```
traeok: `dataSetCrossLPAR` is public, so the cast `as any` is not needed and turns off the type check:

 
```suggestion
                const transformedAttrs = zosfiles.Copy.dataSetCrossLPAR({}, rest);
```

Note that you'll need to update the parameters once removing the `as any` cast, as there are 5 required parameters for the function and only 2 are provided here

traeok (changes requested):

packages/zowe-explorer/__tests__/__mocks__/@zowe/zos-files-for-zowe-sdk.ts
```diff
@@ -10,7 +10,7 @@
  */
 
 import * as imperative from "@zowe/imperative";
-import type { IListOptions } from "@zowe/zos-files-for-zowe-sdk";
+import type { IDataSet, IListOptions } from "@zowe/zos-files-for-zowe-sdk";
```
t1m0thyj: This added import seems unused

packages/zowe-explorer/src/trees/dataset/DatasetTree.ts
traeok: While this comment _is_ valid, I would advise not calling this function here because we already have logic that performs the cross-LPAR copy, so this would result in a duplicate copy attempt.

To clarify, I think `generateDatasetOptions` is the right function to use here - since its private, we'd have to either make the function public in the SDK, or keep the `zosfiles.Copy as any` cast in place. From there, we can pass those options to the `createDataSet` API like what was done in https://github.com/zowe/zowe-explorer-vscode/pull/3731/commits/89c2420e2a788a239e36cfc8965893a757f49df5.

packages/zowe-explorer/src/trees/dataset/DatasetTree.ts
```diff
@@ -112,16 +112,26 @@ export class DatasetTree extends ZoweTreeProvider<IZoweDatasetTreeNode> implemen
...
+                const isBinary = transformedAttrs.recfm === "U" && transformedAttrs.lrecl === 0;
                 try {
                     dsname = sourceNode.getLabel() as string;
                     await ZoweExplorerApiRegister.getMvsApi(destinationInfo.profile).createDataSet(
-                        zosfiles.CreateDataSetTypeEnum.DATA_SET_PARTITIONED,
+                        isBinary ? zosfiles.CreateDataSetTypeEnum.DATA_SET_BINARY : zosfiles.CreateDataSetTypeEnum.DATA_SET_PARTITIONED,
                         dsname,
-                        {}
+                        transformedAttrs
                     );
```
t1m0thyj: Agree with Trae's suggestion here - I think we want to keep `generateDatasetOptions`.

Also the `isBinary` check may be unnecessary. We could use a blank data set template if all the attributes are specified on the `transformedAttrs` object.

Something like this should work:
```suggestion
                // create a PDS on remote with the same attributes as source PDS
                const transformedAttrs = (zosfiles.Copy as any).generateDatasetOptions({}, rest);

                try {
                    dsname = sourceNode.getLabel() as string;
                    await ZoweExplorerApiRegister.getMvsApi(destinationInfo.profile).createDataSet(
                        zosfiles.CreateDataSetTypeEnum.DATA_SET_BLANK,
                        dsname,
                        transformedAttrs
                    );
```

packages/zowe-explorer/src/trees/ZoweTreeProvider.ts
```diff
@@ -412,6 +412,10 @@ export class ZoweTreeProvider<T extends IZoweTreeNode> {
             ? await Profiles.getInstance().fetchAllProfilesByType(profileType)
             : await Profiles.getInstance().fetchAllProfiles();
         for (const profile of profiles) {
+            // Skip profiles without a name or data (invalid profiles)
+            if (!profile || !profile.name) {
+                continue;
```
traeok: Is this change necessary for this fix? Also, if we're seeing `undefined` profiles or profiles without a name, we should address this as a separate issue as we shouldn't be seeing nullish profiles being returned by `fetchAllProfiles` or `fetchAllProfilesByType`

packages/zowe-explorer/CHANGELOG.md
```diff
@@ -31,7 +32,9 @@ All notable changes to the "vscode-extension-for-zowe" extension will be documen
 ### Bug fixes
 
 - Fixed an issue where clicking the refresh icon beside a profile in the Unix System Services (USS) view had no effect. [#3693](https://github.com/zowe/zowe-explorer-vscode/issues/3693)
-- Fixed cross-LPAR PDS transfers to preserve original DCB attributes and prevent binary data corruption. Data is now transferred as binary, and encoding is user-selected at file open time. The encoding query parameter is now respected during write operations. [#3731](https://github.com/zowe/zowe-explorer-vscode/pull/3731)
+- Fixed a bug where USS files and data set VS Code editors accessed by an extender (SSH, FTP, etc) are not restored when reloading/reopening a window. [#3678](https://github.com/zowe/zowe-explorer-vscode/pull/3678)
+- Fix issues seen with out of date profile information within Zowe Explorer tree views. [#3664](https://github.com/zowe/zowe-explorer-vscode/issues/3664)
+  > > > > > > > origin/main
```
traeok: Looks like this line was accidentally left behind from a previous merge conflict.


```suggestion
```

ATorrise:
> I'm not sure if I am doing something wrong here but this is what I get after dropping test.pds3 into another LPAR and opening in binary.

Thanks for finding this! Should be fixed now

traeok (commented):
Thanks for addressing feedback, found a couple more things but plan to approve once addressed

packages/zowe-explorer/src/trees/dataset/DatasetTree.ts
```diff
@@ -112,16 +112,25 @@ export class DatasetTree extends ZoweTreeProvider<IZoweDatasetTreeNode> implemen
...
-                // create a PDS on remote
+                // find source PDS attributes and make destination PDS have same attributes
+                const sourceAttributesResponse = await ZoweExplorerApiRegister.getMvsApi(sourceInfo.profile).dataSet(dsname, {
+                    attributes: true,
+                    responseTimeout: sourceInfo.profile?.profile?.responseTimeout,
+                });
+                const { dsname: dsnameSource, ...rest } = sourceAttributesResponse.apiResponse.items[0];
+
+                // create a PDS on remote with the same attributes as source PDS
+                const transformedAttrs = (zosfiles.Copy as any).generateDatasetOptions({}, rest);
```
t1m0thyj, line 126: All the data set attributes appear to be replicated correctly except for `size` which increases for me:
| **Original PDS (size=15)** | **Moved PDS (size=225)** |
|---|---|
| Image | Image |

<sup>Used Space and Volumes are also different but I think that's expected</sup>

t1m0thyj (changes requested):
Thanks for addressing feedback! Found a possible issue related to `size` attribute.

packages/zowe-explorer/src/trees/dataset/DatasetTree.ts
ATorrise: fixed!

packages/zowe-explorer/src/trees/ZoweTreeProvider.ts
ATorrise: i accidentally added this to this PR from an unrelated pr i am working on thank you for catching

packages/zowe-explorer/__tests__/__unit__/trees/dataset/DatasetFSProvider.unit.test.ts
```diff
@@ -537,6 +537,36 @@ describe("DatasetFSProvider", () => {
...
+
+            await provider.writeFile(uri, content, { create: false, overwrite: true });
+
+            // ensure that binary and encoding can be passed
+            expect(mockMvsApi.uploadFromBuffer).toHaveBeenCalledWith(
+                Buffer.from(content),
+                binaryEntry.metadata.dsName,
+                expect.objectContaining({
+                    binary: true,
+                    encoding: "IBM-1047",
```
traeok: - I noticed this test is titled "passes binary: true to uploadFromBuffer when entry.encoding.kind is 'binary'". I would only expect `binary` to be passed to `uploadFromBuffer`, rather than both `binary` and `encoding`. Is it intentional to add `IBM-1047` as an encoding query parameter here? I don't think we should pass both, since this could lead to undefined behavior in the API that implements this function.
- The encoding specified on the query parameter should take precedence over the encoding defined on the entry. This `encoding` parameter is meant to act as an override, allowing the developer to pass the desired encoding for the resource. The encoding on the entry should still be respected if the query parameter is not provided.

packages/zowe-explorer/src/trees/dataset/DatasetTree.ts
```diff
@@ -112,19 +112,33 @@ export class DatasetTree extends ZoweTreeProvider<IZoweDatasetTreeNode> implemen
...
             if (!DatasetFSProvider.instance.exists(destUri)) {
-                // create a PDS on remote
+                // find source PDS attributes and make destination PDS have same attributes
+                const sourceAttributesResponse = await ZoweExplorerApiRegister.getMvsApi(sourceInfo.profile).dataSet(dsname, {
+                    attributes: true,
+                    responseTimeout: sourceInfo.profile?.profile?.responseTimeout,
+                });
+                const { dsname: dsnameSource, ...rest } = sourceAttributesResponse.apiResponse.items[0];
+                // need to transform labels
+                const transformedAttrs = (zosfiles.Copy as any).generateDatasetOptions({}, rest);
```
traeok, line 126: Should we file an issue to make `generateDatasetOptions` public so we can safely use it here? Currently, this strategy bypasses type checks for the function call, although its a necessary evil to use the function at this time.

packages/zowe-explorer/__tests__/__unit__/trees/dataset/DatasetFSProvider.unit.test.ts
ATorrise: sorry!

packages/zowe-explorer/src/trees/dataset/DatasetTree.ts
ATorrise: Yes!

traeok (dismissed):
Going to approve - I think we should update the access modifier for `generateDataSetOptions` as a separate issue on the Zowe CLI repo 😋 LGTM

pujal0909 (commented):
LGTM! will approve after tests are fixed

packages/zowe-explorer/src/trees/dataset/DatasetTree.ts
```diff
@@ -136,6 +150,7 @@ export class DatasetTree extends ZoweTreeProvider<IZoweDatasetTreeNode> implemen
             const children = await sourceNode.getChildren();
             for (const childNode of children) {
                 // move members within the folder to the destination
+                childNode.contextValue = 'member'; //force casting binary members to member context
```
t1m0thyj: Overwriting the `contextValue` of a node could have unintended side effects, since there may be additional data stored there like encoding or profile type.

Was this change needed to work around behavior in `crossLparMove`? If so, perhaps we can update code inside that method to correctly determine whether or not a node is a member.

packages/zowe-explorer/src/trees/dataset/DatasetTree.ts
```diff
@@ -166,7 +181,9 @@ export class DatasetTree extends ZoweTreeProvider<IZoweDatasetTreeNode> implemen
                 }
             } catch (err) {
                 //file might already exist. Ignore the error and try to write it to lpar
-                if (err.errorCode.toString() === "404" || err.errorCode.toString() === "500") {
+                // only bail out on a real 404/500 – guard against undefined
+                const code = err?.errorCode?.toString?.();
```
t1m0thyj: Coverity won't like this, since we assume `err` is defined on line 187, but handle the case of it being undefined here.

We should be consistent - since `err` should always be defined in a catch block, and JS objects should always define a `toString` implementation, I think we can do this:
```suggestion
                const code = err.errorCode?.toString();
```

packages/zowe-explorer/src/trees/dataset/DatasetTree.ts
```diff
@@ -112,19 +112,33 @@ export class DatasetTree extends ZoweTreeProvider<IZoweDatasetTreeNode> implemen
...
+                // find source PDS attributes and make destination PDS have same attributes
+                const sourceAttributesResponse = await ZoweExplorerApiRegister.getMvsApi(sourceInfo.profile).dataSet(dsname, {
+                    attributes: true,
+                    responseTimeout: sourceInfo.profile?.profile?.responseTimeout,
+                });
+                const { dsname: dsnameSource, ...rest } = sourceAttributesResponse.apiResponse.items[0];
+                // need to transform labels
+                const transformedAttrs = (zosfiles.Copy as any).generateDatasetOptions({}, rest);
+                let dataSetTypeEnum = zosfiles.CreateDataSetTypeEnum.DATA_SET_BLANK;
+                // if alcunit is cyl, divide primary by 15 to get the number of cylinders
```
t1m0thyj, line 128: Is the same calculation necessary for `secondary`? Not saying it is, just want to make sure we've tested this case 😋

t1m0thyj (changes requested):

packages/zowe-explorer/src/trees/dataset/DatasetTree.ts
ATorrise: it was to make testing easier but we can talk about a different approach?

packages/zowe-explorer/src/trees/dataset/DatasetTree.ts
ATorrise: im not sure what counts as necessary here 😅

t1m0thyj: Did some manual testing and it seems that `secondary` is fine 👍 

However, there may still be some issues around how we handle `primary`. I observed the following behavior:
1. Create a new DS, using attributes from the Partitioned Default template except primary size is set to 10 CYL.
2. Review attributes of created DS and notice the size is 150 CYL (expected 10).

@ATorrise @zFernand0 do you see the same behavior? If so I think we can handle it in a separate PR 🙂

packages/zowe-explorer/__tests__/__unit__/trees/dataset/DatasetFSProvider.unit.test.ts
```diff
@@ -537,6 +537,33 @@ describe("DatasetFSProvider", () => {
...
+            const parent = { entries: new Map(), metadata: { profile: { name: "profile" }, path: "/" } };
+            const binaryEntry = { ...testEntries.pdsMember, wasAccessed: true, encoding: { kind: "binary" } } as DsEntry;
+            parent.entries.set("MEMBER1", binaryEntry);
+            jest.spyOn(provider as any, "lookupParentDirectory").mockReturnValueOnce(parent);
+
+            const mockMvsApi = {
+                uploadFromBuffer: jest.fn().mockResolvedValue({ apiResponse: { etag: "etag" } }),
+            };
+
+            jest.spyOn(ZoweExplorerApiRegister, "getMvsApi").mockReturnValue(mockMvsApi as any);
```
traeok, line 503: (nitpick) this mock for `getMvsApi` is never restored within this describe block, so it could impact future `writeFile` tests

packages/zowe-explorer/__tests__/__unit__/trees/dataset/DatasetTree.unit.test.ts
```diff
@@ -4694,3 +4694,166 @@ describe("Dataset Tree Unit Tests - Function focusOnDsInTree", () => {
...
+            resourceUri: srcUri,
+            contextValue: "ds"
+        };
+
+        (fakeNode as any).getEncoding = jest.fn().mockResolvedValue({ kind: "text" });
+
+    });
+
+    it("should copy a sequential dataset and delete the source", async () => {
+        await tree["crossLparMove"](fakeNode, srcUri, dstUri, false);
```
traeok, line 5118: Is this to workaround `crossLparMove` being private? Usually, private functions can be tested by writing comprehensive test cases for public functions (that would in turn call the private helper functions), but I know we've done similar workarounds in the past to test private functions, so I'm not requesting any changes for this.

packages/zowe-explorer/CHANGELOG.md
```diff
@@ -32,6 +32,7 @@ All notable changes to the "vscode-extension-for-zowe" extension will be documen
 - Fixed an issue where a `401 Unauthorized` error is thrown by the FileSystemProvider implementations when the user cancels or closes an authentication prompt. Now, the filesystem throws a more explicit `AuthCancelledError` in this scenario. [#3662](https://github.com/zowe/zowe-explorer-vscode/issues/3662)
 - Fixed an issue where renaming or deleting the team configuration file caused an unhandled exception in Zowe Explorer, causing old, non-existent profiles to remain in the tree views. [#3772](https://github.com/zowe/zowe-explorer-vscode/issues/3772)
 - Fixed a regression with the `AuthUtils.syncSessionNode` function that caused Zowe Explorer logic to stop unexpectedly if a session node did not yet have a rich hover tooltip. [#3795](https://github.com/zowe/zowe-explorer-vscode/pull/3795)
+- Fixed cross-LPAR PDS transfers to preserve original DCB attributes and prevent binary data corruption. Data is now transferred as binary, and encoding is user-selected at file open time. The encoding query parameter is now respected during write operations. [#3731](https://github.com/zowe/zowe-explorer-vscode/pull/3731)
```
traeok: I think there's a bit of confusion: in this PR we added the encoding query parameter, so that aspect is an enhancement to the APIs rather than a bug fix. Separate changelog entry?


```suggestion
- Fixed cross-LPAR PDS transfers to preserve original DCB attributes and prevent binary data corruption. Data is now transferred as binary, and the encoding is respected if previously set by the user. [#3731](https://github.com/zowe/zowe-explorer-vscode/pull/3731)
```

ATorrise: i think the way im explaining just doesn't make sense to you but i will change it lol

pujal0909 (commented):
Getting this right now: 

image

packages/zowe-explorer/__tests__/__unit__/trees/dataset/DatasetFSProvider.unit.test.ts
zFernand0: Not to say that we shouldn't address this here, but FWIW...
I've added a `jest.restoreAllMocks();` in the `beforeEach` of the top-level `describe` 🙏 

Some other tests had to be refactored though, see:
- #3791 

My suggestion for now is to leave it as is (if the tests pass) and I can help to resolve this in the other PR 🙏

zFernand0 (commented):
Changes MSTM! 😋 
---
Not approving yet as I can't fully test this at the moment 😓

packages/zowe-explorer/__tests__/__unit__/trees/dataset/DatasetTree.unit.test.ts
zFernand0: > Is this to workaround `crossLparMove` being private? 

I'd be curious to know the answer too 😋 
I dont' recall using this technique. I've mostly looked into a combination of the following:
- Calling the method with `as any` and/or,
- Accessing the method from the prototype of the class. 😋

zFernand0:
The conflicts in this PR are related to #3791
In adition to that... we also need to get the changes in #3810 

So I'll be closing #3810 in favor of this PR 🙏 
Please stay tuned 🙏

zFernand0 (changes requested):
Requesting changes just to hold this PR from being merged! 🙏 

I'm looking into these two items 🙏 
- image
- image

packages/zowe-explorer/__tests__/__unit__/trees/dataset/DatasetFSProvider.unit.test.ts
zFernand0: This should be resolved after merging from the main branch 🙏

packages/zowe-explorer/__tests__/__unit__/trees/dataset/DatasetTree.unit.test.ts
zFernand0: > Is this to workaround crossLparMove being private? 

After some experimentation, I'm guessing the answer is yes 😋 
I agree that we could refactor some of the tests to call the handleDrop directly. 🙏 

Curious if there is time to do so before the 3.3.0 release date 🤔 

@ATorrise, any thoughts?

packages/zowe-explorer/src/trees/dataset/DatasetTree.ts
zFernand0: This could either be environmental, or perhaps it was corrected by followup commit 🙏 

Before (left) After (right) 
image

packages/zowe-explorer/src/trees/dataset/DatasetTree.ts
```diff
@@ -152,7 +162,7 @@ export class DatasetTree extends ZoweTreeProvider<IZoweDatasetTreeNode> implemen
             try {
                 const entry = await DatasetFSProvider.instance.fetchDatasetAtUri(destUri);
                 if (entry == null) {
-                    if (sourceNode.contextValue === Constants.DS_MEMBER_CONTEXT) {
+                    if (SharedContext.isDsMember(sourceNode)) {
```
zFernand0, line 165: This is what resolved the 500 error

packages/zowe-explorer/src/trees/dataset/DatasetTree.ts
```diff
@@ -175,9 +187,14 @@ export class DatasetTree extends ZoweTreeProvider<IZoweDatasetTreeNode> implemen
             const contents = await DatasetFSProvider.instance.readFile(sourceNode.resourceUri);
             //write the contents to the destination LPAR
             try {
+                const encodingInfo = await sourceNode.getEncoding();
+                // If the encoding is binary, we need to force upload it as binary
+                const queryString = `forceUpload=true${
+                    encodingInfo?.kind === "binary" ? "&encoding=binary" : encodingInfo?.kind === "other" ? "&encoding=" + encodingInfo?.codepage : ""
```
zFernand0, line 193: And this helped keep the last "Open with Ending" option chosen by the user.

pujal0909 (approved):
works. LGTM!

Issue #3485, opened 2025-02-26
Rename USS file name conflict


pablocarle:
**Describe the bug**

Renaming a file to a file that existed before in the directory but was deleted in z/OS results in a name conflict with a file that in reality does not exist anymore

**To Reproduce**

Haven't tested but probably:

1. Go to test dir /test
2. List files, rename file1 to file2
3. Delete directory contents in z/OS
4. Recreate the directory and try to rename file1 to file2

**Expected behavior**

**Screenshots**

![Image](https://github.com/user-attachments/assets/bc6d8c6a-4577-4ce8-8443-22b697e4fe78)

**Desktop (please complete the following information):**

- OS: MacOS
- Zowe Explorer Version:

PR #3739, opened 2025-07-01
Fixed Rename USS files bug

pujal0909:
## Proposed changes

Renaming a file to a file that existed before in the directory but was deleted in z/OS results in a name does not throw an error anymore.
https://github.com/zowe/zowe-explorer-vscode/issues/3485

To Test:
1. Delete a file on z/OS using ssh
2. Rename another file on zowe explorer with the same name as the deleted file
3. You will no longer see an error appear

## Release Notes

Milestone:

Changelog:

## Types of changes

## Checklist

**General**

**Code coverage**

**Deployment**

### DIFF
--- a/packages/zowe-explorer/__tests__/__unit__/trees/dataset/DatasetActions.unit.test.ts
+++ b/packages/zowe-explorer/__tests__/__unit__/trees/dataset/DatasetActions.unit.test.ts
@@ -11,7 +11,7 @@
 
 import * as vscode from "vscode";
 import * as zosfiles from "@zowe/zos-files-for-zowe-sdk";
-import { Gui, imperative, Validation, ProfilesCache, ZoweExplorerApiType, Sorting } from "@zowe/zowe-explorer-api";
+import { Gui, imperative, Validation, ProfilesCache, ZoweExplorerApiType, Sorting, PdsEntry, DirEntry } from "@zowe/zowe-explorer-api";
 import { DatasetFSProvider } from "../../../../src/trees/dataset/DatasetFSProvider";
 import { bindMvsApi, createMvsApi } from "../../../__mocks__/mockCreators/api";
 import {
@@ -38,13 +38,14 @@ import { FilterDescriptor } from "../../../../src/management/FilterManagement";
 import { ZoweLogger } from "../../../../src/tools/ZoweLogger";
 import { ZoweDatasetNode } from "../../../../src/trees/dataset/ZoweDatasetNode";
 import { SharedUtils } from "../../../../src/trees/shared/SharedUtils";
-import { mocked } from "../../../__mocks__/mockUtils";
+import { mocked, MockedProperty } from "../../../__mocks__/mockUtils";
 import { DatasetActions } from "../../../../src/trees/dataset/DatasetActions";
 import { AuthUtils } from "../../../../src/utils/AuthUtils";
 import { SettingsConfig } from "../../../../src/configuration/SettingsConfig";
 import { TreeViewUtils } from "../../../../src/utils/TreeViewUtils";
 import { DatasetUtils } from "../../../../src/trees/dataset/DatasetUtils";
 import { ProfileManagement } from "../../../../src/management/ProfileManagement";
+import { SharedTreeProviders } from "../../../../src/trees/shared/SharedTreeProviders";
 import { DataSetAttributesProvider } from "../../../../../zowe-explorer-api/lib/dataset/DatasetAttributesProvider";
 
 // Missing the definition of path module, because I need the original logic for tests
@@ -3048,6 +3049,29 @@ describe("Dataset Actions Unit Tests - Function zoom", () => {
         };
     }
 
+    function setupMocksForZoom(
+        blockMocks,
+        selectionText: string,
+        datasetName: string,
+        memberName: string,
+        isValidDataSet: boolean,
+        isValidMember: boolean,
+        profileNames?: string[]
+    ) {
+        blockMocks.document.getText.mockReturnValue(selectionText);
+        Object.defineProperty(vscode.window, "activeTextEditor", { value: blockMocks.editor, configurable: true });
+        jest.spyOn(DatasetUtils, "extractDataSetAndMember").mockReturnValue({ dataSetName: datasetName, memberName: memberName });
+        jest.spyOn(DatasetUtils, "validateDataSetName").mockReturnValue(isValidDataSet);
+        jest.spyOn(DatasetUtils, "validateMemberName").mockReturnValue(isValidMember);
+
+        if (profileNames !== undefined) {
+            Object.defineProperty(ProfileManagement, "getRegisteredProfileNameList", {
+                value: jest.fn().mockReturnValue(profileNames),
+                configurable: true,
+            });
+        }
+    }
+
     beforeEach(() => {
         jest.clearAllMocks();
     });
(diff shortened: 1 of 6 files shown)


anaxceron (approved):

packages/zowe-explorer/src/trees/uss/ZoweUSSNode.ts
```diff
@@ -393,8 +393,24 @@ export class ZoweUSSNode extends ZoweTreeNode implements IZoweUSSTreeNode {
...
+                        Gui.errorMessage(err.message);
+                        return;
+                    }
+                } catch (err) {
+                    if ((err as any).name === "ImperativeError" && Number(err.errorCode) === imperative.RestConstants.HTTP_STATUS_404) {
+                        const parent = UssFSProvider.instance._lookupParentDirectory(newUri);
+                        parent.entries.delete(path.posix.basename(newUri.path));
+                        await vscode.workspace.fs.rename(oldUri, newUri);
+                    }
+                }
```
t1m0thyj: Would it make sense to handle this error inside the `UssFSProvider.rename` function?

This would allow `_lookupParentDirectory` to be called without needing to make it a public method.

Not necessarily suggesting a change, if this error handling is only needed in the ZE tree and not applicable to every `zowe-uss` rename operation.

pujal0909: Thanks for pointing that out! For this case, I think the error handing is only needed in the tree because of the stale state that occurs in the tree view when deleting a file on z/OS. The breakpoints I had in `USSFsProvider.rename()` were not firing and the error is coming from the vscode file system provider, so I put the logic in the `ZoweUSSNode.rename()` for now.

packages/zowe-explorer-api/src/fs/BaseProvider.ts
```diff
@@ -479,7 +479,7 @@ export class BaseProvider {
         return entry;
     }
 
-    protected _lookupParentDirectory(uri: vscode.Uri, silent?: boolean): DirEntry {
+    public _lookupParentDirectory(uri: vscode.Uri, silent?: boolean): DirEntry {
```
JWaters02: Super nitpick, but now this function is public, the name should not have the underscore before it.
```suggestion
    public lookupParentDirectory(uri: vscode.Uri, silent?: boolean): DirEntry {
```

traeok: While I agree, this could be considered breaking because the function was previously marked as protected, meaning that any provider implementing the `BaseProvider` class would need to update its references to use the new function name.

pujal0909:
@zFernand0 Thanks for pointing that out! It should be fixed now.

packages/zowe-explorer-api/src/fs/BaseProvider.ts
zFernand0: couldn't we do something like this?

```typescript
    /**
     * @deprecated Please use the public version
     */
    protected _lookupParentDirectory(uri: vscode.Uri, silent?: boolean): DirEntry {
        return this.lookupParentDirectory(uri, silent);
    }
    public lookupParentDirectory(uri: vscode.Uri, silent?: boolean): DirEntry {
        return this._lookupAsDirectory(
            uri.with({
                path: path.posix.join(uri.path, ".."),
            }),
            silent ?? false
        );
    }
```

zFernand0: in either scenario, we might want to add a changelog entry for the ze-api package. 😋

zFernand0 (dismissed):
LGTM! 😋 
---
Thanks for addressing the feedback 🙏

ATorrise (approved):

ATorrise (approved):

packages/zowe-explorer-api/src/fs/BaseProvider.ts
traeok: I'm in favor of redefining it under the new name (w/o underscore) while marking the previous version as deprecated. However, this function is used in multiple places, so if this change is made, we should also update our pre-existing references to use the new function.

zFernand0: We should start thinking on cleaning up some things before V4 😉 
I vote for adding this in the (future) list of breaking changes in the V4 changelog. 😋 

All that to say, yeah... We should replace all pre-existing references to use the new function 🙏

JWaters02 (dismissed):
It works

traeok:
I think that these changes resolve the problem outlined in #3485, specifically when the user interacts with the tree view.

But I'm starting to wonder if this resolves all scenarios where this can happen. For example, what if an extender lists USS files and then tries to rename one that no longer exists on the remote system?

1. Extender uses `vscode.workspace.fs.readDirectory` to get the entries for a USS folder
2. A short period of time passes, where the end user removes a file in this folder on the remote system
3. The extender calls `vscode.workspace.fs.rename` to rename a USS file based on some user action that was taken
4. The check to remove the USS file entry (that was deleted on z/OS) is not called before the filesystem rename action

I'll have to do some testing to verify, but if the error occurs here, we may need to look into refactoring the `UssFsProvider.rename` function to completely solve this

anaxceron (approved):

traeok (changes requested):
Thanks Pujal for addressing all the feedback thus far. Please see my comment above about how extenders can encounter this error.

anaxceron (approved):

packages/zowe-explorer/src/trees/uss/UssFSProvider.ts
traeok, line 1: Could some of the comments be restored in this file? I think the logic looks good, but it may help to keep the comment that was above line 705 for clarity

packages/zowe-explorer/src/trees/uss/ZoweUSSNode.ts
```diff
@@ -391,26 +391,10 @@ export class ZoweUSSNode extends ZoweTreeNode implements IZoweUSSTreeNode {
         });
 
         try {
-            await vscode.workspace.fs.rename(oldUri, newUri, { overwrite: false });
+            await UssFSProvider.instance.rename(oldUri, newUri, { overwrite: false });
```
traeok: Is there a reason why this was changed to use `vscode.workspace.fs`? If this call was failing before it gets to the filesystem (as we were seeing offline), we can adjust the logic here so that we delete the entry locally before calling the `...fs.rename` function

packages/zowe-explorer-api/CHANGELOG.md
```diff
@@ -6,6 +6,7 @@ All notable changes to the "zowe-explorer-api" extension will be documented in t
 
 ### New features and enhancements
 
+- Marked `protected _lookupParentDirectory()` method as `@deprecated` and redefined it as a `public lookupParentDirectory()` method. [#3485] (https://github.com/zowe/zowe-explorer-vscode/issues/3485)
```
anaxceron: Is it possible just to say, "Deprecated the `bbbb` method and redefined it as the `zzzz` method."?

anaxceron (commented):
Left a question regarding one of the changelogs.

packages/zowe-explorer-api/CHANGELOG.md
```diff
@@ -6,7 +6,7 @@ All notable changes to the "zowe-explorer-api" extension will be documented in t
 
 ### New features and enhancements
 
-- Marked `protected _lookupParentDirectory()` method as `@deprecated` and redefined it as a `public lookupParentDirectory()` method. [#3485] (https://github.com/zowe/zowe-explorer-vscode/issues/3485)
+- Deprecated the `protected _lookupParentDirectory()` and redefined it as a `public lookupParentDirectory()` method. [#3485] (https://github.com/zowe/zowe-explorer-vscode/issues/3485)
```
anaxceron:
```suggestion
- Deprecated `protected _lookupParentDirectory()` and redefined it as a `public lookupParentDirectory()` method. [#3485] (https://github.com/zowe/zowe-explorer-vscode/issues/3485)
```

anaxceron (approved):
Left a minor edit suggestion for the changelog.

traeok (commented):
Running into some issues while testing, here are the steps I'm taking:

1. List my home directory in Zowe Explorer
2. Then deleted the file `ibm1047.txt` from it
3. In the USS tree, I try to rename it to `ibm1048.txt` and encounter a 404 error
4. The non-existent file remains visible in the USS tree

Image

traeok (approved):
Thanks Pujal, this is working as expected - I tested the scenario again and was mixing up a step based on the "steps to reproduce" in the original issue.

PR #3751, opened 2025-07-08
feat(ds): "Show as Table" for data set list and PDS members

traeok:
## Proposed changes

**ZE:**
- Added "Show as Table" option for profiles with search patterns in the Data Sets tree. Users can now right-click on a profile in the Data Sets tree with a search pattern set and select "Show as Table" to view matching data sets in a tabular format.
- Added "Show as Table" option for partitioned data sets. Users can now right-click on a partitioned data set in the Data Sets tree and select "Show as Table" to view its members in a tabular format. 
- Added "Zowe Explorer: List Data Sets" command. Users can now run this command from the command palette to be prompted for a search pattern and display the matching data sets in a table view.

**ZE API:**
- Added dynamic title support for table view actions through the `DynamicTitle` type, allowing action titles to be computed based on selected row data. Action titles can now be functions that receive row data and return a `string` or `Promise<string>`.
- Added dynamic hide condition support for table view actions through the `hideCondition` property on the `Table.Action` type. Actions can now be conditionally hidden from the UI based on the selected row data.
- Added support for no-selection actions in table views through the `NoSelectionCallback` type. Actions can now be executed without requiring any row selection by using the `"no-selection"` callback type.
- Added row pinning functionality to table views with new APIs: `pinRows()` to pin specific rows to the top of the grid, `unpinRows()` to unpin rows, `getPinnedRows()` to retrieve currently pinned rows, and `setPinnedRows()` to replace all pinned rows with a new set.
- Added `waitForAPI()` function to the `Table.View` class to allow extensions to wait for the AG Grid API to be fully initialized before performing operations on the table view.
- Added grid state management functions `getGridState()` and `setGridState()` to the `Table.View` class to allow extensions to save and restore the complete state of the table view, including column positions, sorting, filtering, and other grid configurations.

## GIFs & Screenshots

<details>
<summary>Listing and selecting data sets in table</summary>
image
</details>

<details>
<summary>Listing data set members in table</summary>
image
</details>

<details>
<summary>Listing data set members through "Focused" view (selecting Focus on PDS)</summary>

</details>

<details>
<summary>Expanding PDS members inline in data set table view</summary>

</details>

## Release Notes

Milestone: 3.3

Changelog: See above

## Types of changes

- [x] Documentation (Markdown, README updates)

## Checklist

**General**

**Code coverage**

**Deployment**

### DIFF
--- a/packages/zowe-explorer/__tests__/__unit__/trees/dataset/DatasetTree.unit.test.ts
+++ b/packages/zowe-explorer/__tests__/__unit__/trees/dataset/DatasetTree.unit.test.ts
@@ -40,7 +40,6 @@ import { ZoweLogger } from "../../../../src/tools/ZoweLogger";
 import { ZowePersistentFilters } from "../../../../src/tools/ZowePersistentFilters";
 import { DatasetFSProvider } from "../../../../src/trees/dataset/DatasetFSProvider";
 import { DatasetTree } from "../../../../src/trees/dataset/DatasetTree";
-import { DatasetUtils } from "../../../../src/trees/dataset/DatasetUtils";
 import { ZoweDatasetNode } from "../../../../src/trees/dataset/ZoweDatasetNode";
 import { SharedActions } from "../../../../src/trees/shared/SharedActions";
 import { SharedTreeProviders } from "../../../../src/trees/shared/SharedTreeProviders";
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -3342,7 +3342,7 @@ packages:
       cheerio: 1.0.0-rc.12
       cockatiel: 3.1.3
       commander: 12.1.0
-      form-data: 4.0.0
+      form-data: 4.0.4
       glob: 11.0.3
       hosted-git-info: 4.1.0
       jsonc-parser: 3.2.1
@@ -4853,6 +4853,14 @@ packages:
       responselike: 3.0.0
     dev: true
 
+  /call-bind-apply-helpers@1.0.2:
+    resolution: {integrity: sha512-Sp1ablJ0ivDkSzjcaJdxEunN5/XvksFJ2sMBFfq6x0ryhQV/2b/KwFe21cMpmHtPOSij8K99/wSfoEuTObmuMQ==}
+    engines: {node: '>= 0.4'}
+    dependencies:
+      es-errors: 1.3.0
+      function-bind: 1.1.2
+    dev: true
+
   /call-bind@1.0.7:
     resolution: {integrity: sha512-GHTSNSYICQ7scH7sZ+M2rFopRoLh8t2bLSW6BbgrtLsahOIB5iyAVJf9GjWK3cYTDaMj4XdBpM1cA6pIS0Kv2w==}
     engines: {node: '>= 0.4'}
@@ -5847,6 +5855,15 @@ packages:
     engines: {node: '>=12'}
     dev: true
 
+  /dunder-proto@1.0.1:
+    resolution: {integrity: sha512-KIN/nDJBQRcXw0MLVhZE9iQHmG68qAVIBg9CqmUYjmQIhgij9U5MFvrqkUL5FbtyyzZuOeOt0zdeRe4UY7ct+A==}
+    engines: {node: '>= 0.4'}
+    dependencies:
+      call-bind-apply-helpers: 1.0.2
+      es-errors: 1.3.0
+      gopd: 1.2.0
+    dev: true
+
   /duplexer2@0.1.4:
     resolution: {integrity: sha512-asLFVfWWtJ90ZyOUHMqk7/S2w2guQKxUI2itj3d92ADHhxUSbCMGi1f1cBcJ7xM1To+pE/Khbwo1yuNbMEPKeA==}
     dependencies:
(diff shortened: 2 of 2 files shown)


JWaters02 (changes requested):
I just tested this, in this order:
1. Installed VSIX
2. Ctrl+shift+p -> Developer: Reload window
3. Do search on zosmf profile, right click on a PDS -> "Show as table" shows the zowe resources tab but nothing appears
4. Right click on zosmf profile -> "Show as table" returns error
> Error running command zowe.ds.tableView: Cannot read properties of undefined (reading 'path'). This is likely caused by the extension that contributes zowe.ds.tableView.
5. I checked with "Show as table" on jobs tree and also same issue as above where the pane is blank

image

6. Ctrl+shift+p -> Developer: Reload window
7. Do search on zosmf profile, right click on same PDS as above -> "Show as table" and it _showed the table fine_ (same with jobs too)
8. Checked again on profile show as table, still same error
9. Ctrl+shift+p -> ZE: List data sets, put in a HLQ data set that contains multiple data sets. Table showed fine, but clicking on the arrow next to any PDS in the list fails to show members under the PDS
10. Ctrl+shift+p -> ZE: List data sets, put in a PDS pattern. Table showed fine with just the PDS as one line, again clicking arrow fails to show the members under it
11. Pinning feature is nice. But when I pin a data set, how do I unpin it? Right click menu doesn't show unpin and the selection checkbox is gone on pinned items

packages/zowe-explorer/src/trees/dataset/DatasetTableView.ts
```diff
@@ -0,0 +1,1164 @@
...
+            },
+        },
+    };
+
+    // These fields are typically included in data set metadata.
+    private expectedFields = [
+        {
+            field: "dsname",
+            headerName: l10n.t("Data Set Name"),
+            initialSort: "asc",
```
JWaters02, line 410: Perhaps this could be a future enhancement but one to mention anyway. Could the sort order of the data set you open the table with please apply to the initial column sort order. For now I believe this would just read the default sort order but when I eventually get round to it each PDS (and profile) would have sort persistence. But yeah this initialSort could read defaults setting and get desc/asc and also if possible the sort type like date modified or created date.

traeok: Sure, I'll push up that change shortly 👍

traeok: Implemented in 487d124

traeok:
I found a bug where switching between 2 PDS in the table can cause the rows to be out of sync. I've addressed this in the latest commit, 487d124.

> 5\. I checked with "Show as table" on jobs tree and also same issue as above where the pane is blank

I'm not able to reproduce your issue, starting from step 1 through 8 @JWaters02 - you mentioned an issue with the jobs tree, but I didn't make any changes to that table. I'm wondering if there was an issue with building the webviews or packaging the VSIX that could be causing this. Will continue to test further with the VSIX.

> 9. Ctrl+shift+p -> ZE: List data sets, put in a HLQ data set that contains multiple data sets. Table showed fine, but clicking on the arrow next to any PDS in the list fails to show members under the PDS

After following steps 1-8 I tried this with the latest changes - clicking on the arrow next to a PDS works as expected and the members appear below it

> 10. Ctrl+shift+p -> ZE: List data sets, put in a PDS pattern. Table showed fine with just the PDS as one line, again clicking arrow fails to show the members under it

I don't want to sound like a broken record, but I can't repro this with the latest changes 😓 

> 11. Pinning feature is nice. But when I pin a data set, how do I unpin it? Right click menu doesn't show unpin and the selection checkbox is gone on pinned items

Good question, I don't know why the checkbox does not show for pinned rows, I did some search and found this: https://github.com/ag-grid/ag-grid/issues/3365

But unfortunately its not clear whether this was resolved or if its by design. In the meantime:
- I can add the unpin option to the context menu, or as an alternative..
- You can select those items outside of the pinned region and select "Unpin"

> 12. Ran Ctrl+shift+p -> ZE: List data sets and put in qualifier. Pinned a few things, changed column sorts, changed paginations etc. Reran Ctrl+shift+p -> ZE: List data sets and put in the same qualifier on same profile. Table "forgot" all settings. Probably future enhancement to have table persistence for any one qualifier

This is currently by design, but we could definitely pursue a more persistent table state. I think it would be best to do this as a follow-up enhancement since there's already a lot of code changes in this PR and I don't want to increase scope 😋

JWaters02:
> I'm not able to reproduce your issue, starting from step 1 through 8 @JWaters02 - you mentioned an issue with the jobs tree, but I didn't make any changes to that table. I'm wondering if there was an issue with building the webviews or packaging the VSIX that could be causing this. Will continue to test further with the VSIX.

I can't replicate the jobs table not showing at all any more, I think it was probably just that one off session after installing the build or something. 

As for step 4/8, with local build I still get the same error. 

image

> After following steps 1-8 I tried this with the latest changes - clicking on the arrow next to a PDS works as expected and the members appear below it

> I don't want to sound like a broken record, but I can't repro this with the latest changes 😓

Yep drop down arrow now also works for me fine on local build.

> Good question, I don't know why the checkbox does not show for pinned rows, I did some search and found this: https://github.com/ag-grid/ag-grid/issues/3365
But unfortunately its not clear whether this was resolved or if its by design. In the meantime:
I can add the unpin option to the context menu, or as an alternative..
You can select those items outside of the pinned region and select "Unpin"

Okay I see it's a problem with underlying library now. Then I think unpin option in context menu as well as the other workaround as the select -> unpin button was not obvious to me to try that (as pinning something usually means "moving" it to pinned rather than "copying" it there) - the obvious first thing to try was right click to unpin. Maybe others would disagree so best to ask on that one 🙂 

> This is a side-effect of the current design - a new table is built every time you call the command. But we could definitely pursue a more persistent table state. It might be a good idea to do this as a follow-up enhancement, since there's already a lot of code changes in this PR 😋

Yep sounds good to me!

anaxceron (dismissed):

anaxceron (dismissed):

traeok:
> Similarly, if you set a filter and show as table, then change the filter to something else, the "display in tree" no longer works, even if the item exists in favourites.

> If you show a favourites filter or PDS as table and your profile filter is not set or is set to something different, "display in tree" does not show the item in favourites.

These 2 should be addressed in 05bade9 - I just was missing some logic to search through favorites.

> If you set a filter and show as table, then change the filter to something else (but importantly do not show new filter as table), the dropdown arrows on the table no longer works (focus is a workaround that works though).

Still looking into this

jace-roell (approved):
Really cool feature @traeok. LGTM!

PR #3775, opened 2025-07-24
fix(regression): Handle uncaught exception in `ZoweTreeNode.getProfile`

### LINKED ISSUE #3772: Cannot hide or remove profile from tree after deleting config
**Describe the bug**

**To Reproduce**

1. Create a team config in Zowe Explorer UI

The default name of `zosmf` is used:

Image

2. Use VS Code to delete the `zowe.config.json` and `zowe.schema.json` files. 
3. "Manage Profile" and select "Hide"

Image

4. You get an error:

```
Error running command zowe.profileManagement: Zowe Explorer Profiles Cache error: Could not find profile named: zosmf.. This is likely caused by the extension that contributes zowe.profileManagement.
```

**Expected behavior**

I should be able to remove ("hide") a profile from the UI that I deleted.  

**Screenshots**

- OS:
- Zowe Explorer Version:

**Additional context**

traeok:
## Proposed changes

Fixes #3772 and resolves multiple scenarios due to a `ZoweTreeNode.getProfile` regression in the Zowe Explorer API. The `loadNamedProfile` function was called outside of a try/catch block, causing exceptions from the function to be propagated to the caller/API consumer (breaking change).

If the user:
- deletes a profile from their team configuration
- renames a profile in the team configuration
- deletes the team configuration entirely
- renames the team configuration file

The matching profile node is not removed from any of the trees, because the Zowe Explorer core logic abruptly terminated due to the uncaught exception.

## Release Notes

Milestone: 3.3.0

Changelog:

## Types of changes

## Checklist

**General**

**Code coverage**

**Deployment**

### DIFF
--- a/packages/zowe-explorer-api/src/tree/ZoweTreeNode.ts
+++ b/packages/zowe-explorer-api/src/tree/ZoweTreeNode.ts
@@ -83,7 +83,9 @@ export class ZoweTreeNode extends vscode.TreeItem {
             } catch (err) {
                 // Profile does not exist. Log and return last known profile for backwards compatibility
                 Logger.getAppLogger().error(
-                    `[ZoweTreeNode.getProfile] Profile ${this.profile.name} does not exist for node ${this.label}, returning last known profile`
+                    `[ZoweTreeNode.getProfile] Profile ${
+                        this.profile.name
+                    } does not exist for node ${this.label?.toString()}, returning last known profile`
                 );
             }
         }
--- a/packages/zowe-explorer-api/src/tree/ZoweTreeNode.ts
+++ b/packages/zowe-explorer-api/src/tree/ZoweTreeNode.ts
@@ -14,6 +14,7 @@ import * as imperative from "@zowe/imperative";
 import { IZoweTreeNode } from "./IZoweTreeNode";
 import type { BaseProvider } from "../fs/BaseProvider";
 import type { ProfilesCache } from "../profiles";
+import { Logger } from "@zowe/imperative";
 
 /**
  * Common implementation of functions and methods associated with the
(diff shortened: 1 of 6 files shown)


packages/zowe-explorer-api/src/tree/ZoweTreeNode.ts
```diff
@@ -76,8 +77,17 @@ export class ZoweTreeNode extends vscode.TreeItem {
      * @returns {imperative.IProfileLoaded}
      */
     public getProfile(profilesCache?: ProfilesCache): imperative.IProfileLoaded {
-        if (this.profile != null && profilesCache?.loadNamedProfile != null) {
-            return profilesCache.loadNamedProfile(this.profile.name);
+        if (this.profile != null && profilesCache != null) {
+            try {
+                return profilesCache.loadNamedProfile(this.profile.name);
```
zFernand0, line 82: Thanks for catching this issue.
In theory, there is one more function that's calling `loadNamedProfile`

CallStack:
`ssoLogin|ssoLogout > getServiceProfile > getServiceProfileForAuthPurposes`

Not sure that we should be super concerned with that other use, since it's using the `cache` 😋 

However, we should warn extenders that may rely on the `loadNamedProfile` 😅

zFernand0 (dismissed):
LGTM! 😋 
---
Thanks for addressing the issue so quickly.
The profiles just go away now (as they used to before) 😋

packages/zowe-explorer-api/src/tree/ZoweTreeNode.ts
traeok: I was tempted to refactor this as well, but after looking into the logic, I realized the function has always thrown an error if the profile does not exist. Since this code has thrown an error for 5 years in this scenario, I think its best to keep it as-is to preserve this behavior (as we're already past the major version boundary).

I would expect the `ssoLogin` function to throw an error if the profile does not exist, so I believe this thrown exception is acceptable.

JillieBeanSim (dismissed):
thanks for working on this @traeok this does resolve the delete config and add to workspace issues but I do see with the profile name change scenario the tree does get updated but I end up in a 401 credential loop and not sure why when token auth in place even after re-logging in after the name change I get successful login then a 401 prompt

traeok:
@JillieBeanSim Thanks for reviewing. The scenario you mentioned definitely sounds problematic. But, this also means that the same behavior would be seen in `main` before #3673 was merged, correct? 

If so, in the interest of quickly reverting this major breaking change & regression, I think its best to fix that scenario in a different PR.

zFernand0 (approved):
LGTM! 😋 
---
Thanks for the clear changelog too! 🙏

PR #3791, opened 2025-07-31
Fix: Prevent Data Loss - Handle Long Records in Save Process

### LINKED ISSUE #3745: zowe explorer should be smarted when editing members with lrecl
**Describe the bug**

**To Reproduce**

**Expected behavior**

**Screenshots**

- OS:
- Zowe Explorer Version:

**Additional context**

I edited a data set with lrecl 80, and created a member.
I entered a long line o txt and tried to save it.  It gave me

Failed to save 'COLINDEL': Unable to write file 'zowe-ds:/zosmfuserid/SYS1.ZOSR24.PARMLIB/COLINDEL' (Error: Rest API failure with HTTP(S) status 500 EDC5003I Truncation of a record occurred during an I/O operation. (errno2=0xC0440016) __amrc: __last_op=151 __code=0x00000000 fwrite() error)

I was expecting the edit window where I entered text to have a window width of the lrecl, either a hard stop or a feint line to mark the  width of the file.

____________________

I was confused by the error message having zosmfuserid.   What is this userid.  Having just SYS1.ZOSR24.PARMLIB/COLINDEL would be clearer

zFernand0:
## Proposed changes

This change prevents data loss by modifying the save process to handle data set records exceeding the logical record length.

## Release Notes

Fix #3745

Milestone: 3.3.0

Changelog: Prevented data loss by modifying the save process to handle data sets with records that exceeded the logical record length

## Types of changes

## Checklist

**General**

**Code coverage**

**Deployment**

## Further comments

Double-byte characters:
- They can be problematic because they may look like a single character. 
- We are still able to determine if they cause the line to go over the `lrecl`
- Encoding continues to work as expected and doesn't trigger the error of going over the `lrecl`
    image


zFernand0:
Since we do not want to impact performance by checking each and every line on the client, we set out to get some benchmarks on a few different approaches. The results prove that we don't have to worry too much about it! 😅 

- `iterate through bytes`: this is the most basic approach which loops through the buffer/uint8array and calculates each line length based on the presence of a new line (byte value: 10)
- `toString().split("\n")`: this is likely the most intuitive approach when given a buffer. 
  - Note: using a regex `/\r?\n/g` will significantly slow things down here.
- `await openTextDocument + lineAt().text`: is leveraging vscode APIs and it seems to be the most performant
- `await openTextDocument + lineAt().range`: is leveraging vscode APIs and it seems to comparable to the previous one.

Here are the results:
image

packages/zowe-explorer/src/trees/dataset/DatasetFSProvider.ts
```diff
@@ -589,6 +593,29 @@ export class DatasetFSProvider extends BaseProvider implements vscode.FileSystem
...
+                        safeUpload = false;
+                    }
+                });
+            }
+
+            if (!safeUpload) {
+                const optionCancel = vscode.l10n.t("Cancel");
+                const optionContinue = vscode.l10n.t("Continue");
+                const selectedOption = await Gui.showMessage(
+                    vscode.l10n.t("This upload operation may result in data loss. Are you sure you want to continue?"),
```
JWaters02: Could we put some more info in this warning? For example, what line number the problematic data is in. Because otherwise how will the user know how to fix their file so the upload will be safe (if intellisense is not active on the file)?

zFernand0: This should be resolved in 2e6bcc1b66c34db7fd9e6c94c45fcdcf0d5d5a18

---

We discussed some of the UX on this and here are some outcomes we settled on:
- Better error message with line numbers (2e6bcc1b66c34db7fd9e6c94c45fcdcf0d5d5a18)
- Prevent the save operation: (f29c5f6379bcdd461a8617391bc2def27b4b1265)
  - Meaning, don't even ask to continue, instead present the error.
- Use the error-correlator class for better visualization (2e6bcc1b66c34db7fd9e6c94c45fcdcf0d5d5a18)
- Make sure we handle both, PS and PO/PDS (f29c5f6379bcdd461a8617391bc2def27b4b1265)

packages/zowe-explorer/src/trees/dataset/DatasetFSProvider.ts
```diff
@@ -254,7 +255,7 @@ export class DatasetFSProvider extends BaseProvider implements vscode.FileSystem
         }
     }
 
-    private async fetchDataset(uri: vscode.Uri, uriInfo: UriFsInfo): Promise<PdsEntry | DsEntry> {
+    private async fetchDataset(uri: vscode.Uri, uriInfo: UriFsInfo, forceFetch: boolean = false): Promise<PdsEntry | DsEntry> {
```
traeok: (nitpick) Since we provide a default, falsy value for `forceFetch`, we should be more explicit to define it as optional without a value:


```suggestion
    private async fetchDataset(uri: vscode.Uri, uriInfo: UriFsInfo, forceFetch?: boolean): Promise<PdsEntry | DsEntry> {
```

This would be consistent with the VS Code API, which follows and recommends the same practice (optional w/o default as opposed to default, falsy value for boolean): https://github.com/microsoft/vscode/wiki/Extension-API-guidelines#undefined-is-false

packages/zowe-explorer/src/trees/dataset/DatasetFSProvider.ts
```diff
@@ -613,7 +619,8 @@ export class DatasetFSProvider extends BaseProvider implements vscode.FileSystem
      */
     public async writeFile(uri: vscode.Uri, content: Uint8Array, options: { readonly create: boolean; readonly overwrite: boolean }): Promise<void> {
         const basename = path.posix.basename(uri.path);
-        const parent = this.lookupParentDirectory(uri);
+        const parent = this.lookupParentDirectory(uri) as PdsEntry;
```
traeok: I don't think we can always assume that the parent of `uri` is a `PdsEntry` here.. The scenario that comes to mind is writing to a sequential data set under a profile - the profile would not be an instance of `PdsEntry`.

packages/zowe-explorer/src/trees/dataset/DatasetFSProvider.ts
```diff
@@ -625,14 +632,37 @@ export class DatasetFSProvider extends BaseProvider implements vscode.FileSystem
...
+                    if (document.lineAt(i).text.length > dsStats?.lrecl) {
+                        longLines.push(i + 1);
+                    }
+                }
+            } catch (err) {
+                // do nothing since we may be trying to create an entry in the FS that doesn't exist yet
+            }
+            if (longLines.length > 0) {
+                // internal error code to indicate unsafe upload
+                throw new Error(`Zowe Explorer: Unsafe upload | ${longLines.join(",")}`);
```
traeok: More of a question for clarification: Would it help to throw an ImperativeError here and tuck the long lines into additionalDetails? Not required, just wondering if that could help with error formatting.

packages/zowe-explorer/src/trees/dataset/DatasetFSProvider.ts
```diff
@@ -660,38 +691,50 @@ export class DatasetFSProvider extends BaseProvider implements vscode.FileSystem
                     entry.mtime = Date.now();
                     entry.size = content.byteLength;
                     entry.inDiffView = true;
+
+                    parent.entries.set(basename, entry);
```
traeok: (nitpick) Please correct me if I'm missing something here, but since entry is already in `parent.entries`, updating properties on `entry` should propagate without calling `parent.entries.set` (as entry is a reference to the object in `parent.entries`).

packages/zowe-explorer/src/trees/dataset/DatasetFSProvider.ts
```diff
@@ -660,38 +691,50 @@ export class DatasetFSProvider extends BaseProvider implements vscode.FileSystem
...
-            // Prompt the user with the conflict dialog
-            await this._handleConflict(uri, entry);
-            return;
+            if (err.message.includes("Zowe Explorer: Unsafe upload")) {
+                const longLines = err.message.split(" | ")[1].split(",");
+                const dataLossMsg = vscode.l10n.t("This upload operation may result in data loss.");
+                const linesToReview = longLines.length > 5 ? longLines.slice(0, 5).join(", ") + "..." : longLines.join(", ");
+                const shortMsg = vscode.l10n.t("Please review the following lines:");
+                const newErr = new Error(`${dataLossMsg} ${shortMsg} ${linesToReview}`);
+                newErr.stack = shortMsg + "\n - " + longLines.join("\n - ");
```
traeok: What are your thoughts on grouping several lines into ranges for readability?

Currently, if I have an upload with several lines over the LRECL, this is what would be shown:

```
Please review the following lines:
1
2
3
4
5
6
7
8
9
10
11
12
16
...
```

For around 10 lines or so, this is fairly straightforward, but what if the user has >100 lines over LRECL? For example, the following output could be reduced to: 

```
Please review the following lines:
1-12
16
...
```

Which (in my opinion) is a lot easier to digest and resolve. But folks can chime in if they disagree 😋

.eslintrc.yaml
```diff
@@ -31,6 +31,7 @@ overrides:
       "@typescript-eslint/explicit-function-return-type": off
       "@typescript-eslint/unbound-method": off
       zowe-explorer/no-floating-promises: warn
+      curly: off
```
traeok, line 34: Was this lint rule changed in this PR? Just want to make sure I understand the purpose of this change ^^

zFernand0: yeah, mostly to avoid a lot of one-line `if (expression) return x;` in test 🙏

packages/zowe-explorer/src/trees/dataset/DatasetFSProvider.ts
zFernand0: I agree with that premise. However... 😓 
I wasn't seeing that behavior during unit testing (once they are executed in isolation). That indicates that without setting the entry (again), the original entry in the diff could be left untouched (even after the user has made changes to it)

traeok: > I wasn't seeing that behavior during unit testing (once they are executed in isolation)

I think that could be an issue with how the test objects are initialized in that file, and since this is a nitpick I don't want to suggest making changes for this one since it's probably going to introduce scope creep. Thanks for the context though 😋

.eslintrc.yaml
traeok: Ah ok, personally I'm okay with the change but we might want to mention it in ZE standup to get the team's thoughts ^^

packages/zowe-explorer/src/trees/dataset/DatasetFSProvider.ts
JWaters02: Seems fair to me.

packages/zowe-explorer/src/trees/dataset/DatasetFSProvider.ts
zFernand0: resolved in aee7dc8f2a46b18f13369b524c55903c7bb3253d

packages/zowe-explorer/src/trees/dataset/DatasetFSProvider.ts
zFernand0: resolved in aee7dc8f2a46b18f13369b524c55903c7bb3253d

packages/zowe-explorer/src/trees/dataset/DatasetFSProvider.ts
zFernand0: resolved in aee7dc8f2a46b18f13369b524c55903c7bb3253d

packages/zowe-explorer/src/trees/dataset/DatasetFSProvider.ts
zFernand0: Grouped ranges:
image

resolved in 0da6742f692efc8185939b1178486ccfc9e5d847

packages/zowe-explorer/src/trees/dataset/DatasetFSProvider.ts
zFernand0: I don't mean to be difficult, I just want to make sure we are on the same page on this 😓 

@traeok, question...

Generally speaking...
Should we have `jest.restoreAllMocks();` in the `beforeEach` of the main/overall `describe`?

From my recollection, doing this will force each `it` to run in "isolation". Meaning, no getting away with mock-results from previous tests making new tests pass. (That's confusing, sorry) 🫤❓ 

If that's indeed what we want, then I believe the we need to call `parent.entries.set` 😰

traeok: I think in the context of the tests, this call is necessary because the entry doesn't persist between test cases - as you've mentioned, the tests are running in isolation. Some of the mock objects are pre-defined at the top of the file, so I am wondering if there's some pollution going on between the test cases..

But in a real-world scenario with the FSP, the entry is a reference to the object stored in the `entries` map - so that when entry is changed, the changes are reflected for that entry in the map without needing to do another `set`.

Again, just nitpicking as this won't impact the behavior in production - I just wanted to see if we could avoid it 😋

zFernand0: That makes more sense, 🤔 

I'm not a fan of making changes only because of tests 😓 
So, I'm wondering if it should be removed 🤔 
will leave a TODO item 😓

zFernand0: I take it back... it was due to a possible forceFetch (which was added in this PR) 
so, if we do end up fetching... we are now updating the entry for follow-up operations 🙏 

Resolved in 11488f7f5

.eslintrc.yaml
JWaters02: Might need this discussion as I have a few changes of my own in .eslintrc.yaml file (that I permanently keep in a local stash and a cloud patch file) - ever since I first started making contributions here.

JWaters02 (commented):
Great feature, seems to work really well for me! I made a really big file (10k lines with 1k lines over lrecl) and it errored instantly with good info. I like that you can pick a place on MF or locally to save the file to or revert the changes.

However I do think that maybe the revert button should have some popup warning like "Reverting this file will lose your changes. Are you sure you want to continue?" and then the same error again maybe. Cos I accidentally clicked this (even though I'm testing the feature) and I can imagine someone else might do this...

Also one small thing, in troubleshooting window, "code: Not available". Should this be here? Should it show a code that could be referenced to in the docs later on? 

image

zFernand0:
Hey Joshua, 
Thanks for taking a look 🙏 

>  I do think that maybe the revert button should have some popup warning like "Reverting this file will lose your changes. Are you sure you want to continue?" and then either confirm or cancel.

I believe we could do this, but I'm not a fan of too many prompts.
Also, people should be able to undo (`Ctrl/Cmd + Z`) the _revert_ operation 😅 

> "code: Not available". Should this be here? Should it show a code that could be referenced to in the docs later on?

I like the idea of having some sort of code that people can reference in the docs 🙏 

Since we don't go out to the mainframe and let the 500 error happen (from z/OSMF, FTP, ...), I believe we may have to come up with error codes ourselves.

We could use something like this:
`EDC5003I Truncation of a record occurred during an I/O operation.`

Would it be ok to handle as a separate PR? 😅 
Trying to focus on other PRs before code-freeze 😓

JWaters02 (approved):

zFernand0:
Hey Joshua, 
I'm re-requesting your review because I've made some nuanced changes in how we handle PS data sets.
Please let me know if you've experienced any data loss in previous commits :pray:
Also, I will be looking into the revert confirmation message later tonight or tomorrow morning! 🙏

packages/zowe-explorer/CHANGELOG.md
```diff
@@ -32,6 +32,7 @@ All notable changes to the "vscode-extension-for-zowe" extension will be documen
 - Fixed an issue where cancelling multiple commands in rapid succession could render the terminal unusable. [#3640](https://github.com/zowe/zowe-explorer-vscode/issues/3640)
 - Fixed an issue where a `401 Unauthorized` error is thrown by the FileSystemProvider implementations when the user cancels or closes an authentication prompt. Now, the filesystem throws a more explicit `AuthCancelledError` in this scenario. [#3662](https://github.com/zowe/zowe-explorer-vscode/issues/3662)
 - Fixed an issue where renaming or deleting the team configuration file caused an unhandled exception in Zowe Explorer, causing old, non-existent profiles to remain in the tree views. [#3772](https://github.com/zowe/zowe-explorer-vscode/issues/3772)
+- Fixed a regression with the `AuthUtils.syncSessionNode` function that caused Zowe Explorer logic to stop unexpectedly if a session node did not yet have a rich hover tooltip. [#3795](https://github.com/zowe/zowe-explorer-vscode/pull/3795)
 - Prevented data loss by modifying the save process to handle data sets with records that exceeded the logical record length. [#3791](https://github.com/zowe/zowe-explorer-vscode/issues/3791)
 
 ### Bug fixes
```
traeok: Seeing a duplicate `Bug fixes` header here:


```suggestion
```

packages/zowe-explorer/src/trees/dataset/DatasetFSProvider.ts
```diff
@@ -710,13 +715,39 @@ export class DatasetFSProvider extends BaseProvider implements vscode.FileSystem
...
+                    }
+                }
+                for (const lines of groupedLines) {
+                    let lineRange = "";
+                    if (lines.lines.length > 1) {
+                        lineRange = `Lines: ${lines.lines[0]}-${lines.lines[lines.lines.length - 1]}`;
+                    } else {
+                        lineRange = `Line: ${lines.lines[0]}`;
+                    }
+                    newErr.stack += `\n${lineRange}\n${lines.text}\n`;
```
traeok, line 749: thanks for adding this 🙏

traeok (approved):
Left a comment about the changelog, will re-approve once resolved. Thanks Fernando

JWaters02 (approved):

JillieBeanSim:
@zFernand0 yes I am experiencing it on main as well, also had shared recording via slack

https://github.com/user-attachments/assets/d0c4280a-5cb2-4f40-a2f3-f5f751c1844d

PR #3793, opened 2025-08-04
Added VSCode setting to use default tso profile

pujal0909:
## Release Notes

Milestone:

Changelog:

## Types of changes

## Checklist

**General**

**Code coverage**

**Deployment**

### DIFF
--- a/packages/zowe-explorer/src/commands/TsoCommandHandler.ts
+++ b/packages/zowe-explorer/src/commands/TsoCommandHandler.ts
@@ -20,6 +20,7 @@ import { Definitions } from "../configuration/Definitions";
 import { ZowePersistentFilters } from "../tools/ZowePersistentFilters";
 import { SettingsConfig } from "../configuration/SettingsConfig";
 import { Constants } from "../configuration/Constants";
+import { Profiles } from "../configuration/Profiles";
 
 /**
  * Provides a class that manages submitting a TSO command on the server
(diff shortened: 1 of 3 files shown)


packages/zowe-explorer/__tests__/__integration__/bdd/features/dialogs/IssueCommandDefaultProfile.feature
```diff
@@ -0,0 +1,13 @@
...
+    Given a user who is looking at the Zowe Explorer tree views
+    And the default tso profile is set to true
+    When a user selects <opt> from the command palette
+    Then a quick pick appears to select a profile
+    When a user selects a profile
+    Then a user can enter in status as the command and submit it
+
+    Examples:
+      | opt |
+      | Issue TSO Command |
```
traeok: Since we only have one example in this scenario, we could remove the scenario outline in favor of a standard "Scenario":


```suggestion
Scenario: User wants to issue a tso command
    Given a user who is looking at the Zowe Explorer tree views
    And the default tso profile is set to true
    When a user selects Issue TSO Command from the command palette
    Then a quick pick appears to select a profile
    When a user selects a profile
    Then a user can enter in status as the command and submit it
```

packages/zowe-explorer/__tests__/__unit__/commands/ZoweCommandProvider.unit.test.ts
```diff
@@ -60,6 +60,7 @@ describe("ZoweCommandProvider Unit Tests", () => {
             jest.spyOn(SharedContext, "isSessionNotFav").mockReturnValue(true);
         });
         it("should check current profile and perform the case when status is 'active'", async () => {
+            jest.spyOn(SettingsConfig, "getDirectValue").mockReturnValue(true);
```
traeok: (nitpick) if we only check the value once, we should use `mockReturnValueOnce` to prevent this mock from carrying over to other tests in this describe block. Or, we can keep the `mockReturnValue` call as-is and use `mockRestore` on the mock at the end of the test.

pujal0909: good catch thanks!

traeok (approved):
LGTM, thanks Pujal for the enhancement!

packages/zowe-explorer/src/commands/TsoCommandHandler.ts
```diff
@@ -160,15 +160,21 @@ export class TsoCommandHandler extends ZoweCommandProvider {
...
+            const profiles = profileInfo.getAllProfiles("tso");
+            if (profiles.length > 0) {
+                tsoProfile = await this.selectServiceProfile(profiles.map((p) => imperative.ProfileInfo.profAttrsToProfLoaded(p)));
             }
         }
+        if (tsoProfile != null) {
+            const prof = profileInfo.mergeArgsForProfile(tsoProfile.profile as imperative.IProfAttrs);
+            iStartTso.forEach((p) => (tsoProfile.profile[p] = prof.knownArgs.find((a) => a.argName === p)?.argValue));
+        }
         if (tsoProfile) {
```
t1m0thyj, line 178: (nit) These `if` branches seem like they could be combined into one

packages/zowe-explorer/src/commands/ZoweCommandProvider.ts
```diff
@@ -164,7 +164,6 @@ export abstract class ZoweCommandProvider {
                 return;
             }
             const profile = this.profileInstance.allProfiles.find((tempProfile) => tempProfile.name === sesName);
-            await this.profileInstance.checkCurrentProfile(profile);
```
t1m0thyj, line 167: Is removing this intentional? It seems like we're losing profile validation here.

t1m0thyj: Discussed offline, this should be fine - validation was happening twice before, so this removes duplicate logic 👍

PR #3802, opened 2025-08-08
Update datasets.ts

NageshVG:
Updating plx as extension for PLX datasets.

## Proposed changes
Updating dataset with name PLX use plx as extension.

## Types of changes

- [ x] Bug fix (non-breaking change which fixes an issue)


adam-wolfe:
We plan to remove the .PLX extension from the PL/I filetype assignment. Not planning to add support for PLX at this time.

PR #3809, opened 2025-08-14
fix: null-like access in `AuthUtils.handleProfileAuthOnError`

traeok:
## Proposed changes

Issue introduced from changes in #3673 (surfaced by security scanning tool), where `profile` could be used even if it was `undefined`. This is because `profile` was optional, and the new condition added to this check caused the `profile` parameter to be accessed as long as that condition was satisfied. However, we always provide a profile when this function is called in ZE, so this PR refactors the function to make `profile` required.

## Release Notes

Milestone: 3.3.0?

Changelog: No changelog since code change is in ZE only and has no user-facing impact

## Types of changes

## Checklist

**General**

**Code coverage**

**Deployment**


t1m0thyj (approved):
LGTM, thanks @traeok!

t1m0thyj (approved):

zFernand0 (approved):
LGTM! 😋 
---
We do seem to pass profile every time we call this function in ZE 🙏 

Edge case, but not a real problem...
There could be a very rare case where the cache itself is not defined by the time we want to get info from the URI, but in that case I think we have error handling that throws an error if it can't find the profile, instead of letting it bubble up with a `unable to read property of undefined (reading 'name')`.

Issue #2536, opened 2023-10-30
Sorting PDS members in the descending order causes unwanted behavior


adam-wolfe:
**Describe the bug**

When PDS members are sorted by date modified or user ID in the descending order, when the PDS is collapsed and re-expanded, the order of items with identical search values (i.e., same date last modified or same user ID) reverses (see screen recording)

**To Reproduce**

1. Sort PDS members by user ID or date modified in descending order
2. Collapse the PDS
3. Expand the PDS
4. Notice that the order of items with the same user id or date modified has reversed

**Expected behavior**

The order should not change when the PDS is collapsed and re-expanded.

**Screenshots**
Sorry for the huge gif.
![SortDescendingIssue](https://github.com/zowe/vscode-extension-for-zowe/assets/122040687/8f46cd1e-93e8-430a-8f47-f8e17f9adced)

**Desktop (please complete the following information):**

- OS: Windows 10
- Zowe Explorer Version: 
- (Optional) Zowe CLI Version:
- (Optional) Are you using Secure Credential Store?

**Additional context**

This does not happen when the order is set to ascending.

zFernand0:
Is this due to the fact that we do not sort down to the millisecond in which something was created?
If we want, we could introduce a second sort criteria (e.g. `name` -when sorting by date, `date`-when sorting by name/user)

traeok:
> Is this due to the fact that we do not sort down to the millisecond in which something was created? If we want, we could introduce a second sort criteria (e.g. `name` -when sorting by date, `date`-when sorting by name/user)

Technically we do sort by the smallest unit, as the date comparison was `dateA < dateB`. I've decided to adjust this slightly by using Day.js to compare the dates down to the very second, as it is the smallest unit of time that z/OSMF provides from their API responses.

But I agree, I think a user would expect it to fall back to sorting by name when the values are equal. Since it didn't take long to clean up the sort function, and the updated function is easier to read, I decided to address that in #2538

traeok:
Fixed in #2538

PR #3843, opened 2025-09-05
feat: Data set and USS download commands

JWaters02:
## Proposed changes

Adds 5 new commands:
1. Download member
2. Download data set
3. Download all members
4. Download USS file
5. Download USS directory

Didn't add command for download all data sets even though its in the API because of [#2536](https://github.com/zowe/zowe-cli/issues/2536) but happy to add it if something thinks it would be useful to have it in ZE... 

https://github.com/user-attachments/assets/532628cd-5770-4f05-b8f7-0362c5bce2fa

## Release Notes

Milestone: 3.4.0

Changelog:

## Types of changes

## Checklist

**General**

**Code coverage**

**Deployment**

## Further comments

Please could someone (ideally someone on zowe cli team who worked on downloads api) have a look at the options I am passing into downloads and if I am missing any/getting some wrong etc. Thanks


packages/zowe-explorer/src/trees/uss/USSActions.ts
```diff
@@ -312,6 +315,253 @@ export class USSActions {
...
+                    stageName: 0, // TaskStage.IN_PROGRESS
+                };
+
+                const options: zosfiles.IDownloadOptions = {
+                    directory: downloadOptions.generateDirectory
+                        ? path.join(downloadOptions.selectedPath.fsPath, node.fullPath)
+                        : downloadOptions.selectedPath.fsPath,
+                    overwrite: downloadOptions.overwrite,
+                    binary: downloadOptions.encoding?.kind === "binary",
+                    encoding: downloadOptions.encoding?.kind === "other" ? downloadOptions.encoding.codepage : undefined,
```
traeok: I think for downloading directories, you'll want to fall back to the encoding specified on the user's profile when the codepage is not provided. 

```suggestion
                    encoding: downloadOptions.encoding?.kind === "other" ? downloadOptions.encoding.codepage : profile.profile?.encoding,
```

packages/zowe-explorer/src/trees/uss/USSActions.ts
```diff
@@ -312,6 +315,253 @@ export class USSActions {
...
+                title: vscode.l10n.t("Downloading USS file..."),
+                cancellable: true,
+            },
+            async () => {
+                const options: zosfiles.IDownloadSingleOptions = {
+                    file: downloadOptions.generateDirectory
+                        ? path.join(downloadOptions.selectedPath.fsPath, node.fullPath)
+                        : path.join(downloadOptions.selectedPath.fsPath, path.basename(node.fullPath)),
+                    binary: downloadOptions.encoding?.kind === "binary",
+                    encoding: downloadOptions.encoding?.kind === "other" ? downloadOptions.encoding.codepage : undefined,
```
traeok: Same suggestion as for downloading USS directories - we should download the file with the profile encoding if a codepage is not provided.

JWaters02:
Should we have USS directory download allow user to choose the default encoding method for USS files being downloaded? I see in the API that if none is provided, each USS file will be downloaded with an estimate if it has tags. This might be best as if user is downloading big directories of files, USS files might be using all sorts of encoding. 

On the other hand, I suppose they could choose encoding, but the prompt could have an extra selection for "don't use any encoding" for directories so the user could both choose an encoding for all files or choose not to have any chosen by them/their profile. But this would be tricky to add as the prompt for encoding is used by upload and other download functions. Plus the current prompt for encoding does not support nodes being directories right now (I'd need to figure out something...).

Any thoughts?

packages/zowe-explorer/l10n/bundle.l10n.json
```diff
@@ -341,6 +305,42 @@
   "initializeUSSFavorites.error.buttonRemove": "initializeUSSFavorites.error.buttonRemove",
   "File does not exist. It may have been deleted.": "File does not exist. It may have been deleted.",
   "Pulling from Mainframe...": "Pulling from Mainframe...",
+  "The 'move' function is not implemented for this USS API.": "The 'move' function is not implemented for this USS API.",
```
anaxceron:
```suggestion
  "The 'Move' function is not implemented for this USS API.": "The 'move' function is not implemented for this USS API.",
```

packages/zowe-explorer/l10n/poeditor.json
```diff
@@ -741,6 +732,15 @@
   "initializeUSSFavorites.error.buttonRemove": "",
   "File does not exist. It may have been deleted.": "",
   "Pulling from Mainframe...": "",
+  "The 'move' function is not implemented for this USS API.": "",
```
anaxceron:
```suggestion
  "The 'Move' function is not implemented for this USS API.": "",
```

packages/zowe-explorer/l10n/bundle.l10n.json
```diff
@@ -341,6 +305,42 @@
...
+  "Failed to move {0}/File path": {
+    "message": "Failed to move {0}",
+    "comment": [
+      "File path"
+    ]
+  },
+  "Profile does not exist for this file.": "Profile does not exist for this file.",
```
anaxceron: Should we say "**A** profile ... "?

packages/zowe-explorer/src/trees/dataset/DatasetActions.ts
```diff
@@ -611,6 +613,333 @@ export class DatasetActions {
...
+        const children = await node.getChildren();
+        if (!children || children.length === 0) {
+            Gui.showMessage(vscode.l10n.t("The selected data set has no members to download."));
+            return;
+        }
+
+        if (children.length > Constants.MIN_WARN_DOWNLOAD_FILES) {
+            const proceed = await Gui.showMessage(
+                vscode.l10n.t(
+                    "This data set has {0} members. Downloading a large number of files may take a long time. Do you want to continue?",
```
anaxceron:
```suggestion
                    "This data set has {0} members. Downloading a large number of files can take a long time. Do you want to continue?",
```

davenice:
I haven't tried the UX myself yet, but from the video I wasn't quite sure what I expected to happen about choosing encodings. The wording suggests I can choose the encoding of every file somehow ... but I can't quite envisage the GUI for this!

A couple of suggestions on the notifications....

I feel like when I get a few notifications that are quite generic like "File downloaded" or "New copy completed" and I find myself wondering if that was a notification I got just now, or one I got a while back that should have cleared. My stance for CICS has been to try and be a bit more descriptive with what the notification is e.g. "New copy completed on MYPROG1" or similar. We could be more descriptive in the notifications here, although I'm not set on that.

We could also provide a link in the notification to the file, or the directory we downloaded to. Again I don't think this is a must-have if there is a time pressure.

JWaters02:
Thanks a lot for the feedback dave, I really appreciate it🙂 

The video is actually super out of date now I really should post a new one for better feedback, apologies!

> I haven't tried the UX myself yet, but from the video I wasn't quite sure what I expected to happen about choosing encodings. The wording suggests I can choose the encoding of every file somehow ... but I can't quite envisage the GUI for this!

Instead of putting encoding options right into the download options quick pick, I have an option for "Choose Encoding" and if it is checked, it will also enter the existing choose encoding quick pick used for upload/open with encoding.

For downloading members of a PDS, this encoding will be applied to all members, which does make sense given they all follow the same characteristics - but for a USS directory, files in a directory can have all kinds of encodings so this is much tricker. So for downloading USS directories, I have the same "Choose Encoding" option but the resulting encoding selection quick pick is almost the same as the existing one but with an additional "Auto-detect file encoding" option which essentialls selects no encoding and relies on the encoding detection in zowe-cli which uses USS file tags and stuff.

> A couple of suggestions on the notifications....
>
> I feel like when I get a few notifications that are quite generic like "File downloaded" or "New copy completed" and I find myself wondering if that was a notification I got just now, or one I got a while back that should have cleared. My stance for CICS has been to try and be a bit more descriptive with what the notification is e.g. "New copy completed on MYPROG1" or similar. We could be more descriptive in the notifications here, although I'm not set on that.

This is a good point, I think I will implement that. Currently, I have it only saying what type of download has occured but I definately should include the path of the thing downloaded and/or resulting file path too.

What I do have now though is when a warning or error occurs, the user can click "View Details" and it will open the Troubleshooting webview with the command response from zowe-cli which contains a lot more details about the download. These types of details can't be shown in notifcations because of being too verbose. A warning can occur, for example, if overwrite is off and a file was skipped because of the overwrite and in the details it prints which files were skipped and not skipped (same as in zowe-cli (whch I have an open PR for to add overwrite support to all commands except USS directories (which it already exists for), so you can currently only test this with those)).

> We could also provide a link in the notification to the file, or the directory we downloaded to. Again I don't think this is a must-have if there is a time pressure.

A link is a good idea!

traeok:
> I added a new test file USSUtils.unit.test.ts but when I run it, I get this error:
> 
> <snip>
> Unable to import package.json to get app Options : TypeError [ERR_INVALID_ARG_TYPE]: The "paths[0]" argument must be of 
> type string. Received undefined

Still debugging this, it's a strange error with the Jest JUnit reporting where it fails to locate the package JSON. Plan to read over some of the changes to get a better grasp of why this is happening

traeok:
> Only vague thought I had is that I tried to use MockedProperty as much as possible instead of Object.defineProperty and maybe that usage has caused some weird side effect?

Skimming through the tests, I do see a few `new MockedProperty` calls that are not calling the `Symbol.dispose` function, that could have an impact depending on what's being mocked

packages/zowe-explorer/src/trees/uss/UssFSProvider.ts
```diff
@@ -433,10 +433,17 @@ export class UssFSProvider extends BaseProvider implements vscode.FileSystemProv
...
-        await this.autoDetectEncoding(file);
+        try {
+            const file = this._lookupAsFile(uri) as UssFile;
+            await this.autoDetectEncoding(file);
 
-        return file.encoding;
+            return file.encoding;
+        } catch (error) {
+            if (error instanceof vscode.FileSystemError && error.code === "FileIsADirectory") {
+                throw new Error(vscode.l10n.t("Cannot fetch encoding for directories. Encoding options are only available for files."));
```
t1m0thyj: Curious why this extra error check is needed, since we were already returning a clear error code (`FileIsADirectory`) previously.

JWaters02: No idea, it's been long enough I don't remember how most of this code works 😂  probably some unthoughtful jank. I wanted to make this PR draft again because it still needs more work to polish it off (bits n bobs) but the button to convert draft back to PR is not there for me. I plan to look at this sometime this weekend or maybe weekend after that, and try to refactor the bad bits with fresh set of eyes.

davenice: Fernando had the magic button and has moved it back to draft :-)

JWaters02:
> Skimming through the tests, I do see a few `new MockedProperty` calls that are not calling the `Symbol.dispose` function, that could have an impact depending on what's being mocked

Hey @traeok so I tried a few things:
1. Adding proper disposals to MockedProperty (in all places in test files were I was using it). Had no change
2. Replacing MockedProperty with Object.defineProperty. Had no change. So it surely can't be anything to do with MockedProperty? (a thought at the time of writing)
3. Running `pnpm test:unit -- --testPathPattern="USSUtils.unit.test"` did not work, which rules out some dodginess with npx and jest.
4. Made new test file from scratch with simple assertions tests worked, but as soon as I added the global mocks function (without even any tests using it), the error re-appeared on that file.
5. Made new test file and this time did not include any global mocks function, just doing all mocks in each test suite. I added a few tests at a time, all none without doing any MockedProperty/Object.defineProperty and all the tests were running WITHOUT the error from appearing. I got to the point where I had rewritten 23 of 23 test cases in my temp file without using global mocks and converting property mocks to spy ones in all but one test where I kept the mocked property on the constants mock (as I realised many property mocks didn't need to be).
6. So then I made a second test file which took my learnings and rewrote the original file again with global mocks to remove code duplication, but this time without mocking on built in node libraries and only mocking on stuff that really needed it. I also consulted chatgpt with original file and the first test file to ask it what it thought about the original file getting the error but the new one not, and its response was something about mocked property being instantiated before jest's mocking initialisation is done. I have no idea how any of that works so I asked it for a workaround for setting up the mocks properly and it seemed to work but again I don't understand this fact so it could be hallucinating and I wouldn't know. 

So off the back of points 5 and 6 I now replaced the original USSUtils with my temp one that is working so you can see what I eventually ended up with. Now I remember that when I was oriignally making the test file it was working up to a certain point when I started using the mocked properties inglobal mocks so thats also another factor I forgor about because it's been so long since I touched this,

packages/zowe-explorer/src/trees/uss/UssFSProvider.ts
JWaters02: thanks mr fernando and mr dave!

looked at this again yes I think this was just confusion on my part, good spot

packages/zowe-explorer/l10n/bundle.l10n.json
JWaters02: refers to a function name which is lowercase so won't make it upper

packages/zowe-explorer/l10n/poeditor.json
JWaters02: Same as above

packages/zowe-explorer/l10n/bundle.l10n.json
JWaters02: Yess

Except looking at it now this is from code unrelated to my changes, so probably not correct thing to do to change it in here? But I will anyway because its such a tiny thing shhh

packages/zowe-explorer-api/src/extend/MainframeInteraction.ts
```diff
@@ -235,6 +250,16 @@ export namespace MainframeInteraction {
          */
         getContents(dataSetName: string, options?: zosfiles.IDownloadSingleOptions): Promise<zosfiles.IZosFilesResponse>;
 
+        /**
+         * Retrieve all members of a partitioned data set and save them to a directory.
+         *
+         * @param {string} dataSetName
+         * @param {string} directoryPath
```
awharn: This parameter doesn't seem to exist on `downloadAllMembers`

packages/zowe-explorer-api/CHANGELOG.md
```diff
@@ -12,6 +12,8 @@ All notable changes to the "zowe-explorer-api" extension will be documented in t
 - Added support to delete VSAM data sets for z/OSMF type profiles. [#3824](https://github.com/zowe/zowe-explorer-vscode/issues/3824)
 - Added support for loading credential manager options from the `imperative.json` file. Added a `credentialManagerOptions` object in the JSON object in `imperative.json` to specify options for the current credential manager. [#3935](https://github.com/zowe/zowe-explorer-vscode/pull/3935)
 - Added support for custom persistence levels for Windows (persist option) to support the credential manager in less permissive environments. For more information on how to configure this option, see the "Troubleshooting Zowe CLI credentials" on [Zowe Docs](https://docs.zowe.org/stable/troubleshoot/cli/troubleshoot-cli-credentials/#secrets-sdk-persistence-level-for-windows). [#3935](https://github.com/zowe/zowe-explorer-vscode/pull/3935)
+- Added support for `downloadDirectory` and `downloadAllMembers` to the `MainframeInteraction` namespace. [#3843](https://github.com/zowe/zowe-explorer-vscode/pull/3843)
```
anaxceron: We should describe what `downloadDirectory` and `downloadAllMembers` are

packages/zowe-explorer-api/CHANGELOG.md
```diff
@@ -12,6 +12,8 @@ All notable changes to the "zowe-explorer-api" extension will be documented in t
...
+- Added support for calling `UssApi.fileList` with `zosfiles.IUSSListOptions`. [#3843](https://github.com/zowe/zowe-explorer-vscode/pull/3843)
```
anaxceron: We should describe what `UssApi.fileList` with `zosfiles.IUSSListOptions` are

packages/zowe-explorer-ftp-extension/CHANGELOG.md
```diff
@@ -5,6 +5,7 @@ All notable changes to the "zowe-explorer-ftp-extension" extension will be docum
 ### New features and enhancements
 
 - Updated minimum VS Code version from 1.79 to 1.90. We are dropping support for VS Code releases that bundle versions of Node.js no longer receiving security updates. [#3684](https://github.com/zowe/zowe-explorer-vscode/issues/3684)
+- Added unimplemented functions for `downloadDirectory` and `downloadAllMembers`, that may be implemented in the future. [#3843](https://github.com/zowe/zowe-explorer-vscode/pull/3843)
```
anaxceron: We should describe what `downloadDirectory` and `downloadAllMembers` are

packages/zowe-explorer/CHANGELOG.md
```diff
@@ -16,6 +16,7 @@ All notable changes to the "vscode-extension-for-zowe" extension will be documen
 - Added support for loading credential manager options from the imperative.json file. Add a `credentialManagerOptions` object in the JSON object in `imperative.json` to specify options for the current credential manager. [#3935](https://github.com/zowe/zowe-explorer-vscode/pull/3935)
 - Added support for custom persistence levels for Windows (persist option) to support the credential manager in less permissive environments. For more information on how to configure this option, refer to the ["Troubleshooting Zowe CLI credentials"](https://docs.zowe.org/stable/troubleshoot/cli/troubleshoot-cli-credentials/#secrets-sdk-persistence-level-for-windows) page on Zowe Docs. [#3935](https://github.com/zowe/zowe-explorer-vscode/pull/3935)
 - Updated minimum VS Code version from 1.79 to 1.90. We are dropping support for VS Code releases that bundle versions of Node.js no longer receiving security updates. [#3684](https://github.com/zowe/zowe-explorer-vscode/issues/3684)
+- Added support for downloading data sets, data set members, USS files and USS directories. [#3843](https://github.com/zowe/zowe-explorer-vscode/pull/3843)
```
```suggestion
- Added support for downloading data sets, data set members, USS files, and USS directories. [#3843](https://github.com/zowe/zowe-explorer-vscode/pull/3843)
```

packages/zowe-explorer-ftp-extension/CHANGELOG.md
JWaters02: If I explain what they are, wouldn't it be confusing to the end user if the function is actually implemented or not? I only added this changelog entry to not fail the changelog check, so not sure what to do.

packages/zowe-explorer-api/src/extend/MainframeInteraction.ts
traeok, line 1: The new functions in this file must be marked as optional to prevent breaking changes for extenders. Otherwise extenders will get an error when building with these API changes.

packages/zowe-explorer-api/src/extend/MainframeInteraction.ts
```diff
@@ -119,9 +119,24 @@ export namespace MainframeInteraction {
...
 
+        /**
+         * Download a USS directory to the local file system.
+         *
+         * @param {string} ussDirectoryPath The path of the USS directory to download
+         * @param {zosfiles.IDownloadOptions} fileOptions Download options including local directory path
+         * @param {zosfiles.IUSSListOptions} listOptions Options for listing files in USS
+         * @returns {Promise<zosfiles.IZosFilesResponse>}
+         */
+        downloadDirectory(
```
traeok:
```suggestion
        downloadDirectory?(
```

packages/zowe-explorer-api/src/extend/MainframeInteraction.ts
```diff
@@ -235,6 +250,15 @@ export namespace MainframeInteraction {
...
         getContents(dataSetName: string, options?: zosfiles.IDownloadSingleOptions): Promise<zosfiles.IZosFilesResponse>;
 
+        /**
+         * Retrieve all members of a partitioned data set and save them to a directory.
+         *
+         * @param {string} dataSetName
+         * @param {zosfiles.IDownloadOptions} [options]
+         * @returns {Promise<zosfiles.IZosFilesResponse>}
+         */
+        downloadAllMembers(dataSetName: string, options?: zosfiles.IDownloadOptions): Promise<zosfiles.IZosFilesResponse>;
```
traeok:
```suggestion
        downloadAllMembers?(dataSetName: string, options?: zosfiles.IDownloadOptions): Promise<zosfiles.IZosFilesResponse>;
```

packages/zowe-explorer-ftp-extension/src/ZoweExplorerFtpUssApi.ts
```diff
@@ -101,6 +101,23 @@ export class FtpUssApi extends AbstractFtpApi implements MainframeInteraction.IU
...
+    /**
+     * Download a USS directory to the local file system.
+     * Currently not supported by FTP extension.
+     *
+     * @param ussDirectoryPath The path of the USS directory to download
+     * @param fileOptions Download options including local directory path
+     * @param listOptions Options for listing files in USS
+     * @returns A file response with the results of the download operation.
+     */
+    public downloadDirectory(
```
traeok: This implementation stub would not be necessary if the API was marked optional.

packages/zowe-explorer-ftp-extension/CHANGELOG.md
traeok: Following up on my comments, we could remove this changelog (and the changes to FTP extension altogether) if we marked the new API functions as optional on the interface and removed the newly-added stubs.

packages/zowe-explorer-ftp-extension/src/ZoweExplorerFtpMvsApi.ts
```diff
@@ -371,6 +371,10 @@ export class FtpMvsApi extends AbstractFtpApi implements MainframeInteraction.IM
         }
     }
 
+    public downloadAllMembers(_dataSetName: string, _options?: zosfiles.IDownloadOptions): Promise<zosfiles.IZosFilesResponse> {
```
traeok: This implementation stub can also be removed once the function declaration on the `MainframeInteraction.IMvs` interface is marked optional

packages/zowe-explorer/src/trees/dataset/DatasetActions.ts
```diff
@@ -618,6 +620,446 @@ export class DatasetActions {
...
+            overwrite,
+            generateDirectory,
+            preserveCase: preserveOriginalLetterCase,
+            encoding,
+            selectedPath,
+            overrideExtension,
+            fileExtension,
+        } = dataSetDownloadOptions;
+
+        await DatasetActions.executeDownloadWithProgress(
```
traeok, line 869: Once the APIs are marked optional, we'll need to check if the API exists before starting the download. The same applies for the USS logic in `USSActions` that handles directory downloads.

packages/zowe-explorer-api/src/extend/MainframeInteraction.ts
JWaters02: Okay thanks a lot trae this makes sense. Will make things more straight forward as well.

packages/zowe-explorer/src/trees/dataset/DatasetActions.ts
JWaters02: @traeok do you mean something like this?
```
mvsApi = ZoweExplorerApiRegister.getMvsApi(profile) 
if !mvsApi.downloadAllMembers then error "this API not supported by this profile type"
```

traeok: Yes, sorry for the late reply.

packages/zowe-explorer-api/src/extend/MainframeInteraction.ts
```diff
@@ -119,9 +119,24 @@ export namespace MainframeInteraction {
          *
          * @param {string} ussFilePath
          * @param {zosfiles.IDownloadOptions} options
+         * @returns {Promise<zosfiles.IZosFilesResponse>}
```
zFernand0, line 122: Thanks for adding the return value to the TSDoc. 🙏 
It would be nice if all functions were consistently returning the proper `Promise<...>` value 😅 
This is not a request for changes though 😋

packages/zowe-explorer/__tests__/__unit__/trees/dataset/DatasetUtils.unit.test.ts
```diff
@@ -130,3 +130,120 @@ describe("Dataset utils unit tests - function extractDataSetAndMember", () => {
...
+        const result = await DatasetUtils.getExtensionMap(mockNode, false);
+
+        expect(result["jcl"]).toBe("jcl");
+        expect(result["member"]).toBe("jcl");
+    });
+
+    it("should apply override extension with preserveCase", async () => {
+        const mockNode = createMockNode("TEST.PDS", [{ label: "MEMBER1" }, { label: "MEMBER2" }]);
+
+        const result = await DatasetUtils.getExtensionMap(mockNode, true, "xml");
```
zFernand0, line 243: For the `preserveCase` tests, should we have a random value with a number up upper/lower -case characters?
For example; `.f@K3`

packages/zowe-explorer/src/trees/dataset/DatasetInit.ts
```diff
@@ -235,6 +235,20 @@ export class DatasetInit {
...
+            )
+        );
+
+        context.subscriptions.push(
+            vscode.commands.registerCommand("zowe.ds.downloadMember", async (node: IZoweDatasetTreeNode) => DatasetActions.downloadMember(node))
+        );
+
+        context.subscriptions.push(
+            vscode.commands.registerCommand("zowe.ds.downloadDataSet", async (node: IZoweDatasetTreeNode) => DatasetActions.downloadDataSet(node))
+        );
```
zFernand0, line 254: dumb question: 🤚 

Should we `await` these calls?

packages/zowe-explorer/src/trees/dataset/DatasetUtils.ts
```diff
@@ -158,4 +158,58 @@ export class DatasetUtils {
...
+
+    /**
+     * Gets a map of file extensions for all members of a PDS to be used for IDownloadOptions.
+     */
+    public static async getExtensionMap(
+        node: IZoweDatasetTreeNode,
+        preserveCase: boolean,
+        overrideExtension?: string
+    ): Promise<{ [key: string]: string }> {
+        ZoweLogger.trace("dataset.utils.getExtensionMap called.");
```
zFernand0: i know we are not very consistent, but should `trace` log messages match the `class.method` names exactly? 😅 
No need to do this here though 😋

packages/zowe-explorer/src/trees/dataset/DatasetUtils.ts
```diff
@@ -158,4 +158,58 @@ export class DatasetUtils {
...
+            if (overrideExtension) {
+                extension = overrideExtension;
+            } else {
+                for (const [ext, matches] of DS_EXTENSION_MAP.entries()) {
+                    if (ext === ".c") {
+                        // Special case for ".c" extension, skip the following logic
+                        // As it's not unique enough and would otherwise match on anything containing "C"
+                        // TODO: rrevisit later if this can be handled better while still allowing c files
+                        continue;
+                    }
```
zFernand0, line 186: Thanks for adding this `TODO` comment. 
Should we create an issue about this?

packages/zowe-explorer/src/trees/dataset/DatasetUtils.ts
```diff
@@ -158,4 +158,58 @@ export class DatasetUtils {
...
+                    if (matches.some((match) => (match instanceof RegExp ? match.test(label) : label.includes(match)))) {
+                        extension = ext;
+                        break;
+                    }
+                }
+
+                // If no extension found, fall back to using the PDS name as extension
+                if (!extension) {
+                    const parentExtension = DatasetUtils.getExtension(node.label as string);
```
zFernand0, line 195: Curious if we should handle the `.c` pattern-matching in the getExtension function?

That way we can avoid some code duplication (i.e. the regex match done in the section above. 😋 

It'd be good to discuss, as such change may be considered breaking. 🙏

packages/zowe-explorer/src/trees/uss/UssFSProvider.ts
```diff
@@ -716,13 +720,13 @@ export class UssFSProvider extends BaseProvider implements vscode.FileSystemProv
                         Gui.errorMessage(err.message);
                         return;
                     }
-                } catch (err) {
-                    if (err.name === "Error" && Number(err.errorCode) === imperative.RestConstants.HTTP_STATUS_404) {
+                } catch (listError) {
+                    if (listError.name === "Error" && Number(listError.errorCode) === imperative.RestConstants.HTTP_STATUS_404) {
```
zFernand0, line 725: 🤔 Variable shadowing? 

Thanks for fixing! 🙏

packages/zowe-explorer/package.json
```diff
@@ -1724,6 +1782,18 @@
...
+          "when": "never"
+        },
+        {
+          "command": "zowe.ds.downloadAllMembers",
+          "when": "never"
+        },
+        {
+          "command": "zowe.ds.downloadMember",
+          "when": "never"
+        },
```
zFernand0, line 1796: I know this is outside of the scope of this PR, but at some point we should consider exposing all of these commands int he command palette 🙏

packages/zowe-explorer/src/trees/uss/USSUtils.ts
```diff
@@ -109,4 +110,47 @@ export class USSUtils {
...
+                        totalCount += await this.countAllFilesRecursively(child, maxDepth, currentDepth + 1);
+                    }
+                } else {
+                    totalCount += 1;
+                }
+
+                // Return early to avoid unnecessary recursion
+                if (totalCount > Constants.MIN_WARN_DOWNLOAD_FILES) {
+                    return totalCount;
+                }
```
zFernand0: I love this! 🙏 

I do have one request though. 

Could we pass the current value of totalCount to the deeper calls?
Not to add it, but to check for the `totalCount > (warning_number = 100)`

Consider the scenario of me downloading this "dummy" folder:
- dumy
  - A01
  - A02
  - `...`
  - A99
  - B (directory)
    - B01
    - B02
    - `...`
    - B99
    - C (directory)
      -  C01
      - `...`

zFernand0 (changes requested):
LGTM! (so far) 😋 
---
I have yet to try the functionality, plus i'm missing about 6 files to review 😅 

I may have a second round of comments later today.

The only reason for my requested changes was due to some edge cases arroud recursion 🙏

packages/zowe-explorer/src/trees/uss/USSUtils.ts
JWaters02: This is a good suggest, but honestly I don't think there is really a need for this function anymore. It was my first idea to count number of files in tree to warn of max files, but then when I added support for filters the count would be come entirely inaccurate so I swapped to doing the call on list API with this recursive count as a "fallback". But if the list API call fails then we probably have bigger problems than trying to do a recursive count then a download API call which would probably also fail.

packages/zowe-explorer/src/trees/shared/SharedUtils.ts
```diff
@@ -588,6 +712,156 @@ export class SharedUtils {
...
+                correlation: {
+                    summary: `${downloadType} download completed with warnings`,
+                },
+            });
+
+            await SharedUtils.showDownloadMessage(warningMessage, downloadedPath, false, true, () => {
+                void vscode.commands.executeCommand("zowe.troubleshootError", correlatedWarning, detailedInfo.join("\n\n"));
+            });
+        } else {
+            await SharedUtils.showDownloadMessage(message, downloadedPath, false, false);
```
zFernand0: Nitpick: 
What do you think about being more "explicit" about what the booleans represent when calling the showDownloadMessage? Can we use the hasErrors and hasWarnings instead of true false? 😋 

And if we do, it feels like som of this can be comined and perhaps we just call the showDownloadMessage only twice (1. if (!response) 2. after the `if/elif/else`) 😋 

No need to actually make these changes though. I'm just looking for ways to reduce some mini-duplication, but it's probably insignificant since most things are already combined in the the private method 🙏

packages/zowe-explorer/src/trees/shared/SharedUtils.ts
```diff
@@ -170,6 +172,30 @@ export class SharedUtils {
...
+     * @param profileName The profile name to save encoding for
+     * @param encoding The encoding preference to save
+     */
+    public static setCachedDirectoryEncoding(profileName: string, encoding: "auto-detect" | ZosEncoding): void {
+        const encodingMap =
+            ZoweLocalStorage.getValue<Record<string, "auto-detect" | ZosEncoding>>(Definitions.LocalStorageKey.USS_DIRECTORY_ENCODING) ?? {};
+        encodingMap[profileName] = encoding;
+        ZoweLocalStorage.setValue(Definitions.LocalStorageKey.USS_DIRECTORY_ENCODING, encodingMap);
+    }
+
```
zFernand0: Dumb question:
Where are these functions used?

packages/zowe-explorer/src/trees/shared/SharedUtils.ts
```diff
@@ -265,9 +301,10 @@ export class SharedUtils {
      *
      * @param {string} response - The response from the user
      * @param {string} contextLabel - The context label of the node
-     * @returns {Promise<ZosEncoding | undefined>} The {@link ZosEncoding} object or `undefined` if the user dismisses the prompt
+     * @returns {Promise<ZosEncoding | null | undefined>} The {@link ZosEncoding} object, `null` for auto-detect,
+     *   or `undefined` if the user dismisses the prompt
      */
-    private static async processEncodingResponse(response: string | undefined, contextLabel: string): Promise<ZosEncoding | undefined> {
+    private static async processEncodingResponse(response: string | undefined, contextLabel: string): Promise<ZosEncoding | null | undefined> {
```
zFernand0: I believe we've attempted to give `null` a different behavior than `undefined` in some cases. However, it's always good to avoid doing this. Mostly because of the subtle issues that tend to creep up over the years when we miss purposefully avoid a `=` in a `== null` check.

Why no just create a new type of ZosEncoding and add it the the ZosEncoding type?

If we feel that it might be tricky for extenders to adopt, we could just use `{kind: "other", codepage: "auto-detect"}` and check directly for `?.codepage === auto-detect"`.

packages/zowe-explorer/src/trees/uss/USSActions.ts
zFernand0, line 1: I love all the functionality that is coming in! 🙏 
I do have one question... 😅 
Can we try to keep PRs a bit smaller next time? 😓 
This one is kind of on me for waiting to long to review 😢

packages/zowe-explorer-api/src/extend/MainframeInteraction.ts
JWaters02: Confused on this one. Which ones aren't returning the proper Promise<...> value? Thanks

packages/zowe-explorer/src/trees/dataset/DatasetInit.ts
JWaters02: No idea, does it matter? It already returns a promise, the await wouldn't change anything other than be more explicit I suppose? But please, not dumb question, it made me question reality. I am likely missing something in your motivation to ask 😅

packages/zowe-explorer/src/trees/dataset/DatasetUtils.ts
JWaters02: You are right, is not consistent everywhere haha. But in this case I do recall having a problem where getExtensionMap wasn't being called and I benefited from the trace log - but on the other hand, I probably didn't even use the log, I probably just put a breakpoint in it. So my excuse now is well almost every other function in the file is using it! 😂

packages/zowe-explorer/src/trees/dataset/DatasetUtils.ts
JWaters02: Eeeh, yeah I didn't want to touch getExtension because of it being breaking changes. Maybe an issue you are right, but probably a V4 thing?

packages/zowe-explorer/package.json
JWaters02: Agreed, would be cool to have an enhancement where you just download by dataset search filter or enter the file or directory or w/e name in command palette quick pick of file you wanna download.

packages/zowe-explorer/src/trees/uss/USSUtils.ts
JWaters02: @zFernand0 would be you opposed if I yeeted this function out of existence? due to the above comment. thanks

packages/zowe-explorer/src/trees/uss/USSActions.ts
JWaters02: Ok firstly, this is not on anyone for not reveiwing. Cos ya'll are paid to work on this whole ecosystem so you got resource allocation priorities and ZRS is fucking awesome project and you need to spend time on that (and I'm excited to ditch z/OSMF when ZRS is stable and has parity). Plus adhering to your SLAs from customers, that's surely gotta take up a lot of your time. So yeah, no harm done, don't worry about it. 

My excuse is I just didn't want to ship a half baked feature because I get to see first hand every single day at work how much people complain about the smallest things in their development tools - and yes, unfortunately, ZE and vscode is at the top of the list by far. Not because it's bad, but just because everyone's used to working in greenscreen and eclipse IDEs for the last couple decades+ and now they're being forced by the company to adopt ZE because it's free and free always beats the £30k/year per user cost of IDz. So my motivation is to ship features that are fullly complete and not the butt of a complaint. 

However I could have also done the data set part first then USS commands, but the problem is with ADHD programming is I can't stick to one area for a while and I'll get stuck on something and need something else to move to. Then the stuff I figured out in USS commands really helped me with the data set commands part. Plus fixing bugs/issues in the CLI part, where I just felt giga blocked on the ZE impl until those CLI fixes got merged. So that's also where some of the time went. Bit of a loaded comment but there we have it

JWaters02: Also just fyi I am always up for an adhoc call on slack or wherever with whoeever, so please just send me a DM whenever

packages/zowe-explorer/src/trees/shared/SharedUtils.ts
JWaters02: Not dumb, they from legacy implementation. good spot

packages/zowe-explorer/src/trees/uss/USSActions.ts
JWaters02: The problem I think is that all this reviewing on the code does not matter if people actually try to use the frontend download functionality and give feedback that it's bad UX. Because if it's bad UX, then the code is going to have to change anyway, thus all the time spent on the reviewing is going to be partially wasted. So I just wished that I would've gotten more people spending 20 mins checking out the branch and testing out the actual user facing functionality and giving feedback on it, than spending the several hours it takes to review the code. Because now there's a chance you've spent all this time reviewing and then you'll try it and its bad. That's why I wanted the feedback requested label - it's feedback on the frontend stuff, not on the code! Hope this makes sense. Cheers

packages/zowe-explorer-api/src/extend/MainframeInteraction.ts
```diff
@@ -96,7 +96,7 @@ export namespace MainframeInteraction {
          *     as well as the list of results in apiResponse.items with
          *     minimal properties name, mode.
          */
-        fileList(ussFilePath: string): Promise<zosfiles.IZosFilesResponse>;
+        fileList(ussFilePath: string, options?: zosfiles.IUSSListOptions): Promise<zosfiles.IZosFilesResponse>;
```
JillieBeanSim, line 99: brought up in another issue, with adding a new optional parameter should we create a new API and mark the other as deprecated?

packages/zowe-explorer-api/src/profiles/ZoweExplorerZosmfApi.ts
```diff
@@ -114,8 +114,11 @@ import { IDataSetCount } from "../dataset/IDataSetCount";
      * An implementation of the Zowe Explorer USS API interface for zOSMF.
      */
     export class UssApi extends CommonApi implements MainframeInteraction.IUss {
-        public fileList(ussFilePath: string): Promise<zosfiles.IZosFilesResponse> {
-            return zosfiles.List.fileList(this.getSession(), ussFilePath, { responseTimeout: this.profile?.profile?.responseTimeout });
+        public fileList(ussFilePath: string, options?: zosfiles.IUSSListOptions): Promise<zosfiles.IZosFilesResponse> {
+            return zosfiles.List.fileList(this.getSession(), ussFilePath, {
+                responseTimeout: this.profile?.profile?.responseTimeout,
+                ...options,
```
JillieBeanSim, line 120: would we need to check for existence of options here to avoid issues like seen [here](https://github.com/zowe/zowe-explorer-vscode/issues/4039) or have a new API deprecating the old for adoption of new feature

packages/zowe-explorer/src/trees/dataset/DatasetInit.ts
JillieBeanSim: I would compare with others that are async which usually do have an await but if we would like it to run in background then no await, would just need to make sure the popup message for success isn't shown before the download is complete I would think

packages/zowe-explorer-api/src/extend/MainframeInteraction.ts
JWaters02: Hmm @JillieBeanSim I do not understand why this would be a breaking change or not backwards compatible, as I thought simply making a parameter optional is enough to get by.

packages/zowe-explorer-api/src/profiles/ZoweExplorerZosmfApi.ts
JWaters02: In the case of the previous issue where the optional parameter was added, the mistake was that they were using it without checking if it was existent or not. In this case however, the `zosfiles.List.fileList` API also lists `options` as optional, so if `options` does not exist, it does not matter as fileList will continue to work as before but without the specific options. So I don't think any check here needs to be done.

packages/zowe-explorer/src/trees/uss/USSUtils.ts
JWaters02: I'm just going to do it because it's a useless function now.

packages/zowe-explorer/src/trees/dataset/DatasetInit.ts
JWaters02: My understanding is that vscode commands are always aysnchronous so it doesn't matter if there is an await or not, the downloads will always run in the background and not block any other commands from being executed. I did try it with an await but there is no difference in functionality. However I think I will add an await regardless because apparently it should give a better stack trace if things go wrong?

JWaters02:
> Thanks for working on this @JWaters02 i have left a few comments/questions and also with the discussion on the UI I wonder if we should keep compatibility with the jobs tree search without the check boxes and clicking on the item takes them to a way to fill in the value. Any boolean values we need to clearly document the default in description and docs.

Thanks for the review @JillieBeanSim in terms of the jobs search UI part which part of the UI are you specifically talking about? The USS advanced filters part? I think maybe it might make sense there as currently if you select multiple filters to change it brings up each one in a row, but with jobs search design they would change each part in turn. So it would more clicks/actions but would probably make more sense UX wise.

As for the other parts (top level option quick picks) just having the checkboxes makes most sense to me, because they are just booleans and don't need some extra string or number box to specify value.

JWaters02:
As for the documentation comment, I have a WIP PR open in zowe/docs-site for downloads which I will do a pass over once this one is merged into main as some things have changed since I initially wrote that.

adam-wolfe:
The UX looks fine to me and the things I've tried have worked as expected.

The only confusing thing I found was that it is not clear what effect `Override File Extension` has when downloading members. If I go to download members from a PDS named `ADAMW.TEST.CBL` the quick pick says "current: jcl", however, regardless of whether the Override File Extension checkbox is checked, the downloaded files will be given a .cbl extension (which is what I would want; but what is the purpose of the checkbox then?)
image

packages/zowe-explorer/src/trees/dataset/DatasetInit.ts
traeok: These two (before/after changes) are functionally equivalent:

```typescript
// Before
vscode.commands.registerCommand("zowe.ds.downloadDataSet", async (node: IZoweDatasetTreeNode) => DatasetActions.downloadDataSet(node));
// After
vscode.commands.registerCommand("zowe.ds.downloadDataSet", async (node: IZoweDatasetTreeNode): Promise<void> => {
    await DatasetActions.downloadDataSet(node);
});
```

Command execution logic in VS Code indicates that all promises are awaited from command handlers:

https://github.com/microsoft/vscode/blob/7e715cdc2d29afed47950ed6d7ee66f528480d82/src/vs/workbench/api/common/extHostCommands.ts#L248

```typescript
		const stopWatch = StopWatch.create();
		try {
			return await callback.apply(thisArg, args);
		} catch (err) {
```

Adding the `await` actually creates an extra microtask, but the impact is negligible

packages/zowe-explorer/src/configuration/Definitions.ts
```diff
@@ -59,9 +59,7 @@ export namespace Definitions {
         dirFilterOptions?: UssDirFilterOptions;
     };
     export type UssDirOptions = {
-        filesys?: boolean;
-        symlinks?: boolean;
-        includeHidden?: boolean;
+        followSymlinks?: boolean;
         chooseFilterOptions?: boolean;
         directoryEncoding?: "auto-detect" | ZosEncoding;
```
traeok: `ZosEncoding` was already updated to include `{ kind: "auto-detect" }` - is this still needed?

packages/zowe-explorer/src/configuration/Definitions.ts
```diff
@@ -157,5 +192,8 @@ export namespace Definitions {
         V1_MIGRATION_STATUS = "zowe.v1MigrationStatus",
         DS_SEARCH_OPTIONS = "zowe.dsSearchOptions",
         DISPLAY_RELEASE_NOTES_VERSION = "zowe.displayReleaseNotes",
+        DS_DOWNLOAD_OPTIONS = "zowe.dsDownloadOptions",
+        USS_DOWNLOAD_OPTIONS = "zowe.ussDownloadOptions",
+        USS_DIRECTORY_ENCODING = "zowe.ussDirectoryEncoding",
```
traeok: This local storage key is not used in the code, is this expected?

packages/zowe-explorer/src/trees/dataset/DatasetActions.ts
```diff
@@ -625,6 +627,425 @@ export class DatasetActions {
...
+
+            optionsQuickPick.onDidHide(() => {
+                if (!wasAccepted) {
+                    resolve(null);
+                }
+            });
+
+            optionsQuickPick.show();
+        });
+        optionsQuickPick.dispose();
```
traeok: This pattern is identical to the one in `USSActions` - consider extracting this logic into a shared utility to reduce code duplication.

packages/zowe-explorer/src/trees/uss/USSActions.ts
```diff
@@ -314,6 +317,541 @@ export class USSActions {
...
+                }
+
+                const listOptions: zosfiles.IUSSListOptions = {
+                    ...listFilterOpts,
+                    type: listFilterOpts.type ?? "f", // fallback to files only
+                    symlinks: !downloadOptions.dirOptions.followSymlinks,
+                };
+
+                try {
+                    if (token.isCancellationRequested) {
```
traeok: - `downloadAllMembers` does not check this token
- (nitpick) if the SDKs aren't cancelling internally (we aren't passing any token to the call), then the cancellation token is only cosmetic. Consider adding a comment to investigate cancelling the request at the SDK level

packages/zowe-explorer/src/trees/dataset/DatasetInit.ts
JWaters02: Makes sense, good to have expert opinion! haha!

packages/zowe-explorer/src/trees/uss/USSActions.ts
JWaters02: I'll make a feature request issue in CLI for downloads to support cancellation mid download for ussDir, downloadAllMembers and downloadAllDataSets. For now I have set cancellable to false in the progress bars on downloads.

JWaters02: https://github.com/zowe/zowe-cli/issues/2693

PR #3882, opened 2025-09-30
fix(ds): Prevent deleting members for the wrong profile

### LINKED ISSUE #3883: Delete action incorrectly assumes parent PDS is in tree view selection
**Describe the bug**

Say that I have the same PDS on two different LPARs: Profile A and Profile B.

If I have the same PDS name selected under profile _A_, but then right click on a PDS member under the _same_ PDS name in profile _B_ to delete it... the data set member under profile A is deleted. 🫨 

This seems pretty severe and I ran into it while doing manual tests for another issue 😅 

**To Reproduce**

1. Create the same PDS name on two different systems with the same set of members
2. Search for that PDS under both profiles
3. Expand the PDS under both profiles
4. Select the PDS under profile A
5. Right-click on a member in profile B and select "Delete"
6. Notice that the member in profile A is deleted (actual) rather than profile B (expected)

**Expected behavior**

The delete action for members should not consider my tree view selection, only the item that was selected. The parent should be trivial to determine based on the selected PDS member(s).

- OS: Windows 11
- Zowe Explorer Version: build from `main` branch / 3.3.0 / 3.4.0-snapshot

t1m0thyj:
## Proposed changes

Fixes #3883

## How to Test

1. In Zowe Explorer, open a PDS member across two separate profiles. The data set name and member name must both match. For example, `IBMUSER.PDS(TESTMEM)` on both LPAR1 and LPAR2.
2. Click the member node associated with LPAR1 to make it the focused node in Data Sets tree. Then right-click (without selecting) the member node associated with LPAR2, and select the "Delete" action.
3. Verify that the member was deleted on LPAR2, not LPAR1.

## Release Notes

Milestone: 3.1.1

Changelog: Fixed an issue where the Delete action in the Data Sets tree could delete a PDS member previously selected in the tree rather than the one that was right-clicked

## Types of changes

## Checklist

**General**

**Code coverage**

**Deployment**

### DIFF
--- a/packages/zowe-explorer-api/CHANGELOG.md
+++ b/packages/zowe-explorer-api/CHANGELOG.md
@@ -15,6 +15,7 @@ All notable changes to the "zowe-explorer-api" extension will be documented in t
 - Updated Zowe SDKs to `8.27.0` for technical currency. [#3848](https://github.com/zowe/zowe-explorer-vscode/pull/3848)
 - Added support to delete VSAM data sets for z/OSMF type profiles. [#3824](https://github.com/zowe/zowe-explorer-vscode/issues/3824)
 - Fixed an issue where the `IMvs.putContents` function did not properly strip carriage returns from chunks during upload operations on Windows. Now, the function correctly converts CRLF sequences to LF, even if the sequence lands on a chunk boundary. [#3853](https://github.com/zowe/zowe-explorer-vscode/issues/3853)
+- Fixed an issue where the `vscode.workspace.fs.readFile` function would return empty contents for a USS file when an error occurred while fetching the file from the mainframe, instead of displaying an error message. [#3894](https://github.com/zowe/zowe-explorer-vscode/issues/3894)
 
 ## `3.3.0`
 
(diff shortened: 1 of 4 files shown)


packages/zowe-explorer/src/trees/dataset/DatasetActions.ts
```diff
@@ -653,17 +645,10 @@ export class DatasetActions {
...
-                (selectedNode) => selectedNode.getParent() && !SharedContext.isSession(selectedNode) && !SharedContext.isInformation(selectedNode)
-            ) as IZoweDatasetTreeNode[];
-        } else {
-            if (node.getParent() && !SharedContext.isSession(node) && !SharedContext.isInformation(node)) {
-                nodes = [];
-                nodes.push(node);
-            }
-        }
+        // Filter out sessions and information messages
+        let nodes = selectedNodes.filter(
```
traeok: (nitpick) `nodes` can be made `const` if lines 703-710 were refactored to return early, instead of setting `nodes` to an empty array. That said, I'm not expecting you to change existing code that's not entirely related to this fix, just thinking of ways to optimize where possible 🙂 

It is a bit odd that those aforementioned lines have a pattern of `await Gui.infoMessage(...).then(...)` - but this code is pre-existing, so not requesting changes 😋

traeok (approved):
LGTM, thanks for the fix!

pujal0909 (changes requested):
Thanks @t1m0thyj! This resolves the issue of the member being deleted off the wrong profile, but I am getting this error when selecting a PDS on one LPAR and right clicking without selecting a PDS (named the same) on another LPAR. It seems to delete off the correct profile, but this error appears. Not sure if this is related to your changes or not.

image

packages/zowe-explorer/src/trees/dataset/DatasetActions.ts
t1m0thyj: Addressed in https://github.com/zowe/zowe-explorer-vscode/pull/3882/commits/60f8fa3684c4f6eda03243f6888c348087170ad2

PR #3954, opened 2025-12-01
Safely move PDS members across LPARs

ATorrise:
## Proposed changes

**TLDR: Fixes drag-and-drop target resolution and makes cross-LPAR moves delete the source only after the destination write call succeeds. Any delayed appearance in member listings is treated as UI/listing propagation, not a failed write.**

Problems: 
  - Drop target was not correctly resolved to the PDS container
    - Dropping onto a PDS child node (member / placeholder) could resolve to the wrong target. 
  - Move logic relied on local/FS assumptions about immediate visibility
    - After a successful write, the member may not show up immediately in list APIs / tree refresh due to caching / eventual consistency (especially for empty/newly opened PDS), which made behavior confusing. 

Fixes:
1. Corrected drop target
    -  Dropping onto non-container nodes (members, sequential datasets, or the “No data sets found” placeholder) now re-maps to the correct parent node in the DS tree
    - Prevents invalid drop targets while preserving intuitive drag-and-drop behavior
2. Delete source only after transfer succeeds
    - For cross-LPAR moves, the source member is deleted only after the destination write completes successfully.
    - We treat the REST write success response as the authoritative completion signal for the move
3. Defensive error handling
    - If the destination write fails (or throws), the source is preserved and the user sees a clear error
    - Preevents partial moves 
4. Corrected data set / member naming
    - Fixes incorrect URI/path slicing that produced malformed API calls
5. Same profile member moves allowed
6. `handleDrop` restructuring and test adjustments

**Note: member list delay**
After a successful upload, the destination member may not be immediately discoverable via list APIs (ie allMembers()) due to caching / propagation timing. We refresh the UI after the move, but we do not determine source deletion because of list visibility anymore

## How to Test
1. Configure two different LPARs
2. On LPAR1 create: `PDS.SOURCE(MEMBER)`
3. On LPAR2 create: `PDS.TARGET`
5. In ZE, drag `MEMBER` from `PDS.SOURCE` into `PDS.TARGET` -> drop DIRECTLY onto node that says "No data sets found"

Expected Results
  - `PDS.TARGET(MEMBER)` is available on LPAR2
  - The content matches exactly
  - The source member is deleted only after the destination is verified
  - A success message is shown

Further testing that IS needed!:
1. Reload your test ZE environment
2. Drag `MEMBER` from what was just `PDS.TARGET` back into its original source
3. Ensure that existing members can be moved as well!! Not just newly created members
4. Test both same-profile and cross-profile moves

## Release Notes

Milestone:

Changelog: Stabilized cross-LPAR PDS member moves to prevent data loss and race conditions.
Member deletion now occurs only after verified persistence on the target system. [#3954](https://github.com/zowe/zowe-explorer-vscode/pull/3954)

## Types of changes

## Checklist

**General**

**Code coverage**

**Deployment**

### DIFF
--- a/packages/zowe-explorer/CHANGELOG.md
+++ b/packages/zowe-explorer/CHANGELOG.md
@@ -16,6 +16,7 @@ All notable changes to the "vscode-extension-for-zowe" extension will be documen
 - Added support for loading credential manager options from the imperative.json file. Add a `credentialManagerOptions` object in the JSON object in `imperative.json` to specify options for the current credential manager. [#3935](https://github.com/zowe/zowe-explorer-vscode/pull/3935)
 - Added support for custom persistence levels for Windows (persist option) to support the credential manager in less permissive environments. For more information on how to configure this option, refer to the ["Troubleshooting Zowe CLI credentials"](https://docs.zowe.org/stable/troubleshoot/cli/troubleshoot-cli-credentials/#secrets-sdk-persistence-level-for-windows) page on Zowe Docs. [#3935](https://github.com/zowe/zowe-explorer-vscode/pull/3935)
 - Updated minimum VS Code version from 1.79 to 1.90. We are dropping support for VS Code releases that bundle versions of Node.js no longer receiving security updates. [#3684](https://github.com/zowe/zowe-explorer-vscode/issues/3684)
+- Expose commands to drive search in the USS and Data Sets views. [#3933](https://github.com/zowe/zowe-explorer-vscode/issues/3933)
 
 ### Bug fixes
 
--- a/packages/zowe-explorer/__tests__/__unit__/extension.unit.test.ts
+++ b/packages/zowe-explorer/__tests__/__unit__/extension.unit.test.ts
@@ -207,6 +207,7 @@ async function createGlobalMocks() {
             "zowe.uss.pasteUssFile",
             "zowe.uss.copyUssFile",
             "zowe.uss.copyRelativePath",
+            "zowe.uss.setUssPath",
             "zowe.jobs.deleteJob",
             "zowe.jobs.runModifyCommand",
             "zowe.jobs.runStopCommand",
--- a/packages/zowe-explorer/__tests__/__unit__/trees/uss/USSActions.unit.test.ts
+++ b/packages/zowe-explorer/__tests__/__unit__/trees/uss/USSActions.unit.test.ts
@@ -1056,3 +1056,664 @@ describe("USS Action Unit Tests - function copyRelativePath", () => {
         expect(mocked(vscode.env.clipboard.writeText)).toHaveBeenCalledWith("usstest");
     });
 });
+
+describe("USS Action Unit Tests - function filterUssTreePrompt", () => {
+    function createBlockMocks(globalMocks) {
+        const newMocks = {
+            testUSSTree: undefined as unknown as USSTree,
+            ussNode: createUSSNode(globalMocks.testSession, createIProfile()),
+            testTreeView: createTreeView(),
+            quickPick: {
+                items: [] as any[],
+                placeholder: "",
+                ignoreFocusOut: false,
+                activeItems: [] as any[],
+                value: "",
+                show: jest.fn(),
+                hide: jest.fn(),
+                dispose: jest.fn(),
+                onDidAccept: jest.fn(),
+                onDidHide: jest.fn(),
+            },
+        };
+        newMocks.testUSSTree = createUSSTree(
+            [createFavoriteUSSNode(globalMocks.testSession, globalMocks.testProfile)],
+            [newMocks.ussNode],
+            newMocks.testTreeView
+        );
+
+        return newMocks;
+    }
+
+    it("should prompt for profile and USS path, then call filterUssTree", async () => {
+        const globalMocks = createGlobalMocks();
+        const blockMocks = createBlockMocks(globalMocks);
+
+        const getRegisteredProfileNameListSpy = jest
+            .spyOn(require("../../../../src/management/ProfileManagement").ProfileManagement, "getRegisteredProfileNameList")
+            .mockReturnValue(["profile1", "profile2"]);
+        const createQuickPickSpy = jest.spyOn(vscode.window, "createQuickPick").mockReturnValue(blockMocks.quickPick as any);
+
+        blockMocks.quickPick.onDidAccept.mockImplementation((callback) => {
+            blockMocks.quickPick.activeItems = [{ label: "profile1" }];
+            callback();
+            return { dispose: jest.fn() };
+        });
+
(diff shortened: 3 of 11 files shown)


ATorrise:
ok im finding that this error happens when you move a member that already existed.. there seems to be some kind of lapse in awareness from ZE knowing what members exist, thinking that those that were created before its start up arent recognized, only those after its startup? tbd... investigating.. ty @pujal0909

packages/zowe-explorer/CHANGELOG.md
```diff
@@ -29,6 +29,7 @@ All notable changes to the "vscode-extension-for-zowe" extension will be documen
 - Fixed an issue where user attempts to perform a search on an individual PDS from the Zowe-Explorer extension using the "Search PDS members" action and query a search string, it brings out additional dataset as a part of the result.[#3682](https://github.com/zowe/zowe-explorer-vscode/issues/3682)
 - Fixed an issue where loading of virtual workspaces interfered with the activation of some Zowe Explorer extensions. [#3923](https://github.com/zowe/zowe-explorer-vscode/pull/3923)
 - Fixed issue seen by extender profile types where a type error was returned with 401 instead of notification with update option. [#3943](https://github.com/zowe/zowe-explorer-vscode/issues/3943)
+- Stabilized cross-LPAR dataset move for data integrity. [#3954](https://github.com/zowe/zowe-explorer-vscode/pull/3954)
```
```suggestion
- Stabilized data set moves across LPARs for data integrity. [#3954](https://github.com/zowe/zowe-explorer-vscode/pull/3954)
```

packages/zowe-explorer/src/trees/dataset/DatasetTree.ts
```diff
@@ -175,74 +272,120 @@ export class DatasetTree extends ZoweTreeProvider<IZoweDatasetTreeNode> implemen
...
+                    if (err.name !== "EntryNotFound" && err.name !== "FileNotFound" && err.name !== "FileSystemError") {
+                        throw err;
+                    }
                 }
-                return;
+                if (i === maxRetries - 1) {
+                    throw new Error(vscode.l10n.t("Failed to verify data write to {0}. Content size mismatch or member empty after multiple retries.", dsname));
+                }
+                await new Promise((resolve) => setTimeout(resolve, retryDelay));
+                retryDelay *= 2;
```
t1m0thyj: Could you please provide more details for how to reproduce the scenario where the polling mechanism is needed?

ATorrise: will be removing the verification code for file transfer to target system in favor of trusting the success of write even if newly transferred data is not yet visible via member listing / UI refresh

in the old implementation, handleDrop() didn’t consistently pivot the drop target to the PDS container. When dropping onto a PDS child context (member / placeholder), destUri could be constructed incorrectly, so we could successfully write to an unintended/invalid destination URI and then immediately delete the source. The fix is: normalize the drop target to the PDS container + correct destination member URI construction, so the write goes to the correct place before deleting the source.

packages/zowe-explorer/src/trees/dataset/DatasetTree.ts
```diff
@@ -113,136 +112,278 @@ export class DatasetTree extends ZoweTreeProvider<IZoweDatasetTreeNode> implemen
...
+                Gui.errorMessage(vscode.l10n.t("Failed to move {0}: {1}", dsname, err.message));
             }
         }
     }
 
     public async handleDrop(
         targetNode: IZoweDatasetTreeNode | undefined,
         dataTransfer: vscode.DataTransfer,
-        _token: vscode.CancellationToken
+        _token?: vscode.CancellationToken
```
traeok: The VS Code API doc suggests that `token` is a required parameter:

https://code.visualstudio.com/api/references/vscode-api#TreeDragAndDropController&lt;T&gt;

I realize that we don't do anything with it (we probably should in case user cancels the drop, but that's a separate issue) - but we can remove the `?` here:



```suggestion
        _token: vscode.CancellationToken
```

packages/zowe-explorer/src/trees/dataset/DatasetTree.ts
```diff
@@ -113,136 +112,278 @@ export class DatasetTree extends ZoweTreeProvider<IZoweDatasetTreeNode> implemen
...
+            if (SharedContext.isDs(target)) {
+                Gui.errorMessage(vscode.l10n.t("Cannot drop items into a sequential dataset."));
                 return;
             }
 
-            // 2. PDS member collision check (only if both are PDS)
+            // 1) Same-object check
+            const nodeProfile = node.getProfile?.();
+            const targetProfile = target.getProfile?.();
+            const nodeParentUri = node.getParent?.()?.resourceUri?.path;
```
traeok: Are there cases where the `getProfile` is `undefined`? The optional chaining seems redundant here.

ATorrise: tests but ill make the tests better instead

packages/zowe-explorer/src/trees/dataset/DatasetTree.ts
```diff
@@ -274,6 +274,12 @@ export class DatasetTree extends ZoweTreeProvider<IZoweDatasetTreeNode> implemen
                 (SharedUtils.getNodeProperty(sourceNode as any, "label") as string) ??
                 "";
 
+            // FIX: Safety check for empty dataset name (Requested by reviewer)
```
traeok: Please remove this comment as its redundant

packages/zowe-explorer/src/trees/dataset/DatasetTree.ts
```diff
@@ -113,105 +112,220 @@ export class DatasetTree extends ZoweTreeProvider<IZoweDatasetTreeNode> implemen
...
             }
-            // read the contents from the source LPAR
-            const contents = await DatasetFSProvider.instance.readFile(sourceNode.resourceUri);
-            //write the contents to the destination LPAR
+        } else {
+            dsname =
+                (sourceNode as any).getLabel?.() ??
+                (sourceNode as any).label ??
+                (SharedUtils.getNodeProperty(sourceNode as any, "label") as string) ??
+                "";
```
traeok: - `getLabel` is a required function and should exist on `sourceNode` as long as its not `null` or `undefined`
- The `label` property is what's returned by the core Zowe Explorer nodes for `getLabel`
- `getNodeProperty` call is redundant as it is effectively the same as accessing `sourceNode.label`

```suggestion
            dsname = sourceNode.getLabel().toString();
```

t1m0thyj:
@ATorrise Based on discussion offline, I opened #3999 which focuses solely on fixing the linked issue. I suggest that we merge that PR first, then test if any of the race conditions that would require additional fixes still exist.

ATorrise:
Ty timothy, #3999 seems to be sufficient!!

PR #3962, opened 2025-12-05
fix: Return clearer error if profile type unavailable in FS; skip child favorite nodes if type missing 

traeok:
## Proposed changes

Fixes the following issues:

- With recent changes to the file systems (around time of 3.3/3.3.1) we now perform extra checks in `stat` to get a profile's session. This introduced an issue in the favorites logic, where an unhandled error occurs if the profile type that contributed those nodes is no longer available at time of favorites initialization.
- If the profile type is unavailable within the file system, the thrown error had code "Unknown" and an unclear error message for any caller that encounters it. Now, the file system uses a generic helper function called `getApiOrThrowUnavailable` where any non-existing profile type error is transformed into a more developer-friendly `Unavailable` error with specific details on why the resource failed to load.

### How to test

- Favorite a data set, USS file/directory or job under an extender profile type (`ssh`, `zftp`, `rse`, etc.).
- Disable the extension that contributes the type and reload the VS Code window
- ZE should not throw any errors during activation even if the extender profile type is unavailable during favorites initialization
- Enabling the extension & reloading the window should not cause errors either. The favorited resource(s) for that profile type should re-appear in the Favorites view once it is registered.

To retain backwards compatibility I am keeping stale favorites in local storage. At some point we should consider cleaning up stale favorites - but is out of scope for this PR

## Release Notes

Milestone: 3.5.0

Changelog: See details above

## Types of changes

## Checklist

**General**

- [x] New ZE APIs are tested with extender types that haven't adopted yet to determine breaking changes. Can use Zowe zFTP marketplace extension.

**Code coverage**

**Deployment**

### DIFF
--- a/packages/zowe-explorer-api/CHANGELOG.md
+++ b/packages/zowe-explorer-api/CHANGELOG.md
@@ -18,6 +18,7 @@ All notable changes to the "zowe-explorer-api" extension will be documented in t
 
 - Fixed an issue where secure credentials and headers were being logged to the Zowe logger and VSCode output channel. [#3848](https://github.com/zowe/zowe-explorer-vscode/pull/3848)
 - Updated Zowe SDKs to version `8.28.0` to address an issue where copying a PDS member to a data set across LPARs failed. This occurred when the target PDS already contained members, but none matched the name of the PDS member being copied. [#3848](https://github.com/zowe/zowe-explorer-vscode/pull/3896)
+- Ensure that the `updateCredentials` function adds the given credentials to the `availableCreds` cache of the `Session` object. [#3940](https://github.com/zowe/zowe-explorer-vscode/pull/3940)
 
 ## `3.3.1`
 
--- a/packages/zowe-explorer-api/__tests__/__unit__/vscode/ZoweVsCodeExtension.unit.test.ts
+++ b/packages/zowe-explorer-api/__tests__/__unit__/vscode/ZoweVsCodeExtension.unit.test.ts
@@ -694,6 +694,8 @@ describe("ZoweVsCodeExtension", () => {
             });
             const showInputBoxSpy = jest.spyOn(Gui, "showInputBox").mockResolvedValueOnce("fakeUser").mockResolvedValueOnce("fakePassword");
             const saveCredentialsSpy = jest.spyOn(ZoweVsCodeExtension as any, "saveCredentials");
+            const addCredsToSessionSpy = jest.spyOn(imperative.AuthOrder, "addCredsToSession");
+
             const profileLoaded: imperative.IProfileLoaded = await ZoweVsCodeExtension.updateCredentials(
                 promptCredsOptions,
                 undefined as unknown as Types.IApiRegisterClient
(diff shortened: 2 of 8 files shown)


anaxceron (approved):
Nicely done changelogs!

packages/zowe-explorer-api/src/fs/utils/FsAbstractUtils.ts
```diff
@@ -78,4 +79,49 @@ export class FsAbstractUtils {
...
+     * into a `FileSystemError.Unavailable` error for users and extenders.
+     *
+     * @param profile The profile used to resolve the API
+     * @param apiGetter A callback that returns the API instance (or throws)
+     * @param options.apiName Optional friendly name for the API (e.g. "Common API")
+     * @param options.registeredTypes Optional list of registered types to short-circuit early
+     * @returns The value from `apiGetter` when successful
+     * @throws vscode.FileSystemError.Unavailable when the profile type is not registered or the API getter throws a "non-existing" error
+     */
+    public static getApiOrThrowUnavailable<T>(
```
t1m0thyj, line 94: Initially I was thinking it feels weird to add a utility method that wraps the API getters, versus updating the signature of `getMvsApi`, `getJesApi`, and friends to have an additional boolean parameter (either `optional` defaulting to true or `throwIfNotFound` defaulting to false).

But after further consideration, I think this is a clever way to implement the "throw if unavailable" behavior, that is less impactful since it avoids having to change the signature of methods on the ZE API Register 👍

packages/zowe-explorer/src/trees/dataset/DatasetTree.ts
```diff
@@ -417,13 +417,20 @@ export class DatasetTree extends ZoweTreeProvider<IZoweDatasetTreeNode> implemen
...
+    public async createProfileNodeForFavs(profileName: string, targetProfileType?: string): Promise<ZoweDatasetNode | null> {
         ZoweLogger.trace("DatasetTree.createProfileNodeForFavs called.");
         let favProfileNode: ZoweDatasetNode;
         try {
             const profile = Profiles.getInstance().loadNamedProfile(profileName);
+            if (targetProfileType && profile.type !== targetProfileType) {
+                return null;
+            }
+            if (!ZoweExplorerApiRegister.getInstance().registeredMvsApiTypes().includes(profile.type)) {
+                return null;
```
t1m0thyj, line 432: Trying to understand what scenario would cause this to be true - is it when old favorites are left around in local storage for a profile type that is no longer installed? Curious if we should log an INFO or WARN level message here. Or if we should delete "stale" favorites so they don't accumulate.

traeok: Yes, this scenario is when old favorites are left around in local storage for a profile type that's no longer installed.

I hesitated to introduce the deletion of old favorites because we weren't doing so previously, and that could be perceived as breaking. One scenario is where a user has to uninstall an extender (possibly due to upgrades, an issue, or some other unknown reason) and if they happen to use Zowe Explorer once before reinstalling it, all of their extender favorites are gone from local storage. 

I'm not opposed to the cleanup of stale favorites but this seems out of scope and may warrant further research.

traeok: Regarding the log message, I considered adding one but it may be confusing for extenders if they have to review logs. It may mislead developers into believing that their favorites aren't being loaded correctly when its really just a matter of activation timing. But I'm curious to get your thoughts, do you think there would be benefit(s) in having a log here?

packages/zowe-explorer-api/src/fs/utils/FsAbstractUtils.ts
traeok: > versus updating the signature of getMvsApi, getJesApi, and friends to have an additional boolean parameter (either optional defaulting to true or throwIfNotFound defaulting to false

FWIW, we already have a pattern of catching errors & exiting early for `get###Api` calls if the profile type is not found - this is just meant to translate it into a nicer filesystem error 😋 we could definitely revisit our approach to return `undefined` if the API doesn't exist in v4.

packages/zowe-explorer/src/trees/dataset/DatasetTree.ts
t1m0thyj: Good point, I agree that cleaning up stale favorites would be outside the scope of this PR. Probably no need to have a log message (or if we do, it should be DEBUG level to avoid confusion).

PR #3963, opened 2025-12-05
Stat fetch default

jace-roell:
## Proposed changes

Modifies the behavior of FileSystem calls in regards to fetching and caching values, aswell as caching similar in-flight requests to avoid redundant calls.

**Feature Flags**

- FeatureFlag.ts class: new class that allows feature flags to be set in the feature-flags.json file.
~~Currently is used for the fetchByDefault feature flag, this flag will be removed prior to merging of this PR, but it is kept in order to allow for toggling of behavior when testing~~

- Proposed new behavior
`fetch=false` (default): will do a lookup in the fileSystem cache but will fallback to a remote lookup in the case where one is not found
`fetch=true`: will ignore values in the FileSystem cache, does a remote lookup regardless

**Request caching**

- When a FileSystem request is made, a key is generated for the request. The key and request are stored in a map. If a request with the same key is made, the subsequent request will return the promise of the original to avoid redundant requests. The map entry is cleared upon the resolution of of the promise.
- The key includes the function, URI, and any truthy query parameters.

**Enhanced Dataset Caching**
- When making a FileSystem call on a Dataset member, the request will return all of the members of the parent PDS and find the corresponding member. This logic has been enhanced to cache all of the returned members instead of the just the desired member.

**Catching invalid DS URIs**
- Catches more vscode `stat()` calls with URIs that could not exist as data sets

**Trailing slash fix**
- Fixes a bug where URIs with trailing slashes would throw an error 

Example Script:

```
    const profile = "<profile>";
    const dsn1 = "<DSN>/<MEM1>";
    const dsn2 = "<DSN>/<MEM2>";

        setTimeout(async () => {
            try {
                console.info("Stat1 DS");

                const statResult = await vscode.workspace.fs.stat(vscode.Uri.parse(`zowe-ds:/${profile}/${dsn1}`));
                console.info(`Stat1 success ${JSON.stringify(statResult)}`);
            } catch (err) {
                console.error(`Stat1 ds error ${JSON.stringify(err)}`);
            }
        }, 3000);

        setTimeout(async () => {
            try {
                console.info("Stat2");
                const statResult = await vscode.workspace.fs.stat(vscode.Uri.parse(`zowe-ds:/${profile}/${dsn2}`));
                console.info(`Stat2 success ${JSON.stringify(statResult)}`);
            } catch (err) {
                console.error(`Stat2 error ${JSON.stringify(err)}`);
            }
        }, 3000);
```

Wiki update: https://github.com/zowe/zowe-explorer-vscode/wiki/FileSystemProvider#reuse-similar-ongoing-requests

## Release Notes

Milestone:

Changelog:

## Types of changes

## Checklist

**General**

**Code coverage**

**Deployment**


packages/zowe-explorer/CHANGELOG.md
```diff
@@ -33,6 +34,9 @@ All notable changes to the "vscode-extension-for-zowe" extension will be documen
 
 ### Bug fixes
 
+- Fixed an issue where file system calls incorrectly threw a FileNotFound error for existing files, ensuring they are now fetched correctly. [#3554](https://github.com/zowe/zowe-explorer-vscode/issues/3554)
+- Fixed an issue where file system URIs with trailing slashes failed to resolve. [#3904](https://github.com/zowe/zowe-explorer-vscode/issues/3904)
```
zFernand0: These two should be moved to the recent changes

packages/zowe-explorer/CHANGELOG.md
```diff
@@ -33,6 +34,9 @@ All notable changes to the "vscode-extension-for-zowe" extension will be documen
...
+- Fixed an issue where an error appeared when moving an empty PDS across LPARs. [#4001](https://github.com/zowe/zowe-explorer-vscode/pull/4001)
```
zFernand0: And this one can be removed as it might be a duplicate 😋

zFernand0 (commented):
just commenting on changelogs 😋

packages/zowe-explorer/src/trees/dataset/DatasetFSProvider.ts
```diff
@@ -390,38 +482,66 @@ export class DatasetFSProvider extends BaseProvider implements vscode.FileSystem
...
         } catch (err) {
-            // Errors unrelated to the filesystem cannot be handled here
             if (!(err instanceof vscode.FileSystemError)) {
                 throw err;
             }
 
-            if (err.code === "FileNotFound") {
-                // if the entry doesn't exist in the local file system, first check to see if it exists on the remote before throwing an error.
+            //TODO Feature Flag
+            if (err.code === "FileNotFound" && FeatureFlags.get("fetchByDefault")) {
```
slavek-kucera: `&& !shouldFetch` otherwise the remoteLookupForResource will be called again.

packages/zowe-explorer/src/trees/dataset/DatasetFSProvider.ts
```diff
@@ -130,34 +151,92 @@ export class DatasetFSProvider extends BaseProvider implements vscode.FileSystem
...
+     */
+    public async stat(uri: vscode.Uri): Promise<vscode.FileStat> {
+        const segments = uri.path.split("/").filter((s) => s.length > 0);
+        const isMemberRequest = segments.length === 3; // /PROFILE/DSN/MEMBER
+
+        if (isMemberRequest) {
+            const memberName = segments[2];
+            const parentPath = segments.slice(0, 2).join("/");
+            const parentUri = uri.with({ path: `/${parentPath}` });
+            const parentKey = "list" + this.getQueryKey(uri) + "_" + parentUri.toString();
```
slavek-kucera: With the fetch by default is enabled, is there still a missed opportunity to share the request when two different callers issue `stat` (and other calls) for the same resource - one is with `fetch=true` while the other does not provider the `fetch` parameter?

jace-roell: As noted in the PR description, the behavior for `fetch=true` remains unchanged. The fetchByDefault logic is an optimization of the default behavior (`fetch=false`); therefore, to take advantage of this new lookup logic, the `fetch=true` query parameter must be removed from extender calls.

Due to this, the request reuse case would be when `fetch=false` and no fetch parameter is present, which is currently implemented.

slavek-kucera: The scenario is
- fetchByDefault enabled
- cache is empty

two requests are issued by (possibly different extenders):
- fetch=true (because the caller 1 wants up-to-date results)
- without fetch parameter (caller 2 just wants to know the info and does not require the most up-to-date version)

The result will be two concurrent calls to the backend for the same information.

jace-roell: I see

the current implementation does not have the granularity to separate the remoteLookup and lookup call since they are both within the wrapped promise. It would be a large refactor to separate the calls such that the `fetch=false` fallback remoteLookup call and `fetch=true` remoteLookup call would be reused.

packages/zowe-explorer/src/trees/uss/UssFSProvider.ts
```diff
@@ -978,4 +1045,73 @@ export class UssFSProvider extends BaseProvider implements vscode.FileSystemProv
         const uriInfo = FsAbstractUtils.getInfoForUri(uri, Profiles.getInstance());
         return { profile: uriInfo.profile, path: uriInfo.isRoot ? "/" : uri.path.substring(uriInfo.slashAfterProfilePos) };
     }
+
+    /**
+     * Helper to clean URI string for cache keys (removes trailing slash)
+     */
+    private getCleanUriString(uri: vscode.Uri): string {
+        const s = uri.toString();
+        return s.endsWith("/") ? s.slice(0, -1) : s;
```
slavek-kucera: Was this meant to check just `uri.path`?

packages/zowe-explorer/src/trees/uss/UssFSProvider.ts
```diff
@@ -978,4 +1045,73 @@ export class UssFSProvider extends BaseProvider implements vscode.FileSystemProv
...
+    }
+
+    private async executeWithReuse<T>(
```
slavek-kucera: This is nearly identical to the version in datasetftprovider.ts.

packages/zowe-explorer-api/__tests__/__unit__/fs/BaseProvider.unit.test.ts
```diff
@@ -700,3 +700,174 @@ describe("onCloseEvent", () => {
...
+            checkLocalSpy.mockReturnValue(false);
+            keyGenSpy.mockImplementation((u: vscode.Uri) => u.query);
+
+            await prov.executeWithReuse(testUri, {
+                keyGenerator: keyGenSpy,
+                checkLocal: checkLocalSpy,
+                execute: executeSpy,
+            });
+
+            expect(executeSpy).toHaveBeenCalled();
```
t1m0thyj: (nit) Should we expect the request cache to contain an entry for the URI?
```typescript
expect(prov.requestCache.has(testUri.toString())).toBe(true);
```

packages/zowe-explorer-api/src/utils/FeatureFlags.ts
```diff
@@ -0,0 +1,98 @@
...
+    /**
+     * Sets a flag value and persists it to the JSON file.
+     * Updates in-memory cache immediately.
+     */
+    public static async set(key: string, value: any, save: boolean = false): Promise<void> {
+        this.flags[key] = value;
+        if (save) {
+            await this.saveToDisk();
+        }
+    }
```
t1m0thyj, line 79: I believe this allows any ZE extender to change the value of any feature flag. If so, are we ok with this or should we restrict access to editing feature flags with an ACL like we do for local storage?

packages/zowe-explorer/__tests__/__unit__/trees/dataset/DatasetFSProvider.unit.test.ts
```diff
@@ -95,7 +95,7 @@ const testUris: TestUris = {
 
 describe("DatasetFSProvider", () => {
     let mockedProperty: MockedProperty;
-    beforeEach(() => {
+    beforeEach(async () => {
```
t1m0thyj: (nit) The `async` keyword seems unnecessary here.

packages/zowe-explorer/__tests__/__unit__/trees/dataset/DatasetFSProvider.unit.test.ts
```diff
@@ -1089,6 +1353,151 @@ describe("DatasetFSProvider", () => {
...
+            const remoteLookupForResourceSpy = jest.spyOn(DatasetFSProvider.instance, "remoteLookupForResource");
+            const lookupSpy = jest.spyOn(DatasetFSProvider.instance, "lookup");
+            await DatasetFSProvider.instance.stat(fetchUri);
+            await DatasetFSProvider.instance.stat(localUri);
+
+            expect(remoteLookupForResourceSpy).toHaveBeenCalledWith(fetchUri.with({ path: "/sestest/USER.DATA.PDS" }));
+            expect(lookupSpy).toHaveBeenCalledWith(fetchUri.with({ path: "/sestest/USER.DATA.PDS" }), false);
+        });
+
+        it("should make a system call and if fetchByDefault is enabled and the entry is not in the cache", async () => {
```
t1m0thyj:
```suggestion
        it("should make a system call if fetchByDefault is enabled and the entry is not in the cache", async () => {
```

packages/zowe-explorer/__tests__/__unit__/utils/AuthUtils.unit.test.ts
```diff
@@ -259,7 +259,14 @@ describe("AuthUtils", () => {
...
                 const runSequentialMock = jest.spyOn(AuthHandler, "runSequentialIfEnabled").mockImplementation((_profile, action) => action());
                 try {
-                    const statResult = await DatasetFSProvider.instance.stat(testUris.ps);
+                    const statResult = await DatasetFSProvider.instance.stat(
+                        Uri.from({
+                            scheme: ZoweScheme.DS,
+                            path: "/sestest/USER.DATA.PS",
+                        }).with({
+                            query: "fetch=true",
+                        })
```
t1m0thyj: (nit) Curious, why not:
```suggestion
                        testUris.ps.with({
                            query: "fetch=true",
                        })
```

packages/zowe-explorer/src/trees/dataset/DatasetFSProvider.ts
```diff
@@ -40,6 +42,8 @@ import { DatasetUtils } from "./DatasetUtils";
 import { AuthUtils } from "../../utils/AuthUtils";
 import { ProfilesUtils } from "../../utils/ProfilesUtils";
 
+const EXPECTED_MEMBER_LENGTH = 2; // /DATA.SET/MEMBER
```
t1m0thyj: (nit) If this constant is only used by `DatasetFSProvider` then it would be good practice to define it as a `private readonly` variable inside the class, rather than at the global level.

packages/zowe-explorer/src/trees/uss/UssFSProvider.ts
```diff
@@ -301,6 +344,7 @@ export class UssFSProvider extends BaseProvider implements vscode.FileSystemProv
     }
 
     public async remoteLookupForResource(uri: vscode.Uri): Promise<UssDirectory | UssFile> {
+        console.log("remoteLookupCalled");
```
t1m0thyj: This line doesn't have a TODO comment but should eventually be removed 😋

packages/zowe-explorer/src/trees/uss/UssFSProvider.ts
```diff
@@ -135,6 +154,19 @@ export class UssFSProvider extends BaseProvider implements vscode.FileSystemProv
...
     }
 
+    /**
+     * Returns file statistics about a given URI.
+     * @param uri A URI that must exist as an entry in the provider
+     * @returns A structure containing file type, time, size and other metrics
+     */
+    public async stat(uri: vscode.Uri): Promise<vscode.FileStat> {
+        return this.executeWithReuse<vscode.FileStat>(uri, {
+            keyGenerator: (u) => "stat_" + this.getQueryKey(u) + this.getCleanUriString(u),
```
t1m0thyj: (nit) Why do stat operations for data sets use the verb "list" in the key, but for USS files they use the verb "stat" instead?

packages/zowe-explorer-api/src/utils/FeatureFlags.ts
```diff
@@ -0,0 +1,192 @@
...
```
zFernand0, line 79: Can we add TSDoc for the parameters in this function?
I guess I would like to let people know about what the `save` parameter will do. 😋 

Without it, people don't really have to await this `set` call 😋 

---

Also, what was the resolution of Timothy's comment?

>I believe this allows any ZE extender to change the value of any feature flag. If so, are we ok with this or should we restrict access to editing feature flags with an ACL like we do for local storage?

Any security concerns around this?

packages/zowe-explorer-api/src/utils/FileManagement.ts
```diff
@@ -73,6 +73,7 @@ export class FileManagement {
         );
         for (const folder of foldersWithProfile) {
             try {
+                // TODO: FEATURE-FLAG(fetchByDefault): remove fetch=true
                 await workspace.fs.stat(folder.uri.with({ query: "fetch=true" }));
```
zFernand0, line 77: Having this FeatureFlag class will be beneficial for future endeavors 🙏

packages/zowe-explorer-api/src/utils/FeatureFlags.ts
jace-roell: I have implemented an ACL to the `FeatureFlags` class to specifically allow reads and writes for specific keys. Also, the tsDoc has been updated. Thanks @zFernand0 !

zFernand0 (approved):

packages/zowe-explorer/src/trees/uss/UssFSProvider.ts
```diff
@@ -135,6 +147,49 @@ export class UssFSProvider extends BaseProvider implements vscode.FileSystemProv
...
+                if (parentPath && parentPath !== cleanUri && parentPath !== "/") {
+                    const parentKey = "list_" + queryKey + parentPath;
+                    if (this.requestCache.has(parentKey)) {
+                        return parentKey;
+                    }
+                }
+
+                return selfKey;
+            },
+            checkLocal: () => !!this.lookup(uri, false),
```
slavek-kucera: This throws on failure.

packages/zowe-explorer/src/trees/uss/UssFSProvider.ts
```diff
@@ -527,6 +615,26 @@ export class UssFSProvider extends BaseProvider implements vscode.FileSystemProv
...
+     * @param uri The URI pointing to a valid file on the remote system
+     * @returns The file's contents as an array of bytes
+     */
+    public async readFile(uri: vscode.Uri): Promise<Uint8Array> {
+        return this.executeWithReuse<Uint8Array>(uri, {
+            keyGenerator: (u) => "readFile_" + this.getQueryKey(u) + this.getCleanUriString(u),
+            checkLocal: () => {
+                try {
+                    const entry = this._lookupAsFile(uri, { silent: true }) as UssFile;
+                    return entry && entry.wasAccessed;
```
slavek-kucera: This does not throw... but also it does not work correctly in the non-throwing case.

packages/zowe-explorer/src/trees/uss/UssFSProvider.ts
```diff
@@ -979,4 +1087,11 @@ export class UssFSProvider extends BaseProvider implements vscode.FileSystemProv
         const uriInfo = FsAbstractUtils.getInfoForUri(uri, Profiles.getInstance());
         return { profile: uriInfo.profile, path: uriInfo.isRoot ? "/" : uri.path.substring(uriInfo.slashAfterProfilePos) };
     }
+
+    /**
+     * Helper to clean URI string for cache keys (removes trailing slash)
+     */
+    private getCleanUriString(uri: vscode.Uri): string {
```
slavek-kucera:
```suggestion
    private pathWithoutTrailingSlash(uri: vscode.Uri): string {
```

packages/zowe-explorer/src/trees/uss/UssFSProvider.ts
```diff
@@ -218,6 +273,17 @@ export class UssFSProvider extends BaseProvider implements vscode.FileSystemProv
         };
     }
 
+    private _createDirectoryRecursive(uri: vscode.Uri): void {
```
slavek-kucera, line 278: Just out of curiosity, what happens when I ask for `/a/b/c/d/../../../e`?

packages/zowe-explorer/src/trees/uss/UssFSProvider.ts
```diff
@@ -57,15 +59,25 @@ export class UssFSProvider extends BaseProvider implements vscode.FileSystemProv
         return UssFSProvider._instance;
     }
 
-    /* Public functions: File operations */
+    protected async lookupWithCache(uri: vscode.Uri): Promise<UssDirectory | UssFile | IFileSystemEntry> {
+        try {
+            // Check cache for resource
+            const localLookup = this.lookup(uri);
+            if (localLookup) return localLookup;
+        } catch {}
```
slavek-kucera:
```suggestion
        // Check cache for resource
        const localLookup = this.lookup(uri, true);
        if (localLookup) return localLookup;
```

packages/zowe-explorer/src/trees/dataset/DatasetFSProvider.ts
```diff
@@ -620,14 +670,34 @@ export class DatasetFSProvider extends BaseProvider implements vscode.FileSystem
...
+    public async readFile(uri: vscode.Uri): Promise<Uint8Array> {
+        return this.executeWithReuse<Uint8Array>(uri, {
+            keyGenerator: (u) => "readFile" + this.getQueryKey(u) + "_" + u.toString().replace(/\/$/, ""),
+            checkLocal: () => {
+                try {
+                    const entry = this._lookupAsFile(uri, { silent: true }) as DsEntry;
+                    return entry && entry.wasAccessed;
+                } catch {
+                    return false;
+                }
```
slavek-kucera: The same here, this does not throw and does not work.

packages/zowe-explorer/src/trees/dataset/DatasetFSProvider.ts
```diff
@@ -393,38 +438,47 @@ export class DatasetFSProvider extends BaseProvider implements vscode.FileSystem
...
+
+    /**
+     * Reads a directory located at the given URI.
+     * @param uri A valid URI within the provider
+     * @returns An array of tuples containing each entry name and type
+     */
+    public async readDirectory(uri: vscode.Uri): Promise<[string, vscode.FileType][]> {
+        const dirEntry = await this.executeWithReuse<DirEntry>(uri, {
+            keyGenerator: (u) => "list" + this.getQueryKey(u) + "_" + u.toString().replace(/\/$/, ""),
+            checkLocal: () => !!this._lookupAsDirectory(uri, true),
```
slavek-kucera: returns `true` for files.

packages/zowe-explorer/src/trees/dataset/DatasetFSProvider.ts
```diff
@@ -131,34 +140,64 @@ export class DatasetFSProvider extends BaseProvider implements vscode.FileSystem
...
+        const isMemberRequest = segments.length === 3;
+
+        if (isMemberRequest) {
+            const memberName = segments[2];
+            const parentPath = segments.slice(0, 2).join("/");
+            const parentUri = uri.with({ path: `/${parentPath}` });
+
+            const pdsEntry = await this.executeWithReuse<DirEntry>(parentUri, {
+                keyGenerator: (u) => "list" + this.getQueryKey(u) + "_" + u.toString().replace(/\/$/, ""),
+                checkLocal: () => !!this._lookupAsDirectory(parentUri, true),
```
slavek-kucera: returns `true` for files.

packages/zowe-explorer/src/trees/uss/UssFSProvider.ts
jace-roell: It seems that the request will throw an error due to the file system `type` property being improperly set. So I would not recommend using dot segments in FS calls.

zFernand0: I'm about done with reviewing this PR.
Your recommendation here makes sense to me! 🙏

packages/zowe-explorer/src/trees/dataset/DatasetFSProvider.ts
```diff
@@ -66,33 +68,42 @@ export class DatasetFSProvider extends BaseProvider implements vscode.FileSystem
         return new vscode.Disposable(() => {});
     }
 
+    protected async lookupWithCache(uri: vscode.Uri): Promise<DirEntry | DsEntry | IFileSystemEntry> {
+        try {
+            // Check cache for resource
+            const localLookup = this.lookup(uri);
+            if (localLookup) return localLookup;
+        } catch {}
```
traeok: The USS counterpart of this function does not use an empty catch block - it just passes `silent: true` to the `lookup` call. Should we do that here as well?

packages/zowe-explorer/src/trees/dataset/DatasetFSProvider.ts
```diff
@@ -131,34 +140,65 @@ export class DatasetFSProvider extends BaseProvider implements vscode.FileSystem
...
+        const segments = uri.path.split("/").filter((s) => s.length > 0);
+        const isMemberRequest = segments.length === 3;
+
+        if (isMemberRequest) {
+            const memberName = segments[2];
+            const parentPath = segments.slice(0, 2).join("/");
+            const parentUri = uri.with({ path: `/${parentPath}` });
+
+            const pdsEntry = await this.executeWithReuse<DirEntry>(parentUri, {
+                keyGenerator: (u) => "list" + this.getQueryKey(u) + "_" + u.toString().replace(/\/$/, ""),
```
traeok, line 180: The key generator is a bit different than the USS one, although it may be necessary due to the nuances between the two providers 😋 

Is there opportunity to standardize the key prefix format in some way across both?

packages/zowe-explorer-api/src/fs/BaseProvider.ts
```diff
@@ -32,6 +32,8 @@ export class BaseProvider {
 
     protected constructor() {}
 
+    public requestCache = new Map<string, Promise<any>>();
```
traeok: (nitpick) I would recommend using `Promise<unknown>` instead of `Promise<any>` if possible. It's nothing urgent, but technically you could await any request in this cache and treat the result as any type. `unknown` forces you to check or assert the type.

packages/zowe-explorer/src/trees/uss/UssFSProvider.ts
```diff
@@ -57,15 +59,23 @@ export class UssFSProvider extends BaseProvider implements vscode.FileSystemProv
...
-     * @param uri A URI that must exist as an entry in the provider
-     * @returns A structure containing file type, time, size and other metrics
+     * Executes the core logic for the stat operation on a given URI.
+     * This is separated to facilitate caching and testing.
+     * @param uri The URI of the resource to stat.
+     * @returns A promise that resolves to a vscode.FileStat object.
      */
-    public async stat(uri: vscode.Uri): Promise<vscode.FileStat> {
-        ZoweLogger.trace(`[UssFSProvider] stat called with ${uri.toString()}`);
+    private async statImplementation(uri: vscode.Uri, isActiveEditor: boolean = false): Promise<vscode.FileStat> {
```
traeok: (nitpick) Would it make sense to call the boolean parameter `isVisibleEditor` for consistency? The DS `stat` uses `isVisibleEditor` but the USS `statImplementation` parameter is named `isActiveEditor`.


```suggestion
    private async statImplementation(uri: vscode.Uri, isVisibleEditor: boolean = false): Promise<vscode.FileStat> {
```

packages/zowe-explorer/src/trees/dataset/DatasetFSProvider.ts
```diff
@@ -131,34 +138,72 @@ export class DatasetFSProvider extends BaseProvider implements vscode.FileSystem
...
+    public async stat(uri: vscode.Uri): Promise<vscode.FileStat> {
+        const segments = uri.path.split("/").filter((s) => s.length > 0);
+        const isMemberRequest = segments.length === 3;
+
+        const targetPrefix = uri.path.split("/").slice(0, 3).join("/") + "/";
+
+        const isVisibleEditor = vscode.window.visibleTextEditors.some((editor) => {
+            const editorUri = editor.document.uri;
+            return editorUri.toString() === uri.toString() || editorUri.path.startsWith(targetPrefix);
+        });
```
traeok: The scope of what is considered a "visible editor" is a bit broader here than USS, which could be expected. Do we intend this to just cover any member of a PDS? Or could the check be more targeted?

For context, if a PS data set is cached locally, it would always return the cached entry here without checking for `mtime` changes. Could the same `isVisibleEditor` pattern be applied to the PS path?

packages/zowe-explorer-api/CHANGELOG.md
```diff
@@ -6,6 +6,7 @@ All notable changes to the "zowe-explorer-api" extension will be documented in t
 
 ### New features and enhancements
 
+- Implemented request caching capabilities for filesystem providers that extend the `BaseProvider` class. [#3844](https://github.com/zowe/zowe-explorer-vscode/issues/3554)
```
t1m0thyj, line 9: Should we also mention in changelog the addition of the `FeatureFlags` class?

Issue #3859, opened 2025-09-15
Submit job with polling


adam-wolfe:
**Is your feature request related to a problem? Please describe.**

To use the new job polling functionality introduced in 3.3.0, I need to ensure that the job that I want to poll is shown in the tree views and then right-click on the profile node and select polling. It would be convenient if I could submit a job with polling directly from the context menu.

**Describe the solution you'd like**

Add an item to the data set/member context menu and editor context menu to submit job/jcl with polling so that the user is notified when the job completes and is able to jump to the job in the tree.

Maybe something like "Submit Job (open when complete)". Not sure of wording, but the idea being that the user can submit a job and be notified when the job completes with a link that lets them jump to the job in the tree.

**Describe alternatives you've considered**

**Additional context**

JWaters02:
> For the "poll active jobs" behaviour which was introduced in zowe explorer 3.3, I think from experimenting that it looks for new jobs starting that match the current jobs filter (It doesn't seem to poll all of the spool files within the matching jobs).

Yup sorry I'm adding in docs for poll active jobs now, but it does not poll any spool files within jobs, only the statuses of the jobs themselves (I mean, all it does, is just automatically refresh the jobs tree for the set filter on the poll interval so you dont have to do it manually, and sends notification of ret code when a job completes). You can combine it with polling spool files within jobs being polled though.

> So I don't think being able to open the just started job (which sets the filter to a very specific job ID) and start polling would achieve much, because another job with that filter will never start up.

I understand what you are saying and I think this is a fair point but from experience sometimes when I submit a job I want to get a notification when it completes so I don't have to keep refreshing the tree manually. So being able to poll on an individually submitted job would be nice - and even if you wanted to poll on different filters (while polling on a specific job ID filter), you can still poll on favourited jobs filters at the same time!

PR #3967, opened 2025-12-08
feat(jobs): Submit job with polling

JWaters02:
## Proposed changes

Implements #3859 

Also shows `JOBNAME(JOBID)` in the job submitted notification.

https://github.com/user-attachments/assets/3b828d7e-9d41-49e1-b433-5f0b6b9b12c3

## Release Notes

Milestone: 3.5.0

Changelog: Added a new **Start Polling** button to the job submitted notification which polls on the job status and notify when the job is complete. The poll interval can be changed in the setting `Zowe > Jobs: Poll Interval`

## Types of changes

## Checklist

**General**

**Code coverage**

**Deployment**

### DIFF
--- a/packages/zowe-explorer/CHANGELOG.md
+++ b/packages/zowe-explorer/CHANGELOG.md
@@ -34,6 +34,8 @@ All notable changes to the "vscode-extension-for-zowe" extension will be documen
 - Fixed issue where the Show Attributes view displayed "Open virtualization format" under data set attributes instead of "Space overflow" for the `ovf` attribute key. [#3985](https://github.com/zowe/zowe-explorer-vscode/issues/3985)
 - Fixed release notes displaying when Zowe Explorer is installed for the first time. [#3966](https://github.com/zowe/zowe-explorer-vscode/pull/3966)
 - Fixed release notes displaying when the Zowe Explorer version is downgraded. [#3966](https://github.com/zowe/zowe-explorer-vscode/pull/3966)
+- Fixed issue where favorited items caused a failure during initialization, specifically when the favorited item belongs to a profile type that's no longer registered or accessible. [#3962](https://github.com/zowe/zowe-explorer-vscode/pull/3962)
+- Fixed issue where the filesystem throws a vague `Unknown` FileSystemError if an API cannot be loaded for the given profile type. Now, if the API is inaccessible for the given profile type, an `Unavailable` error is thrown from the filesystem with more context. [#3962](https://github.com/zowe/zowe-explorer-vscode/pull/3962)
 
 ## `3.3.1`
 
--- a/packages/zowe-explorer-api/CHANGELOG.md
+++ b/packages/zowe-explorer-api/CHANGELOG.md
@@ -12,6 +12,7 @@ All notable changes to the "zowe-explorer-api" extension will be documented in t
 - Added support to delete VSAM data sets for z/OSMF type profiles. [#3824](https://github.com/zowe/zowe-explorer-vscode/issues/3824)
 - Added support for loading credential manager options from the `imperative.json` file. Added a `credentialManagerOptions` object in the JSON object in `imperative.json` to specify options for the current credential manager. [#3935](https://github.com/zowe/zowe-explorer-vscode/pull/3935)
 - Added support for custom persistence levels for Windows (persist option) to support the credential manager in less permissive environments. For more information on how to configure this option, see the "Troubleshooting Zowe CLI credentials" on [Zowe Docs](https://docs.zowe.org/stable/troubleshoot/cli/troubleshoot-cli-credentials/#secrets-sdk-persistence-level-for-windows). [#3935](https://github.com/zowe/zowe-explorer-vscode/pull/3935)
+- Added a new filesystem helper function named `getApiOrThrowUnavailable` for generalized error handling with API access. When the given API getter fails due to a non-existing profile, an `Unavailable` FileSystemError is thrown with additional context on the encountered error. [#3962](https://github.com/zowe/zowe-explorer-vscode/pull/3962)
 
 ### Bug fixes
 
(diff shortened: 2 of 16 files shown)


davenice:
This is nice. I would use it! Did you consider something like "Notify on job completion" or "Notify me when the job completes" as the button?

I think "Start polling" focusses on the mechanism we're using under the covers, but I don't think as a user that I'd necessarily understand that I'll be notified when the job ends. I half expected that when I clicked on that perhaps you'd open the job and poll somehow 🤔 

I can understand that there might be some caveats around this. Or that possibly you've tried to tie it in with other uses of polling around the interface. What do you think?

JWaters02:
> This is nice. I would use it! Did you consider something like "Notify on job completion" or "Notify me when the job completes" as the button?
> 
> I think "Start polling" focusses on the mechanism we're using under the covers, but I don't think as a user that I'd necessarily understand that I'll be notified when the job ends. I half expected that when I clicked on that perhaps you'd open the job and poll somehow 🤔
> 
> I can understand that there might be some caveats around this. Or that possibly you've tried to tie it in with other uses of polling around the interface. What do you think?

I actually had it called "Notify When Complete" at first, but the naming felt wrong, so I changed it to what is consistent around the rest of the UI. I do agree with you that the term "polling" is a bit technical and people might not know what it means, but also we use the term in several other places so idk. I'm fine with w/e. More feedback from others on what they think would be good too.

packages/zowe-explorer/CHANGELOG.md
```diff
@@ -16,6 +16,7 @@ All notable changes to the "vscode-extension-for-zowe" extension will be documen
...
 - Added support for custom persistence levels for Windows (persist option) to support the credential manager in less permissive environments. For more information on how to configure this option, refer to the ["Troubleshooting Zowe CLI credentials"](https://docs.zowe.org/stable/troubleshoot/cli/troubleshoot-cli-credentials/#secrets-sdk-persistence-level-for-windows) page on Zowe Docs. [#3935](https://github.com/zowe/zowe-explorer-vscode/pull/3935)
 - Updated minimum VS Code version from 1.79 to 1.90. We are dropping support for VS Code releases that bundle versions of Node.js no longer receiving security updates. [#3684](https://github.com/zowe/zowe-explorer-vscode/issues/3684)
+- Added a new **Start Polling** button to the job submitted notification which polls on the job status and notify when the job is complete. The poll interval can be changed in the setting `Zowe > Jobs: Poll Interval`. [#3967](https://github.com/zowe/zowe-explorer-vscode/pull/3967)
```
traeok: (nitpick) the bolded text for "Start Polling" seems out of place for our changelogs, capitalizing it & leaving it as unformatted should be fine

packages/zowe-explorer/src/trees/dataset/DatasetActions.ts
```diff
@@ -2434,4 +2444,86 @@ export class DatasetActions {
...
+                const statusMsg = Gui.setStatusBarMessage(
+                    `$(sync~spin) ${vscode.l10n.t({
+                        message: "Polling job: {0}...",
+                        args: [displayName],
+                        comment: ["Job display name"],
+                    })}`,
+                    Constants.STATUS_BAR_TIMEOUT_MS
+                );
+
+                try {
```
traeok, line 2485: You may be able to avoid calling dispose multiple times by using a `finally` block with this try/catch

packages/zowe-explorer/src/trees/dataset/DatasetActions.ts
```diff
@@ -1476,13 +1482,17 @@ export class DatasetActions {
...
-                    vscode.l10n.t({
-                        message: "Job submitted {0}",
-                        args: [`[${job.jobid}](${setJobCmd})`],
-                        comment: ["Job ID and set job command"],
-                    })
-                );
+                const notifyButton = vscode.l10n.t("Start Polling");
+                const jobDisplay = `${job.jobname}(${job.jobid})`;
+                const message = vscode.l10n.t({
+                    message: "Job submitted {0}",
```
traeok: (nit) I know this "job submitted" text isn't new, but I wonder if we should add a colon after "Job submitted" and then the info message would look a little more polished:

```suggestion
                    message: "Job submitted: {0}",
```

packages/zowe-explorer/src/trees/dataset/DatasetActions.ts
```diff
@@ -1238,13 +1240,17 @@ export class DatasetActions {
...
+                    args: [`[${jobDisplay}](${setJobCmd})`],
+                    comment: ["Job name and ID with set job command"],
+                });
+                const selection = await Gui.showMessage(message, { items: [notifyButton] });
+                if (selection === notifyButton) {
+                    DatasetActions.pollSubmittedJob(sessProfile, sessProfileName, job.jobid, job.jobname);
+                }
```
traeok: The `ZoweLogger.info` call should be moved above this dialog. Now that we are awaiting the `showMessage` call, it won't write to the logger until the user responds w/ their selection (if they ever do)

JWaters02:
> Poll for job completion

Poll for job completion sounds like a great idea :D, I will go with that unless anyone else has anything to add

packages/zowe-explorer/src/trees/dataset/DatasetActions.ts
```diff
@@ -1240,24 +1240,24 @@ export class DatasetActions {
...
                     vscode.l10n.t({
                         message: "Job submitted {0} using profile {1}.",
                         args: [job.jobid, sessProfileName],
                         comment: ["Job ID", "Profile name"],
                     })
                 );
+                const selection = await Gui.showMessage(message, { items: [notifyButton] });
+                if (selection === notifyButton) {
+                    DatasetActions.pollSubmittedJob(sessProfile, sessProfileName, job.jobid, job.jobname);
+                }
```
traeok: (nitpick) The current code blocks this function until the user interacts with the notification. Since the job has already been submitted at this point, consider using `.then()` instead of `await` for the `Gui.showMessage` call - this allows the function to return immediately while still handling the user's response if they click the button. The same applies to the `submitMember` change on line 1499.

JWaters02: Ooh yeah good point, will change this

JillieBeanSim (approved):
LGTM, tested submit job and JCL. thanks @JWaters02

Issue #3948, opened 2025-11-24
Moving a source PDS to target PDS with the same name and volume does not prompt with overwrite option


pujal0909:
**Describe the bug**
Moving a PDS from system A to system B fails when the source PDS and target PDS have the same name and volume. 
This also happens when moving one member of a PDS to another PDS on the same system.
This error appears.

<img width="541" height="162" alt="Image" src="https://github.com/user-attachments/assets/5c5fab1e-f7a9-4883-b587-141bfaf140af" />

**To Reproduce**

1. Create PDSs with the same name and volume on two different systems
2. Move PDS over from one system to another.

**Expected behavior**

I would expect to see a prompt giving me the option to overwrite the target PDS similar to when moving a sequential data set.

**Screenshots**

**Desktop (please complete the following information):**

- OS:
- Zowe Explorer Version:
- (Optional) Zowe CLI Version:
- (Optional) Do you have secure credentials enabled?
- (Optional) Are you using a custom credential manager?

**Additional context**

ArghadeepDas-ibm:
Hi Team,
I’d like to work on resolving the issue.

JillieBeanSim:
closing with delivery of PR https://github.com/zowe/zowe-explorer-vscode/pull/3972

PR #3827, opened 2025-08-25
Preventing Same-LPAR / Same-Dataset Drag-and-Drop

ATorrise:
## Release Notes

Milestone:

Changelog:  
Bug Fix: Prevent drag-and-drop between profiles pointing to the same LPAR/dataset. The operation is now blocked and an error message is shown instead; users should refresh the target profile to see any changes.

## Types of changes

- [X] Bug fix (non-breaking change which fixes an issue)
- [ ] Enhancement (non-breaking change which adds or improves functionality)
- [ ] Breaking change (a change that would cause existing functionality to not work as expected)
- [ ] Documentation (Markdown, README updates)
- [ ] Other (please specify above in "Proposed changes" section)

## Checklist

**General**

- [X] I have read the [CONTRIBUTOR GUIDANCE](https://github.com/zowe/zowe-explorer-vscode/wiki/Best-Practices:-Contributor-Guidance) wiki
- [X] All PR dependencies have been merged and published (if applicable)
- [X] A GIF or screenshot is included in the PR for visual changes
- [X] The pre-publish command has been executed:
  - **v2 and below:** `yarn workspace vscode-extension-for-zowe vscode:prepublish`
  - **v3:** `pnpm --filter vscode-extension-for-zowe vscode:prepublish`

**Code coverage**

- [ ] There is coverage for the code that I have added
- [ ] I have added new test cases and they are passing
- [ ] I have manually tested the changes

**Deployment**

- [ ] I have tested new functionality with the FTP extension and profile verifying no extender profile type breakages introduced
- [ ] I have added developer documentation (if applicable)
- [ ] Documentation should be added to Zowe Docs
  - If you're an outside contributor, please post in the [#zowe-doc Slack channel](https://openmainframeproject.slack.com/archives/CC961JYMQ) to coordinate documentation.
  - Otherwise, please check with the rest of the squad about any needed documentation before merging.
- [ ] These changes may need ported to the appropriate branches (list here):
image

PR #3972, opened 2025-12-08
fix : moving PDS with same name and volume #3948

ArghadeepDas-ibm:
## Proposed changes

Closing : [#3948]

To test:
- Create PDSs with the same name and volume on two different systems
- Move PDS over from one system to another.
- Prompt comes up to confirm whether to complete the move 
- Select the "Confirm" option.
- Dataset should move to the target system without the error being generated. 

## Release Notes

Milestone: 3.4.1

Changelog: Fixed issue where moving a PDS from one system to another which already has a dataset with the same name and volume is bringing up an error.

## Types of changes

## Checklist

**General**

**Code coverage**

**Deployment**

### DIFF
--- a/packages/zowe-explorer-api/__tests__/__unit__/profiles/ProfilesCache.unit.test.ts
+++ b/packages/zowe-explorer-api/__tests__/__unit__/profiles/ProfilesCache.unit.test.ts
@@ -722,7 +722,7 @@ describe("ProfilesCache", () => {
                 subject: "CN=Test",
                 fingerprint256: "mockedFingerprint",
                 validFrom: "2022-01-01",
-                validTo: "2026-01-01",
+                validTo: `${new Date().getFullYear() + 1}-01-01`,
             }));
             const profCache = new ProfilesCache(fakeLogger as unknown as imperative.Logger, __dirname);
             expect(profCache.isCertFileValid("mockedCertFilePath")).toBe(true);


.eslintrc.yaml
```diff
@@ -83,7 +83,7 @@ rules:
     - warn
     - code: 150
   # no-array-constructor: off
-  no-console: error
```
awharn, line 86: I don't believe we want to allow console statements in the code. Perhaps this was disabled during testing and needs to be re-enabled?

packages/zowe-explorer/src/trees/dataset/DatasetTree.ts
```diff
@@ -233,12 +233,6 @@ export class DatasetTree extends ZoweTreeProvider<IZoweDatasetTreeNode> implemen
             if (!node) continue;
             const srcDsn = SharedUtils.getNodeProperty(node, "label");
 
-            // 1. Same-object check
```
awharn, line 236: This seems to re-introduce an issue where, if a user has two profiles that refer to the same system and the same DASD, and a data set is copied from one profile tree node to another, it causes the dataset to be deleted without any confirmation.

packages/zowe-explorer/src/trees/dataset/DatasetTree.ts
```diff
@@ -233,12 +233,6 @@ export class DatasetTree extends ZoweTreeProvider<IZoweDatasetTreeNode> implemen
...
-            if (node.getProfile && target.getProfile && (await SharedUtils.isSamePhysicalDataset(node.getProfile(), target.getProfile(), srcDsn))) {
-                Gui.errorMessage(vscode.l10n.t(SharedUtils.ERROR_SAME_OBJECT_DROP));
-                return;
-            }
```
t1m0thyj, line 240: Thanks @ArghadeepDas-ibm for looking into this!

Agree with @awharn that we want to keep the prompt associated with this check (that was added in #3827), and add a button to allow users to bypass/override the same-DASD warning.

packages/zowe-explorer/src/trees/shared/SharedUtils.ts
```diff
@@ -628,8 +628,11 @@ export class SharedUtils {
 
             const volsAreEqual = srcVols.length === dstVols.length && srcVols.every((vol: any, idx: number) => vol === dstVols[idx]);
 
-            // if both name and vols match, they're the same dataset
-            return namesAreEqual && volsAreEqual;
+            //Compare system config to determine target and source datasets are on same system
+            const systemIsEqual = srcProfile.profile?.host === dstProfile.profile?.host;
+
+            // if name, vols and host match, they're the same dataset
+            return namesAreEqual && volsAreEqual && systemIsEqual;
```
traeok, line 639: (nitpick) What if one profile is going through API ML and the other is direct to service? The hostnames could be different but the API ML instance may point to the same system.

We discussed this in the past and I think we can accept that possibility, but we should consider having a comment above the declaration of `systemIsEqual` for clarity.

traeok (commented):
Thanks for looking into this @ArghadeepDas-ibm, left a quick suggestion regarding the system check

packages/zowe-explorer/CHANGELOG.md
```diff
@@ -29,6 +29,7 @@ All notable changes to the "vscode-extension-for-zowe" extension will be documen
 - Fixed an issue where user attempts to perform a search on an individual PDS from the Zowe-Explorer extension using the "Search PDS members" action and query a search string, it brings out additional dataset as a part of the result.[#3682](https://github.com/zowe/zowe-explorer-vscode/issues/3682)
 - Fixed an issue where loading of virtual workspaces interfered with the activation of some Zowe Explorer extensions. [#3923](https://github.com/zowe/zowe-explorer-vscode/pull/3923)
 - Fixed issue seen by extender profile types where a type error was returned with 401 instead of notification with update option. [#3943](https://github.com/zowe/zowe-explorer-vscode/issues/3943)
+- Fixed issue where moving a PDS from one system to another which already has a dataset with the same name and volume is bringing up an error. [#3948](https://github.com/zowe/zowe-explorer-vscode/issues/3948)
```
```suggestion
- Fixed issue where moving a PDS from one system to another, which already had a data set with the same name and volume, resulted in an error. [#3948](https://github.com/zowe/zowe-explorer-vscode/issues/3948)
```

packages/zowe-explorer/src/trees/shared/SharedUtils.ts
ArghadeepDas-ibm: Hi @traeok ,
Should we only add a clarifying comment above systemIsEqual, or should we also update the logic to explicitly handle the edge case where one profile connects via API ML and the other connects directly to the service?

zFernand0: It would be nice to be able to handle more hostname related edge cases, but it's going to be hard to cover them all, e.g. proxies, dns resolution, aliases... 😅 
I think it's fine for now to keep this best-effort check with the known limitations 🙏

zFernand0 (approved):
LGTM! 😋

Issue #3911, opened 2025-10-24
Add option to auto-select TSO/SSH profile for issuing commands


adam-wolfe:
**Is your feature request related to a problem? Please describe.**

If I frequently issue TSO or SSH commands with Zowe Explorer (via right-clicking on a profile), I need to select the TSO/SSH profile to use each time. 

**Describe the solution you'd like**

Frequently we see configs that include profiles nested by LPAR followed by zosmf (e.g., `dev1.zosmf`). If I right-click on `dev1.zosmf` and choose to issue a TSO or SSH command, it would be helpful if Zowe Explorer could check for TSO or SSH profiles starting with `dev1.` to automatically pick the right TSO or SSH profile from my config.

We would need to consider behavior for cases where:
- there are more than 1 TSO or SSH profile within a parent profile
  - Prompt for user to select a profile (or fallback to default profile?)
- there are no TSO or SSH profiles within a parent profile
  - Prompt for user to select a profile (or fallback to default profile?)
- the command is being issued through an extension that uses different profile information (e.g., for Zowe Native Proto, we know which SSH profile to use.)

**Describe alternatives you've considered**

The option to always use the default TSO profile for issuing TSO commands is useful in cases where I do not need system/lpar-specific settings, but there is no such option for issuing SSH commands. 

**Additional context**

vtushar06:
@adam-wolfe @JillieBeanSim, I have gone through this issue and reproduced in local, should I raise PR for fix as changes are done.

JillieBeanSim:
Hey @vtushar06 yes please feel free to raise a PR with the fixes. Thank you for the contributions to the project ❤️

PR #3995, opened 2025-12-22
feat: Add option to auto-select TSO/SSH profile for issuing commands

vtushar06:
## Proposed changes

Implements auto-selection of TSO/SSH profiles based on parent profile naming convention. When a user right-clicks on a parent profile (e.g., "dev1.zosmf") to issue TSO or Unix commands, the extension will automatically select the corresponding service profile (e.g., "dev1.tso" or "dev1.ssh") if exactly one match exists. This eliminates the need for users to manually select profiles when following consistent naming conventions.

**Related Issue**: #3911

## Release Notes

Milestone: v3.5.0

Changelog:Add option to auto-select TSO/SSH profiles based on parent profile naming convention for improved command workflow

## Types of changes

## Checklist

**General**

**Code coverage**

**Deployment**

- [x] These changes may need ported to the appropriate branches (list here):

## Further comments

The implementation is backward compatible with no breaking changes. Users with non-standard profile naming conventions are unaffected—the extension falls back to prompting for manual selection. Auto-selection only occurs when exactly one matching profile is found.

### DIFF
--- a/packages/zowe-explorer/release-notes.md
+++ b/packages/zowe-explorer/release-notes.md
@@ -1,3 +1,39 @@
+## `3.4.0`
+
+### VS Code engine support change
+
+Updated minimum VS Code version from 1.79 to 1.90. We are dropping support for VS Code releases that bundle versions of Node.js no longer receiving security updates.
+
+### Credential Manager Updates
+
+#### Loading credential manager options
+
+Added support for loading credential manager options from the imperative.json file. Added a `credentialManagerOptions` object in the JSON object in `imperative.json` to specify options for the current credential manager.
+
+#### Windows custom persistence levels
+
+Added support for custom persistence levels for Windows (persist option) to support the credential manager in less permissive environments. For more information on how to configure this option, refer to the ["Troubleshooting Zowe CLI credentials"](https://docs.zowe.org/stable/troubleshoot/cli/troubleshoot-cli-credentials/#secrets-sdk-persistence-level-for-windows) page on Zowe Docs.
+
+### MVS data set enhancements
+
+#### Copy across LPAR updates
+
+Added support for copying data sets from multiple source LPARs at once in the cross-LPAR copy/paste functionality. Updated Zowe SDKs to version `8.28.0` to address an issue where copying a PDS member to a data set across LPARs failed. This occurred when the target PDS already contained members, but none matched the name of the PDS member being copied.
+
+#### VSAM support updates
+
+Added support to delete VSAM data sets via right-click action.
+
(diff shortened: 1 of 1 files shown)


packages/zowe-explorer/src/commands/ZoweCommandProvider.ts
```diff
@@ -174,6 +174,50 @@ export abstract class ZoweCommandProvider {
         }
     }
 
+    /**
+     * Try to auto-select a profile based on parent profile naming.
+     * E.g., if parent is "dev1.zosmf", look for "dev1.tso" or "dev1.ssh"
+     */
+    protected autoSelectProfile(
```
davenice: Left a comment in Slack. We have similar code in the CICS extension and initially started by breaking down the name of the profile to identify the parent, before switching to use `ProfileManagement.getProfilesCache().fetchBaseProfile(profile.name)`

https://github.com/zowe/cics-for-zowe-client/blob/245f904f8b9d5387546edff6d0e70a12faba2935/packages/vsce/src/utils/commandUtils.ts#L78

I'm wondering if we could align with the intention of making this an API. I think a common integration use case would be identifying connections that "are related" so it would be useful generally for extenders.

vtushar06:
@davenice Good point Switched to using fetchBaseProfile() instead of parsing the profile name manually.

The logic now matches what you're doing in the CICS extension - checking if profiles share the same base profile rather than string matching.

One small difference: I only auto-select when there's exactly one match. So if someone has both dev1.tso and dev1.ssh with the same base profile, they'll get prompted to choose. Seemed safer than just picking the first one.

Thanks for pointing this out - makes the code cleaner and more reliable.

vtushar06:
Hii @JillieBeanSim @adam-wolfe, guys can you please have a quick look into these changes and please let me know if any more modifications required

traeok:
Thanks @vtushar06, your changes sound reasonable and should address the concerns highlighted earlier. But there are some build failures that prevent me from compiling the latest changes. Please address those and I'll take a look once resolved.

vtushar06:
Hii @traeok @JillieBeanSim, any update over this guys whenever possible?

JillieBeanSim:
Hey @vtushar06 could you add more codecov to have that check pass please? the other failing checks will be handled by the squad as they are seen across many PRs

traeok (commented):
Thanks for addressing feedback. As Billie mentioned the patch coverage is a bit shy of the expected minimum (90%) - please address the gap in test coverage and I'll take a look as soon as possible.

PR #4001, opened 2026-01-05
fix: Moving an empty PDS no longer shows error

pujal0909:
## Proposed changes

When moving an empty PDS across LPARs, it no longer displays the following error.

Screenshot 2026-01-05 at 10 51 36 AM

## Release Notes

Milestone:

Changelog:

## Types of changes

## Checklist

**General**

- [ x I have read the [CONTRIBUTOR GUIDANCE](https://github.com/zowe/zowe-explorer-vscode/wiki/Best-Practices:-Contributor-Guidance) wiki

**Code coverage**

**Deployment**

### DIFF
--- a/packages/zowe-explorer/__tests__/__unit__/trees/dataset/DatasetTree.unit.test.ts
+++ b/packages/zowe-explorer/__tests__/__unit__/trees/dataset/DatasetTree.unit.test.ts
@@ -5651,7 +5651,7 @@ describe("DatasetTree.crossLparMove", () => {
             resourceUri: srcUri.with({
                 path: `${srcUri.path}/${vscode.l10n.t("No data sets found")}`,
             }),
-            contextValue: "member",
+            contextValue: "information",
             getEncoding: jest.fn().mockResolvedValue({ kind: "text" }),
             getChildren: jest.fn().mockResolvedValue([]),
         };


anaxceron (approved):

packages/zowe-explorer/src/trees/dataset/DatasetTree.ts
```diff
@@ -146,16 +146,18 @@ export class DatasetTree extends ZoweTreeProvider<IZoweDatasetTreeNode> implemen
...
+                        childNode,
+                        sourceUri.with({
+                            path: path.posix.join(sourceUri.path, childNode.getLabel() as string),
+                        }),
+                        destUri.with({
+                            path: path.posix.join(destUri.path, childNode.getLabel() as string),
+                        }),
+                        true
+                    );
+                }
```
traeok: (nitpick) you can avoid wrapping all the logic within the `for` loop by continuing early when the label is "No data sets found", e.g.:

```typescript
for (const childNode of children) {
    if ((childNode.getLabel() as string) === vscode.l10n.t("No data sets found")) {
        continue;
    }
    // perform move if we made it here
}
```

This would avoid an extra level of nesting, making the code easier to maintain

packages/zowe-explorer/src/trees/dataset/DatasetTree.ts
```diff
@@ -146,16 +146,18 @@ export class DatasetTree extends ZoweTreeProvider<IZoweDatasetTreeNode> implemen
...
-                    childNode,
-                    sourceUri.with({
-                        path: path.posix.join(sourceUri.path, childNode.getLabel() as string),
-                    }),
-                    destUri.with({
-                        path: path.posix.join(destUri.path, childNode.getLabel() as string),
-                    }),
-                    true
-                );
+                if ((childNode.getLabel() as string) !== vscode.l10n.t("No data sets found")) {
```
t1m0thyj: This check makes sense, but could unexpectedly break in the future if we change the wording of "No data sets found". It may be safer to check context value of node instead:

```suggestion
                if (!SharedContext.isInformation(childNode)) {
```

traeok (approved):
Tested and LGTM, thanks Pujal

packages/zowe-explorer/src/trees/dataset/DatasetTree.ts
traeok: Alternatively, we could assign the localized result of "No data sets found" to a single constant and reuse elsewhere 😋 That way if we do change the wording, the check would still pass. But I think this is also a good approach.

PR #4060, opened 2026-02-05
feat: Zowe Explorer setting that allows users to show or hide hidden files in the USS tree view

ArghadeepDas-ibm:
## Proposed changes

Introduced new Zowe Explorer setting that allows users to show or hide hidden files in the USS tree view. Default behaviour is not to show the hidden file in the USS tree, once we toggle the setting the USS tree refreshes and the hidden files come up.

Related issue :[3912](https://github.com/zowe/zowe-explorer-vscode/issues/3912)

image

## Release Notes

Milestone: 3.5.0

Changelog: Added a new VS Code toggle setting that allows users to show or hide hidden files in the USS tree view.

## Types of changes

## Checklist

**General**

**Code coverage**

**Deployment**


JillieBeanSim (approved):
LGTM! thanks @ArghadeepDas-ibm for adding this feature.

packages/zowe-explorer/__tests__/__unit__/trees/uss/ZoweUSSNode.unit.test.ts
```diff
@@ -1713,3 +1719,132 @@ describe("ZoweUSSNode Unit Tests - Function getUssFiles", () => {
...
+                parentPath: "/u",
+            }),
+            mockListFiles: jest.fn(),
+            mockFilterHiddenFiles: jest.fn(),
+        };
+
+        return newMocks;
+    }
+
+    it("should filter hidden files when ShowHiddenFiles.disabled is true", async () => {
```
JWaters02, line 1744: Still references old setting name

packages/zowe-explorer/__tests__/__unit__/trees/uss/ZoweUSSNode.unit.test.ts
```diff
@@ -1713,3 +1719,132 @@ describe("ZoweUSSNode Unit Tests - Function getUssFiles", () => {
...
+        await blockMocks.rootNode.getChildren();
+
+        expect(getDirectValueSpy).toHaveBeenCalledWith("zowe.files.showHiddenFiles");
+        expect(filterHiddenFilesSpy).not.toHaveBeenCalled();
+
+        getDirectValueSpy.mockRestore();
+        filterHiddenFilesSpy.mockRestore();
+    });
+
+    it("should not filter hidden files when ShowHiddenFiles.disabled is false", async () => {
```
JWaters02, line 1787: Still references old setting name

packages/zowe-explorer/src/extension.ts
JWaters02, line 1: This migration function still references the old setting name. It also still relies on the inverted setting. The case is also incorrect (it's currently `ShowHiddenFiles` but setting is `showHiddenFiles`). 

I also have some other comments about this function (sorry I didn't notice it before)!

1. The location of the function being in extension.ts is not ideal because there is migration logic in SettingsConfig, so I think it would make most sense to have it as a static method in that file and call it from the standardizeSettings function
2. If something breaks in the function, it stops the entire extension from loading properly because the rest of the code in extension.ts fails to load. So wrapping the logic in a try catch would allow it to fail gracefully.This happened in this instance where it was still referencing the old setting name so I got this error and the extension failed to activate:

Image

3. I think there is a bug where once on V4, every time the extension activates, it won't check the state of the migration key, so will _always_ override the setting to be false if the user hasn't explicitly set the value before. So it needs to check if the user has already migrated first by reading the migration key. 
4. The migration key value shouldn't be hardcoded, I think it should be in `Definitions.LocalStorageKey`

packages/zowe-explorer/src/configuration/Definitions.ts
```diff
@@ -157,5 +157,6 @@ export namespace Definitions {
         V1_MIGRATION_STATUS = "zowe.v1MigrationStatus",
         DS_SEARCH_OPTIONS = "zowe.dsSearchOptions",
         DISPLAY_RELEASE_NOTES_VERSION = "zowe.displayReleaseNotes",
+        SHOW_HIDDEN_FILES_MIGRATED = "zowe.settings.showHiddenFilesMigrated",
```
zFernand0: I'm curious if we should have a broader migration status (like with V1 above)

By the time V4 comes around, we could have more than just hidden files to migrate over 🙏

zFernand0 (dismissed):
Is it possible to make the migration setting a bit more generic? 

It could be done in a future PR, but I would not want us to have to support multiple, feature-specific migration settings at the same time 😅

packages/zowe-explorer/src/configuration/Definitions.ts
ArghadeepDas-ibm: Hi Fernando,
I have addressed the above comments and introduced a new enum:
```
export enum V3MigrationStatus {
        None,
        JustMigrated,
    }
```
And currently used this to determine the migration status.

PR #4075, opened 2026-02-21
feat: Add Name and DateCreated to PDS filter

### LINKED ISSUE #2986: Missing local PDS member filter based on member name pattern
Hi,

ZE allows you to filter (locally) the members of a PDS on the modification date or on the User ID, but it does not allow you to filter on the name of the member itself.

image

On the other hand, the filtering of members on the PDS during the search / fetch from z/OS works:

image
image
image

It would be necessary to be able to filter (and unfilter) locally on the list of members, if possible with the same syntax as that used under ISPF/PDF option "3.4":
image

Note: VS Code offered a search at the Explorer level (Ctrl+F), but it seems to have disappeared in the latest versions (currently version 1.90.2).

Thanks.

### LINKED ISSUE #3780: Problems with  search PDS member and others
**Describe the bug**

I have a dataset displaying members.

I want to search the PDS members ...
I right click on the data set, and search PDS members, enter a string.

It then displays a pop up saying Are you sure you wan to search 125 data sets and members.!

NO... I want to search just the single data set I selected.

It needs to be clearer how to do this for one data set. The description search pds members implies single PDS.
Perhaps two options
search in the selected PDS
search in ?   What ever it is going to search.   

 I have 105 datasets beginning with COLIN, so I dont know where 125 comes from.
______________________

To save me the tedious scrolling, I wanted to display a subset of members.... beginning with RD... 
I could not see how to do this.  Filter didnt work.
_____________________________

I cannot find a way of displaying the members I want (RD..) there are over 100 members, and I have to do 
pgdown, pgdown,pgdown, pgdown,pgdown, press enter - to get to the second page.  This is really tedious if you have lots of members.

putting 'previous page' and 'next page' as the top two items in the list would be very helpful.

I could not find a shortcut key for"next page".

Does this exist ?  

**To Reproduce**

**Expected behavior**

**Screenshots**

- OS:
- Zowe Explorer Version:

**Additional context**

JWaters02:
## Proposed changes

Closes #2986
Closes #3780 

Adds Name and DateCreated options to the PDS member filter quick pick. 

Supports names specified with wildcard `*`.

To avoid breaking changes but to make sense UX wise, I also reordered the filter options to be in the same order as the sorting quick pick but without changing `Sorting.DatasetFilterOpts` in ZE API.

https://github.com/user-attachments/assets/287add6b-b3b9-4371-87f6-c735854b75b6

Add support to filters for comma seperated values for all types e.g.
image

Also works with pagination within each page (as expected - pagination is requests to the server side which only has support for returning results sorted by name in ascending, so sorting and filtering only works per page in that base order). 

Before (ds pagination set to 100):
image

After (with ds pagination set to 2):
image
image
image
image

@FALLAI-Denis works as you expect?

## Release Notes

Milestone: 3.5.0

Changelog: 
ZE: Add support for filtering PDS members by name and by date created.
ZE API: Added `Name` and `DateCreated` to `Sorting.DatasetFilterOpts`.

## Types of changes

## Checklist

**General**

**Code coverage**

**Deployment**


JWaters02:
> Hi,
> 
> This produces the expected result: the ability to filter by member names after retrieving the list of all members. But why apply the filter to all PDSs present in the Dataset view? Is this also the behavior of the other filters? (I generally only work on one PDS at a time).
> 
> Thanks.

Thanks for the prompt reply! This PR does not add the ability to filter on all data sets at once - that is already a thing (you can try it out yourself). In the video, I was showing that the feature works as expected with the new filter types. Hope this makes sense 🙂

packages/zowe-explorer/CHANGELOG.md
```diff
@@ -28,6 +29,7 @@ All notable changes to the "vscode-extension-for-zowe" extension will be documen
 - Updated Zowe SDKs to `8.30.1` for technical currency. [#4079](https://github.com/zowe/zowe-explorer-vscode/pull/4079)
 - Fixed an issue where screen readers could not read dataset attributes from the dataset attribute webview.[#4023](https://github.com/zowe/zowe-explorer-vscode/issues/4023)
 - Fixed tooltip for Unix System Services profiles to display relevant text. [#4019](https://github.com/zowe/zowe-explorer-vscode/issues/4019)
+- Fixed an issue where filtering a data set by Date Modified showed members without date attributes. [#4075](https://github.com/zowe/zowe-explorer-vscode/pull/4075/)
```
ATorrise: seeing the same thing @pujal0909 has seen. but otherwise the filter works. maybe there should be a check to see if date modified should even appear as a filter option if all members of the pds don't have the date attribute. seems like if one member doesn't have date modified they all don't

JWaters02: Hmm, did you try on latest commit? I fixed the bug pujal reported in [3ecdc66](https://github.com/zowe/zowe-explorer-vscode/pull/4075/commits/3ecdc669e4630c5762c16ceddb7c32d9a56c489a) and it works for me now. 
e.g.
image
filtered on date created:
image

and for a pds with only members without dates on them:
image

image

> maybe there should be a check to see if date modified should even appear as a filter option if all members of the pds don't have the date attribute

I think this is overengineered, the user simply won't try to filter on a date type if none of the members have any dates - and if they don't know this (due to not having a sort on a date type), the filter result not returning anything will make it clear immediately

> seems like if one member doesn't have date modified they all don't

this is not the case, I have seen many data sets with a mix of both (usually members created by scripts or automated processes like endevor pipelines won't have a date associated, but then when a user edits them, the date will appear).

packages/zowe-explorer/src/trees/dataset/DatasetTree.ts
```diff
@@ -2007,17 +2010,26 @@ Would you like to do this now?`,
             return;
         }
 
-        const dateValidation = (value): string => {
-            return dayjs(value).isValid() ? null : vscode.l10n.t("Invalid date format specified");
+        const dateValidation = (value: string): string => {
+            const parts = value.split(",").map((v) => v.trim());
+            return parts.every((p) => p.length > 0 && dayjs(p).isValid()) ? null : vscode.l10n.t("Invalid date format specified");
```
zFernand0, line 2015: It seems that we allow for milliseconds since epoch to be specified, but the invalid date format doesn't show up 😅 

All the options below reflect today's date:
- `2024-03-19T14:00:00Z`
- `19/03/2026`
- `mAr-19-2026`
- `1773946168211` (no red box saying this is invalid)

JWaters02: This isn't a regression in this PR (sanity checked against 3.4.2 release), still using existing dayjs() function to check validity as before. Do you think we should only allow dates and not time/date times then?

traeok: > It seems that we allow for milliseconds since epoch to be specified

Isn't that technically a valid timestamp? Sure, its highly unlikely someone would use it, but its still considered a valid date format 😅

zFernand0: I'm ok with the "Milliseconds Since Epoch" (`MSE`) being supported, but I'm afraid that I can't seems to match them with the actual dataset member creation 😢 

For ISO (2024-03-19T14:00:00Z) I can use any time, and the member created that day will still show 🙏 
However, I'm not able to match the `MSE` with the member created 😅 

> This isn't a regression in this PR (sanity checked against 3.4.2 release), still using existing dayjs() function to check validity as before. Do you think we should only allow dates and not time/date times then?

I'd vote for removing the time from the time/date comparison, as long as it makes the MSE work 🙏 

> Sure, its highly unlikely someone would use it

I agree, so let's not worry about it 😋

JWaters02: > For ISO (2024-03-19T14:00:00Z) I can use any time, and the member created that day will still show 🙏
However, I'm not able to match the MSE with the member created 😅

Ah I see, that is strange, as the code is shared for both date types. I can change it to just care about the date part. 👍

zFernand0: No need to worry about it.
it works as expected for all other cases 🙏 

I doubt anyone will be using MSE 😋

JWaters02: Yep ok cool let's just leave it haha thanks

Issue #3712, opened 2025-06-15
Adding item to favourites does not change "add to favourites" to "remove from favourites"


JWaters02:
**Describe the bug**

Easiest to just show:

![Image](https://github.com/user-attachments/assets/63ff918f-7459-4209-89fa-50c3809e2c14)

Click "Add to favorites" but star is not changed to remove. Clicking on the star again does not remove (so not just icon bug).

![Image](https://github.com/user-attachments/assets/97c18580-66f2-4f04-bbdf-6a3db685dc10)

Note that the remove from favorites star is correct in favorites tree

![Image](https://github.com/user-attachments/assets/78d7fc6a-b610-4788-9423-fc32f081482b)

**To Reproduce**

1. Add data set to favorites
2. See that it is added, but star is still "add" rather than "remove" 

**Expected behavior**

**Screenshots**

**Desktop (please complete the following information):**

- OS:
- Zowe Explorer Version: 3.3.0-SNAPSHOT (commit SHA `e4ef56e75b73310e688106c6fd73e4386c63b049`)
- (Optional) Zowe CLI Version:
- (Optional) Do you have secure credentials enabled?
- (Optional) Are you using a custom credential manager?

**Additional context**

JWaters02:
Note to anyone in the future who tries to look at this: this bug is a trap! It looks easy enough to fix, maybe the node context is not being updated correctly or maybe the when clause on the package.json is wrong? Nope, when clicking add to favourites the node context gets `_fav` appended to it, and then when clause is correct for both. In fact this code and when clause has not been touched for years but this bug only showed up in the last few months. So clearly there is something else going on, and I have no idea how to fix!

Issue #2339, opened 2023-06-14
Adding the possibility to favorites PDS member filter and enhancing how it is handle


Peariforme:
**Is your feature request related to a problem? Please describe.**

In Zowe Explorer v2.8.2, when using a PDS member filter for example DTOPT.TOP.SRC(TOP0*) the star button to put the filter in favorite is not present.

![image](https://github.com/zowe/vscode-extension-for-zowe/assets/47952322/0437cffb-5f70-4191-a3e3-3a073043d02d)

![image](https://github.com/zowe/vscode-extension-for-zowe/assets/47952322/abc1d0af-04ba-4336-bc37-6634e77d63ff)

Moreover we see that there is a member filter thanks to the icon but I can't find a way to see what is the filter after having setting it. Plus when I set two filter in one for example :

DTOPT.TOP.SRC(TOP0*), DOCMT.OCM.SRC(O*)

Then only the first one appears in the list.

**Describe the solution you'd like**

When the filter has a wildcard in the member section
- Show in the explorer the exact filter and not only the PDS name to avoid confusion
- Add the possibility to put the filter in favorites
- When using a filter with multiple PDS with member filter, show all of them in the explorer

JWaters02:
Apologies - thought I could set multiple parents of this issue!

Issue #1356, opened 2021-06-11
Data set member favorite add the entire data set with all members to favorites


JillieBeanSim:
**Describe the bug**

When adding a data set member to favorites the entire data set is added with all of it's members vs just the member that was intentionally added.
Members that are in favorites still have the option to `Add to Favorites` in the right-click menu of favorited item and within the expanded session .

**To Reproduce**

1. Go to Data Sets View and enter a search for data sets
2. Click on a PDS that has multiple members to expand 
3. Either select the star next to a member or do the right-click action to `Add to Favorites`
4. Expand Favorites tree and see that the entire contents of that Data Set was added vs just the member selected.
5. In both, the Favorites and Session with the Data Set, right-click any member that was added to Favorites and see that the option to `Add to Favorites` is still there even though it is already in Favorites.

**Expected behavior**

Only the data set member chosen to be in Favorites shows up in the Favorites and when a right-click is done on that data set member the favorites option should be `Remove from Favorites`

**Screenshots**

How the DS is saved via Settings
![image](https://user-images.githubusercontent.com/49491949/121683154-fc2c6e80-ca8a-11eb-9b1b-b6c3730f5971.png)

![strange-fav-behavior](https://user-images.githubusercontent.com/49491949/121683163-ff275f00-ca8a-11eb-9266-9c13c2ee6ab3.gif)

**Desktop (please complete the following information):**

- OS: Windows 10
- Zowe Explorer Version: 1.15.1
- (Optional) Zowe CLI Version:
- (Optional) Are you using Secure Credential Store?

PR #4077, opened 2026-02-22
fix: favoriting a single member in a PDS now only adds that member to the favorites tree

JWaters02:
## Proposed changes

Introduces the ability to favorite and unfavorite specific members in a PDS. Fix favorites bugs in that also exist in main.

Demo video covering most cases/edge cases (there were several more that I have found and fixed too):

https://github.com/user-attachments/assets/fbb9d163-a8ed-45f8-94d4-29c6571b37d7

This PR was originally intended to be about fixing the long standing and highly requested issue of #1356 (and part of #2339 where the wording is kind of bad but I think most people upvoted it due to it saying "add favorites PDS member" in the title), but it ended up being a pretty major overhaul of how favorites are implemented to support them as it wasn't an easy fix! Also I ended up fixing a few bugs related to favorites, one of which being #3712. That bug fix ended up being a whole pain in of itself and is explained in way more detail below.

## Release Notes

Milestone: 3.5.0

Changelog: 
- Fixed an issue where favoriting a single member in a PDS was adding the entire PDS with all members to the favorites tree. 
- Fixed an issue where adding an item to favorites was still incorrectly showing the "Add to Favorites" icon on it. 

## Types of changes

## Checklist

**General**

**Code coverage**

**Deployment**

## Further comments

Since it is a big change and I've had it stashed for months slowly working on it since May 2025!!, I wrote some docs to help myself and now you understand the changes.

Firstly, a new enum `Definitons.PdsFavoriteState` is introduced with states:

- `NotFavorited` - The PDS has no favorite status
- `EntirePds` - The entire PDS and all its members are favorited (the same as existing behaviour in main)
- `SpecificMembers` - Only specific members within the PDS are favorited

Each PDS node in the favorites tree now carries two new properties:

- `pdsFavoriteState` - one of the three enum values above
- `favoritedMemberNames` - a string array of member names that are favourited, only meaningful when `pdsFavoriteState` is `SpecificMembers`

Member specific favorites are stored in persistance as separate entries in the following format:

```
[profile]: PDS.NAME(MEMBER1){contextValue}
[profile]: PDS.NAME(MEMBER2){contextValue}
```

A fully favorited PDS is stored as a single entry:

```
[profile]: PDS.NAME{contextValue}
```

When favorites are restored from persistance, entries with member names are grouped into a single PDS node with `pdsFavoriteState = SpecificMembers` and the corresponding `favoritedMemberNames` array.

Session nodes now use a seperate `_filterFav` context suffix instead of the standard `_fav` suffix. This isbecause:

1. The existing `isSessionNotFav` check explicitly excludes nodes with `_fav` in their context. When `isSessionNotFav` returns false, `datasetFilterPrompt` treats the node as a saved search from the favorites tree and tries to read its label as a filter pattern, then calls `addSession`. So if it had `_fav` added, it would try to set the filter to an undefined value and thus cause everything to break.

2. Multiple `package.json` when clauses use `^(?!.*_fav.*)session.*` to control visibility of several other unrelated commands. Adding `_fav` to a search-tree session node would hide all of these

3. The `_fav` suffix on session nodes would show a "remove from favorites" star icon on the session/profile node itself (not on the favorites tree), which I believe is intentionally not being shown to keep the number of buttons on the profile node limited to only those necessary. Using `_filterFav` allows the context menu to show "remove from favorites" for the saved search without adding the inline icon to it.

The changes also introduced several bugs regarding filtered PDS results (e.g. where filter is `HLQ.FILTER(MEM*)`), so for now I fixed them so that the behaviour is the same as on main - where the filtered PDS results _cannot_ be added to favorites. I consider this feature out of scope of this PR as the scope is already way larger than I anticipated and I think the change to support filtered data sets in favorites would also be a lot of work. Note that it directly goes against _one of_ the requests in the community upvoted issue #2339 so I added a comment in there to clarify this.

---

Fixes the following bugs (that exist on main already):

### 1. Favoriting a member adds the entire PDS - #1356

Currently: Clicking "add to favorites" on a member called `addFavorite(node.getParent())`, which added the entire PDS with all members (so seems intentional rather than a bug but I'd say it's not expected UX)

Now: Member favorites are handled by `addMemberToFavorites`, which creates or updates a PDS node in favorites with `pdsFavoriteState = SpecificMembers` and only the selected member in `favoritedMemberNames`.

### 2. Favorited items still show "add to favorites" in context menu - #3712

Currently:

1.  Members inside favorites,
2.  members in the search tree that were already favorited,
3.  session filters,
4.  and PDS/DS nodes,
    in the search tree after being added to favorites all showed "add to favorites" because their context values never received the `_fav` suffix (I thought they were when looking at it last year, but the changed node themselves weren't being updated in the tree so it was using the stale nodes).

Now: Members and PDS/DS nodes in the favorites tree always get `_fav` context via `getChildren`. In the search tree, `updateMemberFavContext` and `updateNodeFavContext` are called after adding or removing favorites to add or remove the `_fav` suffix on matching nodes. Also `getChildren` on session nodes marks returned PDS/DS children with `_fav` if they exist in favorites, and `getChildren` on PDS nodes marks member children with `_fav` when the PDS is expanded.

As for session, `updateSessionFilterFavContext` adds or removes the `_filterFav` suffix on session nodes based on whether their current pattern matches a saved search in favorites. The when clauses in `package.json` are updated to check for `_filterFav` alongside `_fav`.

### 3. Adding entire PDS when member favorites exist was blocked

Currently: If some members of a PDS were already favorited, trying to add the entire PDS showed "PDS already in favorites" (basically an extension of the first bug)

Now: Adding an entire PDS when member favorites exist upgrades the favorite from `SpecificMembers` to `EntirePds`, clears `favoritedMemberNames`, and refreshes the tree.

---

Some call flow shenanigans to help understand how things are being wired together (like my brain trying to spell favourites in the murican way):

### 1. Adding a member to favorites

`addFavorite(memberNode)` detects `isDsMember` and delegates to `addMemberToFavorites`.

`addMemberToFavorites` has the following branches:

1. Member is in the favorites tree (`isFavoriteDescendant` and parent `isFavoritePds`):

   - If state is `EntirePds`, show notification message that "PDS already in favorites" (like before)
   - If state is `SpecificMembers` and member exists: show notification message that "Member already in favorites"
   - Either way, return early as we don't need to do anything else
   - Note, this branch technically isn't needed as we should never show the add to favorites command for this state, however it could still happen as the bug has appeared before and its an extra guard just in case

2. Member is in the session tree and the PDS already exists in favorites (looked up by label from `profileNodeInFavorites.children`):

   - If state is `EntirePds`, show notification message that "PDS already in favorites" (like before)
   - If member already in `favoritedMemberNames`, show notification message that "Member already in favorites"
   - Otherwise, push member name to `favoritedMemberNames`, update persistance, call `updateMemberFavContext`

3. PDS not in favorites yet: create new PDS node with `SpecificMembers` state, set `favoritedMemberNames` to the single member, add to favorites, call `updateMemberFavContext`

Note: branch 1 uses `isFavoriteDescendant` to distinguish between favorites-tree and session-tree members. There was an edge case where it only checked `isFavoritePds(parentPds)`, which incorrectly matched session tree PDS nodes that had `_fav` in their context (from whole PDS favoriting). This caused the session tree path to read `pdsFavoriteState` from the session tree PDS (always `undefined`) instead of the actual favorites PDS, resulting in "PDS already in favorites" when trying to re-add an unfavorited member.

### 2. Adding an entire PDS to favorites

Call `addFavorite(pdsNode)` when node is a PDS.

1. If PDS exists in favorites with `SpecificMembers`, upgrade to `EntirePds`, clear `favoritedMemberNames`, call both `updateMemberFavContext` and `updateNodeFavContext`
2. If PDS already fully favorited, show notif "already in favorites"
3. If PDS not in favorites, create favorited PDS node with `EntirePds` state (default), call `updateMemberFavContext` and `updateNodeFavContext`

### Removing a member from favorites

`removeFavorite(memberNode)` detects `isDsMember` and resolves the actual favorites PDS:

- If the member is in the favorites tree (`isFavoriteDescendant` and parent `isFavoritePds`), the parent node is used directly
- If the member is in the session tree, the favorites PDS is looked up by label from `profileNodeInFavorites.children`
- If no favorites PDS is found, return early as member is not favorited

There was an edge case caused by the code checking `isFavoritePds(parentPds)` on the session-tree PDS, which only matched if the session PDS had `_fav` in its context. When only specific members were favorited (not the entire PDS), `addMemberToFavorites` doesn't call `updateNodeFavContext`, so the session-tree PDS never gets `_fav`. This caused the entire member-removal block to be skipped, making unfavoriting single-member PDS favorites a no-op.

Once the favorites PDS is resolved:

1. If favorites PDS has `SpecificMembers` state:

   - Filter out the member from `favoritedMemberNames`
   - If no members remain, remove the entire PDS node from favorites
   - If members remain, update `favoritedMemberNames` with the filtered list

2. If favorites PDS has `EntirePds` state:

   - Enumerate all current children except the removed member
   - When called from the session tree, the favorites PDS may not have children loaded, so use the session-tree PDS (original parent) to enumerate members
   - If no members remain, remove the PDS from favorites
   - If members remain, convert to `SpecificMembers` with the remaining list

Now call `updateFavorites`, refresh, and call `updateMemberFavContext`

### Removing an entire PDS/data set from favorites

Call `removeFavorite(favoritedNode)` for PDS or data set nodes in the favorites tree.

After removal, call `updateMemberFavContext` (for PDS) and `updateNodeFavContext` to strip `_fav` from the search tree node

### Removing a filter favorite from session node

Call `removeFavorite(sessionNode)` where the session node has `_filterFav` context but is not in the favorites tree.

Finds the matching saved search in favorites by comparing the session's current pattern, removes it, and calls `updateSessionFilterFavContext` to strip `_filterFav`

### on getChildren

When `getChildren` is called:

1. On a session node, marks returned PDS/DS children with `_fav` if they exist in favorites. Also strips stale `_fav` from any filtered PDS nodes (`isFilterSearch`) since filtered PDS nodes are not favoritable
2. On a non-favorites-tree PDS node (checked via `isFavoriteDescendant` rather than `isFavorite` to handle session-tree PDS nodes that have `_fav`), bidirectionally updates member context: adds `_fav` to members that should be favorited and removes stale `_fav` from members that should not be (this handled an edge case where members were unfavorited but their cached nodes still carried `_fav`). Skips filtered PDS nodes (`isFilterSearch`) as their members are not favoritable
3. On a favorited PDS node, filters children to only favorited members (if `SpecificMembers`), marks all returned children with `_fav`
4. In the `finalResponse` loop, handles member `isFilterSearch` context:
   - For filtered PDS (has `memberPattern`): strips `_fav` from members and adds `isFilterSearch` if not already present
   - For non-filtered PDS: strips stale `isFilterSearch` from reused member nodes (handles the case where a PDS filter was reset but cached member nodes still carry `isFilterSearch`)


JWaters02:
@zFernand0 

If we have a favorited PDS, and then we delete all the members from the PDS, should the PDS be removed from the favorites or should the PDS stay in favorites with no members?

I was starting to feel that the PDS should stay in favorites with no members, but now I'm thinking it should be removed, because who will be using an empty PDS?

JWaters02:
While I did develop the majority of this code change using a TDD campaign (with over 50 new test cases and many existing tests refactored to account for the changes), I still get the feeling that there are more scenarios I didn't think of yet that edge cases need to be developed for, so ideally I just need people to daily drive this version of the extension and let me know if they find any regressions or edge cases. Or just see how many cases you can think of and try them out - maybe the case is already covered by a test and works, but maybe it isn't and doesn't do what you expect.

The only scenario I am currently aware of that I don't want to fix in this PR is the one explained previously - if you favorite a whole PDS, unfavorite one member, then create a new member, it doesn't add the new member to the PDS' favorites.

JWaters02:
> Something to consider regarding accessibility, based on my initial code review: When an PDS is in SpecificMembers mode, there's no UI indicator. Consider adding a description to the PDS node, such as "3 members" or "Partial" to help users understand what they're viewing at a glance.

I had quite a few musings about this implementation (I had several versions of comments written out here to rubber duck it) because it turned out to be a pretty difficult problem as I was thinking about what would actually be intuitive, and then when implementing it I came across some issues that I very almost gave up on but in the end found a simple solution that actually works.

So what I decided was to have it say "x/y members" where x is the number of favorited members and y is the actual number of members it has - but only show when the PDS node is expanded and when the PDS is in `SpecificMembers` state. I considered your two suggestions but they are not intuitive enough/not easy to understand what they are referring to.

image

There is one noticable drawback with this solution: because we only get the total number of actual members when we call getChildren, we set the node description there, but the node then needs to be refreshed to actually display the updated description, which as a result getChildren is called by the tree provider again. Even if dirty is false, if paginate is true, this line will not be true, so therefore it does not just return early:
https://github.com/zowe/zowe-explorer-vscode/blob/4fe5cd11d8a495a6dce095f5cdb0c892fe07848b/packages/zowe-explorer/src/trees/dataset/ZoweDatasetNode.ts#L295

So for a large data set, this could take quite some time to refresh for the second time even though we are doing nothing new except updating the parent node's description. So what I am unsure on, is how could we refresh the PDS node to update the description *without* also calling getChildren for everything?

JWaters02:
So as per trae's suggestion we could have a check near the top of getChildren where 
```ts
if (!this.dirty && SharedContext.isFavoritePds(this) && this.favoritedMemberNames != null) {
   return this.children;
}
``` 
and it seems to work fine in all cases... until i turn pagination on. for this example I set pagination count to 10 and the PDS contains 11 members.

case 1:
1. first i favorite a whole pds contianing 11 members. favorited pds has no description (correct) and shows 10 members in the first page and the last member in the second page
2. i go to second page and click to unfavorite that member. the entire pds is removed from favorites (wrong)

case 2:
1. step 1 from case 1.
2. i unfavorite the first member of the first page. description now shows 9/10 members (wrong, it should be 10/11 members or maybe 10/10 members since the pagination would technically only fetch first 10) and pagination is gone with the favorited member in the second page also being gone (and it is not favorited if we look at the non-favorited tree it has add to favorites star). 
3. now in non-favorited tree click to add the last member in the second page (which is missing on favorited one from step 2) and description still says 9/10 members with pagination showing with 2 pages.
4. click on next page and suddenly the pages are gone, last member we just favorited is gone, but it is still favorited by checking it in non-favorites tree
5. now click to remove it from favorites. PDS description says 1/1 member, but shows no children even though it is still expanded. even if we collapse and rexpand the tree, it stays the same
6. now click to re-add it to favorites again. desc still says 1/1 member but it is added to the tree with pagination arrows in the second page.
7. click previous page. pagination arrows now disappear and only the last member is left with still no other members there even though they are still all favorited. 
and these broken behaviour continues. so the whole PDS node is in a broken state with the pagination.

also note that these broken issues with the pagination do not occur without the check

pagination is handled basically at the end of getChildren, and I don;t understand what it's doing properly, so I'm not sure of a solid workaround to this issue. 

I did also try messing with the call back in DatasetTree getChildren depending on if we should paginate or not using a cached list of the children (so we can return just the cached list of children instead of doing all of getChildren again), but this creates even worse issues where you can't even click on the next page button as it throws an error and looking into it it feels like the completely the wrong track, so I'm not sure where to go from here.
```ts
private childrenCache = new WeakMap<IZoweDatasetTreeNode, IZoweDatasetTreeNode[]>();
...
const prevDescription = element.description;
const shouldPaginate = SettingsConfig.getDirectValue<number>(Constants.SETTINGS_DATASETS_PER_PAGE, Constants.DEFAULT_ITEMS_PER_PAGE) > 0;
if (!element.dirty && shouldPaginate && this.childrenCache.has(element)) {
   return this.childrenCache.get(element);
}
const response = await element.getChildren(shouldPaginate);
if (element.description !== prevDescription) {
   this.childrenCache.set(element, response);
   this.nodeDataChanged(element);
} else {
   this.childrenCache.delete(element);
}
```

JWaters02:
Also, can people test migrating their favorites between this PR version and main version? It should all just work (it usually does) but recently on main version I've noticed that sometimes I cannot open the favorites tree due to a REST error that a data set qualifier is too long (with the workaround being to remove the whole profile from favorites), but I'm not sure if it's an issue with migrating from these changes to old favorites or a bug in main itself. Obviously we need to make sure that migrating between new and old favorites works as expected as it is a pretty regular occurance of needing to downgrade ZE versions to certain bugs or changes that don't work on a certain site's ecosystem.

PR #4122, opened 2026-03-19
fix: downloadAllMembers to fetch member count much faster

### LINKED ISSUE #4109: Issues with Download All Members command
**Describe the bug**

I can't download members without expanding the PDS in the tree first, and while they are downloading the progress message doesn't update.

**To Reproduce**

1. Search for data sets in ZE.
2. Right-click on a PDS (do not expand it) and select "Download All Members".
3. Notice that profile is validated but nothing happens.
4. Expand the PDS and repeat the download command.
5. Notice that downloading works, and the progress bar moves, but the progress message stays at "1 of 100".

**Expected behavior**

Expanding a PDS shouldn't be required to download its members.

**Screenshots**

Image

- OS: MacOS 26.3
- Zowe Explorer Version: Main branch

**Additional context**

The data set I was trying to download is very large with over 100k members, but I don't think this is necessary to reproduce.

JWaters02:
## Proposed changes

Fixes #4109 by calling list allMembers API to get member count for the warning popup.

Tested on:
- PDS with no children
- PDS with ~120 children
- PDS with ~10k children

## Release Notes

Milestone:

Changelog: N/A (it fixes an issue not in a release yet)

## Types of changes

## Checklist

**General**

**Code coverage**

**Deployment**


packages/zowe-explorer/src/trees/dataset/DatasetActions.ts
t1m0thyj: Currently the way that progress is reported may lag for a large PDS. When we display `X of Y`, the `X` only updates every 1% or 1k members for a data set with 100k members.

Since the `value` param provided by the Zowe SDKs is an integer, it doesn't give a granular percentage value (0.0%, 0.1%, etc.) that could be used to infer a more precise member count. However I noticed that the `percentComplete` callback gets triggered only once per member, so we should be able to do the following:

```suggestion
                let numDownloaded = 0;
                const task: imperative.ITaskWithStatus = {
                    set percentComplete(value: number) {
                        realPercentComplete = value;
                        Gui.reportProgress(progress, realTotalEntries, ++numDownloaded, "");
```

JWaters02: Huh interesting I'll have to try it and see, that's cool though, would indeed be slightly nicer QoL to have the download updated per member instead of in groups when doing a lot.

PR #4144, opened 2026-03-26
Fix: submit job confirmation dialog box not coming up when we submit jobs

### LINKED ISSUE #4097: Confirmation for submitting jobs doesn't appear
When submitting a job, there should be a dialog that appears to prompt for confirmation, but I don't see it in the latest release.

https://github.com/zowe/zowe-explorer-vscode/blob/97872e01e811fb273ee68300687b550e2b1d17e2/packages/zowe-explorer/src/trees/dataset/DatasetActions.ts#L1534

ArghadeepDas-ibm:
## Proposed changes

Fixes: https://github.com/zowe/zowe-explorer-vscode/issues/4097
The submit job confirmation dialog was not appearing when submitting dataset members because the code couldn't match the configuration value (localization keys like %zowe.jobs.confirmSubmission.allJobs%) against the translated strings in the array.

## Release Notes

Milestone:

Changelog: Fixed submit job confirmation dialog box not coming up when we submit jobs.

## Types of changes

## Checklist

**General**

**Code coverage**

**Deployment**


packages/zowe-explorer/CHANGELOG.md
```diff
@@ -27,6 +27,7 @@ All notable changes to the "vscode-extension-for-zowe" extension will be documen
 - Fixed an accessibility issue where screen readers could not access the job ID link after job submission. Replaced the inaccessible markdown link with keyboard-accessible "Open Job" button that can be navigated using Tab and activated with Enter or Space keys. [#4030](https://github.com/zowe/zowe-explorer-vscode/issues/4030)
 - Fixed an issue where focus was lost on a previously selected tree item after opening a Zowe resource (Data set, USS file, or Job) in the editor and closing it using `Ctrl+W`. Screen readers will now announce when focus is restored. [#4025](https://github.com/zowe/zowe-explorer-vscode/issues/4025)
 - Fixed an issue where loading profiles could wrongly assume token authentication if parent profile defines multiple auth types. [#4105](https://github.com/zowe/zowe-explorer-vscode/issues/4105)
+- Fixed submit job confirmation dialog box not coming up when we submit jobs.[#4097](https://github.com/zowe/zowe-explorer-vscode/issues/4097)
```
```suggestion
- Fixed the submit job confirmation dialog box failing to display when a job is submitted. [#4097](https://github.com/zowe/zowe-explorer-vscode/issues/4097)
```

anaxceron (approved):

traeok (changes requested):
Thanks for looking into this Arghadeep. I believe there is a simpler fix that's easier to maintain. 

See `Constants.ts` where `JOB_SUBMIT_DIALOG_OPTS` is defined:
https://github.com/zowe/zowe-explorer-vscode/blob/434dff168060d5a6445850301256550e7688ff78/packages/zowe-explorer/src/configuration/Constants.ts#L273-L278

The core change that stands out to me is changing the switch statement to look for the placeholders instead of the `vscode.l10n.t` calls.

Could we not just change this array to use the placeholders instead? For example:
```ts
public static JOB_SUBMIT_DIALOG_OPTS = [
    "%zowe.jobs.confirmSubmission.disabled%",
    "%zowe.jobs.confirmSubmission.yourJobs%",
    "%zowe.jobs.confirmSubmission.otherUserJobs%",
    "%zowe.jobs.confirmSubmission.allJobs%",
];
```

This would keep most of the code untouched, while having the same effect as what's been done here.

packages/zowe-explorer/CHANGELOG.md
```diff
@@ -31,7 +31,11 @@ All notable changes to the "vscode-extension-for-zowe" extension will be documen
...
 - Fixed an issue where focus was lost on a previously selected tree item after opening a Zowe resource (Data set, USS file, or Job) in the editor and closing it using `Ctrl+W`. Screen readers will now announce when focus is restored. [#4025](https://github.com/zowe/zowe-explorer-vscode/issues/4025)
 - Fixed an issue where loading profiles could wrongly assume token authentication if parent profile defines multiple auth types. [#4105](https://github.com/zowe/zowe-explorer-vscode/issues/4105)
+<<<<<<< 4097_job_submit_confirmation_dialog_box
+- Fixed the submit job confirmation dialog box failing to display when a job is submitted. [#4097](https://github.com/zowe/zowe-explorer-vscode/issues/4097)
+=======
 - Fixed an issue where the USS tree collapsed down to the top level set from the initial filter.[#3932](https://github.com/zowe/zowe-explorer-vscode/issues/3932)
+>>>>>>> main
```
t1m0thyj: Please remove the extra lines from resolving merge conflict.

packages/zowe-explorer/src/trees/dataset/DatasetActions.ts
```diff
@@ -1898,24 +1898,49 @@ export class DatasetActions {
             return selection != null && selection?.title === "Submit";
         };
 
-        const confirmationOption: string = vscode.workspace.getConfiguration().get("zowe.jobs.confirmSubmission");
+        const confirmationOption: string | boolean = vscode.workspace.getConfiguration().get("zowe.jobs.confirmSubmission");
```
t1m0thyj: Should we change the default value of this setting in package.json as suggested in the issue - https://github.com/zowe/zowe-explorer-vscode/issues/4097#issuecomment-4004441978

"Other User Jobs" would be a less aggressive default value for the setting than "All Jobs".

anaxceron (approved):

JillieBeanSim (commented):
Thanks for working on this @ArghadeepDas-ibm I do see the confirmation module when expected per setting value but I do notice that the default changes on this branch vs the marketplace.  Not sure if that could be considered a user-facing breaking change or not but curious why change the default value?
Marketplace:
Image
Branch:
Image

traeok:
Even though the confirmation dialog was broken for a period of time, I second Billie's thoughts that the change to "Other user's jobs" could be perceived as a breaking change, even if a small one. If a user is still leveraging an older version of Zowe Explorer where the feature _was_ working, then they could be caught off guard when it no longer confirms them to submit their own jobs.

I think it would be best to table this change to v4 and in the meantime, we can keep the PR in scope of the reported issue (not seeing dialog)

I see that the issue discusses changing the default, but I'm not sure that we should assume that folks aren't relying on this default behavior as we even admit it could be breaking (violates Zowe Conformance Criteria):

> Technically changing the default value of the setting constitutes a breaking change, but no one seems to have noticed the setting being broken for quite some time

ArghadeepDas-ibm:
Hi @JillieBeanSim ,
The default value was updated to “Other user jobs,” as requested in the issue, to align with the expected behavior. https://github.com/zowe/zowe-explorer-vscode/issues/4097

JillieBeanSim:
> Hi @JillieBeanSim , The default value was updated to “Other user jobs,” as requested in the issue, to align with the expected behavior. #4097

Hey @ArghadeepDas-ibm as discussed in scrum that could be considered a breaking change and will have to wait until v4.  Pleases change the default back to original for now and add a comment in https://github.com/zowe/zowe-explorer-vscode/issues/4063 to remember to tackle in v4. Can also leave a comment in the issue linked to this PR that the default change is a breaking change and will need to wait until the next major release sharing the link to v4 Planning issue. Thanks

JillieBeanSim (approved):
LGTM! thanks for the fix @ArghadeepDas-ibm

packages/zowe-explorer/src/trees/dataset/DatasetActions.ts
```diff
@@ -1913,22 +1913,31 @@ export class DatasetActions {
...
+        if (!confirmationOption || typeof confirmationOption !== "string") {
+            return true; // Default behavior: allow submission without confirmation
+        }
+
+        const camelCase = confirmationOption
+            .toLowerCase()
+            .split(" ")
+            .map((word, index) => (index === 0 ? word : word.charAt(0).toUpperCase() + word.slice(1)))
+            .join("");
+        const dialogOption = Constants.JOB_SUBMIT_DIALOG_OPTS.indexOf(`%zowe.jobs.confirmSubmission.${camelCase}%`);
```
traeok, line 1926: I'm still a bit confused by the point of this logic. The logic implies that previously, we _were_ storing the localized values themselves (e.g. "Your jobs", "Other user jobs", etc.). So if this is the case, why do we need to transform them into an l10n placeholder and compare the placeholders? Why can't we just compare against the l10n string values themselves?

For example, let's say `confirmationOption` is "Your jobs":
- camelCase transforms it from `"Your jobs" -> "your jobs" -> ["your", "jobs"] -> "yourJobs"`.
- then, we substitute it in a template string `%zowe.jobs.confirmSubmission.${camelCase}%` and compare it against the l10n placeholders specified in `Constants.JOB_SUBMIT_DIALOG_OPTS`.

This is functionally equivalent to: `confirmationOption === vscode.l10n.t("zowe.jobs.confirmSubmission.yourJobs")`.

Since these two are equivalent, why not just compare what we saved in `settings.json` with the localized value?

This also tracks with what I'm seeing in marketplace release of Zowe Explorer: if I select "Your jobs" for this option in VS Code settings, the `settings.json` file contains:

```json
 "zowe.jobs.confirmSubmission": "Your Jobs",
```

So with what's out there now, we are already storing localized values. If this is the case, we can just compare what they have saved to the `vscode.l10n.t(...)` elements previously contained in `Constants.JOB_SUBMIT_DIALOG_OPTS`.

Maybe there is some edge case where the placeholder gets stored instead of the localized value. In that case, I would consider the storing of that placeholder as a bug, and not something we should aim to support.

traeok (commented):
One thing that's still unclear to me that I think needs to be addressed before merging: I think this method of comparing the confirmation option feels a bit over-engineered. Given what I'm seeing with marketplace versions, there's an opportunity to optimize this logic and remove the unconventional text transformations.

packages/zowe-explorer/src/configuration/Constants.ts
```diff
@@ -276,10 +276,10 @@ export class Constants {
...
     public static JOB_SUBMIT_DIALOG_OPTS = [
-        "%zowe.jobs.confirmSubmission.disabled%",
-        "%zowe.jobs.confirmSubmission.yourJobs%",
-        "%zowe.jobs.confirmSubmission.otherUserJobs%",
-        "%zowe.jobs.confirmSubmission.allJobs%",
+        vscode.l10n.t("Disabled"),
+        vscode.l10n.t("Your Jobs"),
+        vscode.l10n.t("Other user Jobs"),
+        vscode.l10n.t("All Jobs"),
     ];
```
zFernand0, line 283: Not suggesting changes...

But we should find a better way to avoid having to keep this array in-sync (same order) with the enum `JobSubmitDialogOpts` 😅

zFernand0 (approved):
LGTM! 😋

traeok (approved):
LGTM, thanks Arghadeep for addressing my feedback

packages/zowe-explorer/src/configuration/Constants.ts
traeok: Not to play devil's advocate, but is it really worth pursuing a better way for one array? If we change the options (which we shouldn't do in a breaking fashion anyway), we'll just have to change the array - it's probably more trouble than its worth to optimize that process 😋

PR #4160, opened 2026-04-02
Fix for extender break in USS filter search

JillieBeanSim:
## Proposed changes

fix extender break with filter search in uss tree view

## Release Notes

Milestone:

Changelog: n/a bug was not published yet

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] Enhancement (non-breaking change which adds or improves functionality)
- [ ] Breaking change (a change that would cause existing functionality to not work as expected)
- [ ] Documentation (Markdown, README updates)
- [ ] Other (please specify above in "Proposed changes" section)

## Checklist

**All checks must pass before this pull request is reviewed by the Zowe Explorer squad.**

**General**

- [x] I have read the [CONTRIBUTOR GUIDANCE](https://github.com/zowe/zowe-explorer-vscode/wiki/Best-Practices:-Contributor-Guidance) wiki
- [x] All PR dependencies have been merged and published (if applicable)
- [ ] A GIF or screenshot is included in the PR for visual changes
- [x] The pre-publish command has been executed:
  - **v2 and below:** `yarn workspace vscode-extension-for-zowe vscode:prepublish`
  - **v3:** `pnpm --filter vscode-extension-for-zowe vscode:prepublish`
- [ ] New ZE APIs are tested with extender types that haven't adopted yet to determine breaking changes. Can use Zowe zFTP marketplace extension.

**Code coverage**

- [ ] There is coverage for the code that I have added
- [x] I have added new test cases and they are passing
- [ ] I have manually tested the changes

**Deployment**

- [ ] I have tested new functionality with the FTP extension and profile verifying no extender profile type breakages introduced
- [ ] I have added developer documentation (if applicable)
- [ ] Documentation should be added to Zowe Docs
  - If you're an outside contributor, please post in the [#zowe-doc Slack channel](https://openmainframeproject.slack.com/archives/CC961JYMQ) to coordinate documentation.
  - Otherwise, please check with the rest of the squad about any needed documentation before merging.
- [ ] These changes may need ported to the appropriate branches (list here):

PR #4162, opened 2026-04-06
clean up DS and USS file system notifications

jace-roell:
## Proposed changes

Updated DS and USS file systems to generate notifications that align with the remote resource updates rather than local file system changes

## How to test
1. Add the following code snippet to the `extension.ts` file to log statements based on events fired:
<details>
<summary><b>extension.ts code snippet</b></summary>

```typescript
    const dsWatcher = vscode.workspace.createFileSystemWatcher(
        new vscode.RelativePattern(vscode.Uri.from({ scheme: "zowe-ds", path: "/" }), "**/*"),
        false,
        false,
        false
    );

    dsWatcher.onDidDelete((uri) => {
        uri.scheme == "zowe-ds" && console.log("[DELETED]", uri.toString());
    });

    dsWatcher.onDidCreate((uri) => {
        uri.scheme == "zowe-ds" && console.log("[CREATED]", uri.toString());
    });

    dsWatcher.onDidChange((uri) => {
        uri.scheme == "zowe-ds" && console.log("[CHANGED]", uri.toString());
    });

    const uss = vscode.workspace.createFileSystemWatcher(
        new vscode.RelativePattern(vscode.Uri.from({ scheme: "zowe-uss", path: "/" }), "**/*"),
        false,
        false,
        false
    );

    uss.onDidDelete((uri) => {
        uri.scheme == "zowe-uss" && console.log("[DELETED]", uri.toString());
    });

    uss.onDidCreate((uri) => {
        uri.scheme == "zowe-uss" && console.log("[CREATED]", uri.toString());
    });

    uss.onDidChange((uri) => {
        uri.scheme == "zowe-uss" && console.log("[CHANGED]", uri.toString());
    });
```
</details>

2. Via the tree, do basic CRUD operations on PDSes. PDS members, Sequential DSes, USS directories and USS files. `created`, `updated`, and `deleted` notifications should be generated based on actions that happen on the remote system and not on the local cache.
3.  Upon doing actions such as creating a PDS member when the resource has already been created out of the scope of ZE (via ISPF, Zowe CLI, etc), it will generate a created notification upon learning that this resource was created in the mean time.
4. Resolving conflicts for contents of a PDS member or USS file will generate create/changed events based on whether the contents were overwritten or not.

## Types of changes

## Checklist

**General**

**Code coverage**

**Deployment**


packages/zowe-explorer/src/trees/uss/UssFSProvider.ts
```diff
@@ -916,7 +924,7 @@ export class UssFSProvider extends BaseProvider implements vscode.FileSystemProv
         parent.mtime = Date.now();
         parent.size -= 1;
 
-        this._fireSoon({ type: vscode.FileChangeType.Changed, uri: parentUri }, { uri, type: vscode.FileChangeType.Deleted });
+        this._fireSoon({ uri, type: vscode.FileChangeType.Deleted });
```
traeok: Should we keep the fire event for the parent URI? Technically the contents of the parent directory have changed when a file inside it is deleted.

packages/zowe-explorer/src/trees/uss/UssFSProvider.ts
```diff
@@ -1048,6 +1056,32 @@ export class UssFSProvider extends BaseProvider implements vscode.FileSystemProv
         }
     }
 
+    public createEntry(uri: vscode.Uri, ussType: "file" | "directory"): UssFile | UssDirectory {
```
traeok, line 1098: Consider adding TSDoc for this to provide clarity on what the function does and its expected return values

packages/zowe-explorer/src/trees/dataset/DatasetFSProvider.ts
```diff
@@ -1075,4 +1074,25 @@ export class DatasetFSProvider extends BaseProvider implements vscode.FileSystem
             throw vscode.FileSystemError.FileNotFound(uri);
         }
     }
+
+    public createEntry(uri: vscode.Uri, dsType: DsType): PdsEntry | DsEntry {
```
traeok, line 1084: Consider adding TSDoc for this function as well (see comment on USS impl)

traeok (commented):
Thanks for working on this Jace, I had a question and 2 suggestions but changes make sense based on my initial review. Going to do some manual testing now

packages/zowe-explorer-api/CHANGELOG.md
```diff
@@ -13,6 +13,7 @@ All notable changes to the "zowe-explorer-api" extension will be documented in t
 
 ### Bug fixes
 
+- Fixed an issue in `BaseProvider._handleConflict` where the editor would incorrectly close during file conflict resolution. [#4162](https://github.com/zowe/zowe-explorer-vscode/pull/4162)
```
anaxceron: We should describe what `BaseProvider._handleConflict` is: Fixed an issue in the `BaseProvider._handleConflict` method/property/cat/whatever where the ... .

anaxceron (approved):
Fantastic, much like a cat 🐈

JillieBeanSim (changes requested):
thanks for working on this @jace-roell 
when I attempt to kick off the diff with extender by opening file in editor to make changes then upload changed file via cli I am not getting the conflict notification or prompt for diff view.  It shows a save was done via editor but file changes if I follow up with pull from mainframe.

jace-roell:
> thanks for working on this @jace-roell when I attempt to kick off the diff with extender by opening file in editor to make changes then upload changed file via cli I am not getting the conflict notification or prompt for diff view. It shows a save was done via editor but file changes if I follow up with pull from mainframe.

I did some more testing with this based on our conversation today. With the following steps:

1. Open a PDS member in a vscode editor
2. use the CLI/ISPF to edit the contents of the member
3. refocus the editor.

I am seeing the contents update immediately based on the stat call made by refocusing the window. I see this behavior with zosmf and the ssh extension. I do not see this behavior with the ftp extension. I am wondering if this is an RSE related issue and whether it is a regression or present in the latest release.

traeok:
One thing to note is that if the extender does not return `mtime` in their response, we do not update the editor as we can't determine whether the resource has been changed. For example, in USS:

https://github.com/zowe/zowe-explorer-vscode/blob/490195beccee428d12ab2340e6446a632f6b0cc4/packages/zowe-explorer/src/trees/uss/UssFSProvider.ts#L135-L140

It's likely that FTP does not always return this key and the same could also apply for extenders. I would recommend using Pull from Mainframe with those profile types until this info is returned as part of the list response for that file.

traeok (approved):
Tested and LGTM, not seeing any issues while testing w/ SSH, z/OSMF
Thanks Jace

JillieBeanSim:
> One thing to note is that if the extender does not return `mtime` in their response, we do not update the editor as we can't determine whether the resource has been changed. For example, in USS:
> 
> https://github.com/zowe/zowe-explorer-vscode/blob/490195beccee428d12ab2340e6446a632f6b0cc4/packages/zowe-explorer/src/trees/uss/UssFSProvider.ts#L135-L140
> 
> It's likely that FTP does not always return this key and the same could also apply for extenders. I would recommend using Pull from Mainframe with those profile types until this info is returned as part of the list response for that file.

I'm sorry but with retesting I am still seeing the issue with extender types and it is mentioned here in the comments seeing the same with zFTP.  This sounds like a regression/breaking change introduced in the code as it worked previously for extenders using the etag value and not requiring `mtime`

traeok:
> I'm sorry but with retesting I am still seeing the issue with extender types and it is mentioned here in the comments seeing the same with zFTP. This sounds like a regression/breaking change introduced in the code as it worked previously for extenders using the etag value and not requiring `mtime`

Gotcha, thanks for clarifying - I can verify w/ 3.4.2 but I discussed w/ Jace and it does sound like a regression. Maybe we can revisit the `mtime` check to see if we can restore this behavior.

jace-roell:
@JillieBeanSim @traeok Since this is not a regression of this PR, I have opened a separate draft PR. https://github.com/zowe/zowe-explorer-vscode/pull/4180

JillieBeanSim (changes requested):
with quick testing it looks good except the extender fix that I just got merged is sorta here lol, I get the pop up error I fixed that resulted in nothing showing in the uss tree but then the tree expands and files show, maybe a timing issue. https://github.com/zowe/zowe-explorer-vscode/pull/4160 

Then I see possible regressions:

packages/zowe-explorer/src/trees/uss/USSActions.ts: canceling replace can leave a phantom local entry; directory “replace” path also appears to do nothing remotely
packages/zowe-explorer/src/trees/uss/UssFSProvider.ts: conflict cancel path mutates cached file contents before aborting, which can desync local state from remote

jace-roell:
> with quick testing it looks good except the extender fix that I just got merged is sorta here lol, I get the pop up error I fixed that resulted in nothing showing in the uss tree but then the tree expands and files show, maybe a timing issue. #4160
> 
> Then I see possible regressions:
> 
> packages/zowe-explorer/src/trees/uss/USSActions.ts: canceling replace can leave a phantom local entry; directory “replace” path also appears to do nothing remotely packages/zowe-explorer/src/trees/uss/UssFSProvider.ts: conflict cancel path mutates cached file contents before aborting, which can desync local state from remote

I believe the issue occured due to an empty directory entry, I pushed a fix for this that I also observed on the main branch. Let me know if this fixes the scenario you were seeing.

packages/zowe-explorer-api/src/fs/BaseProvider.ts
```diff
@@ -63,7 +63,17 @@ export class BaseProvider {
             this.FS_PROVIDER_UI_TIMEOUT
         );
         fsEntry.conflictData = null;
-        vscode.commands.executeCommand("workbench.action.closeActiveEditor");
+        if (closeEditor) {
+            vscode.commands.executeCommand("workbench.action.closeActiveEditor");
```
zFernand0, line 67: Dumb Q:
- Why hide this behind a boolean flag?

It seems that this was the expected behavior. So, do we want to give people the ability to leave the editor open when they decide to overwrite the remote contents?

packages/zowe-explorer-api/src/fs/BaseProvider.ts
```diff
@@ -63,7 +63,17 @@ export class BaseProvider {
...
+
+            // Find the main document and revert it to clear the dirty flag
+            const mainUri = uri.with({ query: "" });
+            const doc = vscode.workspace.textDocuments.find((d) => d.uri.toString() === mainUri.toString());
+            if (doc && doc.isDirty) {
```
zFernand0:
```suggestion
            if (doc?.isDirty) {
```

packages/zowe-explorer-api/src/fs/BaseProvider.ts
```diff
@@ -92,6 +102,13 @@ export class BaseProvider {
         );
         fsEntry.conflictData = null;
         vscode.commands.executeCommand("workbench.action.closeActiveEditor");
+
+        const mainUri = uri.with({ query: "" });
+        const doc = vscode.workspace.textDocuments.find((d) => d.uri.toString() === mainUri.toString());
+        if (doc && doc.isDirty) {
+            await vscode.window.showTextDocument(doc);
+            await vscode.commands.executeCommand("workbench.action.files.revert");
+        }
```
zFernand0: Should we encapsulate this in a little (could be private) utility function?

packages/zowe-explorer-api/src/fs/BaseProvider.ts
```diff
@@ -353,14 +370,18 @@ export class BaseProvider {
...
+        await this.diffOverwrite(uri, false);
         return ConflictViewSelection.Overwrite;
     }
 
+    //TODO (v4): rename _fireSoon -> fireSoon since it is no longer private
     /**
      * Internal VSCode function for the FileSystemProvider to fire events from the event queue
      */
-    protected _fireSoon(...events: vscode.FileChangeEvent[]): void {
+    public _fireSoon(...events: vscode.FileChangeEvent[]): void {
```
zFernand0: In addition to the TODO, can we create the `fireSoon` and deprecate this one? 
And in doing so, can we replace all uses of `_fireSoon` with `fireSoon`?

Quick steps to deprecate: 
- Rename the function to `fireSoon` (desired name - use `F2` so that VSCode finds all the references- if possible)
- Copy the function signature and name it `_fireSoon` (old name)
- Call `fireSoon` (new) function from the `_fireSoon` (old) function

packages/zowe-explorer/src/trees/dataset/DatasetActions.ts
```diff
@@ -2820,6 +2833,12 @@ export class DatasetActions {
             const member = name.split("(")[1].slice(0, -1);
             const res = await mvsApi.allMembers(dsname, options);
             if (res?.success && res.apiResponse?.items.some((m) => m.member == member.toUpperCase())) {
+                if (uri) {
+                    if (!DatasetFSProvider.instance.exists(uri)) {
```
zFernand0: (nit) Can we remove some indentation?
```suggestion
                if (uri && !DatasetFSProvider.instance.exists(uri)) {
```

packages/zowe-explorer/src/trees/dataset/DatasetActions.ts
```diff
@@ -2828,6 +2847,12 @@ export class DatasetActions {
             // Filter results to ensure exact match (e.g., A.B should not match A.B.C)
             const exactMatch = res?.success && res.apiResponse?.items.some((item) => item.dsname?.toUpperCase() === name.toUpperCase());
             if (exactMatch) {
+                if (uri) {
+                    if (!DatasetFSProvider.instance.exists(uri)) {
```
zFernand0: Same as previous comment 🙏 
Can we remove some indentation? 😋

packages/zowe-explorer/src/trees/dataset/DatasetActions.ts
```diff
@@ -1277,10 +1293,6 @@ export class DatasetActions {
             }
 
             if (newNode != null) {
-                if (replace === "notFound") {
-                    await vscode.workspace.fs.writeFile(newNode.resourceUri, new Uint8Array());
-                }
```
zFernand0, line 1282: I'm assuming we don't need to do this anymore since we are checking to see if the memberUri exists in the FS (a few lines above... `!DatasetFSProvider.instance.exists(memberUri)`)

Is that right?

packages/zowe-explorer/src/trees/dataset/DatasetFSProvider.ts
```diff
@@ -830,56 +829,60 @@ export class DatasetFSProvider extends BaseProvider implements vscode.FileSystem
...
+                entry.inDiffView = true;
+                return;
+            }
 
-                if (entry.wasAccessed || content.length > 0) {
-                    const resp = await this.uploadEntry(entry as DsEntry, content, uri, forceUpload, encodingParam);
-                    entry.etag = resp.apiResponse.etag;
-                }
-                entry.data = content;
+            if (!isNew || content.length > 0) {
```
zFernand0, line 868: Dumb question:
Why did we need the `entry.wasAccessed` check before? Should we keep it?

packages/zowe-explorer/src/trees/dataset/ZoweDatasetNode.ts
```diff
@@ -428,16 +428,15 @@ export class ZoweDatasetNode extends ZoweTreeNode implements IZoweDatasetTreeNod
...
-                            DatasetFSProvider.instance.createDirectory(dsNode.resourceUri);
-                        }
-                    } else {
-                        // Create an entry for the data set if it doesn't exist.
-                        if (!DatasetFSProvider.instance.exists(dsNode.resourceUri)) {
-                            await DatasetFSProvider.instance.writeFile(dsNode.resourceUri, new Uint8Array(), { create: true, overwrite: false });
-                        }
+                    const isCollapsible = dsNode.collapsibleState != vscode.TreeItemCollapsibleState.None;
+                    let dsType = isCollapsible ? DsType.Pds : DsType.PdsMember;
+                    const isSequential = SharedContext.isSession(this);
```
zFernand0, line 437: is `SharedContext.isSession(this);` verifying that `this` (assuming it's the parent?) is the actual profile/session node? In which case, the entry/dsNode is a sequential data set?

Should we leave some comments explaining the new checks here? I'm mostly wondering how is it that we determine PS vs Member 😅

packages/zowe-explorer/src/trees/job/JobFSProvider.ts
zFernand0, line 1: We removed the same sort of `_fireSoon` calls from the DatasetFSProvider, and we added a new `_fireSoon`  call for `entry.etag`.

I guess we don't need to do the same for Jobs since there is no concept of updating/saving a job (that could result in a conflict AND we don't have etags for job entries) 😋 

Please let me know if any of my assumptions are wrong 😅

packages/zowe-explorer/src/trees/uss/ZoweUSSNode.ts
```diff
@@ -284,7 +284,7 @@ export class ZoweUSSNode extends ZoweTreeNode implements IZoweUSSTreeNode {
             } else {
                 // Create an entry for the USS file if it doesn't exist.
                 if (!UssFSProvider.instance.exists(ussNode.resourceUri)) {
-                    await vscode.workspace.fs.writeFile(ussNode.resourceUri, new Uint8Array());
+                    UssFSProvider.instance.createEntry(ussNode.resourceUri, "file");
```
zFernand0, line 287: Does `fs.writeFile` make a remote call?

If so, we are avoiding a remote call here, which means we won't be creating an empty USS file anymore on the mainframe, just a local entry.

Follow-up: Do we create it (`fs.writeFile`) later on with actual contents?

packages/zowe-explorer/src/trees/uss/UssFSProvider.ts
```diff
@@ -742,43 +749,46 @@ export class UssFSProvider extends BaseProvider implements vscode.FileSystemProv
...
+                entry.size = content.byteLength;
+                return;
+            }
 
-                if (entry.wasAccessed || content.length > 0) {
-                    const resp = await this.uploadEntry(entry as UssFile, content, { forceUpload });
-                    entry.etag = resp.apiResponse.etag;
-                }
-                entry.data = content;
+            if (!isNew || content.length > 0) {
```
zFernand0, line 783: same as previous Q:
should we still check for wasAccessed here?

zFernand0 (commented):
Changes MSTM! 😋 
---

It looks like the same test failed across the board 🤔 
```
FAIL  __tests__/__unit__/trees/dataset/DatasetTree.unit.test.ts (22.013 s)
    ● Dataset Tree Unit Tests - Function datasetFilterPrompt › updates stats with modified date and user ID if provided in API
```

---

Also left some comments/questions, but mostly nitpicks and for my own understanding of things 🙏

packages/zowe-explorer-api/src/fs/BaseProvider.ts
jace-roell: This was just meant to distinguish behavior between overwrite via the notification prompt and from the editor toolbar overwrite in the diff view.

JillieBeanSim (changes requested):
Thanks for working on feedback @jace-roell, here are a few findings (ordered by severity) with code review

1) High: Failed remote writes can leave ghost entries in the local FS cache (no rollback on error)

- In dataset writes, a new entry is inserted into parent.entries before upload starts, then generic errors are thrown without removing that optimistic entry.
- Evidence:
  - packages/zowe-explorer/src/trees/dataset/DatasetFSProvider.ts:836
  - packages/zowe-explorer/src/trees/dataset/DatasetFSProvider.ts:848
  - packages/zowe-explorer/src/trees/dataset/DatasetFSProvider.ts:920
  - packages/zowe-explorer/src/trees/dataset/DatasetFSProvider.ts:934
- Same issue exists in USS writes.
- Evidence:
  - packages/zowe-explorer/src/trees/uss/UssFSProvider.ts:755
  - packages/zowe-explorer/src/trees/uss/UssFSProvider.ts:764
  - packages/zowe-explorer/src/trees/uss/UssFSProvider.ts:793
  - packages/zowe-explorer/src/trees/uss/UssFSProvider.ts:804
- Impact: after auth/network/permission failures, users can see files that were never created remotely; later operations may treat them as real entries.

2) High: Canceling replace prompts can still create local placeholder nodes and fire Created notifications

- Dataset flow creates a local entry and fires a Created event before asking the user whether to replace.
- Evidence:
  - packages/zowe-explorer/src/trees/dataset/DatasetActions.ts:2836
  - packages/zowe-explorer/src/trees/dataset/DatasetActions.ts:2839
  - packages/zowe-explorer/src/trees/dataset/DatasetActions.ts:2850
  - packages/zowe-explorer/src/trees/dataset/DatasetActions.ts:2853
  - packages/zowe-explorer/src/trees/dataset/DatasetActions.ts:1261
  - packages/zowe-explorer/src/trees/dataset/DatasetActions.ts:1283
  - packages/zowe-explorer/src/trees/dataset/DatasetActions.ts:1296
- USS flow has the same pattern.
- Evidence:
  - packages/zowe-explorer/src/trees/uss/USSActions.ts:102
  - packages/zowe-explorer/src/trees/uss/USSActions.ts:104
  - packages/zowe-explorer/src/trees/uss/USSActions.ts:114
  - packages/zowe-explorer/src/trees/uss/USSActions.ts:115
- Impact: user cancels replacement but still gets a locally-created node/event, causing UI state to diverge from the remote system.

3) Medium: New test coverage currently locks in the cancel-path regression instead of preventing it

- The new USS unit test explicitly expects createEntry and _fireSoon to run even when user selects Cancel.
- Evidence:
  - packages/zowe-explorer/__tests__/__unit__/trees/uss/USSActions.unit.test.ts:415
  - packages/zowe-explorer/__tests__/__unit__/trees/uss/USSActions.unit.test.ts:416
- Impact: this makes the faulty behavior harder to catch in future changes and masks intended cancel semantics.

Open questions / assumptions

- Assumption: Cancel should mean no local node creation and no Created event emission.
- If optimistic placeholders are intentional for UX, explicit rollback is still needed on cancel and failed upload.

packages/zowe-explorer/src/trees/dataset/DatasetFSProvider.ts
jace-roell: I am honestly unsure, though I dont believe the check was/is necessary.

packages/zowe-explorer/src/trees/uss/ZoweUSSNode.ts
jace-roell: `writeFile` when called from `ZoweDatasetNode` would have a buffer with a length of zero which would not result in an `uploadEntry` call (```content.length > 0```). We needed a helper function to just create the node to avoid generating a `created` notification when we are really just creating a placeholder node.

packages/zowe-explorer/src/trees/uss/USSActions.ts
```diff
@@ -88,15 +89,45 @@ export class USSActions {
...
+                    }
+                    if (!UssFSProvider.instance.exists(uri)) {
+                        UssFSProvider.instance.createEntry(uri, nodeType as "file" | "directory");
+                        UssFSProvider.instance.fireSoon({ type: vscode.FileChangeType.Created, uri: uri.with({ query: "" }) });
+                    }
+                }
+
+                if (!replace) {
+                    await ussApi.create(filePath, nodeType);
+                }
```
traeok: If an entry is created on line 114 in the filesystem, but `ussApi.create` fails on line 120, the local entry can remain in the cache. Granted, this scenario is quite uncommon but I thought of a couple situations where this may arise:

- insufficient user permissions to create new file/dir in target directory
- out of space on mountpoint where new USS item is being created

It may be worth adding a try/catch to handle this edge-case.

packages/zowe-explorer/src/trees/uss/UssFSProvider.ts
```diff
@@ -742,46 +749,53 @@ export class UssFSProvider extends BaseProvider implements vscode.FileSystemProv
...
+
+            this.fireSoon({ type: isNew ? vscode.FileChangeType.Created : vscode.FileChangeType.Changed, uri: uri.with({ query: "" }) });
         } catch (err) {
             if (!err.message.includes("Rest API failure with HTTP(S) status 412")) {
-                // Some unknown error happened, don't update the entry
+                // Some unknown error happened, rollback optimistic entry creation
+                if (isNew && parentDir.entries.has(fileName)) {
+                    parentDir.entries.delete(fileName);
+                    parentDir.size -= 1;
+                }
```
traeok, line 798: It seems like there could be an opportunity to de-duplicate rollback logic across DS & USS, but as discussed offline it seems that we do not accurately track size on filesystem entries for DS. I could file a new issue for this but in the meantime, it seems like we could still de-duplicate this "if check & delete" pattern as a new helper function and then add the `.size` change to USS - as a temporary measure until we address inconsistencies w/ DS provider.

packages/zowe-explorer/src/trees/uss/USSActions.ts
```diff
@@ -88,15 +89,63 @@ export class USSActions {
...
+                    const q = vscode.l10n.t({
+                        message: "The {0} already exists.\nDo you want to replace it?",
+                        args: [nodeType === "file" ? "file" : "directory"],
+                        comment: ["Node type"],
+                    });
+                    replace = stringReplace === (await Gui.showMessage(q, { items: [stringReplace, stringCancel] }));
+                    if (!replace) {
+                        return;
+                    }
+                    if (!UssFSProvider.instance.exists(uri)) {
```
JillieBeanSim, line 114: there is a potential race condition in optimistic cache creation. The code creates a cache entry optimistically (lines 115-117) before the API call completes. If multiple operations occur simultaneously on the same file, this could lead to inconsistent state or duplicate entries.

packages/zowe-explorer/src/trees/uss/UssFSProvider.ts
```diff
@@ -280,7 +281,8 @@ export class UssFSProvider extends BaseProvider implements vscode.FileSystemProv
                     rawItems.length === 1 && rawItems[0].name === uriBasename && !(rawItems[0].mode as string | undefined)?.startsWith("d");
                 const isDirectoryResponse =
                     (hasSelfEntry && (rawItems.find((item) => item.name === ".")?.mode as string | undefined)?.startsWith("d")) ||
-                    (rawItems.length > 0 && !hasSelfEntry && !isSingleFileMatch);
+                    (rawItems.length > 0 && !hasSelfEntry && !isSingleFileMatch) ||
```
JillieBeanSim, line 284: The condition rawItems.length === 0 && !hasSelfEntry assumes an empty response means an empty directory, but this could also indicate an API error or permission issue. Add validation to distinguish between empty directories and API failures.

packages/zowe-explorer/src/trees/uss/USSActions.ts
```diff
@@ -88,15 +89,63 @@ export class USSActions {
...
+                        if (!replace) {
+                            await vscode.workspace.fs.createDirectory(uri);
+                            UssFSProvider.instance.fireSoon({ type: vscode.FileChangeType.Created, uri: uri.with({ query: "" }) });
+                        }
+                    }
+                } catch (error) {
+                    // If we created a cache entry optimistically but subsequent operations failed,
+                    // clean up the cache entry to avoid leaving stale data
+                    if (cacheEntryCreated) {
+                        try {
```
JillieBeanSim, line 139: The catch block silently swallows all errors during cache cleanup. Add logging: catch (err) { ZoweLogger.trace("Failed to clean up cache entry:", err); }

packages/zowe-explorer/src/trees/uss/USSActions.ts
```diff
@@ -88,15 +89,63 @@ export class USSActions {
...
+                        message: "The {0} already exists.\nDo you want to replace it?",
+                        args: [nodeType === "file" ? "file" : "directory"],
+                        comment: ["Node type"],
+                    });
+                    replace = stringReplace === (await Gui.showMessage(q, { items: [stringReplace, stringCancel] }));
+                    if (!replace) {
+                        return;
+                    }
+                    if (!UssFSProvider.instance.exists(uri)) {
+                        UssFSProvider.instance.createEntry(uri, nodeType as "file" | "directory");
```
JillieBeanSim, line 115: The createEntry method is called but its return value is not validated. Add validation to ensure the entry was created successfully.

packages/zowe-explorer-api/src/fs/BaseProvider.ts
```diff
@@ -48,7 +48,7 @@ export class BaseProvider {
      * Action for overwriting the remote contents with local data from the provider.
      * @param remoteUri The "remote conflict" URI shown in the diff view
      */
-    public async diffOverwrite(uri: vscode.Uri): Promise<void> {
+    public async diffOverwrite(uri: vscode.Uri, closeEditor: boolean = true): Promise<void> {
```
JillieBeanSim, line 58: The method signature changed from diffOverwrite(uri) to diffOverwrite(uri, closeEditor = true). While backward compatible for direct calls, this may break code using it as a callback.

packages/zowe-explorer/src/trees/uss/UssFSProvider.ts
```diff
@@ -742,46 +749,53 @@ export class UssFSProvider extends BaseProvider implements vscode.FileSystemProv
...
+            entry.data = content;
+            entry.mtime = Date.now();
+            entry.size = content.byteLength;
+
+            this.fireSoon({ type: isNew ? vscode.FileChangeType.Created : vscode.FileChangeType.Changed, uri: uri.with({ query: "" }) });
         } catch (err) {
             if (!err.message.includes("Rest API failure with HTTP(S) status 412")) {
-                // Some unknown error happened, don't update the entry
+                // Some unknown error happened, rollback optimistic entry creation
+                if (isNew && parentDir.entries.has(fileName)) {
```
JillieBeanSim: The rollback logic assumes the parent directory still exists. Add null check: if (isNew && parentDir?.entries.has(fileName)) { ... }

JillieBeanSim (commented):
thanks for addressing feedback @jace-roell  looking over I see improved performance, improved conflict resolution, better event management, and proper cleanup of optimistically created entries. I left a couple comments about a few concerns.

packages/zowe-explorer/src/trees/uss/UssFSProvider.ts
traeok: If there was an API error, either `response.success` would be `false` (therefore not reaching this code) or the API would throw an error (meaning it won't get past line 264).

If the API ran into a permission issue, it should either throw an error or return `success: false`, which is already covered in the pre-existing logic

packages/zowe-explorer/src/trees/uss/USSActions.ts
jace-roell: Since `createUSSNode` is ran specifically for creation in USS, I don't believe there would realistically be multiple simultaneous operations on the resource being created.

Do you have any specific scenarios in mind that would result in an inconsistent state?

packages/zowe-explorer/src/trees/uss/USSActions.ts
jace-roell: Since `createEntry` is guaranteed to return a `UssFile | UssDirectory` or to throw an error from `lookupParentDirectory`, I dont believe validation in `USSActions` is necessary.

I have added a `@throws` to the tsdoc for `createEntry`.

packages/zowe-explorer/src/trees/uss/USSActions.ts
JillieBeanSim: no specific scenario, just mentioning concerns I came across while reviewing could just be an edge case that we can wait to see if users experience.  I see we have a catch that will delete if already created so looks like covering from our side I suppose.

JillieBeanSim (approved):
thanks @jace-roell for answering and addressing feedback/concerns.  tested with extender type and looks good

PR #4182, opened 2026-04-16
fix(audit): use workaround for pnpm audit in CI

traeok:
## Proposed changes

- Uses workaround suggested in https://github.com/pnpm/pnpm/issues/11265 to allow audit to succeed in CI, without upgrading to pnpm 11 (as its still an RC)
- Fixes dompurify vulnerability by updating expected version range to ^3.3.4

## Types of changes

- [x] Other (please specify above in "Proposed changes" section)

## Checklist

**General**

**Code coverage**

### DIFF
--- a/.github/workflows/audit.yml
+++ b/.github/workflows/audit.yml
@@ -21,6 +21,7 @@ jobs:
           npm install -g pnpm
           pnpm install
 
+      # https://github.com/pnpm/pnpm/issues/11265#issuecomment-4258309621 Can change back to pnpm audit once pnpm 11.0.0 is released
       - name: Check Vulnerabilities
         run: pnpm dlx --config.minimumReleaseAge=0 pnpm@11.0.0-rc.1 --config.manage-package-manager-versions=false audit --prod
   audit-all:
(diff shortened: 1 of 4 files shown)


zFernand0 (commented):
LGTM! 😋 
---

As long as we limit the use of `--config.minimumReleaseAge=0 ` to commands that won't install any dependencies, I think we should be fine. 

For example, `pnpm audit fix` could be a bad one to add this option 😋 

---

Overall, I like this clever trick to try and force pnpm to use a specific version 🙏 

---

Thanks for updating `dompurify` so we might be able to close out #4181.
However there are some references to `@types/node@22.x` that should not be in the lockfile which that other PR fixes 🙏 

Could you also look into that, or can we merge that other PR first?

.github/workflows/audit.yml
```diff
@@ -21,8 +21,9 @@ jobs:
           npm install -g pnpm
           pnpm install
 
+      # https://github.com/pnpm/pnpm/issues/11265#issuecomment-4258309621 Can change back to pnpm audit once pnpm 11.0.0 is released
```
zFernand0, line 24: Q: Should we create an issue (to change it back when 11 comes out)?
I don't think that will be necessary since we use them daily 😋

zFernand0 (approved):
LGTM! 😋

.github/workflows/audit.yml
traeok: Good call, I can create one 👍

t1m0thyj (approved):

.github/workflows/audit.yml
traeok: Filed as #4183

PR #4201, opened 2026-04-23
Accessibility Fix (L1): Add tooltips to zowe resource table icons

likhithanimma1:
## Proposed changes

Solves https://github.com/zowe/zowe-explorer-vscode/issues/4020

## Release Notes

Milestone:

Changelog:

## Types of changes

## Checklist

**General**

**Code coverage**

**Deployment**


anaxceron (approved):

traeok (changes requested):
Thanks for working on this Likhitha! However, I'm seeing some unexpected alignment issues with the tooltip: 

Image

It seems shifted off to the left and down, which makes the tooltip look like its for the previous column 🤔

likhithanimma1:
> Thanks for working on this Likhitha! However, I'm seeing some unexpected alignment issues with the tooltip:
> 
> Image
> It seems shifted off to the left and down, which makes the tooltip look like its for the previous column 🤔

Thanks @traeok . I have adjusted the alignment of tooltip, can please review now.

JillieBeanSim (commented):
thanks for working on this @likhithanimma1 I do see that when I am moving through the table via the keyboard I cannot select a row for actions, I only have the option to tab through the table items

packages/zowe-explorer/src/webviews/src/table-view/TableView.tsx
```diff
@@ -590,6 +602,93 @@ export const TableView = ({ actionsCellRenderer, baseTheme, data }: TableViewPro
   const onGridReady = () => {
     messageHandler.send("api-ready");
 
+    // ---------------------------------------------------------------
+    // Keyboard Navigation Setup for Row Selection
+    // ---------------------------------------------------------------
+    const handleKeyDown = (event: KeyboardEvent) => {
```
traeok: It appears that we're rebuilding the entirety of keyboard navigation because the tooltip changes break AG Grid's built-in keyboard navigation. I'm concerned about the robustness of these changes and curious to know if there's a simpler alternative that doesn't break this built-in navigation experience. As a last resort, I'm in favor of keeping pre-existing behavior with the trade-off of less accessibility around icons.

traeok (changes requested):

packages/zowe-explorer/src/webviews/src/table-view/TableView.tsx
likhithanimma1: Hey @traeok . Sorry i didn't check it before but the row selection is still in place as before with the tooltip changes without anything adding additionally. So i did revert my changes regarding keyboard navigation.

likhithanimma1:
> thanks for working on this @likhithanimma1 I do see that when I am moving through the table via the keyboard I cannot select a row for actions, I only have the option to tab through the table items

Hey @JillieBeanSim . Noticed that can select row with the space button.

JillieBeanSim (approved):
thanks for clarifying @likhithanimma1 I can select via the space key in the zowe resources table but not via the enter key. I notice in other views we have like the Edit History the enter key and space key can be used to select items.  I see that the marketplace also only allows the space key as selector key in zowe resources table so not a regression but maybe we should open an issue to get consistency between webviews and actions allowed.

traeok:
"Space to select" was implemented as part of AG Grid keyboard support: https://www.ag-grid.com/javascript-data-grid/row-selection/

> Configure the grid to allow users to select rows by clicking them, focusing a row and pressing ␣ Space or via API

That said I think an enhancement request would be good to also support Enter 👍

traeok (approved):
LGTM, thanks for quickly addressing feedback

zFernand0 (approved):
LGTM! 😋 
---

No changes requested, but check this out 
💫 
😵 

Unrelated to this PR, but have you noticed that the sorting has 3 different states?
- ascending, 
- descending, 
- no-sorting

This works well for all columns, but feels a bit awkward for the default sort order (by data set name) set by the server. 
Here is a GIF to show what I mean 😅 

Image

PR #4212, opened 2026-04-29
Change password on server side

### LINKED ISSUE #1538: Change Password using Zowe CLI APIs
As a Zowe  Explorer User, I would like to have the ability to change my password for the mainframe through Zowe Explorer. 

dependency on Zowe CLI making use of the z/OSMF APIs:
z/OSMF implemented a REST API auth service that will make this enhancement possible:  https://www.ibm.com/docs/en/zos/2.4.0?topic=services-zosmf-authentication

JWaters02:
## Proposed changes

Closes #1538 

Allows the user to change their password on the server of the selected profile, which then also updates their stored password.

Right click profile -> Manage profile -> Change password
Enter old password
Enter new password
Confirm new password

It tries to update the password on the server side first, before doing anything on client side. 

In the case that the password is updated on the server side but fails to modify it in the profile, it shows a warning telling the user to try doing it again manually with update credentials action. But in any case it shouldn't be too much of an issue as when the server does eventually start to reject the old password (which is usually after midnight for z/OSMF) the user will just be prompted to update credentials as normal.

## Release Notes

Milestone: 3.7.0

Changelog: 

## Types of changes

## Checklist

**General**

**Code coverage**

**Deployment**

## Further comments

Note: Ross Cruickshank said that zxplore system will be disabling the ability for passwords to be changed from anywhere but in the website portal, but I'm not sure when that will be implemented. So once this goes in, there may be a chance that issues are raised by zxplore users who are unable to change their passwords - we should be aware of this so there is no time wasted investigating.


JillieBeanSim (commented):
thanks for working on this @JWaters02 this is a great additional capability.  Should we make this extensible like we have done for many other profile actions, i.e., update credentials and create config? This could be a nice feature for extenders like CICS to be able to use from their view as well.  It could be a separate story but thinking we should develop with this in mind for new features as well to avoid having to always go back and refactor once requested.

JWaters02:
> thanks for working on this @JWaters02 this is a great additional capability. Should we make this extensible like we have done for many other profile actions, i.e., update credentials and create config? This could be a nice feature for extenders like CICS to be able to use from their view as well. It could be a separate story but thinking we should develop with this in mind for new features as well to avoid having to always go back and refactor once requested.

Totally agree with you on that one, and I think it should be done in this PR as we have months until 3.6 code freeze so there is no rush and it can be done properly from the start. I can start to have a look at the context around this (like how update credentials does it) and also talk to Dave about CICS extension to see what thoughts he has about it.

packages/zowe-explorer-api/src/extend/MainframeInteraction.ts
```diff
@@ -80,6 +80,19 @@ export namespace MainframeInteraction {
          * @returns {string} the token type name as defined by a CLI plugin that implements the profile.
          */
         getTokenTypeName?(): string;
+
+        /**
+         * Change the password for a user on the remote system.
+         *
+         * Unlike "Update Credentials", which only updates the locally stored
+         * password, this operation contacts the server to change the password
+         * on the mainframe and then updates the local credential store.
```
traeok: Is this function expected to update the local credential store? That seems a bit unconventional as ZE usually manages changes to credentials in the secure store.
I'd recommend adjusting this comment if ZE is handling the local credential changes.

packages/zowe-explorer-api/src/profiles/ZoweExplorerZosmfApi.ts
```diff
@@ -108,6 +108,13 @@ import { IDataSetCount } from "../dataset/IDataSetCount";
         public logout(session: imperative.Session): Promise<void> {
             return Logout.apimlLogout(session);
         }
+
+        public async changePassword(session: imperative.Session, newPassword: string): Promise<void> {
+            const response = await zosmf.ZosmfChangePassword.changePassword(session, newPassword);
+            if (!response.success) {
+                throw new Error(response.message ?? "Password change failed");
+            }
```
traeok: I'm not opposed to this pattern, but wondering if it would be more consistent to just return the RHS of `response` directly (as we do w/ other APIs) and then check for `response.success` where the API is used?

packages/zowe-explorer/src/management/ProfileManagement.ts
```diff
@@ -219,7 +230,16 @@ export class ProfileManagement {
     }
 
     private static basicAuthQp(node: IZoweTreeNode): vscode.QuickPickItem[] {
+        const profile = node.getProfile();
         const quickPickOptions: vscode.QuickPickItem[] = Object.values(this.basicAuthUpdateQpItems);
+        try {
+            const commonApi = ZoweExplorerApiRegister.getInstance().getCommonApi(profile);
+            if (typeof commonApi.changePassword === "function") {
```
traeok: `commonApi.changePassword` is either `undefined` or an implemented function with type `(session: imperative.Session, newPassword: string) => Promise<void>`. Can we simplify this to a null check to match other places where optional APIs are used?


```suggestion
            const commonApi = ZoweExplorerApiRegister.getInstance().getCommonApi(profile);
            if (commonApi.changePassword != null) {
```

packages/zowe-explorer/src/management/ProfileManagement.ts
```diff
@@ -77,6 +78,12 @@ export class ProfileManagement {
             description: vscode.l10n.t("Update stored username and password"),
         },
     };
+    public static readonly changePasswordQpItems: Record<string, vscode.QuickPickItem> = {
+        [ProfileManagement.AuthQpLabels.changePassword]: {
+            label: `$(lock) ${vscode.l10n.t("Change Password")}`,
+            description: vscode.l10n.t("Change your password on the server and update stored password"),
+        },
+    };
```
traeok, line 86: Do we intend to add other items related to changing passwords to this quick pick?
Otherwise, not sure that the array is necessary.

JWaters02:
Hey, sorry, can you set this to draft PR? I don't have the ability to, and this change is WIP for the suggestion billie made. thanks

JWaters02:
@JillieBeanSim Hi, so I had a chat with dave about if this would be useful for them from an extenders point of view, but we aren't really sure how it would be. For CICS extension, they don't have any need to override the change password on server side functionality as they don't have an API for that at all on the server side (if am I understanding correctly @davenice?)

I don't know of any other extenders that might want to use their own implementations of change password on server side, but if they contribute their own profile type, they can already contribute to it by implementing their side of the change password endpoint. Else, I can't think of cases where the standard pattern of "old password + new password + new password again" would want to be changed, except maybe only in the wording? If they aren't changing "password" but something like a "passphrase". 

Could you elaborate on what you had in mind? Thanks!

packages/zowe-explorer-api/src/extend/MainframeInteraction.ts
```diff
@@ -80,6 +80,19 @@ export namespace MainframeInteraction {
...
+         *
+         * @param {imperative.Session} session The session containing the current (old) credentials.
+         * @param {string} newPassword The new password to set on the remote system.
+         * @returns {Promise<void>} Resolves when the password has been changed.
+         */
+        changePassword?(session: imperative.Session, newPassword: string): Promise<void>;
```
JillieBeanSim: Hey @JWaters02, yes this new method should suffice for the extender support I was requesting so extender's could adopt the functionality for their protocols as well in Zowe Explorer.

packages/zowe-explorer/src/management/ProfileManagement.ts
JWaters02: I'm merely copying what all the other quick picks in manage profile are doing, even though most of them also are not having other items added to them. I've left this be for consistency

packages/zowe-explorer-api/src/profiles/IChangePasswordResponse.ts
JWaters02, line 1: Not sure if this is the right place for this interface.

This exists as a generic name alternative to the SDK's `IZosmfChangePasswordResponse` because it would be confusing to extender developers if they are pulling in or using an interface named "Zosmf" even though it's in the extender API so it can be for any profile type.

packages/zowe-explorer/src/utils/ProfilesUtils.ts
```diff
@@ -505,6 +508,165 @@ export class ProfilesUtils {
...
+        // Actually, we can't do a server side check of old password in case the old password expired and we are just changing it...
+        // so what if we have it do the update credentials flow first instead?
+        // If we remove this check, we are removing the "validation" that the user changing the password is the user that owns the account
+        const storedPassword = session.ISession.password;
+        if (storedPassword != null) {
+            // Passwords are not case sensitive
+            if (oldPassword.toLowerCase() !== storedPassword.toLowerCase()) {
+                Gui.errorMessage(vscode.l10n.t("Current password does not match the stored credentials. Password was not changed."));
+                return;
+            }
```
JWaters02, line 573: What are people's thoughts on this logic? See TODO comment.

packages/zowe-explorer/src/utils/ProfilesUtils.ts
```diff
@@ -505,6 +508,165 @@ export class ProfilesUtils {
...
+            await profInfo.updateProperty({
+                profileName: profile.name,
+                profileType: profile.type,
+                property: "password",
+                value: newPassword,
+                setSecure: profInfo.isSecured(),
+            });
+            profile.profile.password = session.ISession.password = newPassword;
+            imperative.AuthOrder.addCredsToSession(session.ISession, ZoweExplorerZosmf.CommonApi.getCommandArgs(profile));
+            AuthHandler.unlockProfile(profile, true); //TODO: Check if this is necessary, and if so, does it go before or after updateCachedProfile?
```
JWaters02, line 644: I'm not sure whether or not to call `Auth.unlockProfile` - currently I am because it seems reasonable to expect that if the server responded with password changed successfully and the local credential update is done, then the creds should be in sync with the server on the profile so requests can resume if they had been paused by the password expiring. If keep- do I call it before or after `Constants.PROFILES_CACHE.updateCachedProfile(profile, node);`?

packages/zowe-explorer/src/utils/ProfilesUtils.ts
```diff
@@ -505,6 +508,165 @@ export class ProfilesUtils {
...
+            Constants.PROFILES_CACHE.updateCachedProfile(profile, node);
+            ZoweVsCodeExtension.onProfileUpdatedEmitter.fire(profile);
+            SharedTreeProviders.getProviderForNode(node).refreshElement(node);
```
JWaters02, line 647: In `Profiles.promptCredentials`, there is this code:

```ts
// If secure credentials are enabled, the config file won't change after updating existing credentials
// (as the "secure" fields are already set). Fire the event emitter to notify extenders of the change.
if (SettingsConfig.getDirectValue<boolean>(Constants.SETTINGS_SECURE_CREDENTIALS_ENABLED)) {
   ZoweExplorerApiRegister.getInstance().onProfilesUpdateEmitter.fire(Validation.EventType.UPDATE);
}
```

Should I be doing the same thing? Because I'm not sure if `ZoweVsCodeExtension.onProfileUpdatedEmitter.fire(profile);` already "covers" this in terms of notifying the extender that the profile has changed.

JWaters02:
Hi all, sorry for leaving this for a long time (maybe this will become a yearly tradition, like the "navigation history" PR on USS tree 😂 ). 

I've done manual tests for the cases where autostore is true/false in combinations when secure creds is enabled or disabled and when I change password while ZE is in the middle of doing requests (polling a long running job) and everything appears to be working on that front- but have some questions about it that I've written in review comments.

JWaters02:
At work today my mainframe password expired so I took this as an oppurtunity to verify that running the change password command from my zosmf profile would successfully update the expired password to a new one and unlock my account - and it did! 

The thing I noticed is that of course when the password expires, ZE prompts with the update credentials box, so when I wanted to change password I first had to click cancel, then open the change password field. This isn't great user experience so it could be suggested to replace the update credentials with change password, but then sometimes the prompt appears not because the password has expired on the server but because the local credentials are wrong, so keeping the update credentials button is correct. Maybe we could add a change password button to this prompt if the user knows their password has expired? But I'm not sure if this makes the popup more confusing. 

Lmk on your thoughts please.

JWaters02:
While looking at IBM Z Open Editor docs I found this section: https://ibm.github.io/zopeneditor-about/Docs/interact_zos_zopeneditor.html#rse-password-expiry-prompt

it seems that through RSE API they already have a way to change password on server side and update local credentials in one action? If so, then it would probably be a good idea to get their extender POV on this feature.

PR #4243, opened 2026-05-11
fix: do not set `mtime` on etag change (not in API response)

traeok:
### Issue on `main`

On `main` we assign `mtime` on a filesystem entry to `Date.now()` when the e-tag is updated:

https://github.com/zowe/zowe-explorer-vscode/blob/a8d3cf3866da02764bc6efbe33d88888b878db11/packages/zowe-explorer/src/trees/dataset/DatasetFSProvider.ts#L606-L611

The same change was applied to the USS filesystem provider:

https://github.com/zowe/zowe-explorer-vscode/blob/a8d3cf3866da02764bc6efbe33d88888b878db11/packages/zowe-explorer/src/trees/uss/UssFSProvider.ts#L548-L553

This is prone to false positives for a couple reasons:
1. It assumes that `mtime` is the current time, which is not a guarantee
2. It assumes an `mtime` value for a data set or resource that may not offer a valid timestamp from the API

As a result of this `mtime` update, it also surfaces an issue where the `mtime` is immediately updated to `Date.now()` as soon as the file or data set is opened in the editor. At this time, the e-tag on the entry is `undefined` and the e-tag from the API response (when pulling its contents) is some stringified hex value, thus satisfying the `if` condition above. When saving, this time is compared against the `mtime` cached in the editor modal. The time from the `stat` call is newer, surfacing the conflict message.

For context, we had always assigned the `mtime`, seeing this as far back as [06c5794](https://github.com/zowe/zowe-explorer-vscode/commit/06c5794ccc8999f57831f49407e4d74dc6e16121). However, with fetch by default in place, this logic now has greater implications.

### Fix

Skipping the `mtime` update seems to be the quickest way forward since:
1. We cannot fully leverage VS Code's built-in etag handling: it relies on `mtime` (a timestamp) that we can't always provide to calculate it
2. **We do not set expectations** on an API to return an `mtime` or an `etag`. There is no type-checking for these responses.
3. Keeping the `mtime` old prevents the false positive conflict
4. Existing conflict handling (412 status code over HTTP) continues to work as-is

### Long-term proposal

We **need** to set a standard for what to expect when reading files, data sets and listing them in v4 - this includes all supported attributes, the format they should be in, and when to expect them.

Having this prevents us from playing guessing games of "does this API return an mtime? what about an etag? what about 412 error from conflict?" ... because we have to now since `apiResponse` is **typed as any** (bad practice):

https://github.com/zowe/zowe-cli/blob/d484bd7ffa4f7a4d9a09ce2536e8e41ea9f0064a/packages/zosfiles/src/doc/IZosFilesResponse.ts#L29-L33

Setting these expectations avoids making assumptions about what APIs should do - response types fill the gap.

## How to test

- Install the VSIX from this PR.
- Open Zowe Explorer and search for a pattern matching a PDS that you control OR have write access to.
- Switch to your 3270 emulator.
- Create a new member without any statistics and save some contents into the member. When saving, make sure to skip updating statistics - meaning that no attributes are saved to the data set. (In ISPF you can do this by typing `stats off` in the Command area of the edit view. Then, close out of the edit view.)
- Switch to Zowe Explorer. Refresh the member list for that PDS (or open it if you haven't already) & open the new member.
- Make changes to the member and save.
- Notice that the "unsaved" symbol is removed from the editor tab and the file is saved. Verify that the save succeeds by checking the contents from the 3270 emulator.
  - No conflict dialog should appear
  - On `main`, the conflict dialog appears here and the save is unsuccessful

## Release Notes

Milestone: 3.5.0

Changelog: Pre-existing changelog entry

## Types of changes

## Checklist

**General**

**Code coverage**

**Deployment**

### DIFF
--- a/packages/zowe-explorer/__tests__/__unit__/trees/dataset/DatasetFSProvider.unit.test.ts
+++ b/packages/zowe-explorer/__tests__/__unit__/trees/dataset/DatasetFSProvider.unit.test.ts
@@ -555,10 +555,9 @@ describe("DatasetFSProvider", () => {
             expect(_updateResourceInEditorMock).toHaveBeenCalledWith(testUris.ps);
         });
 
-        it("updates mtime if the fetched e-tag differs from the existing e-tag", async () => {
+        it("should not update mtime as part of e-tag check & sync", async () => {
             const contents = "dataset contents";
             const oldMtime = 1000;
-            const newMtime = 2000;
 
             const mockMvsApi = {
                 getContents: jest.fn((dsn, opts) => {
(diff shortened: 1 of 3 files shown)


jace-roell (approved):

zFernand0 (approved):
LGTM! 😋 
---

Apologies for the delay 😅 
Thanks for updating the test to reflect this behavior 🙏

PR #4244, opened 2026-05-12
copy frontend vsce implementation

MarkAckert:
## Proposed changes

Addresses [Copy Datasets Implementation](https://github.com/zowe/zowex/issues/377)

## Release Notes

Milestone:
- 3.6.0

Changelog:
- Added the functionality to copy datasets and members in the Data Sets tree. [#377](https://github.com/zowe/zowex/pull/377)

## Types of changes

## Checklist

**General**

- ~[ ] New ZE APIs are tested with extender types that haven't adopted yet to determine breaking changes. Can use Zowe zFTP marketplace extension.~

**Code coverage**

**Deployment**

- ~[ ] I have tested new functionality with the FTP extension and profile verifying no extender profile type breakages introduced~
- ~[ ] Documentation should be added to Zowe Docs~
  ~- If you're an outside contributor, please post in the [#zowe-doc Slack channel](https://openmainframeproject.slack.com/archives/CC961JYMQ) to coordinate documentation.~
  ~- Otherwise, please check with the rest of the squad about any needed documentation before merging.~
- ~[ ] These changes may need ported to the appropriate branches (list here):~

## Further comments

Paired with PR in zowex repo: https://github.com/zowe/zowex/pull/995


packages/zowex-for-zowe-explorer/src/api/SshMvsApi.ts
```diff
@@ -245,16 +246,263 @@ export class SshMvsApi extends SshCommonApi implements MainframeInteraction.IMvs
...
+        }
+
+        try {
+            const response = await (await this.client).ds.createDataset({
+                dsname: dataSetName,
+                attributes: attributes,
+            });
+
+            if (response.success) {
+                Gui.infoMessage(`Successfully allocated dataset "${dataSetName}" like "${likeDataSetName}"`);
```
traeok: While this is informative for end users, I wonder whether the notifications can be excessive when copying several sequential or partitioned data sets that need allocated on the target system. What are your thoughts on "condensing" this into a single message or using `setStatusBarMessage` instead to avoid a swarm of allocation notifications?

packages/zowex-for-zowe-explorer/src/api/SshMvsApi.ts
```diff
@@ -245,16 +246,263 @@ export class SshMvsApi extends SshCommonApi implements MainframeInteraction.IMvs
...
+            if (!response.success) {
+                const errorMsg = `Failed to copy ${fromDataSetName} to ${toDataSetName}.`;
+                Gui.errorMessage(errorMsg);
+                return this.buildZosFilesResponse(response, false, errorMsg);
+            }
+            return this.buildZosFilesResponse(response, true);
+        } catch (error) {
+            const errorMsg = error instanceof Error ? error.message : String(error);
+            Gui.errorMessage(`Error while copying ${fromDataSetName} to ${toDataSetName}: ${errorMsg}`);
+            return this.buildZosFilesResponse({ success: false }, false, errorMsg);
```
traeok: While the UI messages are definitely appreciated, can we surface these at the Zowe Explorer level? I noticed there is already some error handling in-place in `DatasetActions.ts` for catching copy/paste errors. I'm curious to know if there's opportunity for code de-duplication here or if that should be a separate effort.

traeok (commented):
Thanks Mark! Doing some manual testing combined w/ the PR on zowex, and I plan to approve if all goes well. In the meantime I left a couple suggestions based on my interpretation of the new UI messages in this PR.

pujal0909 (changes requested):
~thanks @MarkAckert. I did some manual testing and pds to pds and member to member scenarios look good, but getting this error when copy pasting sequential data set:~

Image

packages/zowex-for-zowe-explorer/src/api/SshMvsApi.ts
MarkAckert: I read through the calling code a little more closely and agree it could be overwhelming. I switched to `setStatusBarMessage` since it's easy :smile: but also think we could simply remove this message altogether if that's still too noisy. Part of the value of this message depends on caller context which isn't available in this method - are we copying 1 node, multiple nodes, is the user issuing an allocateLike directly? We should leave such decisions to the calling code which is outside the scope of this specific change.

packages/zowex-for-zowe-explorer/src/api/SshMvsApi.ts
MarkAckert: I removed this message for now, will take a closer look at de-duplication and error handling

pujal0909 (changes requested):
Not seeing an error message anymore, but when I copy paste a sequential data set, the target gets created but the content is not copied over. The data set is empty in ISPF as well.

MarkAckert:
> Not seeing an error message anymore, but when I copy paste a sequential data set, the target gets created but the content is not copied over. The data set is empty in ISPF as well.

Addressed this issue. cc @pujal0909 

**Cause:**
When we're copying a sequential dataset within a profile and the target does not already exist, we first allocate the dataset before invoking the copy. In this case, we didn't pass a `replace=true` parameter to the `mvsApi.copyDataset` function, so the copy function failed in `zowex` (you must specify `--replace` to overwrite a sequential dataset) and we're left with the empty dataset. The exact same code path works with z/OSMF because those APIs always replace the sequential dataset, even when `replace=false` (or is absent).

So we have a difference in behavior between the `zowex` and `z/OSMF` backends. I chose to leave that difference rather than change `zowex` since I think `zowex`'s current behavior is more reasonable, and instead set `replace=true` when we're allocating a brand new dataset as part of a copy operation.

packages/zowex-for-zowe-explorer/src/api/SshMvsApi.ts
traeok: Thanks for the change to use `setStatusBarMessage` - since it replaces the last message with the new one, the noise should be minimal. The updates are still visible without introducing a queue of notifications that need cleared.

I agree on leaving the decisions to the calling code. This is something we can revisit in a future PR around Zowe Explorer's copy UX.

packages/zowex-for-zowe-explorer/src/api/SshMvsApi.ts
traeok: Sounds good, thanks. I think that the de-duplication could be a separate effort now that the message was removed.

pujal0909 (changes requested):
Image

Hey @MarkAckert ,
Thanks for addressing my previous comments. Copy pasting sequential data sets works for me now. The content are being pasted over, but seeing this error when copy pasting a pds or member. The contents seem to copy over though.

MarkAckert:
I'm not able to reproduce this error. I tried a bunch of different copy operations between PDS and PDSE datasets + members. 

Could you provide more info on the PDS/members being copied so I can reproduce it?

pujal0909:
> I'm not able to reproduce this error. I tried a bunch of different copy operations between PDS and PDSE datasets + members.
> 
> Could you provide more info on the PDS/members being copied so I can reproduce it?

never mind, not seeing it anymore. thanks!

pujal0909 (approved):
LGTM! thanks Mark

traeok (approved):
Thanks Mark, LGTM

packages/zowex-for-zowe-explorer/CHANGELOG.md
```diff
@@ -6,6 +6,7 @@ Check [Keep a Changelog](http://keepachangelog.com/) for recommendations on how
 
 ## Recent Changes
 
+- Added the functionality to copy datasets and members in the Data Sets tree. [#377](https://github.com/zowe/zowex/pull/377)
```
anaxceron, line 9:
```suggestion
- Added the functionality to copy data sets and members in the DATA SETS tree. [#377](https://github.com/zowe/zowex/pull/377)
```

PR #4299, opened 2026-05-31
feat: allow multiple selection of members to download

### LINKED ISSUE #4274: Enhance download member to support multi-select

Suggested during the systems demo. Can either download a single member or all members of a PDS, but not a selection of members. 

Support multiselect to download a selection of members

Download all members and delete the ones not needed.

**Additional context**

JWaters02:
## Proposed changes

Implements #4274 
Enhanced data set member download to allow multiple members to be selected and downloaded in the same action.
- supports multiple members across multiple partitioned data sets
- supports members across multiple profiles

image

If a download action on one of the profiles fails, it carries on and notifies the user, for example here to simulate a profile invalidating mid-download, I edited port of `zosmf_invalid` and saved config so it reloaded, then the download continued on as if nothing had happened, eventually giving a notification that the download completed with errors. It had downloaded the first member (`Z02589/CBL/BLAH.cbl`) from the invalid profile right before I changed it to invalid port. Then the `Z02589/JCL/*` members below were from the `zosmf_zxplore` profile.

image

## Release Notes

Milestone:

Changelog: Enhanced data set member download to allow multiple members - including across unique partitioned data sets - to be selected and downloaded in the same action.

## Types of changes

## Checklist

**General**

**Code coverage**

**Deployment**

## Further comments

Originally I had the implementation be just adding support to DatasetActions.downloadMember for node list but then when you right click multi select the command showed as `Download member...` as singular so I opted instead to have an entirely new command just so the command name could use plural (`Download members...`). Perhaps a bit of an overcook but it did also mean a bit less confusing code flow in that now `downloadMember` is split off into `downloadMember` and `downloadSingleMember`, so then `downloadMembers` could also call `downloadSingleMember` as a common function. Ultimately the code deduplication was about the same but I think this reads better as otherwise there was more code paths in a single function.


packages/zowe-explorer/src/trees/dataset/DatasetActions.ts
```diff
@@ -958,45 +998,122 @@ export class DatasetActions {
...
+                        if (response?.commandResponse) {
+                            commandResponses.push(response.commandResponse.toString());
+                        }
+                    } catch (e) {
+                        hasErrors = true;
+                        const nodeLabel = String(selectedNode.getLabel() ?? "");
+                        const profileName = String(profile?.name ?? "");
+                        const profileSuffix = profileName ? ` (${profileName})` : "";
+                        const message = e instanceof Error ? e.message : String(e);
+                        commandResponses.push(`Failed to download ${nodeLabel}${profileSuffix}: ${message}`);
```
traeok, line 1115: Can we log errors from this try/catch using `ZoweLogger`?

packages/zowe-explorer/src/trees/dataset/DatasetActions.ts
```diff
@@ -958,45 +998,122 @@ export class DatasetActions {
...
+                        downloadedCount++;
+
+                        if (response?.success === false) {
+                            hasErrors = true;
+                        }
+
+                        if (response?.commandResponse) {
+                            commandResponses.push(response.commandResponse.toString());
+                        }
+                    } catch (e) {
```
traeok, line 1109: If an error occurs in this try/catch related to authentication, the logic moves onto the next selected node and tries to download. Should we break in event of a 401/authorization error from the API call?

Suggestion: You could try using `AuthUtils.errorHandling` in the try/catch if the error is a 401/auth-related error, allowing user to handle authentication failure.

packages/zowe-explorer/src/trees/dataset/DatasetActions.ts
```diff
@@ -958,45 +998,122 @@ export class DatasetActions {
...
+
+                for (const selectedNode of validSelectedNodes) {
+                    if (token?.isCancellationRequested) {
+                        break;
+                    }
+
+                    const profile = selectedNode.getProfile();
+
+                    try {
+                        const { response } = await DatasetActions.downloadSingleMember(selectedNode, dataSetDownloadOptions);
```
traeok: `downloadSingleMember` is called for every node, and inside this function it calls `DatasetUtils.getExtensionMap` - this calls `node.getChildren()` on the parent PDS, which may fetch all children of that PDS.

Is there a way to potentially optimize this flow to avoid calling `getChildren` or `getExtensionMap` for each node? For example, if you have 20 members selected under a single PDS, it would call `getChildren` 20 times on the same PDS.

packages/zowe-explorer/src/trees/dataset/DatasetActions.ts
```diff
@@ -958,45 +998,122 @@ export class DatasetActions {
...
+                        hasErrors = true;
+                        const nodeLabel = String(selectedNode.getLabel() ?? "");
+                        const profileName = String(profile?.name ?? "");
+                        const profileSuffix = profileName ? ` (${profileName})` : "";
+                        const message = e instanceof Error ? e.message : String(e);
+                        commandResponses.push(`Failed to download ${nodeLabel}${profileSuffix}: ${message}`);
+                    }
+
+                    if (progress) {
+                        Gui.reportProgress(progress, totalCount, processedCount++, "");
```
traeok, line 1126: Q: is the fourth parameter meant to be an empty string? Usually an action label is specified for that parameter (e.g. "Downloading members" or "Fetching members")

packages/zowe-explorer/src/trees/dataset/DatasetActions.ts
JWaters02: I wanted it to keep going if there is an error from one from profile so then members from another profile could keep going without being effected, but then as you say if there is an auth error or something that should be blocked from trying again, then it shouldn't do that. Also this case has such a minute chance of actually happening (since basically no one is going to use it across multiple profiles anyway) that I don't think its necessary anymore

packages/zowe-explorer/src/trees/dataset/DatasetActions.ts
JWaters02: That's a very good point and when I first read it I thought this was going to be irrelevant with getExtensionMap basically being removed in favor of PDS name but now looking at the code again I see that the getChildren is still there as to actually populate the extension map. ~~I think I should just have children be prop drilled down from downloadMember or downloadMembers into downloadSingleMember which passes it into getExtensionMam~~ actually I should have extension map constructed in the top functions based on their requirements and then pass them into download single member

packages/zowe-explorer/src/trees/dataset/DatasetActions.ts
JWaters02: Because if its empty then it just keeps using the same title as when first starting up the progress bar and I didnt want the title to change when downloading - it downloads quickly enough for whatever text is there (like ds/mem name) is not noticable and there is already number/total there so dont need another running count

packages/zowe-explorer/src/trees/dataset/DatasetActions.ts
JWaters02: ended up getting rid of node.getChildren because it wasn't necessary anymore after the other PR to switch back to getting extension from only PDS name. that lead to a lot of refactorings around extensionMap. ultimately kept calling getExtension for each member in downloadMembers as it is basically irrelevant on speed, and the alternative (building an extension map "cache" for tracking profile+pds parent node pairs for each member and checking if that is the same on a new member, but all of the above only when not override extension) is a lot of extra code complexity

anaxceron (approved):

packages/zowe-explorer/src/trees/dataset/DatasetActions.ts
traeok: Thanks for refactoring! I think removing the call to `node.getChildren` should be enough to minimize impact.

packages/zowe-explorer/src/trees/dataset/DatasetActions.ts
traeok: Makes sense - I think that as long as we do not continue requests for a profile after encountering an authentication error, then that should suffice.

packages/zowe-explorer/src/trees/dataset/DatasetActions.ts
traeok: Sounds good to me. I was initially curious about the empty string being passed as it seemed redundant, but then realized that `Gui.reportProgress` has `action` as a required parameter 😅 Separate effort, but it seems like we could have a default in that function and make `action` optional.

traeok (approved):
Tested and LGTM, thanks Joshua for the enhancement

packages/zowe-explorer/__tests__/__unit__/trees/dataset/DatasetUtils.unit.test.ts
```diff
@@ -130,122 +130,3 @@ describe("Dataset utils unit tests - function extractDataSetAndMember", () => {
         expect(result.memberName).toBe("");
     });
 });
-
-describe("Dataset utils unit tests - function getExtensionMap", () => {
```
t1m0thyj, line 134: No objections with moving the `getExtensionMap` method, but I think these tests were valuable to avoid extension-related regressions and should be kept around even if the method is private now.

packages/zowe-explorer/src/trees/dataset/DatasetActions.ts
```diff
@@ -958,45 +1005,142 @@ export class DatasetActions {
...
+                        const profileSuffix = profileName ? ` (${profileName})` : "";
+                        const message = e instanceof Error ? e.message : String(e);
+                        commandResponses.push(`Failed to download ${nodeLabel}${profileSuffix}: ${message}`);
+                        ZoweLogger.error(`Error downloading member ${nodeLabel}${profileSuffix}: ${message}`);
+                        await AuthUtils.errorHandling(e as Error, {
+                            apiType: ZoweExplorerApiType.Mvs,
+                            profile,
+                            scenario: vscode.l10n.t("Unable to download data set member."),
+                        });
+                        break;
```
t1m0thyj, line 1122: Thinking about the usage of `break` here: If a user selects 20 members to download and the 2nd one fails due to a temporary network hiccup or a specific member-level issue (e.g. member migrated), the entire batch download is aborted, leaving the remaining 18 members unattempted.

I guess aborting the batch download is probably good to keep as the safest option, in case the error is due to expired credentials or a permanent network disconnection. Or perhaps a middle ground would be to abort downloads on a per-profile basis and proceed with the rest? Meaning if I download 10 members with Profile A and 10 with Profile B, and first request fails, then abort only the first 10 downloads. But this is probably an edge case - I don't anticipate folks will select members across multiple systems very often, although it's cool we support it when they do.

packages/zowe-explorer/src/trees/dataset/DatasetActions.ts
```diff
@@ -941,8 +955,41 @@ export class DatasetActions {
...
+        const parent = selectedNode.getParent() as IZoweDatasetTreeNode;
+        const datasetName = parent.getLabel() as string;
+        const memberName = selectedNode.getLabel() as string;
+        const fullDatasetName = `${datasetName}(${memberName})`;
+        const fileName = uppercaseNames ? memberName : memberName.toLowerCase();
+
+        const targetDirectory = generateDirectory
+            ? DatasetActions.generateDirectoryPath(datasetName, selectedPath, generateDirectory, uppercaseNames)
+            : selectedPath.fsPath;
+        const filePath = path.join(targetDirectory, `${fileName}.${extension}`);
```
t1m0thyj, line 977: The way that file paths are built, if I have the tree structure shown below and select both members to download, I believe they will overwrite each other. This is an edge case that we probably don't need to handle because in most cases it'd be undesirable to add the profile name somewhere in the path structure.
```
> lpar1
  |- IBMUSER.TEST.PDS
    |- TESTMEM
> lpar2
  |- IBMUSER.TEST.PDS
    |- TESTMEM
```

t1m0thyj (commented):
Thanks for your work on this @JWaters02! Functionality looks good, left a few comments

packages/zowe-explorer/src/trees/dataset/DatasetActions.ts
JWaters02: I had similar thoughts (see https://github.com/zowe/zowe-explorer-vscode/pull/4299#discussion_r3358167627) and don't think its worth making the logic more complex for that unless someone complains about it

packages/zowe-explorer/__tests__/__unit__/trees/dataset/DatasetUtils.unit.test.ts
JWaters02: The thing is these tests are mostly related to the original functionality which was determining extension on member name instead of PDS name, and the PR that changed it to just relying on the PDS name just flipped some tests around to match what the new version did. I don't think most of these tests are relevant anymore because the replacing helpers are already being tested through the test cases of the getExtension function and the tests in the downloading functions above (such as overwrite extension and uppercase names etc). I could re-add some unit tests for the new helpers specifically but I don't think it would be testing anything new.

packages/zowe-explorer/src/trees/dataset/DatasetActions.ts
JWaters02: Hmm you are right, that's something that didn't even come across my mind when first implementing the download functionality. But as you say I don't think its worth it for the tradeoffs for such an unlikely use case.

pujal0909 (approved):
Thanks @JWaters02 
The functionality works as expected. Noticed there's conflicts that need to be resolved.

One thing I did notice was that download doesn't seem to respect the profile encoding. 
For example if my profile encoding is set to 1146, it downloads it with the default 1047 encoding (£ is downloaded as $). 
I think this can be addressed in a separate PR since it is out of the scope of this PR. Created this issue for it: https://github.com/zowe/zowe-explorer-vscode/issues/4362

zFernand0 (approved):
LGTM! 😋 
---
Getting some files changed by prettier and prepublish 😅 

Image


Happy to address in a follow up PR 🙏

PR #4334, opened 2026-06-14
fix(datasets): improve LRECL error message clarity

### LINKED ISSUE #4076: Rework wording of error message when saving ds that is over LRECL
**Describe the bug**

I've had a wave of reports from developers finding that the new "This upload operation may result in data loss" error message when trying to save a member that is over the data sets LRECL is not clear enough of what the actual problem is. 

Image

I'm aware this is mostly the issue that they are not reading the error message or clicking "Troubleshoot", but as one developer explained to me, they see the "This upload operation may result in data loss" error, think it's some entirely unrelated problem (such as their connection to the server being broken) and then get fed up that they can't save their program and go back to editing it on the greenscreen.

Also, when I showed them they can click on the "Troubleshoot" button, they thought there was nothing there, as it is not clear in the webview that you can click on the "Full error summary" dropdown (I mean I thought it looks clear enough but I am clearly biased). 

**To Reproduce**

Save file that has a line over LRECL of data set.

**Expected behavior**

I agree that the error message is quite vague though, and since it's hardcoded only in that one LRECL check on the save, we should change it to something clearer, such as "Line(s) in the file exceed the logical record length of this data set. Please review the following lines: {line nums}. Click "Troubleshoot", then "Full error summary" for more detailed information."

**Screenshots**

- OS: Win11
- Zowe Explorer Version: 3.4.1

**Additional context**

On the bright side, this has at least saved several of them from complete data loss as we were saying before, so very glad this feature exists now, thanks fernando!! But does also go to show how common it is that people are going over LRECL length in file, it really surprises me actually.

BRGOVIND:
Fixes #4076

## Problem
The error shown when saving a dataset with lines exceeding the LRECL was:
> "This upload operation may result in data loss."

This is too vague — users interpreted it as a connection problem and abandoned the save rather than investigating the actual cause.

## Solution
Replaced with a precise message that names the real constraint:
> "Line(s) in this file exceed the logical record length (LRECL) of this data set."

The rest of the error (line numbers, Troubleshoot button) remains unchanged.

## Changes
- `src/trees/dataset/DatasetFSProvider.ts` — updated error message string
- `__tests__/.../DatasetFSProvider.unit.test.ts` — updated assertion to match new message
- `l10n/bundle.l10n.json` — updated l10n key
- `l10n/poeditor.json` — updated l10n key

## Testing
All 114 existing unit tests pass with no changes to test logic beyond updating the expected string.

### DIFF
--- a/packages/zowe-explorer/CHANGELOG.md
+++ b/packages/zowe-explorer/CHANGELOG.md
@@ -10,6 +10,7 @@ All notable changes to the "vscode-extension-for-zowe" extension will be documen
 
 ### Bug fixes
 
+- Improved the error message shown when saving a data set member fails due to an LRECL mismatch, now showing the source and target LRECL values and suggesting the user change the target LRECL setting. [#4076](https://github.com/zowe/zowe-explorer-vscode/issues/4076)
 - Fixed an issue that caused a rename to fail when removing the rightmost qualifier from a data set name. [#4273](https://github.com/zowe/zowe-explorer-vscode/issues/4273)
 - Fixed an issue where copying a USS file across profiles when a profile encoding was configured would corrupt characters in the destination file because the encoding was not passed to the upload call. [#4262](https://github.com/zowe/zowe-explorer-vscode/pull/4262)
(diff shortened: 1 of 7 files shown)


zFernand0 (changes requested):

PR #4335, opened 2026-06-14
fix(tables): rename context menu option from 'Display in Tree' to 'Locate in Tree'

### LINKED ISSUE #3771: Rename "Display in Tree" table option to "Locate in Tree" or "Find in Tree"

Came up during a UX review of the new data sets table view - we should consider renaming the option so its more clear to folks. Since this also exists in the jobs table view, I've separated this request into its own enhancement.

Rename the option "Display in Tree" to "Locate in Tree" or "Find in Tree" (we can decide which is better).

Keep it as-is, its not that different from using the locate/find phrasing

BRGOVIND:
Closes #3771

## What changed

The context menu option in both the Data Sets and Jobs table views has been renamed from **"Display in Tree"** to **"Locate in Tree"**.

## Why

"Display in Tree" reads like a toggle — as if clicking it will show or hide something. The actual behaviour is navigation: it jumps to the item and focuses it in the tree view. "Locate in Tree" describes that action accurately, which is the same conclusion the UX review in the issue reached.

## Files updated

| File | Change |
|---|---|
| `src/trees/dataset/DatasetTableView.ts` | Label updated |
| `src/trees/job/JobTableView.ts` | Label updated |
| `l10n/bundle.l10n.json` | l10n key renamed |
| `l10n/poeditor.json` | l10n key renamed |
| `__tests__/__unit__/.../DatasetTableView.unit.test.ts` | Assertion updated |
| `__tests__/__e2e__/features/.../DatasetTableView.feature` | Step text updated (×2) |
| `__tests__/__e2e__/step_definitions/.../DatasetTableView.steps.ts` | Step definition updated (×2) |

## Testing

- 157/157 unit tests pass
- Zero remaining occurrences of the old string confirmed via grep

### DIFF
--- a/packages/zowe-explorer/CHANGELOG.md
+++ b/packages/zowe-explorer/CHANGELOG.md
@@ -10,6 +10,7 @@ All notable changes to the "vscode-extension-for-zowe" extension will be documen
 
 ### Bug fixes
 
+- Renamed the context menu option "Display in Tree" to "Locate in Tree" in the Jobs table view for better clarity. [#3771](https://github.com/zowe/zowe-explorer-vscode/issues/3771)
 - Fixed an issue that caused a rename to fail when removing the rightmost qualifier from a data set name. [#4273](https://github.com/zowe/zowe-explorer-vscode/issues/4273)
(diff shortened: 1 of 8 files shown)


zFernand0 (changes requested):

PR #4336, opened 2026-06-14
fix(dataset): normalize user ID filter input to uppercase

### LINKED ISSUE #3723: Filtering PDS by user ID should not be case-sensitive

User IDs cannot contain lowercase letters. When filtering members by user ID, users might enter a lowercase value for the user ID filter assuming that the filtering will not be case sensitive.

We should convert the user ID filter to uppercase since there cannot be lowercase characters in a user ID.

The workaround is to enter the filter in uppercase.

**Additional context**

BRGOVIND:
Closes #3723

## What's wrong

When filtering PDS members by user ID, the filter only matched if the user typed the ID in the exact same case as it was stored. On z/OS, user IDs are always uppercase — but there is nothing stopping someone from typing `jsmith` instead of `JSMITH` in the input box, and when they do, the filter returns zero results even though the member is right there.

## What changed

In `filterPdsMembersDialog()` inside `DatasetTree.ts`, the filter value is now normalized to uppercase before it is stored, but only for the User ID filter method. Every other filter type (Name, Date Modified, Date Created) is passed through unchanged.

```ts
// z/OS user IDs are always uppercase; normalize the input so filters match
// regardless of what the user typed
const normalizedFilter =
    filterMethod === Sorting.DatasetFilterOpts.UserId ? filter.toUpperCase() : filter;
```

The unit test mock data was also corrected to use uppercase user IDs, matching what a real z/OS mainframe returns.

## Testing

- 276/276 unit tests pass
- Typed a lowercase user ID in the filter dialog → PDS members whose owner matches (case-insensitively) are shown correctly
- All other filter types (Name, date filters) are unaffected

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/packages/zowe-explorer/CHANGELOG.md
+++ b/packages/zowe-explorer/CHANGELOG.md
@@ -10,6 +10,7 @@ All notable changes to the "vscode-extension-for-zowe" extension will be documen
 
 ### Bug fixes
 
+- Fixed an issue where filtering PDS members by user ID was case-sensitive, causing lowercase input (e.g. `jsmith`) to never match z/OS entries stored in uppercase. The filter value is now normalized to uppercase before comparison. [#3723](https://github.com/zowe/zowe-explorer-vscode/issues/3723)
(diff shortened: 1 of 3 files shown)


zFernand0 (changes requested):

PR #4330, opened 2026-06-11
fix: use POSIX escaping for `cwd` as `executeSshCwd` parameter

traeok:
## Proposed changes

Fixed an issue where executing Unix commands could fail if the current working directory path contained certain special characters.

## Release Notes

**Milestone:** 3.5.1/3.5.2
**Changelog:** Fixed an issue where executing Unix commands could fail if the current working directory path contained certain special characters.

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] Enhancement (non-breaking change which adds or improves functionality)
- [ ] Breaking change (a change that would cause existing functionality to not work as expected)
- [ ] Documentation (Markdown, README updates)
- [ ] Other (please specify above in "Proposed changes" section)

## Checklist

**All checks must pass before this pull request is reviewed by the Zowe Explorer squad.**

**General**

- [x] I have read the [CONTRIBUTOR GUIDANCE](https://github.com/zowe/zowe-explorer-vscode/wiki/Best-Practices:-Contributor-Guidance) wiki
- [ ] All PR dependencies have been merged and published (if applicable)
- [ ] A GIF or screenshot is included in the PR for visual changes
- [ ] The pre-publish command has been executed:
  - **v2 and below:** `yarn workspace vscode-extension-for-zowe vscode:prepublish`
  - **v3:** `pnpm --filter vscode-extension-for-zowe vscode:prepublish`
- [ ] New ZE APIs are tested with extender types that haven't adopted yet to determine breaking changes. Can use Zowe zFTP marketplace extension.

**Code coverage**

- [x] There is coverage for the code that I have added
- [x] I have added new test cases and they are passing
- [x] I have manually tested the changes

PR #4338, opened 2026-06-16
fix: pass cwd as-is to SDK (already escaped)

traeok:
## Proposed changes

Fixes issue surfaced by #4330 - the SDK already escapes the cwd, we don't need any quotes around it - we can pass as-is. 
Also addresses an issue surfaced when exec mode was added to the SDKs. When that work was added, escaping was implemented, but ZE was never updated. This caused an issue where any CWD is interpreted literally with double-quotes wrapped around it, which does not match the CWD the user expects based on their input. Now, we pass CWD as-is to the SDK and the SDK handles the escaping.

## Release Notes

Milestone:

Changelog:

## Types of changes

## Checklist

**General**

**Code coverage**


t1m0thyj (approved):

zFernand0 (approved):
LGTM! 😋 
---
let's rely on the upstream sdk to quote it once 🙏

PR #4382, opened 2026-07-08
Implement visual enhancement and elements to config-webview

ArghadeepDas-ibm:
## Proposed changes

Feature 1 — Open zowe.config.json button (always visible, beside Profiles heading)
A codicon-go-to-file icon button sits permanently next to the "Profiles" in Panels.tsx — visible whether the panel is collapsed or not. Clicking it calls onOpenRawFile(configPath) which opens the raw zowe.config.json in VS Code's text editor. The same concept was also wired into the "Add Profile to Tree" quick pick in Profiles.ts as a new "Open zowe.config.json" option.
Screenshot 2026-07-08 at 2 17 28 PM

Feature 2 — InfoIcon contextual help tooltips on profile fields
A new InfoIcon.tsx component renders a codicon-info button beside each profile property label in renderConfig.tsx. Clicking shows a smart floating tooltip (auto-flips at viewport edges) with a description, example value, notes, and a "Required field" badge.
Screenshot 2026-07-08 at 2 19 00 PM

Feature 3— TutorialOverlay for first-time users
A 6-step guided modal (TutorialOverlay.tsx) appears the first time a config file is opened. State is persisted to VS Code's globalState (key zowe.configEditor.tutorialSeen) via a new CONFIG_EDITOR_TUTORIAL_SEEN storage key — so it survives webview teardowns and is tracked per config file path.
Screenshot 2026-07-08 at 2 20 18 PM

Feature 4— Added extra entrypoints to open teh config view as well as the config.json file for experienced user.
Screenshot 2026-07-08 at 2 21 59 PM

## Release Notes

Milestone:

Changelog:

## Types of changes

## Checklist

**General**

**Code coverage**

**Deployment**


packages/zowe-explorer/src/configuration/Profiles.ts
```diff
@@ -555,6 +557,16 @@ export class Profiles extends ProfilesCache {
...
+        if (choice === configOpen) {
+            const allProfileAttrs = mProfileInfo.getAllProfiles();
+            if (allProfileAttrs.length > 0) {
+                const firstProfile = await this.getProfileFromConfig(allProfileAttrs[0].profName);
+                if (firstProfile?.profLoc?.osLoc?.[0]) {
+                    await this.openConfigFile(firstProfile.profLoc.osLoc[0]);
+                }
+            }
+            return;
+        }
```
traeok, line 563: If the user doesn't have any profiles, this option silently fails. Can we still open the config file even when no profiles are present?

packages/zowe-explorer/src/webviews/src/config-editor/components/TutorialOverlay.tsx
```diff
@@ -0,0 +1,126 @@
...
+
+import { useState } from "react";
+import { markTutorialCompleted, markTutorialSkipped } from "../utils/FirstTimeUserDetector";
+
+interface TutorialStep {
+  title: string;
+  description: string;
+}
+
+const STEPS: TutorialStep[] = [
```
traeok: These steps should be localized to support languages other than English - but this localization suggestion applies for the tutorial as a whole

packages/zowe-explorer/src/webviews/src/config-editor/help/FieldHelpContent.ts
```diff
@@ -0,0 +1,204 @@
...
+    description: string;
+    exampleValue?: string;
+    validationRules?: string;
+    required?: boolean;
+}
+
+type ProfileFieldHelp = Record<string, FieldHelp>;
+type ProfileTypeHelp = Record<string, ProfileFieldHelp>;
+
+export const PROFILE_FIELD_HELP: ProfileTypeHelp = {
```
traeok: Should these descriptions also be localized? Or, is there a way that we can get descriptions for these fields from the Zowe schema, rather than duplicating them here?

packages/zowe-explorer/src/webviews/src/config-editor/utils/FirstTimeUserDetector.ts
```diff
@@ -0,0 +1,57 @@
...
+        return true;
+    }
+}
+
+/**
+ * Marks the tutorial as completed so it won't show again.
+ */
+export function markTutorialCompleted(): void {
+    try {
+        localStorage.setItem(TUTORIAL_STORAGE_KEY, "true");
```
traeok: This is writing to the browser `localStorage`, whereas the show/hide decision in `App.tsx` is driven by the separate, per-config-path store synced through VS Code `globalState`. Should we consolidate the state into one place in VS Code?

traeok (dismissed):
Thanks for working on this Arghadeep, I left a couple requests around localization and centralizing state

packages/zowe-explorer/CHANGELOG.md
```diff
@@ -11,6 +11,7 @@ All notable changes to the "vscode-extension-for-zowe" extension will be documen
 
 ### Bug fixes
 
+- Renamed 'Display in Tree' to 'Locate in Tree' in table views to improve clarity. [#3771](https://github.com/zowe/zowe-explorer-vscode/issues/3771)
```
```suggestion
- Renamed the 'Display in Tree' option in table views to 'Locate in Tree' to improve clarity. [#3771](https://github.com/zowe/zowe-explorer-vscode/issues/3771)
```

Maanvi212006:
@ArghadeepDas-ibm pls ensure all commits have the "Signed-off-by" tag

packages/zowe-explorer/src/webviews/src/config-editor/help/FieldHelpContent.ts
jace-roell: I agree with parsing the information from the schema. Since we are already pulling this data for an on-hover tooltip, the current setup duplicates functionality. I favor removing the hover tooltip and replacing the text in the new component with the schema contents. Note that we will only be able to provide defaults for fields defined in the schema, and the 'required' fields will either need to be omitted or hard-coded.

Image

packages/zowe-explorer/src/webviews/src/config-editor/components/Panels.tsx
```diff
@@ -82,6 +85,15 @@ export function Panels({
                   <span className={`codicon ${profilesCollapsed ? "codicon-chevron-right" : "codicon-chevron-down"}`}></span>
                   <h2 style={{ margin: 0, fontSize: "16px" }}>{l10n.t("Profiles")}</h2>
                 </button>
+                <button
+                  className="ce-icon-button"
+                  title={l10n.t("Open zowe.config.json")}
```
jace-roell: I'm curious about the purpose of the new button, given that this functionality is already available in the tab's right-click context menu. I'm open to discussing whether we should remove one of these access points to keep the UI clean.

Image

packages/zowe-explorer/src/webviews/src/config-editor/components/TutorialOverlay.tsx
jace-roell, line 1: 1. Should we add a button to trigger the tutorial manually? Even in my own usage I had accidentally closed it and could not find a way to open it again.

A good location might be in the tabs tool bar (`Tabs.tsx`)

2. Others may have thoughts on this, but is there a particular reason that the tutorial seen is based on the config path rather than a global state? Would we anticipate people wanting to see after creating/viewing a config in a new location? Open to discussion on it.

packages/zowe-explorer/src/webviews/src/config-editor/components/TutorialOverlay.tsx
```diff
@@ -0,0 +1,126 @@
...
+    description: "The left panel lists all profiles in your config. " + "Click any profile to view and edit its properties on the right.",
+  },
+  {
+    title: "Profile Properties",
+    description:
+      "Fields shown here map directly to your zowe.config.json. " + "Click the ⓘ icon next to a field name for a description and example values.",
+  },
+  {
+    title: "Defaults",
+    description: "The Defaults section lets you choose which profile is the default for each connection type (zosmf, ssh, rse, etc.).",
```
jace-roell: Should we omit `rse` from this since it is not a core type?

packages/zowe-explorer/package.json
jace-roell, line 1: I like this new functionality, only issue I am seeing is that it will not shift focus to the correct config layer as it does with the profile tree context menu option (`Manage Profile > Edit Profile`)

This can be done pretty easily by enhancing `zowe.configEditor` to have an uri option

```typescript
// SharedInit.ts:268
vscode.commands.registerCommand("zowe.configEditor", async (opts?: vscode.Uri) => {
                // When invoked from the editor/explorer context menu, opts is the clicked config file's URI.
                const configPath = opts instanceof vscode.Uri ? opts.fsPath : undefined;

                // Check if there's already an open ConfigEditor
                if (existingConfigEditor && existingConfigEditor.panel) {
                    // Reuse existing ConfigEditor
                    existingConfigEditor.panel.reveal();

                    if (configPath) {
                        existingConfigEditor.initialSelection = { profileName: "", configPath };
                        await existingConfigEditor.panel.webview.postMessage({
                            command: "INITIAL_SELECTION",
                            profileName: "",
                            configPath,
                        });
                    }

                    return existingConfigEditor;
                } else {
                    // Create new ConfigEditor
                    const configEditor = new ConfigEditor(context);

                    if (configPath) {
                        configEditor.initialSelection = { profileName: "", configPath };
                    }

                    // Track this instance
                    existingConfigEditor = configEditor;

                    // Set up disposal tracking
                    configEditor.panel.onDidDispose(() => {
                        existingConfigEditor = null;
                    });

                    const ret = await configEditor.userSubmission.promise;
                    configEditor.panel.dispose();
                    return ret;
                }
            })
```

packages/zowe-explorer/src/webviews/src/config-editor/components/TutorialOverlay.tsx
```diff
@@ -0,0 +1,126 @@
...
+import { markTutorialCompleted, markTutorialSkipped } from "../utils/FirstTimeUserDetector";
+
+interface TutorialStep {
+  title: string;
+  description: string;
+}
+
+const STEPS: TutorialStep[] = [
+  {
+    title: "Welcome to the Config Editor",
```
jace-roell: (nit) Can we make this `Welcome to the Zowe Config Editor`

packages/zowe-explorer/src/webviews/src/config-editor/components/InfoIcon.tsx
```diff
@@ -0,0 +1,116 @@
...
+          e.stopPropagation();
+          setVisible((v) => !v);
+        }}
+        title="Click for help"
+      >
+        <span className="codicon codicon-info"></span>
+      </button>
+
+      {visible && (
+        <div ref={tooltipRef} className="field-help-tooltip" style={style} role="tooltip">
```
jace-roell, line 90: With the tool tip related changes overflowing property keys now result in strange sizing and truncation. Previous behavior was to have ellipses for horizontal overflow. 


Image


Connecting this to my other comment: if we remove the schema info from the hover state (moving it to the new menu), the property key's on-hover behavior should instead show the full property name. For example, `longNameXyz1234567890` would be visually truncated to `longNameXyz1234...` but show the full string on hover.

packages/zowe-explorer/src/configuration/Profiles.ts
jace-roell: is there a reason we are specifically opening the `osLoc` of the first profile found? Having an option to select which file to open based on the existing config layer might be more straightforward. 

Also, If the osLoc of the first profile is a user config `zowe.config.user.json`, then the `zowe.config.json` title would be slightly inaccurate.

jace-roell (changes requested):
Thanks @ArghadeepDas-ibm for the cool enhancements! 🥇 

Left a few suggestions below. Sorry if I was a bit nit-picky on some of the functionalities 😅

packages/zowe-explorer/src/webviews/src/config-editor/components/TutorialOverlay.tsx
```diff
@@ -0,0 +1,130 @@
...
+    },
+    {
+      title: l10n.t("Saving Changes"),
+      description: l10n.t(
+        "After editing, click Save to write your changes to disk. Use Refresh to discard unsaved changes and reload from the file."
+      ),
+    },
+    {
+      title: l10n.t("You're all set!"),
+      description: l10n.t("Start by selecting a profile on the left. Click the \u24D8 icons for contextual help on any field."),
```
jace-roell: (nit) There is some duplication between step 3 and 6 in regards to the tooltip icon.

packages/zowe-explorer/src/webviews/src/config-editor/components/InfoIcon.tsx
```diff
@@ -0,0 +1,108 @@
...
+          <div className="help-section">
+            <p>{description}</p>
```
jace-roell, line 92: (nit) consider adding wrapping to the description text to avoid a horizontal scrollbar

tso type profile, codePage property description text
Image

packages/zowe-explorer/package.json
```diff
@@ -170,7 +170,7 @@
       },
       {
         "command": "zowe.configEditor",
-        "title": "Edit Zowe Config Files",
+        "title": "Edit in zowe config webview",
```
jace-roell: Maybe "Edit in Zowe configuration editor" would be more direct?

jace-roell (changes requested):
Just doing a quick pass since I happened to notice you had pushed a commit. Sorry if I jumped the gun 😋 I like how the enhancements are lookin! Left a few styling and wording suggestions. Thanks!

ArghadeepDas-ibm:
Hi @traeok  & @jace-roell,
All the above reviews have been addressed in the latest changes. Also enhanced the "Edit Zowe Config Files" context menu option in the file editor. When you use it while editing a zowe.config.json file, it now opens the configuration editor directly to the specific property where your cursor is currently positioned, making it quicker and easier to navigate to the relevant setting.

https://github.com/user-attachments/assets/0d364a8f-9413-42c0-893b-abf26a70b438

Would request you review the changes and let me know if any other changes are required.

JillieBeanSim (commented):
Thanks for working on this @ArghadeepDas-ibm This is really cool! I have a few small comments/questions regarding the changes:
1. with the `+` menu options we now have `Edit Team Configuration File` & `Open Config File`. At just a glance they seem kinda the same, with the `Edit...` taking us to the new webview should it be renamed to `Edit Team Configuration Profiles` using Profiles instead of file or we should add something at the end of both that will let users know they are entering a GUI or JSON, which could be as simple as updating like this `Edit Team Configuration File via GUI` and `Edit Team Configuration File via JSON` for the options. Also, the action name with icon was confusing for me, I expected one thing but got the other after the click on option. Actually both options take me to the GUI so I don't see a need for the second if they will do the same.
2. If we do decide to have 2 options one that takes you to the JSON and one that takes you to the GUI should we also add both options in the command pallete?
3. I'm curious why use of an equalizer icon for the change to json format button in the GUI instead of following VS Codes format of icon in settings for changing between GUI and JSON?
4.  You mentioned that right-clicking on a profile name in the JSON to take you to the GUI, I am not directed to the profile in question but the same home page style as seen from `+`.  IMO, having the option to jump directly to a profile in the GUI via JSON would be the only reason to have this right-click option in the editor, did double check that it isn't showing in all files right-click menu so that is good.

anaxceron (approved):

jace-roell (commented):
Thanks Arghadeep for addressing the feedback. Just a few points in additions to @JillieBeanSim's

1. Please ensure that the wdio integration tests are passing (Ubuntu 22)
      - Most of them can be fixed by changing `Setup.steps.ts:98` to match the new command palette name
      - The remaining failures are due to the "Open file" context menu button being moved. I would recommend moving this step declaration and usage to `profileList.feature` 
      
2. Ensure DCO is passing

ArghadeepDas-ibm:
Hi @JillieBeanSim,
Thank you for reviewing the PR and for the suggestions.
Apologies for the delay in addressing the feedback. It took some additional time to investigate and resolve the Ubuntu build failure. The review comments have now been addressed in the latest set of changes:

1. The quick pick item have been updated to the below:
image

For opening the webview, I have kept the label as "Edit Team Configuration File in Zowe Configuration Editor". This wording is also used in the context menu, which helps maintain consistency across the extension.
Using "Edit Team Configuration File via GUI", would introduce unnecessary variation for the same action. Since the new webview is called the Zowe Configuration Editor, I felt it would be clearer and more intuitive to use the same terminology throughout the UI.

2. A second Command Palette option has been added to allow users to open the zowe.config.json file directly, alongside the existing option to open the Zowe Configuration Editor.
image

3. Since users already have an option to open the zowe.config.json file, the additional equalizer action felt redundant and could potentially create confusion. Removing it helps simplify the UI and reduces clutter.
image

4. The issue with the right-click option on the configuration file has been fixed. Users can now navigate to the Zowe Configuration Editor correctly, and both the profile and property selections are working as expected.

Please let me know if there are any additional changes or refinements you would like me to make.

ArghadeepDas-ibm:
Hi @jace-roell ,
Thank you for reviewing the PR and for the suggestions.
Apologies for the delay in addressing the feedback. It took some additional time to investigate and resolve the Ubuntu build failure. The review comments have now been addressed in the latest set of changes. The Setup.steps.ts now includes the new command pallet names and profileList.feature is updated with the "Open File" functionality.

Please let me know if there are any additional changes or refinements you would like me to make.

jace-roell:
@ArghadeepDas-ibm 

Thanks for addressing the reviews, just a few more small details I noticed

**Blocking approval**

1. Seems like the profile list sorting options has regressed and no longer includes by "type" and "default". I believe this is by no longer using the `getProfileSortOrderDisplayName()` function in `generalUtils.ts` I believe all of the sorting logic is intact but the visual element has been removed.

2. The right-click context menu option in zowe.config.json files to navigate to the profile/property in the config editor seems to work now. Only nit-pick I have is that the flashing indicator is done on the entire profile element rather than just the specific property.

3. Co-pilot reviews

**Non-blocking**

1. If we are removing the "Navigate to config file" button from the top of the profile list. I would recommend adding it back to the context menu of the tab. My reason for this is this allows the config files to be opened without shifting focus to that configs' tab. Also, if a config file does not have any profiles, there will be no way to easily navigate to it.

2. `SharedInit.ts:1038` has a dead else block that can be removed

ArghadeepDas-ibm:
Hi @jace-roell , 
Thank you for the suggestion. I have addressed your review comments and brought back both the Open File option and the sorting option.
I also fixed the "Navigate to Config File" behavior for profile properties. Previously, the configuration file would flash  both profile and property when we hover over a property and execute the operation. With this change, the flashing behavior is now limited to the profile only, which provides a more consistent user experience.

Could you please review the latest changes and let me know if there is anything else that needs to be updated or refined? Your feedback would be greatly appreciated.

jace-roell (approved):
LGTM! Thanks for the enhancements on this @ArghadeepDas-ibm!

JillieBeanSim (approved):
LGTM! thanks @ArghadeepDas-ibm for this new feature

PR #4394, opened 2026-07-14
fix (v2): ran the mDetails through censor

pujal0909:
## Release Notes

Milestone:

Changelog:

## Types of changes

## Checklist

**General**

**Code coverage**

**Deployment**

### DIFF
--- a/packages/zowe-explorer/__mocks__/@zowe/imperative.ts
+++ b/packages/zowe-explorer/__mocks__/@zowe/imperative.ts
@@ -404,7 +404,6 @@ export namespace SessConstants {
 }
 
 export class LoggerUtils {
-    static readonly CENSOR_RESPONSE = "****";
     static censorRawData(data: string, _category?: string): string {
         return data;
     }
--- a/packages/zowe-explorer/__mocks__/@zowe/imperative.ts
+++ b/packages/zowe-explorer/__mocks__/@zowe/imperative.ts
@@ -402,3 +402,10 @@ export class TextUtils {
 export namespace SessConstants {
     export declare const AUTH_TYPE_TOKEN = "token";
 }
+
+export class LoggerUtils {
+    static readonly CENSOR_RESPONSE = "****";
+    static censorRawData(data: string, _category?: string): string {
+        return data;
+    }
+}
--- a/packages/zowe-explorer/__mocks__/@zowe/imperative.ts
+++ b/packages/zowe-explorer/__mocks__/@zowe/imperative.ts
@@ -412,7 +412,8 @@ export namespace SessConstants {
 }
 
 export class LoggerUtils {
-    static censorRawData(data: string, _category?: string): string {
-        return jest.requireActual("@zowe/imperative").LoggerUtils.censorRawData(data, _category);
+    static readonly CENSOR_RESPONSE = "****";
+    static censorRawData(_data: string, _category?: string): string {
+        return LoggerUtils.CENSOR_RESPONSE;
     }
 }
--- a/packages/zowe-explorer/__mocks__/@zowe/imperative.ts
+++ b/packages/zowe-explorer/__mocks__/@zowe/imperative.ts
@@ -413,6 +413,6 @@ export namespace SessConstants {
 
 export class LoggerUtils {
     static censorRawData(data: string, _category?: string): string {
-        return data;
+        return jest.requireActual("@zowe/imperative").LoggerUtils.censorRawData(data, _category);
     }
 }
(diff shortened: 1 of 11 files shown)


packages/zowe-explorer/__mocks__/@zowe/imperative.ts
```diff
@@ -402,3 +402,10 @@ export class TextUtils {
 export namespace SessConstants {
     export declare const AUTH_TYPE_TOKEN = "token";
 }
+
+export class LoggerUtils {
+    static readonly CENSOR_RESPONSE = "****";
+    static censorRawData(data: string, _category?: string): string {
```
traeok: More of a question for my own understanding: Is this mocked function required for the tests to pass? Or can we invoke the original?

I just noticed that the function is returning the string data in the same way its passed in, which caught me off guard 😅  maybe we can just return CENSOR_RESPONSE as the string? because then it reflects some mocked-out, censored value rather than just a no-op.

traeok (dismissed):
Changes look good, just had a question about the tests.
Thanks Pujal!

t1m0thyj (dismissed):
LGTM, thanks @pujal0909!

jace-roell (dismissed):

packages/zowe-explorer/__mocks__/@zowe/imperative.ts
zFernand0: My suggestion is to keep it as-is

---

After some local testing...

In theory we could use the actual function, but since we are not redacting any data as part of the `moreInfo` in the ProfilesUtils unit tests, the results will be the same (all passing tests) 😋 

```
return jest.requireActual("@zowe/imperative").LoggerUtils.censorRawData(data, _category);
```

---

If we do want to see some redacted values in the test, we might as well add a new unit test altogether as there are some nuances around what gets censored 😅 

Image

packages/zowe-explorer/__tests__/__unit__/utils/ProfilesUtils.unit.test.ts
```diff
@@ -89,7 +89,7 @@ describe("ProfilesUtils unit tests", () => {
             await profUtils.errorHandling(errorDetails, label, moreInfo);
             expect(Gui.errorMessage).toBeCalledWith(moreInfo + ` Error: ${errorDetails.message}`);
```
zFernand0: nit & OOS (out of scope)
Could we change instances of `toBeCalledWith` with `toHaveBeenCalledWith`?

zFernand0 (approved):
LGTM! 😋

packages/zowe-explorer/__tests__/__unit__/utils/ProfilesUtils.unit.test.ts
zFernand0, line 1: Please run `yarn pretty`

Image

zFernand0 (changes requested):

jace-roell (dismissed):

ATorrise (dismissed):

packages/zowe-explorer/__mocks__/@zowe/imperative.ts
traeok: I guess what I find odd is that if we don't need to mock what the function returns, why stub it at all? 😅 
Maybe there's some logic that we're trying to avoid, or to save performance. I agree with the theory of using the actual function, I think that was where I was heading with the request but just wasn't being explicit.

That said, I'm not fixated on this request, it was more of a suggestion. I'm fine with the outcome 😋

zFernand0 (dismissed):

zFernand0 (approved):

packages/zowe-explorer/__mocks__/@zowe/imperative.ts
```diff
@@ -410,3 +410,9 @@ export class TextUtils {
 export namespace SessConstants {
     export declare const AUTH_TYPE_TOKEN = "token";
 }
+
+export class LoggerUtils {
+    static censorRawData(data: string, _category?: string): string {
+        return jest.requireActual("@zowe/imperative").LoggerUtils.censorRawData(data, _category);
```
zFernand0: I'm still inclined to revert it and just `return data` 😋 

---

However, feel free to revert this change. or go directly down to the LoggerUtils.


```
    static censorRawData(data: string, category?: string): string {
        // Deep-require ONLY the real LoggerUtils to avoid booting the full imperative
        // framework (yargs/y18n reads locale files via fs, which is mocked in unit tests).
        return jest.requireActual("@zowe/imperative/lib/logger/src/LoggerUtils").LoggerUtils.censorRawData(data, category);
    }
```

zFernand0 (changes requested):
failing unit tests after using the real function 😋

packages/zowe-explorer/__mocks__/@zowe/imperative.ts
traeok: I'm fine with reverting it, but we should at least return a stubbed censored value, not just `data`. Otherwise, this will affect callers of this function in unrelated tests, and by returning the `data` as-is, the stub is not fulfilling the responsibility of the function.

traeok (approved):
Sorry for the trouble around the `censorRawData` mock 😓 
Looks like we have no choice but to stub it out due to side effects around importing from Imperative.

Changes LGTM 😋

zFernand0 (approved):
LGTM! 😋 
---
Thanks for the offline discussion 🙏

PR #4426, opened 2026-07-23
fix creating files in ZE virtual workspace

### LINKED ISSUE #4423: adding new file to uss directory in workspace fails

jace-roell:
## Proposed changes

When creating a new file or writing to a file from a ZE virtual workspace. It would originally error out

**How to test**
1. Add a USS directory or a PDS to the workspace via the ZE tree
2. After loading the workspace, create a file via the context menu
3. After this change, this functionality should work aswell as the ability to write content to files from this view.

## Release Notes

Milestone:

Changelog:

## Types of changes

## Checklist

**General**

**Code coverage**

**Deployment**


packages/zowe-explorer/CHANGELOG.md
```diff
@@ -13,6 +13,7 @@ All notable changes to the "vscode-extension-for-zowe" extension will be documen
 
 ### Bug fixes
 
+- Fix an issue where creating or writing to a file from a Zowe Explorer virtual workspace would fail. [#4426](https://github.com/zowe/zowe-explorer-vscode/pull/4426)
```
```suggestion
- Fixed an issue where creating or writing to a file from a Zowe Explorer virtual workspace would fail. [#4426](https://github.com/zowe/zowe-explorer-vscode/pull/4426)
```

CBforZ:
Conversation from Standup this morning: 
Fernando raised the concern that enabling this may truncate files that already exist on the remote system  if the state of the workspace was not fully up to date (e.g. someone else created the same file and added content before you created the empty file) 

We wondered if  that is risky enough to pursue adding the kind of warning we have elsewhere like "The remote version is newer, compare/overwrite? " 
( Not sure myself)

packages/zowe-explorer/src/trees/uss/UssFSProvider.ts
```diff
@@ -719,7 +719,9 @@ export class UssFSProvider extends BaseProvider implements vscode.FileSystemProv
         let resp: IZosFilesResponse;
         try {
             await AuthUtils.retryRequest(entry.metadata.profile, async () => {
-                await this.autoDetectEncoding(entry);
+                if (!options?.isNew) {
```
traeok, line 722: Based on the caller function, it appears that `isNew` is logically equivalent to the entry being falsy.
If this is the only case where `isNew` is passed, do we need to pass it into this function at all? Could we just key off of whether the `entry` is non-null instead?

packages/zowe-explorer/CHANGELOG.md
```diff
@@ -13,6 +13,7 @@ All notable changes to the "vscode-extension-for-zowe" extension will be documen
 
 ### Bug fixes
 
+- Fixed an issue where creating or writing to a file from a Zowe Explorer virtual workspace would fail. [#4426](https://github.com/zowe/zowe-explorer-vscode/pull/4426)
```
traeok: Please feel free to correct me if I'm missing some context, but the issue seems to be specific to files created through virtual workspaces, correct? It's not happening for all files created or written to with ZE.

```suggestion
- Fixed an issue where creating a new file in a Zowe Explorer virtual workspace would silently fail, causing write operations to fail on that file as a result. [#4426](https://github.com/zowe/zowe-explorer-vscode/pull/4426)
```

packages/zowe-explorer/src/trees/dataset/DatasetFSProvider.ts
```diff
@@ -902,11 +902,10 @@ export class DatasetFSProvider extends BaseProvider implements vscode.FileSystem
                 return;
             }
 
-            if (!isNew || content.length > 0) {
-                const resp = await this.uploadEntry(entry as DsEntry, content, uri, forceUpload, encodingParam);
-                entry = parent.entries.get(basename) as FileEntry;
-                entry.etag = resp.apiResponse.etag;
-            }
+            // Always upload for new entries, even with empty content
+            const resp = await this.uploadEntry(entry as DsEntry, content, uri, forceUpload, encodingParam);
```
traeok: I think we saw with z/OSMF API that a sequential data set will error out here, but we should probably confirm that this is the case for most extenders before assuming that the server will always reject the request.

As an alternative, we could *explicitly* block users from creating new sequential data sets using this method + when `options.create` is `true` (as we don't have any attributes to pass at time of creation/write). One way of doing this is by checking if the parent entry is a profile entry - if it is, we know that the user is trying to upload contents to a non-existent sequential data set. This could be perceived as defensive programming, but I'm just trying to be mindful of what write/upload requests we try to make and when its considered safe to make them.

I still think its acceptable to create new PDS members with this method as they do not require any pre-requisites (aside from a valid parent PDS).

traeok (commented):
Thanks for working on this Jace! I agree with Chris' comments (based on Fernando's suggestion in today's stand-up) - we already do something similar for the "Create New Member" by looking up the member before trying to create it. This avoids a similar scenario where: data sets are listed under ZE in the tree, but haven't refreshed recently and a new member is added on the target system. By doing a list members / list files check, we can see if the resource already exists and handle that scenario w/ a dialog (similar to UX for "Create New Member" conflict).

packages/zowe-explorer/src/trees/dataset/DatasetFSProvider.ts
```diff
@@ -902,11 +902,10 @@ export class DatasetFSProvider extends BaseProvider implements vscode.FileSystem
                 return;
             }
```
ATorrise, line 936: Since the isNew check is going away, does the non replace path in createMember upload twice now, given it already calls createDataSetMember? Might be worth a query flag to skip just the empty upload rather than dropping the writeFile call, since that would also drop the cache entry and Created event.

ATorrise (dismissed):

packages/zowe-explorer/src/trees/uss/UssFSProvider.ts
jace-roell: `isNew` is passed from writeFile() and is computed before the entry is created. If we were to recompute it, the entry would always exist. So I think we need the additional context from the caller directly.

anaxceron (approved):

CBforZ:
Testing note:  No issues observed with USS files, but when adding a PDS to workspace, I typed a new member name in lower case. Creation of the new member was succesful, but when i defocused the editor and came back to it, it synced the  all-caps version of the dataset member and there are now two copies of it in the workspace ... I was then able to edit and save to the all-caps version and in the workspace I don't see a way to "pull from mainframe" , so the lower case version of the file remained empty.  It does at least catch the conflict if you  try to save something to the lower case version of the member and prompts you to compare. 

Wondering if there is a way for us to uppercase everything except the extension when a new member is created in a PDS workspace so this duplication won't occur? 

Everything else looked good though 

image

zFernand0 (approved):
LGTM! 😋 
---
Thanks for addressing the feedback and for letting me bother you offline 😅

packages/zowe-explorer/src/trees/dataset/DatasetFSProvider.ts
```diff
@@ -876,7 +876,24 @@ export class DatasetFSProvider extends BaseProvider implements vscode.FileSystem
                 .substring(uriInfo.slashAfterProfilePos + 1)
                 .split("/")
                 .filter(Boolean);
-            const ds = new DsEntry(basename, uriPath.length === this.EXPECTED_MEMBER_LENGTH);
+            const isPdsMember = uriPath.length === this.EXPECTED_MEMBER_LENGTH;
+
+            if (!options.overwrite) {
+                const mvsApi = ZoweExplorerApiRegister.getMvsApi(parent.metadata.profile);
+                const remoteRes = isPdsMember ? await mvsApi.allMembers(uriPath[0]) : await mvsApi.dataSet(FsDatasetsUtils.trimExtension(uriPath[0]));
```
traeok: I would recommend refactoring this area a bit just to make sure that we're handling these two scenarios:
- awaiting the extender profile type before trying to access the API. This avoids a potential race condition where filesystem tries to access an extender's API before it is registered.
- authentication errors (401/etc.). This ensures we are prompting the user to re-authenticate if needed to fulfill the `writeFile` request. Otherwise the error may propagate to the developer as an `Unknown` filesystem error.

traeok (commented):
Changes look good, thanks Jace - I left a suggestion around 2 scenarios that we should consider for `writeFile` 👍 Once that is resolved, I think I'll be ready to approve

traeok (approved):
Tested and changes LGTM, thanks Jace for the fix

CBforZ:
Should we capture the bug I commented above as a separate issue?

traeok:
> Testing note: No issues observed with USS files, but when adding a PDS to workspace, I typed a new member name in lower case. Creation of the new member was succesful, but when i defocused the editor and came back to it, it synced the all-caps version of the dataset member and there are now two copies of it in the workspace ... I was then able to edit and save to the all-caps version and in the workspace I don't see a way to "pull from mainframe" , so the lower case version of the file remained empty. It does at least catch the conflict if you try to save something to the lower case version of the member and prompts you to compare.
> 
> Wondering if there is a way for us to uppercase everything except the extension when a new member is created in a PDS workspace so this duplication won't occur?

This is a good catch, thanks @CBforZ for the details. Maybe we should address it before merging since we're now enabling the New File functionality through the Explorer pane? I could see users running into this issue if we aren't coercing the member name to uppercase, for example.

traeok (approved):
Thanks for addressing the latest feedback around lowercase members 😋  LGTM

CBforZ (approved):
Thanks for the fixes

PR #4433, opened 2026-07-28
fix: reconnect Zowe Remote SSH after a dropped or untracked session

### LINKED ISSUE #4425: [3.6.0 main] ssh connection failed softlocks zowe explorer and then overloads mainframe lpar
**Describe the bug**

if the mainframe connection blips even for a couple of seconds while logging into ssh profile on a fresh session, the ssh profile will fail to ever connect, zosmf profile will also take a long time to connect (it does eventually connect after a lot of time) etc. until reloading vscode, even if the mainframe connection is fine. clicking "retry and reload" button should do something. currently it seems to do nothing.

Then, if zowe explorer is kept open (without reloading), the ssh profile will keep on retrying over and over and sending keep alive pings that are returning failure, which is using a large amount of CPU and capping our dev lpar (it doesn't have much resource on it, so the CPU load is enough to cap and cause slowdowns for everyone else).

**To Reproduce**

Not sure, but sometimes our dev lpar becomes capped by some high CPU process and causes a "blip" where connection to the mainframe becomes slow for a few seconds. but if this happens during the ssh connection (assuming when it deploys its server side onto the mainframe), zowe explorer becomes "softlocked" even after the ssh request has timed out, as connecting to other profiles take a lot of time.

**Expected behavior**

When the ssh profile request times out, clicking "retry and reload" button should do something. currently it seems to do nothing. I also expect that trying to connect to zosmf profile after still works without it taking a long time. And finally, when the ssh profile fails to retry, it should only retry once if the user clicks retry, and there should be UX feedback that it's retrying. and it should not use a massive amount of CPU on the mainframe.

**Screenshots**

- OS: Win11
- Zowe Explorer Version: 3.6.0-SNAPSHOT (0ffcb29e7f926715a108588329303009f846f7ca)

**Additional context**

I'll send my log file on slack

traeok:
## Proposed changes

`SshClientCache.connect()` registers a session only after `checkIfOutdated()` and any redeploy finish. A connection error in that window called `handleClientError()` with no cached session yet, and the non-null assertion there threw inside an `ssh2` event handler. This silently kills the callback before any 'Reload' prompt could appear.

`reloadClient()` had the same assumption from the other side: if the session was torn down (e.g. a racing `onClose`) before the user clicked 'Reload and Retry', it logged a debug line and returned without reconnecting. This is most of what 'clicking retry does nothing' in #4425 was actually reporting.

- `handleClientError()`: guard the missing-session case instead of asserting on it; show a generic error and retry hint rather than throwing.
- `reloadClient()`: take the profile instead of the `clientId`, and reconnect fresh when no session is cached instead of returning early. Reset status back to DOWN if the reconnect attempt itself fails, since leaving it at RESTARTING permanently swallows further error notifications for that profile.
- `connect()`: guard the same non-null assumption on the `retryRequests` path, reachable independent of `reloadClient`.
- `buildClient()`: pass `responseTimeout` on the post-deploy rebuild, matching the initial call.

## How to test

These changes are complimentary to [zowe/zowex#1076](https://github.com/zowe/zowex/pull/1076), which is what makes the SSH connection fail fast and clean up instead of hanging or leaking. **Note that the changes in this branch do *not* require this SDK build before this PR can be merged as the changes in this PR are backwards-compatible.**

Build that branch's `@zowe/zowex-for-zowe-sdk` package and link it into `packages/zowex-for-zowe-explorer` (`pnpm link` or a `file:` dependency) before testing, and get a `zowex` binary built from that branch onto the test LPAR (either build+deploy it directly, or let `zowe.zowex.serverAutoUpdate` redeploy it for you ... see step 3).

To make timeouts reproducible without waiting on real network flakiness, temporarily set `zowe.zowex.responseTimeout` to something small like `5` in VS Code settings. Restore it to `60` (or remove the override) once you're done.

**1. A failed reload no longer silences the profile forever**

This is the regression I found while fixing #4425 - its worth testing on its own since it's easy to miss.

- Connect an SSH profile, then cut off the LPAR's network access (or point the profile at an unreachable host) so a running command times out and the "A request timed out..." prompt appears.
- Click **Reload and Retry** while the host is still unreachable, so the reload itself fails.
- Restore connectivity, then trigger another timeout or error on the same profile (e.g. run another command).
- **Expected:** a new prompt appears for the second error. Before this fix, the session was left in a `RESTARTING` state, and `handleClientError` swallows all errors while a session is `RESTARTING`. Nothing would ever be shown again for that profile until reloading the VS Code window.

**2. Reload and Retry actually reconnects once the network is healthy**

- Trigger the timeout prompt as in step 1, but restore connectivity *before* clicking a button.
- Click **Reload and Retry**.
- **Expected:** the client reconnects and any in-flight request is resent. Confirm the tree view repopulates and further actions on the profile work normally.

~~**3. The pre-registration crash path**~~

~~This is the harder one to stage, since the vulnerable window only opens when the extension has to redeploy the server:~~

Deferred for later / future issue reports 😋 

**4. `responseTimeout` is honored after a redeploy**

- With `zowe.zowex.responseTimeout` set to `5`, force a redeploy as in step 3 but let it complete normally (don't kill the connection).
- Run a command that legitimately takes longer than 5 seconds (or throttle the connection to force it).
- **Expected:** the timeout prompt appears around 5 seconds. Before this fix, the client rebuilt after a deploy silently dropped back to the default.

**5. No leaked processes on the LPAR**

This is the piece that needs zowex#1076 rather than this PR - but, its worth a sanity check that the two changes compose correctly.

- Run `ps -ef | grep zowex` on the test LPAR before starting.
- Repeat steps 1-3 a few times against the same profile, including reload attempts that fail.
- **Expected:** the process count returns to baseline after each cycle. No accumulating `zowex server` processes or workers should be left behind once the client disconnects or the server hits its request-timeout limit.

## Release Notes

Milestone: 3.6.0

Changelog: Fixed an issue where the "Reload" and "Reload and Retry" actions did nothing after a Zowe Remote SSH connection dropped, and where an error raised before the session was cached could prevent the prompt from appearing at all. Reconnecting now reports progress and confirms on success, and repeated failures no longer stack duplicate prompts for the same profile.

## Types of changes

## Checklist

**General**

**Code coverage**

**Deployment**


anaxceron (approved):
🦓

CBforZ:
I can't seem to test the current version due to  needing the changes from #4436 to compile  :( But I'll try to look through the code 
 ```
 $ tsc -p ./ && pnpm madge
src/SshClientCache.ts:172:102 - error TS2339: Property 'serverChecksums' does not exist on type 'ZSshClient'.

172                 imperative.Logger.getAppLogger().debug(`Server checksums: ${JSON.stringify(newClient.serverChecksums)}`);
                                                                                                         ~~~~~~~~~~~~~~~

src/SshClientCache.ts:173:63 - error TS2339: Property 'serverChecksums' does not exist on type 'ZSshClient'.

173                 if (await ZSshUtils.checkIfOutdated(newClient.serverChecksums)) {
                                                                  ~~~~~~~~~~~~~~~
```

**Comment by traeok** [comment]
> I can't seem to test the current version due to needing the changes from #4436 to compile :( But I'll try to look through the code

Thanks for taking a look @CBforZ - I might try to rebase against your branch today so at least a local build (when linking to zowex#1076) can be done 😋

**Comment by traeok** [comment]
@CBforZ its not ideal, but I added testing instructions at the very top that leverages your PR and the 2 commits that encapsulate the changes from this PR. Once #4436 is merged, I can update this branch with `main` and the code should compile w/o issues

**Comment by CBforZ** [comment]
Thanks! I just saw the new zowex SDK version is available so I'm  updating #4436

**Review by anaxceron — APPROVED** [review]

**Comment by JWaters02** [comment]
@traeok Hi, while using the VSIX you sent me, I noticed that every time I connect to the ssh profile after reloading VS Code, it redeploys the server - is it meant to be doing that even though the server binary is the same version? I went to a previous commit on main to check that previously, once it deploys a binary of a version, it wouldn't redeploy the same one again.

**Comment by traeok** [comment]
> @traeok Hi, while using the VSIX you sent me, I noticed that every time I connect to the ssh profile after reloading VS Code, it redeploys the server - is it meant to be doing that even though the server binary is the same version? I went to a previous commit on main to check that previously, once it deploys a binary of a version, it wouldn't redeploy the same one again.

Hey @JWaters02, ~I believe this is the expected behavior for dev builds - a normal release build would only deploy once and then version checks take place for future connections. I will revisit the branch shortly just to make sure there aren't any new deploy issues introduced here, but I believe this is part of recent changes merged with #4436~

Update: There was an issue with the server PAX contained within the VSIX 😓  The link I sent off-platform should now reuse the 0.7.1 server after it has deployed once.

**Comment by CBforZ** [comment]
> I will revisit the branch shortly just to make sure there aren't any new deploy issues introduced here, but I believe this is part of recent changes merged with #4436

Hmm so that was supposed to be a side effect of the fact that we were going to use  the git hash as a semver "prerelease" string, but now that we're throwing away the git hash in the version comparison, it shouldn't happen every time... there may be a bug here in my code, sorry. 

@JWaters02  Can you look in the output panel > Zowe Explorer and look for "checkIfOutdated" - it should be printing the  remote vs local version and showing why it is considered outdated

**Comment by JWaters02** [comment]
> > I will revisit the branch shortly just to make sure there aren't any new deploy issues introduced here, but I believe this is part of recent changes merged with #4436
> 
> Hmm so that was supposed to be a side effect of the fact that we were going to use the git hash as a semver "prerelease" string, but now that we're throwing away the git hash in the version comparison, it shouldn't happen every time... there may be a bug here in my code, sorry.
> 
> @JWaters02 Can you look in the output panel > Zowe Explorer and look for "checkIfOutdated" - it should be printing the remote vs local version and showing why it is considered outdated

```
[2026/08/06 12:28:35.826] [DEBUG] [main.extension.js:57] Server version: 0.6.0+a91a5a7c
[2026/08/06 12:28:35.827] [DEBUG] [main.extension.js:57] [ZSshUtils] checkIfOutdated: Comparing remote version '0.6.0+a91a5a7c' to bundled server version '0.7.0'
[2026/08/06 12:28:35.829] [DEBUG] [main.extension.js:57] [ZSshUtils] remote version '0.6.0+a91a5a7c' is outdated? true
[2026/08/06 12:28:35.830] [INFO] [main.extension.js:57] Server is out of date, deploying to ssh at ~/.zowe-server
```

image

**Comment by CBforZ** [comment]
Thanks Josh, so that looks right so far, does it go on to  update the server? Or does it say skipping the deployment because you don't have write access to the dir?  If it does deploy, then the next time you restart does it still say the remote version is 0.6.0+a91a5a7c ?   You could also check by logging in to a shell on z/OS UNIX and doing ./zowex -v after the deployment 

If the version is still 0.6.0 after updating , then... @traeok  Could it possibly be that the vsix you bundled has a zowex version of 0.6.0+a91a5a7c, like  you had built zowex before package.json was updated on the zowex side?   Although the mentioned commit is rather old so I'm not sure if you would have such an old local version of zowex  . Basically it's possible there is a mismatch between the compiled version of zowex that is bundled and the typescript  constant depending on the steps you took to build 
image

If there is a bug with the updated logic , it doesn't appear to happen every time, as I just tested by building a 0.8.0 version of zowex and then connecting a few times. For me the first time it updates, and then subsequent connections say this 

```
[2026/08/06 07:29:52.882] [DEBUG] [main.extension.js:57] [ZSshUtils] checkIfOutdated: Comparing remote version '0.8.0+5f7c066d' to bundled server version '0.8.0'
[2026/08/06 07:29:52.883] [DEBUG] [main.extension.js:57] [ZSshUtils] remote version '0.8.0+5f7c066d' is outdated? false
```

File: packages/zowex-for-zowe-explorer/src/SshClientCache.ts [hunk 1]
```diff
@@ -258,22 +264,43 @@ export class SshClientCache extends vscode.Disposable {
...
+        try {
+            // Reconnecting can take up to the server startup timeout, so surface progress
+            // rather than leaving the user with no feedback after they click Reload
+            await Gui.withProgress(
+                {
+                    location: vscode.ProgressLocation.Notification,
+                    title: `Reconnecting to Zowe Remote SSH for profile "${profile.name as string}"...`,
+                },
+                async () => {
+                    const updatedProfile = await this.mProfilesCache.getLoadedProfConfig(profile.name!, profile.type);
```
zFernand0, line 283: Nit: Could the profile.name be undefined? 
Image

Technically yes, but at that point we have bigger problems than printing a few `undefined` in error messages.

packages/zowex-for-zowe-explorer/src/SshClientCache.ts
```diff
@@ -300,7 +327,12 @@ export class SshClientCache extends vscode.Disposable {
 
     private handleClientError(clientId: string, err: Error): void {
         const errorMsg = err.toString();
-        const clientSession = this.mClientSessionMap.get(clientId)!; // a session must exist, since we're handling the client's error
+        const clientSession = this.mClientSessionMap.get(clientId);
+        if (!clientSession) {
+            imperative.Logger.getAppLogger().error(`Received SSH client error for untracked session ${clientId}: ${errorMsg}`);
+            vscode.window.showErrorMessage(`Zowe Remote SSH encountered an error: ${errorMsg}. Try the operation again.`);
```
zFernand0, line 333: Q: Will this message display for every untracked session? 

A: Yeah, but that's probably the desired behavior (at least I'd want to know) 

---

Q: Should we have a similar guard like the `mActivePrompts` map in `promptErrorAndReload`?

A: Sure we can, but that can potentially hide valuable info for the end user 😅

traeok:
> If the version is still 0.6.0 after updating , then... @traeok Could it possibly be that the vsix you bundled has a zowex version of 0.6.0+a91a5a7c, like you had built zowex before package.json was updated on the zowex side

This was the case, thanks @CBforZ for testing and sorry @JWaters02 for the mishap - that VSIX link has been updated to reflect the latest changes

packages/zowex-for-zowe-explorer/src/SshClientCache.ts
```diff
@@ -258,22 +264,43 @@ export class SshClientCache extends vscode.Disposable {
...
+                    await this.connect(updatedProfile, {
+                        restart: clientSession != null,
+                        retryRequests: clientSession != null && retryRequests,
+                    });
+                }
+            );
+            Gui.showMessage(`Reconnected to Zowe Remote SSH for profile "${profile.name as string}".`);
+        } catch (err) {
+            if (clientSession) {
+                clientSession.status = ServerStatus.DOWN;
```
zFernand0, line 300: This `clientSession` object will get deleted as part of the reload `this.connect(updatedProfile, {restart: clientSession != null...`.  So, by the time we reset the serverStatus to `down`, the `clientSession` tied to the `clientId` will become untracked.

Q: Will this operation have any effect on subsequent `handleClientError` calls?

A: I get it, but it's still good practice to reset it, and regardless we already have a guard for these untracked sessions inside the `handleClientError` function.

zFernand0 (approved):
LGTM! 😋 
---
I left some comments which reflect some of my struggles with using AI to review a PR. 😋 
 
Conceptually, I can definitely see how these changes address the "nothing happening" part 🙏

JWaters02:
I simulated a network failure by cycling the internal VPN connection and then connecting to ssh profile (while it was still refreshing) and can see that it handles it well now with a clear error message. No retry button anymore, but once I tried to filter again after the connection was back, it worked fine immediately.

However, I did another test where I first deleted the server binary. Then reloaded VS Code. Then filtered on ssh profile, and while it was deploying the binary (about 20% done) I turned off my internet connection (to simulate my home wifi dying momentarily which happens sometimes). The deployment then got "stuck" and ZE loading - was like this for several minutes because I wanted to see if it would time out (it didn't). So, I re-enabled my internet connection while it was still "deploying" but once the connection to the mainframe was re-established through emulator, still ZE was softlocked/stuck loading and the deployment step wasn't moving (nothing happened in logs either). If I filtered on ssh profile in jobs view for example, stuff would happen in logs, but then still get stuck loading as well as the deployment was still stuck. But a reload of VS Code fixed the issue. ~~I'd say since this case only occurs when deployment, it is very low likelyhood and so I think this PR can still go ahead without fixing this specific issue.~~ Actually, I see that point 3 in the description talks about fixing this case...

I also did a test where I deleted the zowex server binary (through the ssh profile in unix tree) and then refreshed the tree and after a few seconds it showed a clear error message that the connection was shutdown, and then the next time I tried to connect to ssh profile it went to redeploy the server.

traeok:
> I simulated a network failure by cycling the internal VPN connection and then connecting to ssh profile (while it was still refreshing) and can see that it handles it well now with a clear error message. No retry button anymore, but once I tried to filter again after the connection was back, it worked fine immediately.
> 
> However, I did another test where I first deleted the server binary. Then reloaded VS Code. Then filtered on ssh profile, and while it was deploying the binary (about 20% done) I turned off my internet connection (to simulate my home wifi dying momentarily which happens sometimes). The deployment then got "stuck" and ZE loading - was like this for several minutes because I wanted to see if it would time out (it didn't). So, I re-enabled my internet connection while it was still "deploying" but once the connection to the mainframe was re-established through emulator, still ZE was softlocked/stuck loading and the deployment step wasn't moving (nothing happened in logs either). If I filtered on ssh profile in jobs view for example, stuff would happen in logs, but then still get stuck loading as well as the deployment was still stuck. But a reload of VS Code fixed the issue. ~I'd say since this case only occurs when deployment, it is very low likelyhood and so I think this PR can still go ahead without fixing this specific issue.~ Actually, I see that point 3 in the description talks about fixing this case...
> 
> I also did a test where I deleted the zowex server binary (through the ssh profile in unix tree) and then refreshed the tree and after a few seconds it showed a clear error message that the connection was shutdown, and then the next time I tried to connect to ssh profile it went to redeploy the server.

@JWaters02 thanks for the detailed feedback! Do you think that point 3 should be addressed to consider this issue fully closed? Or do you believe this PR as-is resolves the locking issue you previously encountered?

Regardless, I will dig into that scenario a bit more to see if there's a better way of approaching it.

jace-roell (approved):
LGTM. Thanks @traeok !

JWaters02:
> Do you think that point 3 should be addressed to consider this issue fully closed? Or do you believe this PR as-is resolves the locking issue you previously encountered?

I think you should merge this PR as is and leave the point 3 part for later or if someone else reports the issue in the wild.

JWaters02 (approved):

PR #4436, opened 2026-07-29
ZRS/ zowex: version comparison now uses semver instead of checksums

CBforZ:
## Proposed changes

* Requires us to be able to install the published 0.7.0 version of the SDK (TODO)
* Rename `Connect to zowex server on host...` command to `Deploy SSH server on host and connect...`
* Throw errors when validating ssh profiles or when issuing the "deploy zowex" VSC command if the user does not have write access to the serverPath directory. In the case of validating profiles, we will only throw the error if the zowex binary is not found & we are not able to deploy, so that we don't throw an error when the user is connecting to a shared zowex directory.
* Update tests
* Testing currently requires  linking zowex SDK  with at least 0.7.0 which includes https://github.com/zowe/zowex/pull/1073

## Release Notes

Milestone:

Changelog:

## Types of changes

## Checklist

**General**

**Code coverage**

**Deployment**


CBforZ:
Marking ready for review so I can get any initial feedback, but CI will be broken until we can install the new zowex SDK version

anaxceron (approved):

packages/zowe-explorer/package.json
```diff
@@ -2306,12 +2306,6 @@
           "description": "%zowe.zowex.keepAliveInterval%",
           "scope": "window"
         },
-        "zowe.zowex.serverAutoUpdate": {
-          "type": "boolean",
-          "default": true,
-          "description": "%zowe.zowex.serverAutoUpdate%",
-          "scope": "window"
-        },
```
zFernand0, line 2319: Should we add an entry in the changelog (or release note) about this going away?
And potentially turning auto-update back on, until they figure out they need to specify the property `autoUpdate: false` in their profile?

packages/zowe-explorer/src/extension.ts
```diff
@@ -43,6 +43,27 @@ export async function activate(context: vscode.ExtensionContext): Promise<ZoweEx
     await ProfilesUtils.handleV1MigrationStatus();
     await Profiles.createInstance(ZoweLogger.imperativeLogger);
 
+    // todo where to add  autoUpdate to schema?
```
zFernand0: I think we mentioned we won't be adding autoUpdate to the schema just yet.
And when we do, we could add it to the core ssh profile (for ease of implementation) 😓

packages/zowex-for-zowe-explorer/__tests__/Utilities.test.ts
```diff
@@ -297,6 +303,21 @@ describe("Utilities", () => {
...
+            const profile = { name: "p", profile: { host: "h" } };
+            const { VscePromptApi } = await vi.importMock("../src/VscePromptApi");
+            VscePromptApi.mockImplementation(() => ({
+                promptForProfile: vi.fn().mockResolvedValue(profile),
+                promptForDeployDirectory: vi.fn().mockResolvedValue(mockDeployDir),
+            }));
+            vi.mocked(ZSshUtils.lacksWriteAccess).mockResolvedValue(true);
+            const api = (await vi.importMock("@zowe/zowe-explorer-api")).ZoweVsCodeExtension.getZoweExplorerApi().getExplorerExtenderApi();
+            await expect((Utilities as any).connectCallback(api)).rejects.toThrow(
+                vscode.l10n.t(SshClientCache.WRITE_ACCESS_TO_SERVER_PATH_ERR, "/mock/server/path")
```
zFernand0: Since this is for uni-testing purposes, should we check for the error message to contain part of the constant string? e.g. "You do not have write access" 

That way, we avoid having to mock the l10n.t function 😋

packages/zowex-for-zowe-explorer/__tests__/SSHClientCache.test.ts
```diff
@@ -263,17 +266,75 @@ describe("SshClientCache", () => {
...
 
-        it("should deploy a new server if the current one is outdated and autoUpdate is true", async () => {
-            vi.mocked(ZSshUtils.checkIfOutdated).mockResolvedValueOnce(true);
+        it("should throw an error if the current one is missing (ENOTFOUND) but the user does not have write permission", async () => {
+            // Force ZSshClient.create to throw ENOTFOUND on the first try
+            vi.mocked(ZSshClient.create).mockRejectedValueOnce(new imperative.ImperativeError({ msg: "Not found", errorCode: "ENOTFOUND" }));
+            vi.mocked(ZSshUtils.lacksWriteAccess).mockResolvedValue(true);
+            cache.detectServerOnPath = vi.fn().mockResolvedValue(undefined);
+            await expect(cache.connect(mockProfile)).rejects.toThrow(
+                vscode.l10n.t(SshClientCache.WRITE_ACCESS_TO_SERVER_PATH_ERR, "/mock/server/path")
```
zFernand0, line 275: Same as previous comment, mostly wondering if we could avoid having to mock the l10n.t function altogether 😋

zFernand0 (commented):
Changes LGTM! 😋 
---
Waiting for the SDK version to be updated before approving 🙏

packages/zowe-explorer/package.json
CBforZ: Yes this is a great point, I will add a note to the changelog.

packages/zowe-explorer/src/extension.ts
CBforZ: Right, sorry i left this comment behind. I'll remove it.

packages/zowex-for-zowe-explorer/__tests__/SSHClientCache.test.ts
CBforZ: I will double check, I think i could use `importOriginal` and then re-use the original implementation. 
One thing I was wondering though was if I don't mock l10n.t, then will folks running the unit tests in different languages get test failures? Or do we force a certain locale in the tests?

packages/zowex-for-zowe-explorer/__tests__/Utilities.test.ts
CBforZ: Discussed offline but improved the test a bit to make sure the message is coming back

packages/zowe-explorer/src/extension.ts
CBforZ: Removed the todo comment

packages/zowe-explorer/package.json
CBforZ: Added in https://github.com/zowe/zowe-explorer-vscode/pull/4436/changes/3a59a82af2a625d474b28664b17cbc0923f4b591

anaxceron (approved):
🐯

packages/zowe-explorer/CHANGELOG.md
```diff
@@ -6,6 +6,10 @@ All notable changes to the "vscode-extension-for-zowe" extension will be documen
 
 ### New features and enhancements
 
+- Breaking: Removed the setting `zowe.zowex.serverAutoUpdate`, which was a global setting used to control whether the backend SSH server `zowex` would be automatically updated when it was outdated. To prevent such updates going forward, specify `"autoUpdate": false` in each desired `ssh` profile. [4436](https://github.com/zowe/zowe-explorer-vscode/pull/4436)
+- Replaced the comparison of checksums of the backend SSH server `zowex` with comparing version numbers when checking if the remote server version is outdated. [4436](https://github.com/zowe/zowe-explorer-vscode/pull/4436)
+- Renamed the `Connect to zowex server on host...` command to `Deploy zowex on server and connect...`. [4436](https://github.com/zowe/zowe-explorer-vscode/pull/4436)
```
traeok: I noticed the name of the command in the changelog does not match the final rename:

```suggestion
- Renamed the `Connect to zowex server on host...` command to `Deploy zowex server on host and connect...`. [4436](https://github.com/zowe/zowe-explorer-vscode/pull/4436)
```

packages/zowex-for-zowe-explorer/src/SshClientCache.ts
```diff
@@ -216,6 +224,10 @@ export class SshClientCache extends vscode.Disposable {
                             requests: replayRequests,
                             useNativeSsh,
                         });
+                    } else if (serverNotFound) {
+                        const errMsg = vscode.l10n.t(SshClientCache.WRITE_ACCESS_TO_SERVER_PATH_ERR, serverPath);
```
traeok: I noticed a `vscode.l10n.t` call here, but I don't believe the string was captured for translation because I don't see it in the nls or l10n files 🤔  Granted, while I think this is an issue, its a lower-priority one and not strictly related to changes in this PR - so it could be resolved separately. 

Context: It seems that the `generateLocalization` script only scans what is in `packages/zowe-explorer/src` and not this source folder, so it will always end up rendering in English until that is addressed.

packages/zowex-for-zowe-explorer/src/SshClientCache.ts
```diff
@@ -68,6 +69,11 @@ export class SshClientCache extends vscode.Disposable {
         super(() => this.dispose());
     }
 
+    public static readonly WRITE_ACCESS_TO_SERVER_PATH_ERR =
+        `You do not have write access to the deployment directory '{0}' ` +
```
traeok: Nitpick: Consider adding a period after `'{0}'` so that the final string does not look like a run-on sentence 😋

packages/zowe-explorer/CHANGELOG.md
traeok, line 1: Nitpick: The link text for the issues are missing the prefix `#`, comparing it to pre-existing entries in this file 😋

traeok (dismissed):
Thanks for working on this Chris! I'm requesting changes around a couple stale documentation pieces, and plan to manually test this afternoon.

I also had a question about the README but didn't find an appropriate place to leave it inline: Should we update the README in `packages/zowex-for-zowe-explorer` to reflect the new command name? It still references "Connect to zowex server on host..."

packages/zowex-for-zowe-explorer/src/SshClientCache.ts
CBforZ: I'm okay with resolving it in this PR but just trying to think of how to address this ... 
 I assume it's by updating this command to include the ../zowex-for-zowe-explorer  folder?  I will look up the docs for this command
 `pnpm exec vscode-l10n-dev export --o ./l10n ./src`

CBforZ: Unfortunately adding  the zowex src folder  to this command didn't fix it so we can do a follow up to address this point

CBforZ: Created https://github.com/zowe/zowe-explorer-vscode/issues/4439

packages/zowex-for-zowe-explorer/src/SshClientCache.ts
CBforZ: Added

packages/zowe-explorer/CHANGELOG.md
CBforZ: Updated

packages/zowe-explorer/CHANGELOG.md
CBforZ: Updated

packages/zowex-for-zowe-explorer/src/SshClientCache.ts
```diff
@@ -216,6 +224,10 @@ export class SshClientCache extends vscode.Disposable {
                             requests: replayRequests,
                             useNativeSsh,
                         });
+                    } else if (serverNotFound) {
```
zFernand0: Thinking through how `serverNotFound` is being used...
I think we need to assign `serverNotFound = false` after a successful `launchServer()` call in line 204. Otherwise, this variable is still saying that we didn't find a server even if we detected one in the user's `$PATH`. 

And instead of relying only on `serverNotFound`, should this check be looking for `if(onEnvPathServer)`

zFernand0 (approved):
LGTM! 😋 
---
left a small question but curious to get your thoughts before merging 🙏

packages/zowe-explorer/l10n/poeditor.json
```diff
@@ -444,7 +444,7 @@
     "Allow commands to be executed using integrated terminals": ""
   },
   "zowe.commands.zowex.connect": {
-    "Connect to zowex server on host...": ""
+    "Deploy zowex server on host and connect...": ""
```
dkelosky: should this instead say "Deploy server and connect..." (i.e. remove `zowex`)?  A couple of reasons:
1.) `zowex` may be renamed
2.) users may not know or understand what zowex is - do we need to expose them to this tool?

packages/zowex-for-zowe-explorer/src/SshClientCache.ts
CBforZ: Thanks for noticing this. I agree I'm missing serverNotFound = false; if launchServer doesn't throw an error. However I do think that `serverNotFound` is the correct check for this case  because we only want to raise the error if at the end of this workflow, they have no usable `zowex` instance. That can happen even if  there is no `zowex` on the $PATH. I believe the case where serverNotFound would change from true to false would just be if we find it on the user's  $PATH, but if we don't find it on their $PATH then we still want to notify them that we were not able to write to the deploy dir.    Let me know if that doesn't make sense though, pushed a rework of this section of the code.

packages/zowe-explorer/l10n/poeditor.json
CBforZ: I'm good with this, other folks have any thoughts?  If we're all good, I'll update the string, PR description, and changelog

packages/zowex-for-zowe-explorer/src/SshClientCache.ts
traeok: I took a look at the latest code and I think its mostly resolved. However, I'm wondering if we should set `serverNotFound` to false after a successful call to `buildClient`? (before the outdated check)

CBforZ: Oop. Yes you are right @traeok  sorry

traeok: No problem! Thanks for quickly addressing it 😋 looking over it now

packages/zowe-explorer/l10n/poeditor.json
CBforZ: Everyone was good with this idea. I'm okay removing zowex from the message.  I just wonder if we should include "SSH" like this screenshot, because otherwise  it's not clear that this command is related to the SSH / ZRS functionality 
image

CBforZ: I pushed this wording but let me know any further feedback

traeok (approved):
LGTM, thanks Chris!

pujal0909 (approved):
LGTM!

zFernand0 (approved):

Issue #4232, opened 2026-05-06
Alias Data Set Support


zFernand0:
## Description
Consolidate fixes for alias data sets to ensure they function identically to regular data sets in Zowe Explorer.

## Data Sets

**Scenario: Access and Navigation**
- Recognize aliases (do not default to binary) 
  - #1545
- Expand PDS/PDSE aliases to view members
  - #2699
  - #4213

**Scenario: Operations**
- Rename the alias, not the base data set`
  - #4041
  - #4463 
- Delete the alias, not the base data set
  - #4464 
- Show attributes (indicate it's an alias and show target)
  - #4465

**Scenario: Search and Filtering**
- Display aliases in search/filter results and allow expansion 
  - #4213

JWaters02:
Just came across this problem when investigating a similar issue on the 2.5.0 release of COBOL LSP.

If zowex already reads the master catalog (which I think I recall seeing somewhere when browsing the code a while back?) then could it be easier to add support for it starting in ssh profile backend rather than zosmf profile backend?

Issue #4213, opened 2026-04-29
Listing of a PDS


jagbr01:
I bring up a search for SHARE.INTERTST.* and it brings up a list of datasets.  They are PDS or PDSE datasets, and I can't expand the datasets.
 

**To Reproduce**

**Expected behavior**

**Screenshots**

**Desktop (please complete the following information):**

- OS:
- Zowe Explorer Version:
- (Optional) Zowe CLI Version:
- (Optional) Do you have secure credentials enabled?
- (Optional) Are you using a custom credential manager?

**Additional context**

zFernand0:
Hey @jagbr01,

Are any of these data sets in the `*ALIAS*` volser?

Zowe Explorer does not support listing members for alias data sets.

adam-wolfe:
z/OSMF REST APIs do not provide any built-in functionality to find the data set that an alias points to. However, would it be reasonable for Zowe Explorer to run a TSO command if a user clicks or tries to expand an alias? i.e.:
`listcat entries('[alias name]') all`

then parse the output for `NONVSAM--[actual dsn]`

I am not sure how reliable that would be, or how the FileSystemProvider implementation would handle it, but that seems to be the easiest way to locate the data set that an alias points to when working with the z/OSMF REST APIs.

It would also require Zowe Explorer to pick a TSO profile to use for issuing the command.

CBforZ:
Closed via #4438

Issue #2699, opened 2024-01-31
Cannot access a dataset which is an alias


closerthanmost:
I have a dataset on z/os, let's say my.dataset.clist I have lots of REXX members in this which vscode doesn't recognise as REXX because of the dataset name, so I have set up an alias of my.dataset.clist.exec to point to it.

When I navigate to the alias, I get the following 

![image](https://github.com/zowe/vscode-extension-for-zowe/assets/158277804/704141ca-572a-4c60-93c4-453404969870)

**To Reproduce**

**Expected behavior**

Show the members from the my.dataset.clist(.exec)

**Screenshots**
![image](https://github.com/zowe/vscode-extension-for-zowe/assets/158277804/535fbcd0-224f-4cd9-80ed-c92d4ef58dcb)

**Desktop (please complete the following information):**

- OS: Windows
- Zowe Explorer Version: v2.9.0

zFernand0:
Opened the following issue for better support across the board:
- https://github.com/zowe/zowe-cli/issues/2052

CBforZ:
Closed via #4438

Issue #1545, opened 2021-10-24
Cannot access alias files


FALLAI-Denis:
**Describe the bug**

Zowe Explorer cannot access files which are in fact aliases (on PDS / Libraries), whereas Zowe CLI does.

It considers them to be files in binary format.

**Expected behavior**

Zowe Explorer must be able to work with aliases.
Unlike TSO / ISPF, the name of the alias file should be kept and not replaced by the name of the actual file.

**Screenshots**

In this use case, SYS9.TMO.SMTMSAMC is an alias to SYS9.TMO.P022.STTMSAMC. A system symbol is used by the alias to address the actual file.

TSO/ISPF:

![image](https://user-images.githubusercontent.com/63861965/138604601-e4ebd1c9-1348-459b-8f08-dc122cadd5e3.png)

![image](https://user-images.githubusercontent.com/63861965/138604560-b77f4364-88c9-498f-b8a9-2b2214db41f6.png)

![image](https://user-images.githubusercontent.com/63861965/138604712-d8941ac5-3a28-45ad-b7d0-dde0e1562616.png)

Zowe Explorer:

![image](https://user-images.githubusercontent.com/63861965/138604842-c538d34f-bcd2-415d-9cca-1f809596827e.png)

![image](https://user-images.githubusercontent.com/63861965/138604767-4085b6fc-70b8-45e6-9ac9-f7d28153f230.png)

![image](https://user-images.githubusercontent.com/63861965/138604891-416c5c7a-7118-4eea-bba1-5e9da706216d.png)

Zowe CLI:

![image](https://user-images.githubusercontent.com/63861965/138605157-caf1dbed-4e3c-4c3e-a7fc-b151fc22ff45.png)
![image](https://user-images.githubusercontent.com/63861965/138605167-00a46c23-0fd1-4980-b07f-94b73e64e5cb.png)

![image](https://user-images.githubusercontent.com/63861965/138605227-f596d340-c3a2-4aca-8bad-2e88b01dba54.png)

![image](https://user-images.githubusercontent.com/63861965/138605419-c855c6b8-d956-486c-8656-fa4ecaabb769.png)

**Desktop (please complete the following information):**

- OS: Windows 10 x64
- Zowe Explorer Version: 1.19.0
- (Optional) Zowe CLI Version: 6.33.1
- (Optional) Are you using Secure Credential Store? Yes

CBforZ:
Closed via #4438

PR #4438, opened 2026-07-31
Add support for resolving aliases via z/OSMF

CBforZ:
## Proposed changes

Add support for resolving aliases in the Zowe data set tree and when using the Zowe data set filesystem provider. Currently only supported when connecting via z/OSMF. 
Showing attributes of data sets in the data set tree can show the original data set's attributes instead of those of the alias
Alias datasets are denoted with a new icon. If we're able to resolve an alias, it will change to a "symlink folder" icon instead of a "symlink file" icon if we can determine that it's a PDS. 

## Release Notes

Milestone:

Changelog:

## Types of changes

## Checklist

**General**

**Code coverage**

**Deployment**


CBforZ:
Related: #4232  #1545 #4213 #2699
We can review as a team if this is merged to see if any of these issues can be closed

packages/zowe-explorer/__tests__/__unit__/extending/ZoweExplorerZosmfApi.unit.test.ts
```diff
@@ -56,6 +56,21 @@ describe("Zosmf API tests", () => {
...
+            expect(ds).toEqual(dsName);
+            return { apiResponse: { targetDsn: "WOW.THE.ORIGINAL" }, commandResponse: "", success: true };
+        });
+
+        vi.spyOn(zosfiles.List, "resolveAlias").mockImplementation(resolveAlias as any);
+
+        const api = new ZoweExplorerZosmf.MvsApi();
+        api.getSession = vi.fn();
+        const resolved = await api.resolveAlias(dsName);
+        expect(resolved.apiResponse.targetDsn).toEqual("WOW.THE.ORIGINAL");
```
CBforZ: I wonder why this doesn't count as covering the resolveAlias method in codecov 

Image

adam-wolfe:
Not a formal review, but this is pretty slick. The hover text is useful.

CBforZ:
> Not a formal review, but this is pretty slick. The hover text is useful.

 Why thank you 😊

packages/zowe-explorer/CHANGELOG.md
```diff
@@ -6,6 +6,7 @@ All notable changes to the "vscode-extension-for-zowe" extension will be documen
 
 ### New features and enhancements
 
+- Added support for resolving aliases in the Zowe data set tree and when using the Zowe dataset filesystem provider. Currently only supported when connecting via z/OSMF. [#4438](https://github.com/zowe/zowe-explorer-vscode/pull/4438)
```
anaxceron: Is there a reason we have "data set" and "dataset"?

CBforZ: Nope  , I'll fix  to "data set"

anaxceron (approved):

adam-wolfe:
Would it make sense to disable renaming an alias? Seems to do strange things to the target data set when the alias is renamed.

Edit - actually it renames the target data set to the name you specify and updates the alias to point to the new name. For some reason I needed to reload the window to get the new PDS to list the members.

Attempting to delete the alias results in an error message, so could we disable trying to delete an alias as well?

CBforZ:
Open to opinions on  Adam's points here.  Disabling delete might make sense

JWaters02:
I'd say probably don't worry about deleting or renaming right now - I feel that modifying/creating/deleting ALIAS data sets are usually related to sysprog/admin activities on a system, and a dev may only just be using it (in our case for copybooks shared for CICS regions and MQs). Sysprogs aren't typically using ZE for the majority of their actions anyway so I feel that this use case is not really necessary for ALIAS data set support to exist

CBforZ:
Decided on the call to try and disable the rename and delete actions on alias data sets

packages/zowe-explorer/package.json
```diff
@@ -1305,22 +1305,22 @@
...
         },
         {
-          "when": "view == zowe.ds.explorer && viewItem =~ /^member.*/  && !listMultiSelection",
+          "when": "view == zowe.ds.explorer && viewItem =~ /^member.*/ && !listMultiSelection",
           "command": "zowe.ds.renameDataSetMember",
           "group": "099_zowe_dsModification@2"
         },
         {
-          "when": "view == zowe.ds.explorer && viewItem =~ /^ds.*/ && !listMultiSelection",
+          "when": "view == zowe.ds.explorer && viewItem =~ /^ds.*/ && viewItem =~ /.*(?!_alias)/ && !listMultiSelection",
```
CBforZ: I tried for a bit to disable the  right click context menu items  if the data set is an alias, but  I still haven't seemed to do it so far. The rename and delete right click items still appear . Trying to do a negative lookahead to say the item context does not contain `_alias`. would appreciate any insights 🤔  If I get this working I will add the same guard to the keybinds for delete and rename.

CBforZ: Maybe I could set a new context value like `aliasContext` to _alias as a kind of constant and then say `&& aliasContext not in viewItem` but it seems like there should be some way with regex here  https://code.visualstudio.com/api/references/when-clause-contexts#in-and-not-in-conditional-operators

CBforZ: I couldn't get that to work either though

CBforZ: Now working. Needed to move `.*` inside the  lookahead

CBforZ:
Rename and delete right click options now disabled for alias data set nodes 
"Delete data set" if invoked by other means like keybinding now rejects if any selected nodes are aliases. 
image

CBforZ:
It was pointed out that it's not easy to tell at a glance which data sets are aliases in the tree. I added some icons from https://github.com/microsoft/vscode-codicons . This seems to be mostly working except when aliases of PDS are first added to the tree , they show the folder icon rather than the "symlink folder" icon. If you expand  the PDS though it uses the right icon, and then if you collapse again it uses the right one... haven't figured out why it's not initially correct yet 
image

CBforZ:
> If you expand the PDS though it uses the right icon, and then if you collapse again it uses the right one... haven't figured out why it's not initially correct yet 

Fixed this. I think I've addressed everything?

packages/zowe-explorer-api/CHANGELOG.md
```diff
@@ -6,6 +6,7 @@ All notable changes to the "zowe-explorer-api" extension will be documented in t
 
 ### New features and enhancements
 
+- Added `resolveAlias` to the `MainframeInteraction.IMvs` interface, initially only implemented by the `ZoweExplorerZosmf` class. [#4438](https://github.com/zowe/zowe-explorer-vscode/pull/4438)
```
traeok: I'd recommend clarifying that the function is optional so that extenders know that its non-breaking:

```suggestion
- Added an optional `resolveAlias` function to the `MainframeInteraction.IMvs` interface, initially only implemented by the `ZoweExplorerZosmf` class. [#4438](https://github.com/zowe/zowe-explorer-vscode/pull/4438)
```

Also, we generally try to add support to FTP for new features (since the FTP serves as an open-source extender sample)... but I don't know of a way off-hand to resolve aliases via FTP, so this probably doesn't apply here 😋

packages/zowe-explorer/resources/dark/file-symlink-directory.svg
traeok, line 1: thanks for adding symlink/alias icons, these look great ^^

packages/zowe-explorer/src/trees/dataset/ZoweDatasetNode.ts
```diff
@@ -879,6 +909,41 @@ export class ZoweDatasetNode extends ZoweTreeNode implements IZoweDatasetTreeNod
...
+                    );
+                    const matchingOriginal = (originalAttributes.apiResponse.items ?? originalAttributes.apiResponse) as IZosmfListResponse[];
+                    if (matchingOriginal.length > 0) {
+                        return {
+                            ...item,
+                            dsorg: matchingOriginal[0].dsorg,
+                            recfm: matchingOriginal[0].recfm,
+                            blksz: matchingOriginal[0].blksz,
+                            dsname: resolution.apiResponse.targetDsn,
+                        };
```
traeok, line 933: It seems like migrated targets would not be detected for aliases - so its possible for an alias to point to a migrated data set, but we're not forwarding that information back to the caller. Should we return whether the original is migrated as part of this object?

```suggestion
                        return {
                            ...item,
                            dsorg: matchingOriginal[0].dsorg,
                            recfm: matchingOriginal[0].recfm,
                            blksz: matchingOriginal[0].blksz,
                            dsname: resolution.apiResponse.targetDsn,
                            migr: matchingOriginal[0].migr
                        };
```

Might be worth adding a test for this case

packages/zowe-explorer/__tests__/__unit__/extending/ZoweExplorerZosmfApi.unit.test.ts
traeok: I think it may be because the matching source file is in the zowe-explorer-api package. Can you try moving the test to the test file in ZE API and see if that helps? `packages/zowe-explorer-api/__tests__/__unit__/profiles/ZoweExplorerZosmfApi.unit.test.ts`

packages/zowe-explorer/src/trees/dataset/DatasetFSProvider.ts
```diff
@@ -268,10 +269,13 @@ export class DatasetFSProvider extends BaseProvider implements vscode.FileSystem
         }
 
         for (const resp of datasetResponses) {
-            for (const ds of resp.apiResponse?.items ?? resp.apiResponse ?? []) {
+            for (let ds of resp.apiResponse?.items ?? resp.apiResponse ?? []) {
                 let tempEntry = profileEntry.entries.get(ds.dsname);
                 if (tempEntry == null) {
                     let name = ds.dsname;
+                    if (ds.vol === "*ALIAS") {
+                        ds = await this.resolveAlias(ds, mvsApi);
```
traeok, line 277: If this call fails or throws due to an auth error, the remaining elements are lost and the function rejects. Can we wrap this in try/catch and fallback to the plain "unresolved" file entry if it fails?

traeok (dismissed):
Thanks for working on this Chris, I left a few suggestions/requests around how the new API is called and leveraged but the changes look good so far!

CBforZ:
Thanks for the Feedback, Trae, will work on these ASAP

packages/zowe-explorer/src/trees/dataset/ZoweDatasetNode.ts
CBforZ: I tested this out .  z/OSMF actually returns `migr: yes` for aliases of migrated data sets, so that means that block right above the `(item.vol === "*ALIAS")` is triggered, so the new alias code doesn't come into play for z/OSMF. It seems zowex does not return migr : yes for aliases of migrated data sets, but we don't support resolving aliases over zowex yet. For the future, I added the `migr` property to the return value of resolveAlias in ZoweDatasetNode and we will have to test out this scenario when we add alias resolution support via zowex 

image

packages/zowe-explorer/src/trees/dataset/DatasetFSProvider.ts
CBforZ: Added try/catch inside resolveAlias helper which returns the original `item` if there's an error

packages/zowe-explorer/__tests__/__unit__/extending/ZoweExplorerZosmfApi.unit.test.ts
CBforZ: Thanks, that made the coverage show up!

traeok (approved):
- [x] Sequential data set alias w/ edits work as expected: changes reflect on alias, changes made from alias reflect on original
- [x] Writing to a member under a PDS alias works as expected; changes reflect from writes to member under alias <-> original
- [x] Create a member under a PDS alias works as expected and original listing updates to reflect new member
- [x] Delete a member under a PDS alias works and original listing reflects deleted member after refresh

This looks and works great from my testing, thanks Chris for adding alias support!

CBforZ:
Thanks for the feedback and testing @traeok !

jace-roell (approved):

AllanDouglasMM:
Great function implemented! I pointed out the issue to identify an alias on VS Code to a friend last month and suddenly after my last upgrade on the tool there it was, an alias identification!

Had to come by and share my thanks to you all, great work!

CBforZ:
Glad to hear it!  We appreciate you coming to share your feedback on this feature 🥳

PR #4454, opened 2026-08-12
fix(ds): sequential DS no longer overwritten after recall

### LINKED ISSUE #4412: Refreshing tree after requesting data set recall causes I/O abend error
**Describe the bug**

I am looking at GDG generations and most of them are migrated. If I right click -> recall, then refresh the tree, I get this error:
```
Rest API failure with HTTP(S) status 500
EDC5092I An I/O abend was trapped. (errno2=0xC00B0161)
__amrc: __last_op=401 __code=0x09130038
fopen() error

Failed to save MY.CENSORED.PDS.G2547V00: Insufficient write permissions for this data set. The data set may be read-only or locked. (Error Code 500) Received HTTP(S) error 500 = Internal Server Error.

Protocol:          https
Host:              host
Port:              port
Base Path:         
Resource:          /zosmf/restfiles/ds/MY.CENSORED.PDS.G2547V00
Request:           PUT
Headers:           [{"X-IBM-Data-Type":"text;fileEncoding=IBM-1146"},{"Accept-Encoding":"gzip"},{"If-Match":"342C50929DC45D4CAF9EB8AE069D3F0C"},{"X-IBM-Return-Etag":"true"},{"X-CSRF-ZOSMF-HEADER":true}]
Payload:           <Buffer >
Allow Unauth Cert: true
Available creds:   user,password
Your auth order:   basic,token,bearer,cert-pem
Auth type used:    basic
```

Then the data set nodes under the profile's tree disappear.

Clicking "Retry" on the error does nothing. But then if I click refresh again or click on the profile node twice (once to collapse and again to expand), the data sets reappear with the selected one recalled.

Even if I wait more than long enough for the recall to finish before refreshing the tree, I still get the error. 

**To Reproduce**

1. Filter data sets to find a recalled one
2. Right click -> recall
3. Refresh tree
4. Should see the error

**Expected behavior**

Error not to happen, tree to refresh with the recalled data set to appear (if its completed recalling)

**Screenshots**

- OS: Win11
- Zowe Explorer Version: 3.6.0-SNAPSHOT (0ffcb29e7f926715a108588329303009f846f7ca)

**Additional context**

traeok:
## Proposed changes

Fixes #4412 

## Release Notes

Milestone: 3.6.0

Changelog: Fixed an issue where recalling a sequential data set caused its contents to be overwritten, resulting in data loss. Now, when a data set is recalled, its contents are no longer modified. 

## Types of changes

## Checklist

**General**

**Code coverage**

- [x] I have ~~added~~ updated unit test cases and they are passing

**Deployment**


anaxceron (approved):

packages/zowe-explorer/src/trees/dataset/ZoweDatasetNode.ts
```diff
@@ -250,7 +250,6 @@ export class ZoweDatasetNode extends ZoweTreeNode implements IZoweDatasetTreeNod
             this.command = { command: "vscode.open", title: "", arguments: [this.resourceUri] };
             if (!DatasetFSProvider.instance.exists(this.resourceUri)) {
                 DatasetFSProvider.instance.createEntry(this.resourceUri, DsType.Ps);
-                await vscode.workspace.fs.writeFile(this.resourceUri, new Uint8Array());
```
zFernand0, line 253: I can't easily test this fix, but it would seem that we made a wrong assumption here 🙏

zFernand0 (approved):
LGTM! 😋

JWaters02:
This does appear to fix the problem of wiping the recalled file, however there is a new issue that if the data set being recalled is open in the favorites tree (regardless of if it was recalled _from_ the favorites tree or not), no matter how many times you refresh the tree, the data set never updates as recalled. But if I filter to the data set on the profile, I can see that it has been recalled. If I recall a data set from the filtered profile, refreshing the tree still works to update the node as recalled.

JillieBeanSim:
I am seeing similar to Josh, testing recall doesn't update icon making file inaccessible to test if data is changed or not, even after collapse/re-expand of session as well as refresh of VSC
tested with zosmf and extender type with same results for both

CBforZ:
That does sound like a legitimate issue, but can we fix that as a follow up since this PR fixes a pretty severe data loss issue?

CBforZ (approved):
Thanks for fixing the data loss issue. Up to @traeok  if he wants to try to fix the  issue described above in this PR or do a follow up

traeok:
Thanks for the reviews @JWaters02 @JillieBeanSim @CBforZ 

I can look into why the node is no longer updating. For reference, I've only removed a call to the `vscode.workspace.fs.writeFile` function in this PR, which created an empty filesystem entry for the newly-recalled sequential DS. Now that we only have the `createEntry` function call, it's possible that an indirect tree refresh from that `writeFile` function call is now skipped (which properly reflected the recalled state).

packages/zowe-explorer/src/trees/dataset/ZoweDatasetNode.ts
traeok: Not necessarily - this code worked before the behavior of `writeFile` was changed for empty buffers (it was used to create a local entry and did not write to remote for this path until recently). _But,_ the new behavior is the correct behavior for the `writeFile` function - we just forgot to remove the function call from a few places after that behavior was adjusted 😅

traeok:
While validating this fix I confirmed a related, but separate issue described by Joshua: if a data set is favorited and gets recalled (whether the recall is triggered from the favorite itself _or_ from the filtered profile), the Favorites branch never reflects the recalled state, no matter how many times the tree is refreshed. The filtered profile branch does update correctly.

Root cause (as far as I traced it): there are two independent code paths that try to keep a favorite's migration status in sync, and they don't agree.

- `hRecallDataSet`/`refreshMigratedDataSet` patch the favorite node directly via `findEquivalentNode` + `datasetRecalled()`, but only when the counterpart branch has already been expanded (the lookup searches an empty children array otherwise).
- `syncFavoriteMigrationStatus` (run whenever the Favorites profile node's children are loaded) independently re-derives migration status from `DatasetFSProvider`'s cached entry `.stats.migr`. That field is only populated when an entry is created - `statImplementation` fetches fresh attributes for change detection on every `stat()`, but only uses them to update `mtime`, never `stats.migr`. So the cached value doesn't get corrected unless the entry is evicted first via `invalidateCache`.

Net effect: the patch mentioned above doesn't always reach the favorite node, and even when it does, a later favorites refresh can independently derive a stale "still migrated" signal from the FS cache and flip it back. A real fix modifies `DatasetFSProvider.statImplementation`, which backs every dataset stat call, not just migration status... which I feel would warrant its own test coverage and review rather than piggybacking the fix onto this one. 

Unless there are any objections, I plan to file a separate issue for this and keep this PR scoped to the overwrite-after-recall fix.

Here's the test suite that I ~generated~ used to reproduce the error with `statImplementation`: https://gist.github.com/traeok/62bad4add6af3f18b310443d5fd84c67

PR #4472, opened 2026-08-31
Show progress indicator when ZRS server is restarting

ArghadeepDas-ibm:
## Proposed changes

Resolved - https://github.com/zowe/zowe-explorer-vscode/issues/4449

When Zowe Remote SSH (ZRS) starts or reconnects its backend server process, users previously received no visual feedback — the operation happened silently in the background. This PR adds a spinning status bar indicator at the bottom of VS Code during both scenarios, so users always know when ZRS is active.

## Release Notes

Milestone:

Changelog: Added a visual progress indicator in the VS Code status bar while the Zowe Remote SSH server is starting or reconnecting, so users can see when ZRS is active rather than waiting without feedback. 

## Types of changes

## Checklist

**General**

**Code coverage**

**Deployment**


packages/zowex-for-zowe-explorer/src/Utilities.ts
traeok, line 1: can we revert the changes to this file? it looks like only a single empty line was added

anaxceron (approved):

packages/zowex-for-zowe-explorer/src/SshClientCache.ts
```diff
@@ -254,6 +255,7 @@ export class SshClientCache extends vscode.Disposable {
                 startTime: Date.now(),
                 responseTimeoutMillis: responseTimeout * 1000,
             });
+            statusBarMsg.dispose();
```
traeok: This is only reached when the connection succeeds. Any errors thrown above would skip this dispose call and cause the status bar message to persist, at least until the window is reloaded.

We could wrap this area in a `try/finally` statement and call dispose within the `finally` block, so that the status bar message is always disposed regardless of a successful connection or an error.

packages/zowex-for-zowe-explorer/src/SshClientCache.ts
```diff
@@ -295,6 +297,8 @@ export class SshClientCache extends vscode.Disposable {
                     });
                 }
             );
+            Gui.setStatusBarMessage(`$(sync~spin) Restarting zowex server for profile "${profile.name as string}"...`, reconnectPromise);
```
traeok: In some cases, wouldn't this also trigger whenever no cached session is found and an initial connection is being established? Saying that the server is "Restarting" in those cases is misleading, so we may need to key off of `opts.restart` to pick the right verb.

Also, I'm wondering if we should rephrase this to say "Zowe Remote SSH server" instead, since `zowex` is more technical and refers to the binary? I could go either way on this one, so this part of my comment is more of a suggestion.

```suggestion
            Gui.setStatusBarMessage(`$(sync~spin) ${opts.restart ? "Restarting" : "Starting"} Zowe Remote SSH server for profile "${profile.name as string}"...`, reconnectPromise);
```

traeok (changes requested):
Thanks for adding this indicator @ArghadeepDas-ibm! I think this is a nice improvement to the UX, especially in scenarios where the server is being re-established. I left a suggestion around the message wording for possible edge cases. I'm requesting changes regarding the status bar message persisting in event of an error.

ArghadeepDas-ibm:
Hi @traeok ,
The above reviews has been addressed in the latest changes, would request you to review the changes and let me know if any other changes are required.

traeok (approved):
Tested and the status bar indicator LGTM, thanks Arghadeep!

t1m0thyj (approved):
LGTM, thanks @ArghadeepDas-ibm!

PR #4474, opened 2026-08-31
Feat(api): getEncoding method

ddustindd:
**Proposed changes**
Exposes a method on ZoweExplorerApiRegister.getFileApi() for extenders that work with mainframe file system resources:

- getEncodingForUri(uri) returns the encoding selected by the user for a given dataset or USS file URI, so extenders can read and apply the correct encoding without accessing internal Zowe Explorer state directly.

The method is accessible through Types.IZoweExplorerFileApi, which is now part of Types.IApiRegisterClient.

Milestone:

Changelog:
Added getFileApi() to ZoweExplorerApiRegister, exposing getEncodingForUri for extenders to query encoding state for a given URI. 

## Types of changes

## Checklist

**General**

**Code coverage**

- [x] I have added at least one end-to-end test for new features to validate behavior (if applicable)

**Deployment**


jace-roell:
Hey @ddustindd, thanks for opening this PR,

Quick question, is there a specific use case for invalidating the cache vs using the existing `fetch=true` query parameter on the other file system functions?

ddustindd:
> , is there a specific use case for invalidating the cache vs using the existing `fetch=true` query parameter on the other file system functions?

@jace-roell Passing `fetch=true` as a URI query parameter when calling `stat` or `readDirectory` forces a mainframe metadata lookup, which works well for those operations. Whether that then triggers a fresh download in readFile depends on whether the mtime comparison finds the entry stale.

This comparison has failed in practice, particularly for sequential datasets. I suspect that for sequential datasets, the mainframe mtime may not update immediately after an upload, so stat(fetch=true) sees an unchanged mtime and leaves wasAccessed (ZE's internal flag tracking whether a file has been read since it was last fetched) as true. When revert calls readFile, ZE checks wasAccessed, sees true, and returns the cached bytes without touching the mainframe at all. Do you have any insight into why this would happen more often for sequential datasets than PDS members? I did try the `fetch=true` approach you mention.

`invalidateCacheForUri` skips the comparison by removing the cache entry, so the next readFile finds nothing locally and downloads unconditionally, which makes it ideal for extenders that write outside of ZE's normal write flow and need a guaranteed fresh read.

Appreciate you looking into this. Let me try creating a visual example.

ddustindd:
@jace-roell The use case is for sequential datasets. I investigated stat(fetch=true) as the mechanism to force a re-download but it does not work for sequential datasets. When the entry already exists in the cache, remoteLookupForResource calls fetchDataset, which gates the entire mainframe call behind !entryExists || forceFetch (DatasetFSProvider.ts line 420). 

Screenshot 2026-09-01 at 4 41 25 PM

For a previously opened sequential dataset, entryExists is true and forceFetch is undefined, so the block is skipped. The mtime comparison and wasAccessed reset never run. The revert that follows returns stale cached bytes.

invalidateCacheForUri fixes this by deleting the cache entry first. With no entry present, entryExists is false, the block at line 420 executes, the mainframe is called, and the subsequent revert downloads fresh content. PDS members are unaffected because their fetch path goes through fetchEntriesForDataset (DatasetFSProvider.ts line 469), which always calls the API regardless.

Screenshot 2026-09-01 at 4 41 41 PM

jace-roell:
> nvalidateCacheForUri fixes this by deleting the cache entry first. With no entry present, entryExists is false, the block at line 420 executes, the mainframe is called, and the subsequent revert downloads fresh content. PDS members are unaffected because their fetch path goes through fetchEntriesForDataset (DatasetFSProvider.ts line 469), which always calls the API regardless.

Thanks for looking into this, I did validate that `stat` calls on sequential data sets were not properly respecting the `fetch=true` query param. 

In regards to the `mtime`, I don't believe that sequential data sets return an `mtime` value as PDSes do. (atleast via z/OSMF).  Also, I don't think that relying on an extender to clear a cache entry in order to force a fetch is the best solution for this specific scenario.

I have opened a quick PR to resolve the fetch behavior (https://github.com/zowe/zowe-explorer-vscode/pull/4476). Would you be able to test if this fix would resolve your case and we could omit exposing the `invalidateCache` function?

Thanks again!

ddustindd:
Yes, I can confirm it works, appreciate it. I agree we could omit exposing the `invalidateCache` function. Your PR solves the issue in a cleaner way.

packages/zowe-explorer/src/extending/ZoweExplorerApiRegister.ts
```diff
@@ -409,4 +415,49 @@ export class ZoweExplorerApiRegister implements Types.IApiRegisterClient {
...
+
+    /** @internal Called by UssFSProvider during singleton construction. */
+    public setUssFSProvider(provider: IZoweFSProvider): void {
+        this.ussFSProvider = provider;
+    }
+
+    /** @internal Called by JobFSProvider during singleton construction. */
+    public setJobFSProvider(provider: IZoweFSProvider): void {
+        this.jobFSProvider = provider;
+    }
```
traeok: These setters can be removed - see below for context

packages/zowe-explorer/src/extending/ZoweExplorerApiRegister.ts
```diff
@@ -409,4 +415,49 @@ export class ZoweExplorerApiRegister implements Types.IApiRegisterClient {
...
+        switch (scheme as ZoweScheme) {
+            case ZoweScheme.DS:
+                return this.datasetFSProvider;
+            case ZoweScheme.USS:
+                return this.ussFSProvider;
+            case ZoweScheme.Jobs:
+                return this.jobFSProvider;
+            default:
+                return undefined;
+        }
```
traeok: As it currently stands, this function leaks provider access - a caller could do (`(...getApiRegister() as any).providerForScheme("zowe-ds")` to get access to the full provider instance. The interface does not prevent other class members/functions from being accessed. Is it possible to inline this switch statement to prevent this direct access?

packages/zowe-explorer/src/extending/ZoweExplorerApiRegister.ts
```diff
@@ -409,4 +415,49 @@ export class ZoweExplorerApiRegister implements Types.IApiRegisterClient {
...
+     * @returns the Zowe Explorer file API instance
+     */
+    public getFileApi(): Types.IZoweExplorerFileApi {
+        if (this.fileApi) {
+            return this.fileApi;
+        }
+        this.fileApi = {
+            getEncodingForUri: (uri: vscode.Uri): ZosEncoding | undefined => {
+                const provider = this.providerForScheme(uri.scheme);
+                return provider?.encodingMap[uri.path];
```
traeok: Encodings that ZE derives itself never enter this map - most notably, USS file-tag detection, which assigns `entry.encoding = { kind: "binary" }` or `{ kind: "other", codepage: taggedEncoding }` directly. An extender that opens a USS file tagged `IBM-939` (no explicit user selection) gets `undefined` from the new API and would read the file as the default codepage.

I'd recommend resolving the encoding from the entry first and fall back to the map, for example: `provider.getEncodingForFile(uri) ?? provider.encodingMap[uri.path]`, using a silent (no-fail) lookup. 

Note that `getEncodingForFile` currently calls `lookup(uri, false)` and therefore *throws* `FileNotFound` for an uncached URI, so it needs a silent variant or a `try/catch` before it can be used in a public extender API.

Also, if we are assuming that all providers have an `encodingMap`, would it be possible to define it on the `BaseProvider` class and eliminate the need for the new `IZoweFSProvider` interface?

packages/zowe-explorer/src/extending/ZoweExplorerApiRegister.ts
```diff
@@ -35,6 +37,10 @@ export class ZoweExplorerApiRegister implements Types.IApiRegisterClient {
 
     public onProfileUpdatedEmitter: vscode.EventEmitter<imperative.IProfileLoaded> = new vscode.EventEmitter();
     public readonly onProfileUpdated = this.onProfileUpdatedEmitter.event;
+    private datasetFSProvider?: IZoweFSProvider;
+    private ussFSProvider?: IZoweFSProvider;
+    private jobFSProvider?: IZoweFSProvider;
```
traeok: These should be removed even if they are just private, because these are still accessible for anyone that has access to this class instance (which is already exposed to extenders). It leaks access to providers through the API register - we've taken an implicit stance to avoid exposing the providers themselves as its considered a  security risk.

If we absolutely need private member instances of the providers, they need to be prefixed with a pound sign to make them inaccessible to consumers of the API register. See here for more info: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/Private_elements

traeok (dismissed):
Hi Dustin, thanks for working on adding a new `getEncoding` API! I've left a few suggestions, and I'm requesting changes regarding how the providers are currently stored.

packages/zowe-explorer/src/extending/ZoweExplorerApiRegister.ts
ddustindd: Removed all three setters and replaced them with a single registerFSProvider(scheme, provider) method. The data goes into the #fsProviders Map which is # private, so once a provider is registered there is no way to read it back out from outside the class. This also eliminated the three-setter surface that extenders could have called to overwrite our providers.

packages/zowe-explorer/src/extending/ZoweExplorerApiRegister.ts
ddustindd: replaced the three separate private fields with a single #fsProviders = new Map<ZoweScheme, BaseProvider>(). The # prefix makes these runtime-private so they can't be accessed via `as any`.

packages/zowe-explorer/src/extending/ZoweExplorerApiRegister.ts
ddustindd: Deleted the method entirely. With the three separate fields replaced by a single Map, the lookup became a one-liner so there was no reason for a separate method. The lookup now lives inline inside getEncodingForUri.

packages/zowe-explorer/src/extending/ZoweExplorerApiRegister.ts
ddustindd: - getEncodingForUri now calls getEncodingForFile(uri) first to read entry.encoding (auto-detected path, e.g. USS file-tag detection) and falls back to encodingMap[uri.path] (user-explicit selection). 
- The whole thing is wrapped in a try/catch because getEncodingForFile throws FileNotFound for uncached URIs. 
- Moved encodingMap to BaseProvider since all three providers declared it identically, and deleted IZoweFSProvider since BaseProvider now covers everything the interface was providing

traeok (approved):
Changes LGTM, thanks Dustin for the new method!

traeok (changes requested):
I approved but then realized a changelog entry is missing from Zowe Explorer / ZE API. Can you add an entry to reflect the changes? Aside from that, changes look good to me

ddustindd:
Appreciate the review and you catching the security risks. I will keep your feedback in mind for future development.

traeok (approved):

ddustindd:
@jace-roell Thanks for the fix! After testing I found there's still one more layer to the problem. I opened issue [#4481](https://github.com/zowe/zowe-explorer-vscode/issues/4481) with the details. It traces back to the m4date-only precision in statImplementation. Should I open a PR with a suggested fix? Otherwise happy to help test any follow-up fix.

t1m0thyj (approved):
LGTM, thanks @ddustindd! Clever way to work around circular deps by adding the new method to register FS provider 😋

JillieBeanSim (approved):
LGTM thanks @ddustindd for the updates

PR #4482, opened 2026-09-04
fix: clarify wording of the update credentials prompts

### LINKED ISSUE #3744: Wording for updating credentials is confusing
**Describe the bug**

Image

The meaning of "Leave blank to not store." is difficult to decipher. What does "not store" mean? What happens if I already have a user ID or password set and I clear out the textbox and press enter? Do my credentials get removed?

**To Reproduce**

1. Right click on a profile node.
2. Select "Manage profile".
3. Click "Update credentials".

**Expected behavior**

**Screenshots**

- OS:
- Zowe Explorer Version:

**Additional context**

YousufFFFF:
## Proposed changes

Fixes #3744

The credential prompts ended with "Leave blank to not store.", which doesn't say what "not store" means or whether it removes credentials you already have.

Tracing the behaviour: leaving a field blank returns `undefined` from `promptUserPass`, so `updateCredentials` never reaches `updateProperty`. The operation is simply cancelled and **stored credentials are never removed**. The hint now says exactly that: "Leave blank to cancel."

There is also a second, sharper problem on the exact path in the report (`Manage Profile` → `Update Credentials`, which calls `promptCredentials(profile, true)`). `ZoweVsCodeExtension.promptUserPass` already suppresses the hint when `rePrompt` is set, but `Profiles.promptCredentials` appended it unconditionally and its `userInputBoxOptions` are spread *after* the default `prompt`, so the override won. The user was told to leave the field blank while `validateInput` rejects a blank user name with "User name cannot be empty". `Profiles.promptCredentials` is now `rePrompt`-aware to match the API, so the hint no longer appears when updating existing credentials — the only path where credentials could exist to be lost.

Added unit tests covering both branches.

## Release Notes

Milestone: v3.5.2

Changelog: Fixed confusing wording in the "Update Credentials" prompts. The hint "Leave blank to not store." is now omitted when updating existing credentials, and reworded to "Leave blank to cancel." elsewhere, clarifying that a blank value cancels the operation rather than clearing any stored credentials.

## Types of changes

## Screenshot

image


packages/zowe-explorer-api/CHANGELOG.md
```diff
@@ -14,6 +14,7 @@ All notable changes to the "zowe-explorer-api" extension will be documented in t
 
 ### Bug fixes
 
+- Fixed confusing wording in the credential prompts shown by `promptUserPass`. The hint "Leave blank to not store." is now "Leave blank to cancel.", clarifying that submitting a blank value cancels the operation rather than clearing any stored credentials. [#3744](https://github.com/zowe/zowe-explorer-vscode/issues/3744)
```
anaxceron, line 17: We should explain what `promptUserPass` is: "... shown by class/property/cat `promptUserPass`."

anaxceron, line 17:
```suggestion
- Fixed confusing wording in the credential prompts shown by `promptUserPass`. The message "Leave blank to not store" is now "Leave blank to cancel," clarifying that submitting a blank value cancels the operation rather than clearing any stored credentials. [#3744](https://github.com/zowe/zowe-explorer-vscode/issues/3744)
```

packages/zowe-explorer/CHANGELOG.md
```diff
@@ -21,6 +21,7 @@ All notable changes to the "vscode-extension-for-zowe" extension will be documen
 
 ### Bug fixes
 
+- Fixed confusing wording in the "Update Credentials" prompts. The hint "Leave blank to not store." was shown even when updating existing credentials, where a blank user name is rejected outright. It is now omitted when updating, and reworded to "Leave blank to cancel." elsewhere, clarifying that a blank value cancels the operation rather than clearing any stored credentials. [#3744](https://github.com/zowe/zowe-explorer-vscode/issues/3744)
```
anaxceron, line 24: Please apply changes based on my comments on the other changelog.

anaxceron (commented):
The message is clearer now, nicely done. Left some comments for minor edits.

traeok (approved):
Thanks @YousufFFFF changes LGTM

PR #4483, opened 2026-09-09
Fix: Tree refreshes into a bad state on allocate like command

ArghadeepDas-ibm:
## Proposed changes

Resolves - https://github.com/zowe/zowe-explorer-vscode/issues/4444

The tree refresh fired while currSession.dirty was still false, so VS Code re-queried the tree and got stale (empty) children. A second refreshElement() call then triggered another full re-render cycle, causing the visible flash of disappearing nodes.

Fix:

- Set currSession.dirty = true and currSession.collapsibleState = Expanded before calling datasetProvider.refresh(), so VS Code fetches fresh children immediately on the first render.
- Removed the redundant datasetProvider.refreshElement(currSession) call that was causing the second render cycle.

## Release Notes

Milestone:

Changelog: Fixed an issue where using "Allocate Like" on a data set caused the tree nodes to disappear until the profile was collapsed and re-expanded.

## Types of changes

## Checklist

**General**

**Code coverage**

**Deployment**


packages/zowe-explorer/src/trees/dataset/DatasetActions.ts
```diff
@@ -468,9 +468,10 @@ export class DatasetActions {
 
         const theFilter = datasetProvider.createFilterString(newDSName, currSession);
         currSession.pattern = theFilter.toUpperCase();
-        datasetProvider.refresh();
+        datasetProvider.addSearchHistory(currSession.pattern);
```
JWaters02: Why are we adding this to search history?

ArghadeepDas-ibm: createFilterString reads getSearchHistory()[0] as the initial value of theFilter, and then gates the entire append branch on && theFilter. Since allocateLike never updated the search history, getSearchHistory()[0] stayed frozen at the user's original manual search. On the second run, if that stale history entry didn't satisfy && theFilter correctly, the pattern would accumulate incorrectly.

So I added addSearchHistory to keep getSearchHistory()[0] in sync with the pattern that allocateLike had just set — making theFilter always reflect the current accumulated pattern on the next call.

JWaters02: I'm sorry but I don't understand what you are saying. To rephrase my original question: why would we want to add the newly created data set to the search history, since search history implies that it only contains entries from data sets the user searched for?

ArghadeepDas-ibm: Hi @JWaters02.
You're absolutely right. Adding it to search history was incorrect for exactly the reason you stated: search history is meant to track patterns the user explicitly searched for via the filter dialog, not datasets created programmatically. Putting a dataset name there would pollute the user's search history dropdown with entries they never typed.

I am reverting back the changes. Would request you to review the changes and let me know if any more changes are required.

JWaters02: Would it be possible next time to have the human reply to review requests rather than the LLM? Respectfully, I don't want to feel like my comments are being fed right into a prompt and given the output straight from the bot.

JWaters02 (approved):
Functionality is fixed, thanks

anaxceron (approved):

JillieBeanSim (approved):
LGTM! thanks for the fix @ArghadeepDas-ibm

PR #4496, opened 2026-09-15
fix(ds): decode URI component in rename error message 

### LINKED ISSUE #4468: URI encoding in error messages may be confusing
**Describe the bug**
I assume that this may occur for other error messages as well. If I attempt to rename a member to a member name that already exists, Zowe Explorer displays an error message that shows the URI-encoded form of the DS/member, which can be confusing. E.g.:

> Error running command zowe.ds.renameDataSetMember: Unable to move/copy 'zowe-ds:/dev.zosmf/ADAM.PUBLIC.COBOL/TEST.cbl' because target 'zowe-ds:/dev.zosmf/ADAM.PUBLIC.COBOL/%23TEST.cbl' already exists at destination.. This is likely caused by the extension that contributes zowe.ds.renameDataSetMember.

As a nitpick, there is also an extra period after destination.

**To Reproduce**

Rename a member to a name that is already in use by another member in that PDS.

**Expected behavior**

Avoid URI encoding URIs in error messages. It still might be useful to keep the URI encoding in logs.

**Screenshots**

- OS: 
- Zowe Explorer Version: 3.5.1

**Additional context**

snehasingh-25:
**Title:** `fix(ds): decode URI component in rename error message for special-char member names`

---

## Proposed changes

Fixes #4468

When renaming a PDS member to a name that already exists, Zowe Explorer threw an error message containing the **URI-encoded** form of the member name. For example, attempting to rename a member to `#TEST` would show:

> `Rename failed: %23TEST.cbl already exists`

instead of the expected human-readable:

> `Rename failed: #TEST.cbl already exists`

The root cause was in `DatasetFSProvider.rename()` — it used `path.posix.basename(newUri.path)` directly in the error message, but `newUri.path` contains the percent-encoded URI path (e.g. `%23` for `#`).

**Fix:** Wrap the basename with `decodeURIComponent()` before inserting it into the error message.

---

## Release Notes

**Milestone:** 

**Changelog:**
> Fixed an issue where renaming a PDS member to a name that already exists would display the URI-encoded form of the member name (e.g., `%23TEST`) instead of the decoded form (e.g., `#TEST`) in the error message. [#4468](https://github.com/zowe/zowe-explorer-vscode/issues/4468)

---

## Types of changes

---

## Checklist

**General**
- [x] The pre-publish command has been executed (`pnpm --filter vscode-extension-for-zowe vscode:prepublish`)

**Code coverage**

---

## Further comments

The fix is a single `decodeURIComponent()` call wrapping `path.posix.basename(newUri.path)` in `DatasetFSProvider.ts`. A regression test was added that specifically exercises a member name containing `#` (which gets encoded as `%23` in the URI) to prevent future regressions.


packages/zowe-explorer/src/trees/dataset/DatasetFSProvider.ts
```diff
@@ -1117,7 +1118,7 @@ export class DatasetFSProvider extends BaseProvider implements vscode.FileSystem
     public async rename(oldUri: vscode.Uri, newUri: vscode.Uri, options: { readonly overwrite: boolean }): Promise<void> {
```
traeok, line 1118: Line 1128 of this file computes `const newName = path.posix.basename(newUri.path);` from the exact same `newUri.path` that this PR just decided needs `decodeURIComponent()` two lines above. 

`newName` is then used, unguarded, for: 
- the actual mainframe API calls (`renameDataSet`/`renameDataSetMember`, lines 1136-1145)
- the persisted in-memory entry name (`entry.name = newName`, line 1166)
- the new metadata path (`newPath`, line 1169)

If `newUri.path` can carry percent-encoded segments, then a rename to a special-character name could risk renaming the mainframe member/data set to the *encoded* string (e.g. `%23TEST`) while the tree/error UI now shows the decoded form, producing a user-visible mismatch between what was requested and what was actually stored on the mainframe. 

The identical pattern exists in two places in `UssFSProvider.ts`, which feed `getUssApi(profile).rename(...)` directly.

Keep in mind that our SDKs for z/OSMF already perform URI encoding at request time, so there may be implications of doing URI encoding here before making any SDK requests (duplicate encoding efforts, potential parse issues in SDK)

packages/zowe-explorer/__tests__/__unit__/trees/dataset/DatasetFSProvider.unit.test.ts
```diff
@@ -2624,6 +2624,23 @@ describe("DatasetFSProvider", () => {
             ).rejects.toThrow("Rename failed: USER.DATA.PS2 already exists");
         });
 
+        it("shows a decoded member name in the error when target already exists and name contains special characters", async () => {
```
traeok, line 2627: The real `lookup()` function splits `uri.path` on `/` and does a raw, non-decoded `entries.get(part)` lookup against keys that are stored decoded (e.g. `"#TEST.cbl"`, per `new DsEntry(specialMemberName, true)` in the same test). Because the test stubs `lookup()` instead of exercising it, it never confirms that a `newUri` shaped this way would actually make the real `lookup()` return the existing entry and hit the `throw` statement.

I'd recommend validating with an unmocked `lookup()`/`entries` map, or with a manual repro of the actual UI action that produces the encoded `newUri.path` (renaming through the Data Sets tree), to be sure the fix addresses the real code path rather than a hand-crafted string.

packages/zowe-explorer/CHANGELOG.md
```diff
@@ -26,6 +21,7 @@ All notable changes to the "vscode-extension-for-zowe" extension will be documen
 
 ### Bug fixes
 
+- Fixed an issue where renaming a PDS member to a name that already exists would display the URI-encoded form of the member name (e.g., `%23TEST`) instead of the decoded form (e.g., `#TEST`) in the error message. [#4468](https://github.com/zowe/zowe-explorer-vscode/issues/4468)
```
anaxceron, line 24:
```suggestion
- Fixed an issue where renaming a PDS member to a name that already exists would display the URI-encoded form of the member name (for example, `%23TEST`) instead of the decoded form (for example, `#TEST`) in the error message. [#4468](https://github.com/zowe/zowe-explorer-vscode/issues/4468)
```

anaxceron (commented):
Requesting minor edits to align w/ Zowe style

PR #4498, opened 2026-09-15
fix(ds): refresh sequential dataset editor after save and expose notifyFileChanged for extenders

### LINKED ISSUE #4481: stat(fetch=true) does not refresh sequential data sets modified on the same day (day-level m4date precision)
**Describe the bug**

PR #4476 fixed one layer of this issue (fetch=true was not being forwarded to fetchDataset for sequential data sets, so the mainframe was never called at all).

However there is a second layer to the problem. The editor still shows stale content for same-day edits on sequential datasets. The mainframe is now called, but the response is compared against the cached timestamp using only m4date which seems to have day-level precision, so if the data set was already opened earlier the same day, the dates match, the cache is never invalidated, and readFile still serves the old content.

z/OSMF only populates mtime (HH:MM) for data sets last modified via ISPF. Sequential data sets modified outside of ISPF return only m4date. The relevant code in DatasetFSProvider.ts:

```
if (m4date) {
    const newTime = mtime ? dayjs(`${m4date} ${mtime}:${msec || "00"}`).valueOf() : dayjs(m4date).valueOf();
    if (entry.mtime != newTime) {   // same day → same value → wasAccessed never reset
        entry.mtime = newTime;
        entry.wasAccessed = false;
    }
}
```

Since wasAccessed is never reset, readFile short-circuits and the mainframe is never queried for fresh content.

**To Reproduce**

1. Open a sequential data set in the editor via Zowe Explorer.
2. Modify and save the data set from a non-ISPF tool (e.g. a JCL job or batch update) on the same calendar day it was previously opened.
3. Use the Refresh action on the data set node in the Zowe Explorer tree.
4. The editor still shows the old content instead of the updated content from the mainframe.

I confirmed with logs too.

**Expected behavior**

After refreshing, the editor should show the latest content from the mainframe regardless of whether the data set was modified on the same day.

**Screenshots**
Image

- OS: MacOS 26.5.2
- Zowe Explorer Version: v3.5.0

**Additional context**

This only affects sequential data sets (PS). PDS members are not affected because their fetch path goes through fetchEntriesForDataset, which calls mvsApi.allMembers() and always returns full intra-day mtime precision.

The else branch for the no-m4date case in the same function already handles this correctly by unconditionally setting entry.wasAccessed = false without touching mtime. The same treatment should apply when m4date is present but mtime is absent.

ddustindd:
## Proposed changes
More context can be found in this [issue](https://github.com/zowe/zowe-explorer-vscode/issues/4481).

Fix sequential dataset (PS) editor not refreshing after save. stat() was returning the cached result for PS entries even when the file was open in the editor, because the isVisibleEditor network bypass that already existed for PDS members was never applied to the PS branch. The PS branch now matches PDS member behavior.

Add notifyFileChanged(uri) to IZoweExplorerFileApi so extenders that save to the mainframe outside of ZE's writeFile (e.g. custom editors) can invalidate ZE's cache and fire a FileChangeType.Changed event, causing VS Code to re-read the file.

## Release Notes

Milestone:

Changelog:

## Types of changes

## Checklist

**General**

**Code coverage**

**Deployment**


packages/zowe-explorer/src/trees/dataset/DatasetFSProvider.ts
```diff
@@ -223,7 +223,7 @@ export class DatasetFSProvider extends BaseProvider implements vscode.FileSystem
 
         return this.executeWithReuse<vscode.FileStat>(uri, {
             keyGenerator: (u) => "list" + this.getQueryKey(u) + "_" + u.toString().split("/").slice(0, 3).join("/"),
-            checkLocal: () => !!this.lookup(uri, true),
+            checkLocal: () => (isVisibleEditor ? false : !!this.lookup(uri, true)),
```
traeok: The statement here is a bit loaded as it contains ternary operator and boolean logic, can we reduce it to just the boolean logic? This is functionally equivalent to `!isVisibleEditor && !!this.lookup(uri, true)`:

```suggestion
            checkLocal: () => !isVisibleEditor && !!this.lookup(uri, true),
```

traeok (commented):
Thanks Dustin, the changes make sense but I left a suggestion about making the logic more readable.

packages/zowe-explorer/src/trees/dataset/DatasetFSProvider.ts
ddustindd: Makes sense, thanks for the suggestion!

packages/zowe-explorer/src/extending/ZoweExplorerApiRegister.ts
```diff
@@ -445,6 +445,13 @@ export class ZoweExplorerApiRegister implements Types.IApiRegisterClient {
                     return provider.encodingMap[uri.path];
                 }
             },
+            notifyFileChanged: (uri: vscode.Uri): void => {
+                const provider = this.#fsProviders.get(uri.scheme as ZoweScheme);
```
traeok: Sorry, I should have thought of this during my first round of review, but I think it would be useful to add a log here since we're in the context of Zowe Explorer.

Consider adding a `ZoweLogger.info` call to track when this function is called (containing the provided URI), mostly to help with future debugging endeavors 👍

traeok (commented):
Thanks Dustin! One more thing and I think the PR is good to go

packages/zowe-explorer/src/extending/ZoweExplorerApiRegister.ts
ddustindd: Good call

ddustindd: Cool Suicune pfp btw, I should add a pokemon pfp myself.

traeok (approved):
Changes LGTM, thanks Dustin for the fix & new helper function!

traeok (changes requested):
Sorry, did not catch the missing patch coverage. Can you add some unit tests to cover the new `notifyFileChanged` function?

In the past, I've run into some difficulties getting patch coverage to reflect for arrow functions in Codecov - it may help to part out the arrow function into a static method for testing the logic inside it - and then just calling the static method directly, e.g.:

`notifyFileChanged: (...) => ZoweExplorerApiRegister.doNotifyFileChanged(...)`
and then writing the tests around the static function call rather than testing this small function stub

ddustindd:
> Sorry, did not catch the missing patch coverage. Can you add some unit tests to cover the new `notifyFileChanged` function?
> 
> In the past, I've run into some difficulties getting patch coverage to reflect for arrow functions in Codecov - it may help to part out the arrow function into a static method for testing the logic inside it - and then just calling the static method directly, e.g.:
> 
> `notifyFileChanged: (...) => ZoweExplorerApiRegister.doNotifyFileChanged(...)` and then writing the tests around the static function call rather than testing this small function stub

Thanks for the heads up

packages/zowe-explorer-api/CHANGELOG.md
```diff
@@ -6,6 +6,7 @@ All notable changes to the "zowe-explorer-api" extension will be documented in t
 
 ### New features and enhancements
 
+- Added `notifyFileChanged` to `IZoweExplorerFileApi` so extenders can notify Zowe Explorer that a file was changed outside of the filesystem provider, causing VS Code to re-read the file. [#4481](https://github.com/zowe/zowe-explorer-vscode/issues/4481)
```
anaxceron: We should explain what `notifyFileChanged` and `IZoweExplorerFileApi` are

packages/zowe-explorer/CHANGELOG.md
```diff
@@ -26,6 +27,7 @@ All notable changes to the "vscode-extension-for-zowe" extension will be documen
 
 ### Bug fixes
 
+- Fixed an issue where sequential datasets open in the text editor were not refreshing after a remote save, because the `isVisibleEditor` cache bypass in `stat()` was only applied to PDS members and not sequential datasets. [#4481](https://github.com/zowe/zowe-explorer-vscode/issues/4481)
```
anaxceron: - Per Zowe style, "data set(s)" is two words
- We should explain what `stat()` is

anaxceron (commented):
Left some comments regarding style, please address

traeok:
After you have addressed @anaxceron 's changelog suggestions, please run this command: `pnpm --filter vscode-extension-for-zowe vscode:prepublish` 

Pushing the changes from that command should resolve the failing Lint & Pre-publish stage.

ddustindd:
My changes don't add any new vscode.l10n.t() strings, but running vscode:prepublish locally removes around 100 strings from the l10n files. After investigating, the missing strings are SSH/zowex release notes entries that exist in upstream/main's changelog but not in my branch's source. They appear to have come in through the zowex staging branch merge, and since my branch doesn't have those changelog files, prepublish cannot generate those strings.

I've restored the l10n files to exactly match upstream/main. Is that acceptable for this PR, given that my changes don't touch any localizable strings?

traeok:
> My changes don't add any new vscode.l10n.t() strings, but running vscode:prepublish locally removes around 100 strings from the l10n files. After investigating, the missing strings are SSH/zowex release notes entries that exist in upstream/main's changelog but not in my branch's source. They appear to have come in through the zowex staging branch merge, and since my branch doesn't have those changelog files, prepublish cannot generate those strings.
> 
> I've restored the l10n files to exactly match upstream/main. Is that acceptable for this PR, given that my changes don't touch any localizable strings?

Hmm, that's odd, maybe it is something specific to forks? Just to clarify, does this still happen after updating your fork's `main` branch and doing a `git merge main` locally, then running the prepublish script?

traeok (approved):

ddustindd:
Yes. After investigating, the issue was not fork-specific. The generateReleaseNotesL10n.js script reads release-notes.md from src/webviews/dist/resources/, which is a build artifact excluded by .gitignore. My local dist/ copy was stale and missing the 3.6.0 content, so the script generated fewer strings than what was committed, causing the diff check to fail.

Running pnpm build first to populate dist/ and then vscode:prepublish fixed it. The prepublish check passes now. Appreciate the review and approval!

packages/zowe-explorer-api/CHANGELOG.md
```diff
@@ -6,6 +6,7 @@ All notable changes to the "zowe-explorer-api" extension will be documented in t
 
 ### New features and enhancements
 
+- Added `notifyFileChanged` to `IZoweExplorerFileApi` (the file API returned by `getFileApi()`) so extenders that save to the mainframe outside of Zowe Explorer's filesystem provider can invalidate the cached entry and trigger VS Code to re-read the file from the mainframe. [#4481](https://github.com/zowe/zowe-explorer-vscode/issues/4481)
```
anaxceron: We still should explain what `notifyFileChanged` is, and I think we could easily do that w/ an adjective, like:

"Added the message/header/property/whatever `notifyFileChanged` to `IZoweExplorerFileApi` (the file API returned by `getFileApi()`) so extenders ... "

packages/zowe-explorer/CHANGELOG.md
```diff
@@ -6,6 +6,7 @@ All notable changes to the "vscode-extension-for-zowe" extension will be documen
 
 ### New features and enhancements
 
+- Added `notifyFileChanged` to `IZoweExplorerFileApi` (the file API returned by `getFileApi()`) so extenders that save to the mainframe outside of Zowe Explorer's filesystem provider can invalidate the cached entry and trigger VS Code to re-read the file from the mainframe. [#4481](https://github.com/zowe/zowe-explorer-vscode/issues/4481)
```
anaxceron: Same as my previous comment re: `notifyFileChanged`

anaxceron (commented):
Additional comments :)

packages/zowe-explorer-api/CHANGELOG.md
ddustindd: Good catch, will keep that in mind for future logs.

anaxceron (approved):

JillieBeanSim (approved):
Thanks for the fix @ddustindd

JillieBeanSim:
merging, dev-dep vulnerabilities will be addressed outside this branch

